PR #1, opened 2020-06-01
Bug 12973 - Unnecessary NullCommands being generated by Trigger Action

BrentEaston:
Pull Request Test.
Minor refactor of TriggerAction to use the Builder pattern of the Command class properly to allow NullCommands to be correctly replaced by the first non-null appended Command. 

Simplifies the Command sequence emitted by TriggerActions and simplifies debugging.


BrentEaston:
I am going to recreate these

PR #2, opened 2020-06-01
Bugs 12970 12972

BrentEaston:
12970 - Changes to Old... variables not being encoded in a Command or sent to other clients or log files, nor are they saved in the gamestate.  Undo will break any automation that depends on these.

Added new PersistentPropertyContainer interface to BasicPiece and Decorator only. Would have been nice to add it to the GamePiece interface, but that would break any Custom class that implements GamePiece. Results in a bit of ugly 'instanceof' checks in a couple of places.

12972 - backPoint and backMap used for the SendToLocation Back command where being saved in the gamestate, but no changes during play where being encoded and sent to other clients or a log file. Undo will break automation depending on these being up to date.


BrentEaston:
I am closing and recreating

PR #3, opened 2020-06-01
Bug 12973 - Unnecessary NullCommands being generated by Trigger Action

BrentEaston:
New Pull Request for By 12973. 

No change to functionality. Use Builder Pattern correctly to allow atomic NullCommands to be replaced by the Command that is being appended.

Reduces clutter and unnecessary complexity in Command Sequence and greatly simplifies debugging.


uckelman:
Looks good.

PR #4, opened 2020-06-02
Bugs 12970 12972 final before it drives me mad

BrentEaston:
Been learning a lot about Git :) Have had to this about 4 times, but finally have all the changes for these 2 bugs into a single PR. They are combined because SendToLocation is affected by both.

I've still managed to suck in src/VASSAL/tools/image/ImageIOImageLoader.java by mistake. There was an import I couldn't find.

### DIFF
--- a/src/VASSAL/build/module/map/PieceMover.java
+++ b/src/VASSAL/build/module/map/PieceMover.java
@@ -441,25 +441,30 @@ public class PieceMover extends AbstractBuildable
 
   /** Invoked just before a piece is moved */
   protected Command movedPiece(GamePiece p, Point loc) {
-    setOldLocation(p);
-    Command c = null;
+    Command c = new NullCommand();
+    c = c.append(setOldLocation(p));
     if (!loc.equals(p.getPosition())) {
-      c = markMoved(p, true);
+      c = c.append(markMoved(p, true));
     }
     if (p.getParent() != null) {
       final Command removedCommand = p.getParent().pieceRemoved(p);
-      c = c == null ? removedCommand : c.append(removedCommand);
+      c = c.append(removedCommand);
     }
     return c;
   }
 
-  protected void setOldLocation(GamePiece p) {
+  protected Command setOldLocation(GamePiece p) {
+    Command comm = new NullCommand();
     if (p instanceof Stack) {
       for (int i = 0; i < ((Stack) p).getPieceCount(); i++) {
-        Decorator.setOldProperties(((Stack) p).getPieceAt(i));
+        comm = comm.append(Decorator.setOldProperties(((Stack) p).getPieceAt(i)));
       }
     }
-    else Decorator.setOldProperties(p);
+    else {
+      comm = comm.append(Decorator.setOldProperties(p));
+    }
+    return comm;
+    
   }
 
   public Command markMoved(GamePiece p, boolean hasMoved) {
@@ -472,7 +477,7 @@ public class PieceMover extends AbstractBuildable
       if (p instanceof Stack) {
         for (Iterator<GamePiece> i = ((Stack) p).getPiecesIterator();
              i.hasNext();) {
-          c.append(markMoved(i.next(), hasMoved));
+          c = c.append(markMoved(i.next(), hasMoved));
         }
       }
       else if (p.getProperty(Properties.MOVED) != null) {
@@ -480,7 +485,7 @@ public class PieceMover extends AbstractBuildable
           final ChangeTracker comm = new ChangeTracker(p);
           p.setProperty(Properties.MOVED,
                         hasMoved ? Boolean.TRUE : Boolean.FALSE);
-          c = comm.getChangeCommand();
+          c = c.append(comm.getChangeCommand());
         }
       }
     }
@@ -656,22 +661,23 @@ public class PieceMover extends AbstractBuildable
 
     // Apply key after move to each moved piece
     if (map.getMoveKey() != null) {
-      applyKeyAfterMove(allDraggedPieces, comm, map.getMoveKey());
+      comm = comm.append(applyKeyAfterMove(allDraggedPieces, map.getMoveKey()));
     }
 
     tracker.repaint();
     return comm;
   }
 
-  protected void applyKeyAfterMove(List<GamePiece> pieces,
-                                   Command comm, KeyStroke key) {
+  protected Command applyKeyAfterMove(List<GamePiece> pieces, KeyStroke key) {
+    Command comm = new NullCommand();
     for (GamePiece piece : pieces) {
       if (piece.getProperty(Properties.SNAPSHOT) == null) {
         piece.setProperty(Properties.SNAPSHOT,
                           PieceCloner.getInstance().clonePiece(piece));
       }
-      comm.append(piece.keyEvent(key));
+      comm = comm.append(piece.keyEvent(key));
     }
+    return comm;
   }
 
   /**
--- a/src/VASSAL/build/module/map/PieceMover.java
+++ b/src/VASSAL/build/module/map/PieceMover.java
@@ -677,6 +677,14 @@ public class PieceMover extends AbstractBuildable
     return comm;
   }
 
+  /**
+   * Replace by {@link #applyKeyAfterMove(List, KeyStroke)} to return Commands
+   */
+  @Deprecated
+  protected void applyKeyAfterMove(List<GamePiece> pieces, Command comm, KeyStroke key) {
+    comm.append(applyKeyAfterMove(pieces, key));
+  }
+  
   protected Command applyKeyAfterMove(List<GamePiece> pieces, KeyStroke key) {
     Command comm = new NullCommand();
     for (GamePiece piece : pieces) {
--- a/src/VASSAL/build/module/map/PieceMover.java
+++ b/src/VASSAL/build/module/map/PieceMover.java
@@ -442,7 +442,7 @@ public class PieceMover extends AbstractBuildable
   /** Invoked just before a piece is moved */
   protected Command movedPiece(GamePiece p, Point loc) {
     Command c = new NullCommand();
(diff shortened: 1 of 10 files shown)


src/VASSAL/counters/BasicPiece.java
```diff
@@ -338,6 +359,26 @@ public void setProperty(Object key, Object val) {
     }
   }
 
+  @Override
+  public Command setPersistentProperty(Object key, Object val) {
+    if (persistentProps == null) {
+      persistentProps = new HashMap<>();
+    }
+    final Object oldVal = persistentProps.get(key);
```
yanlyub, line 380: This call to Map.get() can be omitted, the Map.put() and Map.remove() calls already return the previous value, e.g.:
``
final Object oldVal = (val == null)
  ? props.remove(key)
  : props.put(key, val);
``

src/VASSAL/counters/BasicPiece.java
```diff
@@ -571,6 +612,13 @@ public String getState() {
     final Point p = getPosition();
     se.append(p.x).append(p.y);
     se.append(getGpId());
+    se.append(persistentProps.size());
+    // Persistent Property values will always be String (for now).
+    for (Object key : persistentProps.keySet()) {
+      se.append(key.toString());
```
yanlyub, line 631: will throw an NPE if key is null, which is a valid key in HashMap

src/VASSAL/counters/BasicPiece.java
```diff
@@ -571,6 +612,13 @@ public String getState() {
     final Point p = getPosition();
     se.append(p.x).append(p.y);
     se.append(getGpId());
+    se.append(persistentProps.size());
+    // Persistent Property values will always be String (for now).
+    for (Object key : persistentProps.keySet()) {
+      se.append(key.toString()); 
+      String val = (String) persistentProps.get(key);
+      se.append(key == null ? "" : val);
```
yanlyub, line 633: key can never be null here else it would have thrown an NPE above

src/VASSAL/counters/BasicPiece.java
```diff
@@ -604,6 +652,21 @@ else if (oldMap != null) {
...
     setGpId(st.nextToken(""));
+    
+    if (persistentProps == null) {
+      persistentProps = new HashMap<>();      
+    }
+    else {
+      persistentProps.clear();
+    }
+      
+    // Persistent Property values will always be String (for now).
```
yanlyub, line 676: could make persistentProps a Map<String, String>

src/VASSAL/counters/BasicPiece.java
BrentEaston: Great, changed.

src/VASSAL/counters/BasicPiece.java
```diff
@@ -571,6 +612,13 @@ public String getState() {
     final Point p = getPosition();
     se.append(p.x).append(p.y);
     se.append(getGpId());
+    se.append(persistentProps.size());
+    // Persistent Property values will always be String (for now).
+    for (Object key : persistentProps.keySet()) {
```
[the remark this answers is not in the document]
BrentEaston: Updated to 

    persistentProps.forEach((key, val) -> {
      se.append(key == null ? "" : key.toString());
      se.append(val == null ? "" : val.toString());
    });

src/VASSAL/counters/BasicPiece.java
BrentEaston: Yep, plus the code was wrong, It should have been checking val for null.

src/VASSAL/counters/BasicPiece.java
BrentEaston: Thanks, same bug. Fixed whole section as per your suggestion
    persistentProps.forEach((key, val) -> {
      se.append(key == null ? "" : key.toString());
      se.append(val == null ? "" : val.toString());
    });
Much neater.

src/VASSAL/counters/BasicPiece.java
BrentEaston: Yes, I thought about that, however the setProperty/getProperty interface is Object, Object and there is no reason why me have a need for other Objects in the future. Better to keep it generalised.

BrentEaston: no reason why we may NOT have a need for other objects :)

BrentEaston:
Thanks Yan, 
Changes incorporated.

BrentEaston:
Final couple of changes to add debug details to SetPersistentPropertyCommand and to prevent Commands being generated if the Property value isn't changed.

src/VASSAL/counters/BasicPiece.java
```diff
@@ -378,7 +378,22 @@ public Command setPersistentProperty(Object key, Object val) {
       persistentProps = new HashMap<>();
     }
     final Object oldVal = (val == null) ? persistentProps.remove(key) : persistentProps.put(key, val);
-    return new SetPersistentPropertyCommand (getId(), key, oldVal, val);
+     
+    return setPersistentProperty(key, oldVal, val);
```
yanlyub: Have a look at java.lang.Objects.equals(Object a, Object b), this looks like the perfect spot to use it:
`return Objects.equals(oldVal, val) ? null : new SPPC(.....)`

BrentEaston: Perfect, thanks Yan. I have some other uses for that as well!

BrentEaston: Code Updated.

src/VASSAL/tools/image/ImageIOImageLoader.java
```diff
@@ -121,7 +121,7 @@ public ImageIOImageLoader(ImageTypeConverter tconv) {
   // stop calls to ProfileDeferralMgr.activateProfiles(), which is where the
   // race happens.
   static {
-    ProfileDeferralMgr.deferring = false;
+    //ProfileDeferralMgr.deferring = false;
```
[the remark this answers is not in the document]
BrentEaston: Yes, that was the solution. There is a place in the Eclipse project setup to define those. Thanks.

src/VASSAL/counters/BasicPiece.java
```diff
@@ -604,6 +662,28 @@ else if (oldMap != null) {
...
+      persistentProps.clear();
+    }
+      
+    // Persistent Property values will always be String (for now).
+    // Create the HashMap as lazily as possible, no point in creating it for pieces that never move
+    if (persistentProps != null) {
+      persistentProps.clear();
+    }
+    final int propCount = st.nextInt(0);
+    for (int i=0; i < propCount; i++) {
```
uckelman, line 679: `i = 0` (spaces)

src/VASSAL/counters/BasicPiece.java
```diff
@@ -338,6 +373,21 @@ public void setProperty(Object key, Object val) {
...
+    if (persistentProps == null) {
+      persistentProps = new HashMap<>();
+    }
+
+    final Object oldValue = (newValue == null) ? persistentProps.remove(key) : persistentProps.put(key, newValue);
+    return Objects.equals(oldValue, newValue) ? null : new SetPersistentPropertyCommand (getId(), key, oldValue, newValue);
+  }
+
+  @Override
+  public Object getPersistentProperty (Object key) {
```
uckelman, line 387: space

src/VASSAL/counters/BasicPiece.java
```diff
@@ -338,6 +373,21 @@ public void setProperty(Object key, Object val) {
     }
   }
 
+  @Override
+  public Command setPersistentProperty(Object key, Object newValue) {
+    if (persistentProps == null) {
+      persistentProps = new HashMap<>();
+    }
+
+    final Object oldValue = (newValue == null) ? persistentProps.remove(key) : persistentProps.put(key, newValue);
```
uckelman, line 382: No need for parentheses around the condition here.

src/VASSAL/counters/BasicPiece.java
```diff
@@ -338,6 +373,21 @@ public void setProperty(Object key, Object val) {
...
+    return Objects.equals(oldValue, newValue) ? null : new SetPersistentPropertyCommand (getId(), key, oldValue, newValue);
```
uckelman, line 383: Just checking to be sure this is what you want, because this surprised me when I first saw it: `Objects.equals()` returns `true` if both arguments are `null`.

src/VASSAL/build/module/map/PieceMover.java
```diff
@@ -441,25 +441,39 @@ protected boolean isMultipleSelectionEvent(MouseEvent e) {
...
   }
-
+  
+  /**
+   * Replaced by {@link #setOldLocations(GamePiece)} to return generated Commands
+   * @param p
+   */
+  @Deprecated
   protected void setOldLocation(GamePiece p) {
+    setOldLocations(p);
```
uckelman, line 462: So, what will happen now if the old function is called? The new function returns a `Command`, which appears to be thrown away?

src/VASSAL/counters/BasicPiece.java
BrentEaston: Yes, that is correct. We do not want to generate a Command if there has been no change in the value. If both are null, then no change, so return null for the result.

src/VASSAL/build/module/map/PieceMover.java
BrentEaston: Yes, that's right. The Old function will only be called by existing custom code subclassing the PieceMover. They are not expecting a Command to be returned.  The Command to be returned is built up in a local variable and there is no way we can 'make' them use our new Commands. The best we can do is keep the old code working in the old broken way.

BrentEaston:
Pushed format changes and missing @Deprecated annotation.

This code review facility is EXCELLENT!

PR #5, opened 2020-06-02
Bugfix 12554: Do not add the whole Stack to the DragBuffer

yanlyub:


yanlyub:
- Changed the code to select all units regardless of selection when Map.MOVING_STACKS_PICKUP_UNITS is set
- Test using Stalingrad 42 module: select multiple stacks, move them 3 times, then undo 3 times. Result: each of the 3 moves results in the stacks staying in the same position relative to each other, the counters in the stacks stay in the same order, each of the 3 undos move the stacks back to their original positions, with the counters staying in the proper order.
- Test using Stalingrad 42 module: select a stack and a single counter, move 3 times, then undo 3 times. Result: the stack and the single counter stay in the same position relative to each other, the counters in the stack stay in the correct order, the undo actions move the stack and the single counter back to their original position, the counters in the stack stay in the proper order.

BrentEaston:
Terrific. I did a few tests (on a Stacking Map) and it looks fine to me as well. I see no problem with the Drag Shadow that was alluded to in the comments.

yanlyub:
Tested using a map with stacking disabled:
- Moved single counter 3 times followed by undo 3 times, working as expected
- Moved several counters in unison 3 times, undo 3 times, counters keep their position relative to each other, each undo worked as expected
- Moved a stack 3 times followed by undo 3 times, working as expected
- Highlight a stack and a single counter, move them in unison 3 times, undo 3 times, working as expected
- Move a stack once, move units out of the stack each to a different position, then undo until the units are back at the original position in 1 stack, working as expected

uckelman:
There's a particular point above I want to understand: Does this change mean that the drag cursor no longer shows whole stacks if you drag a stack?

BrentEaston:
Personally, I could not see any difference between the drag cursor pre and post this change.

src/VASSAL/build/module/map/PieceMover.java
```diff
@@ -244,25 +244,16 @@ public Object visitDeck(Deck d) {
...
-        else {
-          for (int i = 0; i < s.getPieceCount(); i++) {
-            final GamePiece p = s.getPieceAt(i);
-            if (Boolean.TRUE.equals(p.getProperty(Properties.SELECTED))) {
-              DragBuffer.getBuffer().add(p);
-            }
+        for (int i = 0; i < s.getPieceCount(); i++) {
+          final GamePiece p = s.getPieceAt(i);
+          final boolean pieceSelected = Boolean.TRUE.equals(p.getProperty(Properties.SELECTED));
+          if (selectAllUnitsInStackRegardlessOfSelection || pieceSelected) {
```
uckelman, line 255: Calling `p.getProperty(Properties.SELECTED)` in the case where we don't need it means we don't get the advantage of short-circuit evaluation. Better would be:

```
if (selectAllUnitsInStackRegardlessOfSelection ||
    Boolean.TRUE.equals(p.getProperty(Properties.SELECTED)) {
```

uckelman: Aside: It would be great to have `Stack` implement `Iterable<GamePiece>` so all the loops which are by index but don't use the index for anything other than calling `getPieceAt()` could just have the `GamePiece` directly.

yanlyub: The local boolean variable is there to improve readability, `if (selectAllRegardless || pieceSelected)` reads much better than `if (selectAllRegardless || Boolean.TRUE.methodParenthesisVariableDotMethodParenthesisClassDotConstantParenthesis)`. If it is really necessary to optimize performance at this particular spot without having measured any performance issues here, I can make the requested change, but I ask to reconsider. Code should be written to be read by human developers first and by machines second.

The idea of having the Stack hand out its contents in a standard way already crossed my mind, and I feel the optimal way for this is having the Stack "HAVE A" List of GamePieces (composition) instead of "BE A" List of GamePieces (inheritance). A simple `List<GamePiece> Stack.getContents()` or `getGamePieces()` would solve this in a cleaner way without binding the Stack itself to a new type, and `for (GamePiece piece : aStack)` would look a bit weird while with a `for (GamePiece piece : aStack.getContents())` it would be easier to understand what is going on.

uckelman: On reflection, I think the nicest way to do this would be if `Stack` supported `forEach()`:

```
final DragBuffer d = DragBuffer.getBuffer();
if (selectAllUnitsInStackRegardlessOfSelection) {
  s.forEach((p) -> d.add(p));
}
else {
  s.forEach((p) -> {
    if (Boolean.TRUE.equals(p.getProperty(Properties.SELECTED))) {
      d.add(p);
    }
  );
}
```

uckelman: Even more concise:
```
final DragBuffer d = DragBuffer.getBuffer();
s.forEach(
  selectAllUnitsInStackRegardlessOfSelection ?
  (p) -> d.add(p) :
  (p) -> {
    if (Boolean.TRUE.equals(p.getProperty(Properties.SELECTED))) {
      d.add(p);
    }
  }
);
```

yanlyub: For the iterations over the contents of a Stack, I have a proposal for this in PR #23. Why break a leg and have the Stack itself be an Iterable or a foreach-able thing, if the Stack can spit out a List which already has all this?

uckelman:
I'm going to merge this now, and make the change I suggested in the next commit.

PR #7, opened 2020-06-02
Bug 12576 - if At-Start stack has owningBoard "Any", default to first…

Stew-rt:
Bug 12576 - if At-Start stack has owningBoard "Any", default to first selected board, not first allowable board. If no selected board is available, old default remains.

### DIFF
--- a/src/VASSAL/build/module/map/BoardPicker.java
+++ b/src/VASSAL/build/module/map/BoardPicker.java
@@ -36,6 +36,7 @@ import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Vector;
+import java.util.stream.Collectors;
 
 import javax.swing.Box;
 import javax.swing.BoxLayout;
@@ -416,12 +417,8 @@ public class BoardPicker extends AbstractBuildable implements ActionListener, Ga
   /**
    * @return an array of the names of all boards from which have been selected either by the user via the dialog or from reading a savefile
    */
-  public String[] getSelectedBoardNames() {
-    final ArrayList<String> s = new ArrayList<>(currentBoards.size());
-    for (Board b : currentBoards) {
-      s.add(b.getName());
-    }
-    return s.toArray(new String[0]);
+  public List<String> getSelectedBoardNames() {
+    return currentBoards.stream().map(Board::getName).collect(Collectors.toList());
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Stew-rt:
I'm still new with this codebase, so please check this PR makes sense...

src/VASSAL/build/module/map/BoardPicker.java
```diff
@@ -413,6 +413,17 @@ public void actionPerformed(ActionEvent e) {
     }
   }
 
+  /**
+   * @return an array of the names of all boards from which have been selected either by the user via the dialog or from reading a savefile
+   */
+  public String[] getSelectedBoardNames() {
```
yanlyub, line 419: it's common to use java.util.List in Java:
`public List<String> getSelectedBoardNames { List<String> result = new ArrayList<>(); /* ... */ return result; }`

or using the new stream API:
`public List<String> getSelectedBoardNames { boards.stream().map(Board::getName).collect(Collectors.toList()); }`

The caller can then do:
`List<String> boardNames = map.getBoardPicker().getSelectedBoardNames(); if (!boardNames.isEmpty()) owningBoardName = boardNames.get(0);`

Stew-rt: Ok, thanks, will do.
I was just trying to retain the code style the same as the existing getAllowableBoardNames and getAllowableLocalizedBoardNames classes in this file

yanlyub: I am not one of the maintainers and don't really have a say in these matters but I would not add more bad code to already existing bad code, especially if it is a public method that will potentially become public API which module developers will use in their modules.

BrentEaston: Yes, for new methods, use Lists rather than arrays.  If you want to go further and refactor existing methods, just remember that an unknown amount of existing custom code in modules may use any non-private methods, so @Deprecate the existing method that returns an array and create a new method that returns a List.

Stew-rt: Roger Wilco.
I think if refactoring further methods, I'll do so in a separate PR.
I've noticed I've managed to create this PR against my master branch, so I'm probably going to close this PR this evening, and recreate it against a proper local branch... (unless it gets merged before so)

Stew-rt:
Closing due to creation against my local master, will recreate against proper local trunk

PR #8, opened 2020-06-02
Added unit tests for some methods of Stack class

yanlyub:


yanlyub:
Unit tests for what seem to be the main methods of the Stack class.

PR #9, opened 2020-06-03
Eclipse Project Files. 

BrentEaston:
Rarely changes, but are needed for Eclipse project when creating new branches

Could you check these 2 files into the master please Joel. They make life a LOT easier for Eclipse users. Without them. new branches aren't recognised automatically as new Java projects as you have to fiddle around with re-creating them.

Thanks.


BrentEaston:
That should be the final changes.

PR #13, opened 2020-06-04
Bug 12952 - Evalute $..$ strings in Property Match expressions on source pce

BrentEaston:
Turned out the $..$ variables just wheren't being evalated at all on either the source or the targets. So this can't possibly be working in any module.

### DIFF
--- a/src/VASSAL/counters/Labeler.java
+++ b/src/VASSAL/counters/Labeler.java
@@ -154,7 +154,7 @@ public class Labeler extends Decorator implements TranslatablePiece, Loopable {
    */
   protected String clean(String s) {
     // Cannot use $PieceName$ in a label format, must use $pieceName$
-    return s.replaceAll("$"+BAD_PIECE_NAME+"$", "$"+PIECE_NAME+"$");
+    return s.replace("$"+BAD_PIECE_NAME+"$", "$"+PIECE_NAME+"$");
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


Stew-rt:
Seeing this has reminded me, I stumbled across a weird line in counters.Labeler the other day, where `Labeler.clean` returns:
`return s.replaceAll("$"+BAD_PIECE_NAME+"$", "$"+PIECE_NAME+"$");`

according to the docs String.replaceAll rakes a regex... meaning i'm confused as to what it would do with `$ThisBadPieceName$` as dollar is a special for line end.... well, usually, anyhow.

BrentEaston:
Ah, good spot. That is a bug. You are correct, I just did a quick test and the replace as written does not work. It should be s.replace(), not s.replaceAll().

I will add the fix here.

src/VASSAL/counters/Labeler.java
```diff
@@ -154,7 +154,7 @@ public void mySetType(String type) {
    */
   protected String clean(String s) {
     // Cannot use $PieceName$ in a label format, must use $pieceName$
-    return s.replaceAll("$"+BAD_PIECE_NAME+"$", "$"+PIECE_NAME+"$");
+    return s.replace("$"+BAD_PIECE_NAME+"$", "$"+PIECE_NAME+"$");
```
uckelman, line 157: I see why this happened---having functions named `replace` and `replaceAll` suggests that the former will replace the first occurrence while the latter will replace all occurrences---but that's not what they do.

PR #16, opened 2020-06-12
Bug 13060 - Check unit is on a Map before trying to generate the Shape

BrentEaston:
Bug 13060. Shouldn't happen, but we are getting NPE's trying to generate the shape when a piece has no parent Map. Added defensive coding to check map is non-null before trying to use it.

### DIFF
--- a/src/VASSAL/counters/AreaOfEffect.java
+++ b/src/VASSAL/counters/AreaOfEffect.java
@@ -199,13 +199,14 @@ public class AreaOfEffect extends Decorator implements TranslatablePiece, MapSha
           AlphaComposite.SRC_OVER, transparencyLevel));
 
         Area a = getArea();
-
-        if (zoom != 1.0) {
-          a = new Area(AffineTransform.getScaleInstance(zoom,zoom)
-                                      .createTransformedShape(a));
+        
+        if (a != null) {
+          if (zoom != 1.0) {
+            a = new Area(AffineTransform.getScaleInstance(zoom,zoom)
+                        .createTransformedShape(a));
+          }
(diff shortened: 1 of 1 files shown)


src/VASSAL/counters/AreaOfEffect.java
```diff
@@ -199,14 +199,14 @@ public void draw(Graphics g, int x, int y, Component obs, double zoom) {
...
+        
+        if (a != null) {
+          if (zoom != 1.0) {
+            a = new Area(AffineTransform.getScaleInstance(zoom,zoom).createTransformedShape(a));
+          }
+          g2d.fill(a);
         }
 
-        g2d.fill(a);
-
```
uckelman, line 209: Suggestion: If `a` is null, then nothing else in the same scope as it needs to be done. You could do this instead:
```
        Area a = getArea();
        if (a != null) {
          final Graphics2D g2d = (Graphics2D) g;

          final Color oldColor = g2d.getColor();
          g2d.setColor(transparencyColor);

          final Composite oldComposite = g2d.getComposite();
          g2d.setComposite(AlphaComposite.getInstance(
            AlphaComposite.SRC_OVER, transparencyLevel));

          if (zoom != 1.0) {
            a = new Area(AffineTransform.getScaleInstance(zoom,zoom)
                                        .createTransformedShape(a));
          }
  
          g2d.fill(a);
  
          g2d.setColor(oldColor);
          g2d.setComposite(oldComposite);
        }
```

BrentEaston: Good point. Changes Incorporated

PR #19, opened 2020-06-13
Refactoring - fold expressions into stream chains

yanlyub:
Use modern Java's stream chains to improve code readability and maintainability


src/VASSAL/build/module/DiceButton.java
```diff
@@ -312,10 +312,7 @@ public boolean shouldBeVisible() {
 
   @Override
   public VisibilityCondition getAttributeVisibility(String name) {
-    if (N_DICE.equals(name)
-        || N_SIDES.equals(name)
-        || PLUS.equals(name)
-        || ADD_TO_TOTAL.equals(name)) {
+    if (List.of(N_DICE, N_SIDES, PLUS, ADD_TO_TOTAL).contains(name)) {
```
uckelman, line 315: All of these lists are compile-time constants, but it looks like this creates a new unmodifiable list each time the condition is hit. That seems wasteful. Am I misunderstanding what this code does?

yanlyub: This is about syntactic sugar and code readability, and letting the compiler and the JVM take care of performance optimization, which the modern JVMs are really good at.

uckelman: Can you point me to a benchmark for this?

yanlyub: If you mean a benchmark for code readability, I would guess it is generally easier to read "if list of A, B, C contains X" than these equals chains.

If you mean a CPU/memory performance benchmark, there can be no one benchmark for this in Java, the JVM optimizes code during runtime and the garbage collection is also a very important topic that impacts performance. It's a very complex topic but generally, if a certain piece of code is ran often enough, the JVM starts optimizing it, using loop unrolling, method inlining, and even compiling the bytecode to native code. If this particular place in the code is ran often enough, it will be made to run (almost) as fast as native C++ code, during runtime. And if this place in the code is not ran often enough for the JVM to optimize it, then it does not need optimization by the developer. In Java it is common to write code with readability in mind, write it for other developers not for the machine. Then when there are performance concerns, the code should be profiled e.g. with JProfiler and optimized in the specific areas where performance bottlenecks were measured. Proactive optimization sprinkled throughout the code makes the code less readable and maintainable and does not help with the real performance hogs.

PR #21, opened 2020-06-13
Refactoring - decoupled Deck from static GameModule access

yanlyub:
This is a first step towards making the Deck class testable, as right now it is not even possible to do a simple `new Deck()` without an existing GameModule object tree. After this change it will be possible to provide the Deck with a mocked GameModule upon instantiation.

### DIFF
--- a/src/VASSAL/counters/Deck.java
+++ b/src/VASSAL/counters/Deck.java
@@ -166,18 +166,47 @@ public class Deck extends Stack implements PlayerRoster.SideChangeListener {
     }
   };
 
+  private final GameModule gameModule;
+
+  /**
+   * @deprecated use {@link #Deck(GameModule)}
+   */
+  @Deprecated
   public Deck() {
-    this(ID);
+    this(GameModule.getGameModule());
+  }
+
+  /**
+   * @deprecated use {@link #Deck(GameModule, String)}
+   */
+  @Deprecated
+  public Deck(String type) {
+    this(GameModule.getGameModule(), type);
+  }
+
+  /**
+   * @deprecated use {@link #Deck(GameModule, String, PropertySource)}
+   */
+  @Deprecated
+  public Deck(String type, PropertySource source) {
+    this(GameModule.getGameModule(), type, source);
+  }
+
+  public Deck(GameModule gameModule) {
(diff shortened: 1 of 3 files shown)


src/VASSAL/counters/Deck.java
```diff
@@ -166,18 +166,47 @@ public String encode(Command c) {
...
+  @Deprecated
+  public Deck(String type, PropertySource source) {
+    this(GameModule.getGameModule(), type, source);
+  }
+
+  public Deck(GameModule gameModule) {
+    this(gameModule, ID);
+
+    // the constructor above already adds this listener, why do it a second time?
+    PlayerRoster.addSideChangeListener(this);
```
uckelman, line 199: This looks like a bug. @BrentEaston, thoughts?

The change listener was added in 605fb4ca0 back in 2007. I have a hard time seeing how it makes sense to have the same `Deck` receive two notifications for the same change. Could this be the cause of some weird deck problems we've seen?

yanlyub: Bit off topic but wait til I get to write a proper test for `contents.length` in Deck line 250, that one smells 100% like a bug.

BrentEaston: Yes, that is definitely a bug whenever the no-arg constructor is called. 

However, I don't think it is causing any problems. Apart from SideChange events being quite rare in normal play, the SideChange listener is completely harmless, just updating the Deck counts maintained by the Deck. Calling it twice does no damage.

The weird problems we have are due to the KeyListeners for the Shuffle key being added twice when you load multiple save games in the same session. The KeyListeners added in mySetType are never removed and the Deck objects appear to be re-used between games in the same session. And I've just realised why my last fix didn't work (adding keylistener add/remove to game start/end), I didn't remove the addKeyListeners from mySetType().

I think a bigger question is why are Deck objects being re-used across games sessions. The DrawPile is the Vassal component and is persistent, but the associated Deck should be re-created from scratch with a new Load game.

Ref: http://www.vassalengine.org/tracker/show_bug.cgi?id=12538

uckelman: Ok. I'm going to remove that listener after merging this.

PR #22, opened 2020-06-14
Refactoring - Use single map method from Java 8

yanlyub:
Another minor refactoring that simplifies code

### DIFF
--- a/src/VASSAL/build/module/map/MenuDisplayer.java
+++ b/src/VASSAL/build/module/map/MenuDisplayer.java
@@ -159,11 +159,7 @@ public class MenuDisplayer extends MouseAdapter implements Buildable {
         if (keyCommand.getName() != null &&
                 keyCommand.getName().length() > 0 &&
                 item != null) {
-          ArrayList<JMenuItem> l = commandNames.get(keyCommand.getName());
-          if (l == null) {
-            l = new ArrayList<>();
-            commandNames.put(keyCommand.getName(), l);
-          }
+          ArrayList<JMenuItem> l = commandNames.computeIfAbsent(keyCommand.getName(), k -> new ArrayList<>());
           l.add(item);
         }
       }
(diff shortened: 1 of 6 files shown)


src/VASSAL/build/module/map/MenuDisplayer.java
```diff
@@ -159,11 +159,7 @@ public static JPopupMenu createPopup(GamePiece target, boolean global) {
         if (keyCommand.getName() != null &&
                 keyCommand.getName().length() > 0 &&
                 item != null) {
-          ArrayList<JMenuItem> l = commandNames.get(keyCommand.getName());
-          if (l == null) {
-            l = new ArrayList<>();
-            commandNames.put(keyCommand.getName(), l);
-          }
+          ArrayList<JMenuItem> l = commandNames.computeIfAbsent(keyCommand.getName(), k -> new ArrayList<>());
```
uckelman, line 162: I want to check my understanding of what `Map.computeIfAbsent()` does. It the advantage of it over `Map.putIfAbsent()` that the value is never constructed if the key is present, whereas with `putIfAbsent()` one is constructing a value one might not use? (Related question: Is the `Function` you construct instead any cheaper?)

yanlyub: They have slightly different use cases.

`putIfAbsent()` is a little bit like the old `put()`:
- key exists in map: return the previous value, don't put new value
- key absent from map: return `null`, put the new value

`computeIfAbsent()` is different:
- key exists in map: return the previous value, don't put new value
- key absent from map: return _new value_ (`putIfAbsent()` would return a `null` in this case), compute and put the new value

The `Function` constructed for `computeIfAbsent()` or `computeIfPresent()` is only evaluated if the key was absent or present, respectively, so if the time to compute the value is of concern these `computeIfX()` methods should be preferred to the `put*()` family.

Overall, I tend to remember it like this:
- just want to put new value if absent? use `putIfAbsent()`
- want to put _and use_ new value, if absent? use `computeIfAbsent()`
- want to save time on computing the value if there is a chance the value wont be needed? use the `computeIfX()` family

PR #23, opened 2020-06-15
Refactoring - added method Stack.asList

yanlyub:
For now this is only for discussion as it will probably clash with PR #5.

I am not too happy with the naming, but the simpler `getPieces()` is already taken by the old enumeration method.

This change would have some nice side effects, most of the clients of `getPiecesIterator()` can be rewritten to use the new method, and the List.stream() allows other interesting code improvements.

The whole `PieceIterator` and `PieceFilter` can probably be removed altogether, in favor of this new method and `Predicate<GamePiece>` from the stream API.

### DIFF
--- a/src/VASSAL/build/module/map/MovementReporter.java
+++ b/src/VASSAL/build/module/map/MovementReporter.java
@@ -22,6 +22,7 @@ import java.awt.Point;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.function.Predicate;
 
 import VASSAL.build.GameModule;
 import VASSAL.build.module.Chatter;
@@ -119,9 +120,8 @@ public class MovementReporter {
     Command c = null;
     if (p instanceof Stack) {
       c = new NullCommand();
-      for (Iterator<GamePiece> i = ((Stack) p).getPiecesIterator();
-           i.hasNext();) {
-        c.append(markMoved(i.next()));
+      for (GamePiece gp : ((Stack)p).getPiecesAsList()) {
+        c.append(markMoved(gp));
       }
     }
     else if (p.getProperty(Properties.MOVED) != null) {
@@ -262,14 +262,10 @@ public class MovementReporter {
         return false;
       }
       if (target instanceof Stack) {
-        for (Iterator<GamePiece> i = ((Stack) target).getPiecesIterator(); i.hasNext() ;) {
-          GamePiece piece = i.next();
-          if (Boolean.TRUE.equals(piece.getProperty(Properties.INVISIBLE_TO_ME))
-              || Boolean.TRUE.equals(piece.getProperty(Properties.INVISIBLE_TO_OTHERS))) {
-            return true;
-          }
-        }
-        return false;
+        final Predicate<GamePiece> gamePiecePredicate =
+          piece -> Boolean.TRUE.equals(piece.getProperty(Properties.INVISIBLE_TO_ME))
+            || Boolean.TRUE.equals(piece.getProperty(Properties.INVISIBLE_TO_OTHERS));
+        return ((Stack) target).getPiecesAsList().stream().anyMatch(gamePiecePredicate);
       }
       else {
         return Boolean.TRUE.equals(target.getProperty(Properties.INVISIBLE_TO_ME))
@@ -389,8 +385,7 @@ public class MovementReporter {
       boolean first = true;
       for (GamePiece piece : pieces) {
         if (piece instanceof Stack) {
-          for (Iterator<GamePiece> j = ((Stack) piece).getPiecesIterator(); j.hasNext(); ) {
-            GamePiece p = j.next();
+          for (GamePiece p : ((Stack) piece).getPiecesAsList()) {
             if (isInvisible(p)) {
               if (!first) {
                 names.append(", ");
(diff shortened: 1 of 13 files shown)


uckelman:
Nothing else can be in a stack other than pieces. How about `asList()`?

src/VASSAL/counters/Stack.java
```diff
@@ -76,14 +77,25 @@ public Stack(GamePiece p) {
...
    */
   @Deprecated
   public Enumeration<GamePiece> getPieces() {
     return new EnumeratedIterator<>(new AllPieceIterator());
     //    return new AllPieceEnum();
   }
 
+  /**
+   * @return a defensive copy of the {@link List} of {@link GamePiece}s contained in this {@link Stack}
+   */
```
uckelman, line 90: I need to think about whether a copy is the right thing, or if what we want is (instead, or additionally) a way to get an unmodifiable view.

yanlyub:
`asList()` is a good idea, I wanted to go with the current naming scheme where all the accessors are named something with "pieces" but since the method would already return `List<GamePiece>` it would be clear what exactly `asList()` would return.

Unmodifiable List or defensive copy, I have no preference of one vs. the other, I think it's a minor nuance and it can be either, I can make it an unmodifiable List just as well. Both in a way tell the client "you can read from this but don't waste your time changing its contents, it won't accomplish anything".

src/VASSAL/counters/Stack.java
uckelman: I can't find any place that we use `Stack.getPiecesIterator()` just to fill a list of pieces. Returning an unmodifiable view of `contents` is what we want. If something which doesn't exist now really needs its own copy of the list, it can make one itself.

yanlyub:
Forgot that `UnmodifiableList` is somewhat hidden and cannot be a public interface type. Still, telling the client it's unmodifiable and wrapping in a unmodifiable `List` implementation is better.

I guess it's good enough to become a proper PR now, might even make sense to merge it before PR #5 then I could change PR #5 to use these new methods.

uckelman:
> Forgot that `UnmodifiableList` is somewhat hidden and cannot be a public interface type. Still, telling the client it's unmodifiable and wrapping in a unmodifiable `List` implementation is better.

Isn't `Collections.unmodifiableList()` the thing you want for that?

> I guess it's good enough to become a proper PR now, might even make sense to merge it before PR #5 then I could change PR #5 to use these new methods.

Yes, that's what I was hoping.

src/VASSAL/build/module/map/MovementReporter.java
```diff
@@ -119,9 +120,8 @@ public Command markMoved(GamePiece p) {
     Command c = null;
     if (p instanceof Stack) {
       c = new NullCommand();
-      for (Iterator<GamePiece> i = ((Stack) p).getPiecesIterator();
-           i.hasNext();) {
-        c.append(markMoved(i.next()));
+      for (GamePiece gp : ((Stack)p).asList()) {
+        c.append(markMoved(gp));
```
uckelman, line 124: @BrentEaston Is this a bug? Should it be `c = c.append(markMoved(gp));`?

src/VASSAL/counters/Stack.java
```diff
@@ -76,14 +78,25 @@ public Stack(GamePiece p) {
...
+   * @return an unmodifiable {@link List} of {@link GamePiece}s contained in this {@link Stack}
+   */
+  public List<GamePiece> asList() {
+    List<GamePiece> result = new ArrayList<>();
+    for (int i = 0; i < pieceCount; i++) {
+      result.add(contents[i]);
+    }
+    return Collections.unmodifiableList(result);
+  }
+
```
uckelman, line 99: Oh! I just noticed that we're making a copy anyway, because `contents` is an array.

To avoid that, and also because it's much shorter, `asList()` could be:

```
return Collections.unmodifiableList(Arrays.asList(contents).sublist(0, pieceCount));
```

test/VASSAL/counters/StackTest.java
```diff
@@ -464,4 +466,83 @@ public void getPiecesReverseIteratorSholdIterateInReverseOrder() {
...
+    final List<GamePiece> pieces = s.asList();
+
+    // assert
+    assertEquals(gamePiece0, pieces.get(0));
+    assertEquals(gamePiece1, pieces.get(1));
+    assertEquals(gamePiece2, pieces.get(2));
+  }
+
+  @Test
+  public void asListShouldReturnDefensiveCopy() {
```
uckelman, line 509: This test will fail with `asList()` returning an unmodifiable view.

src/VASSAL/counters/Stack.java
yanlyub: Looks good

test/VASSAL/counters/StackTest.java
yanlyub: Test removed

yanlyub:
Not sure how github works exactly, but if it's possible to smash this whole PR into 1 commit when merging, it might look a bit cleaner in the final commit tree.

uckelman:
> Not sure how github works exactly, but if it's possible to smash this whole PR into 1 commit when merging, it might look a bit cleaner in the final commit tree.

I don't know of a way to do that here. (I could do an interactive rebase locally to squash it, but then I'd have to do a force push... Doesn't seem worth it.)

src/VASSAL/build/module/map/MovementReporter.java
BrentEaston: Yes it should, otherwise the initial NullCommand is not replaced by the first append.

uckelman: Thanks. Fixed in 7dcc5ddda.

PR #24, opened 2020-06-15
Refactoring - make the Deck testable

yanlyub:
Another step towards make the `Deck` testable. For now as a draft. The static access to `PlayerRoster` had to be removed, after that the first unit tests can be written.

The static `PlayerRoster.getInstance()` had to be resolved, I decided to make it a method of the `GameModule` as it is the container for a possible `PlayerRoster` instance.

The static `PlayerRoster.addSideChangeListener()` also had to be changed, I decided to make it a method of the `GameModule` as well. It needed access to the PlayerRoster's sideChangeListeners so `PlayerRoster` received a new accessor method for it. And now the `GameModule.addSideChangeListenerToPlayerRoster()` is simply a convenience method, maybe it makes sense to remove it entirely and have the clients do what it does now i.e. get the `PlayerRoster` instance from the `GameModule`, do a null-check on it, and add the listener via the new `PlayerRoster.addSideChangeListenerToInstance()` method.

The naming for `PlayerRoster.addSideChangeListenerToInstance()`, I am not too happy with, but the `addSideChangeListener` is already used for the static method.

Taking this further, the sibling `static PlayerRoster.removeSideChangeListener()` should also change in the same way. Taking this even further, the many static methods of `PlayerRoster` can change into regular methods as well, and the clients can get a possible `PlayerRoster` instance from the `GameModule`, do the null-check themselves and then call these new methods.

I have also added a simple first test for `Deck`.

The change is backwards compatible, I have declared the methods as deprecated and referenced the new methods in their javadoc.

I have thought of two other ways of doing this, this particular way feels the least cumbersome and invasive, and uses the fact that one `GameModule` already is directly related to zero or one `PlayerRoster`s as an advantage. The other ways broke up this structure and either passed both a `GameModule` and a `PlayerRoster` to a new `Deck` via constructor, or passed some new container that contained both a `GameModule` and a `PlayerRoster` via `Deck`s constructor.

This part: `getComponentsOf(PlayerRoster.class).stream().findFirst().orElse(null)` is a replacement for the previous for-loop in `PlayerRoster.getInstance()` that did the same thing, it either returned the first found object, or null. I deduced from this that one `GameModule` is related to (has-a relation) zero or one `PlayerRoster`.

### DIFF
--- a/src/VASSAL/build/GameModule.java
+++ b/src/VASSAL/build/GameModule.java
@@ -1127,4 +1127,21 @@ public abstract class GameModule extends AbstractConfigurable implements Command
     myI18nData.setAttributeTranslatable(MODULE_VERSION, false);
     return myI18nData;
   }
+
+  /**
+   * @return the {@link PlayerRoster} instance, or <code>null</code> if no {@link PlayerRoster} exists
+   * within this {@link GameModule}
+   */
+  public PlayerRoster getPlayerRoster() {
+    return getComponentsOf(PlayerRoster.class).stream()
+                                              .findFirst()
+                                              .orElse(null);
+  }
+
+  public void addSideChangeListenerToPlayerRoster(PlayerRoster.SideChangeListener l) {
+    PlayerRoster r = getPlayerRoster();
+    if (r != null) {
+      r.addSideChangeListenerToInstance(l);
+    }
+  }
 }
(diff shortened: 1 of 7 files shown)


uckelman:
The naming is not my favorite, but as you note the method we're deprecating is the one which has the best name.

uckelman:
It's a thing where it would be nice to round up all the modules which depend on the old method, have them change to the new method, remove the old one, create a new new method with the original name, and then deprecate the new method with the ugly name...

PR #26, opened 2020-06-16
Add initial checkstyle configuration

yanlyub:
I did not know where else to put it, usually these things go in `src/main/resources` but in the absence of the near-standard project structure, I created a `resources` subdirectory in the root.

This is a configuration for checkstyle, it can be used as an IDE plugin during coding to report code style violations, a code formatter can be deduced from this as well. It also works as a CI plugin that would test compliance with the code style during the build.

Plugins for the IDEs can be found here:
- https://checkstyle.org/eclipse-cs
- https://plugins.jetbrains.com/plugin/1065-checkstyle-idea

As for the code style itself, the rules can report violations in different levels, the code can then be adjusted in several steps. I took the commonly used Google Java Style and adjusted some of the rules to fit the current Vassal codebase:

- errors:
use of tabs, wrong left/right curly brace placement, wrong indentation
- warnings:
most of the other rules report as warnings
- info:
I moved some rules that I felt can be applied at the very last to info


uckelman:
How do I run it from the command line?

yanlyub:
This might get too complicated on it's own and might not be worth the trouble, it's much easier to plug it into maven or use it as an IDE plugin. With maven it becomes just another part of the regular build pipeline, and can be run in standalone mode with a simple `mvn checkstyle:checkstyle` on the command line.

Should probably decide about maven first, after that think about this one.

uckelman:
Running it via maven would be a way to run it on the command line. That answers my question.

yanlyub:
This is now combined with the new maven setup. It is ran as part of the build in the validate phase i.e. before compiling the code, and can be triggered manually with either `mvn clean validate` or `mvn checkstyle:check`.

The configuration is set to only report indentation, tabs and curly brace violations as errors, and to not break the build when these errors are encountered.

The violations are reported to stdout in the regular maven output, if that is too much we can also redirect the checkstyle part of the output to a file.

There are currently 2528 violations of the above mentioned rules, once we remove them we can change it so they break the build.

yanlyub:
New plan, the commits that reflect this are already in this PR:

- set the default severity level for all checkstyle checks (except the ones below) to "info"
- set the severity level for the checks of tabs, indentation, curly brace placement to "warning"
- configure checkstyle to break the build when severity level "error" is encountered
- step by step, get rid of the warnings
- once they are gone, raise the severity level for tabs/indentation/curly brace to "error" so they break the build, at the same time raise the severity level for the next set of checks from "info" to "warning"
- follow this iterative process until the code looks good enough and all checkstyle checks are assigned the proper severity level

Also: I give up the idea of having the same code formatters for all developers, this has several reasons:
- checkstyle only enforces a minimal set of rules, while the usual code formatters enforce very concrete rules, so the checkstyle configuration does not map 1:1 to the configuration of a code formatter
- it used to be possible to convert a checkstyle configuration to an Eclipse code formatter but this feature is broken since a while and will be removed in the future
- the code formatters themselves do not map 1:1 to each other, the Eclipse code formatter has different settings from the IntelliJ code formatter and they can not be fully converted between each other
- enforcing a certain code formatter for all developers is not the way to go, the modus operandi "commit, push, have the build server tell you whether you broke the build by violating our minimal code style requirements" is better as it allows the developers to use whatever code formatter they want as long as they comply with the projects code style rules

yanlyub:
Whoopsie....

https://travis-ci.com/github/vassalengine/vassal/builds/174146426#L23423 "The job exceeded the maximum log length, and has been terminated."

uckelman:
Oh that's fun. Thanks for running down a (temporary) fix.

PR #27, opened 2020-06-16
Use maven to compile and run tests

yanlyub:
How about this neat little idea - keep Makefile as the current build system, but delegate it to maven to compile and run tests during a build?

Will obviously need a maven installation or simply the mvn binary in the $PATH.

I tested the full release cycle on my linux machine and it did compile with maven, ran the same amount of tests as the regular Makefile version, and was able to build a working linux release from the compiled classes afterwards.

This can be extended later to do more things during the build like running checkstyle and static code analysis, building the final .jar, the final release package (at least for Linux, maybe for other OSes as well). Could then be put into a proper CI pipeline and automatically build and run tests on every PR, I think they have something for this on Github and/or on Travis CI which is free for open source projects.


yanlyub:
My fingers got itchy, especially after seeing that VASL the grandfather of Vassal already builds with maven.

This is a maven project now: maven manages all dependencies, compiles, builds the jar, puts the Vassal.jar and all dependency jars to a new directory in preparation for the Makefile's VASSAL-{linux|windows|macosx|other} goals. From there on the Makefile takes over and does the same that it used to do.

Some dependencies could not be found in any maven repository, I found the same artifacts but had to guess their version:
- smackx is in version 3.0.4 now, previous version unknown
- smackx-debug is in version 3.2.1 now, previous version unknown
- timingframework is in version 1.0 now, not 100% clear if that is the same what it was before
- org.netbeans:wizard is in version 0.998.1, it supposedly was the same version before but the code is slightly different and I had to fix a bit to keep it working

Two dependencies are new, batik-shared-resources and one other (one of the two logbacks IIRC?), they are transitive dependencies and are pulled in by the existing dependencies.

Several dependencies do not need to be explicitly declared now as they are pulled in by others as their transitives, the whole batik tree is pulled in by only four dependencies for instance.

Overall it seems to work, the Vassal.jar has the same contents that it used to have in the same structure, I can build a linux release, run it, open modules, shove counters around.

Unit tests run fine apart from the tests with images that sometimes have hiccups, possibly due to the separation of code and resources, some images are probably only there for testing and they need to be moved from src/main/resources to src/test/resources.

Next step: use a maven plugin to get the version number and bake it into the release without modifying the source code.

yanlyub:
Version number in Info.java is read at runtime from git.properties file, which is generated by a maven plugin. The result is the same version number that was previously generated by the Makefile's `git describe --tags`. The source code does not need to be changed anymore during a release.

Tests are fixed.

Dependencies smackx, timingframework and wizard have been analyzed and found compatible with previous libs.

Dependency smackx-debug was removed as it was not used.

pom.xml
```diff
@@ -0,0 +1,348 @@
...
+							<goal>enforce</goal>
+						</goals>
+						<configuration>
+							<rules>
+								<banDuplicatePomDependencyVersions />
+								<requireMavenVersion>
+									<version>3.5</version>
+								</requireMavenVersion>
+								<requireJavaVersion>
+									<version>11</version>
```
uckelman, line 228: Shouldn't this be `[11,)`? Otherwise it requires compiling with exactly Java 11.

yanlyub: Good point, the >=11 as the minimum Java version for compilation makes more sense. The 11 as the source/target level is already defined elsewhere.

Makefile
```diff
@@ -86,73 +83,31 @@ LAUNCH4J:=~/java/launch4j/launch4j
 vpath %.class $(shell find $(CLASSDIR) -type d)
 vpath %.java  $(shell find $(SRCDIR) -type d -name .svn -prune -o -print)
```
uckelman, line 84: You could remove the above five lines (`#SOURCES` to `vpath`).

uckelman:
This seems to be swallowing a _lot_ of warnings. What causes that?

Makefile
yanlyub: I have no idea what vpath does, was afraid to break anything.

uckelman: It has to do with where `make` searches for files. It's not something one normally needs, outside of unusual use cases like this one.

yanlyub:
I documented the part of the pom.xml that enables/disables the compile warnings, the maven compiler plugin does this by passing / not passing these args to the compiler:
`
<arg>-Xlint:all</arg>
<arg>-Xmaxwarns</arg> <arg>10000</arg>
`

We can enable them per default, though they clutter up the build output a bit, and the IDEs usually have their own settings for warnings, e.g. in Eclipse each type of warning can be given a custom value out of {error/warning/info}, and they appear in a convenient tree.

uckelman:
> I documented the part of the pom.xml that enables/disables the compile warnings, the maven compiler plugin does this by passing / not passing these args to the compiler:
> `<arg>-Xlint:all</arg> <arg>-Xmaxwarns</arg> <arg>10000</arg>`

Thanks.

> We can enable them per default, though they clutter up the build output a bit, and the IDEs usually have their own settings for warnings, e.g. in Eclipse each type of warning can be given a custom value out of {error/warning/info}, and they appear in a convenient tree.

Ideally we would be steadily reducing the number of warnings. I've tried to do this over the years, which is why there are zero warnings of certain types. Java 8 and later have massively increased the number the number of warnings, so the time is ripe for working on those.
 
> The issue with too many warnings is that it becomes too easy to not take them seriously.

Hiding the warnings also doesn't take them seriously. The way to take them seriously is to work on 
eliminating them.

Is there a way to get a warning count from maven?

uckelman:
@BrentEaston Would you give the maven setup a try? What do you think? This is going to affect you as an Eclipse user differently from how it will affect me.

yanlyub:
> Is there a way to get a warning count from maven?

Not for the regular compiler warnings, at least nothing that I know about, there is only the compiler args that can be passed to the compiler, and the `<showWarnings>` configuration element, which doesn't show nearly as many warnings as the `<arg>-Xlint:all</arg> <arg>-Xmaxwarns</arg> <arg>10000</arg>` args.

What is usually used in the Java world are static code analyzers like findbugs/spotbugs and pmd, sonarqube, code style rules like checkstyle, that all support a customizable ruleset, many of the rules are similar or even the same as the regular compiler warnings, they can create neat reports in txt or xml format for reporting and further processing, and severity levels can be set for every single rule so they can even break the build if there are violations of certain rules. And all of them are relatively easy to plug into maven's build chain.

> This is going to affect you as an Eclipse user differently from how it will affect me.
Ideally, taking a clean Eclipse workspace and doing a "Import Maven project" then selecting the pom.xml should get everything setup in Eclipse, but you never know with computers. I can help with that if needed, I have fought many battles against the Eclipse+Maven combination, they used to be not good friends at all but it got much better in the recent years.

One other possible advantage of the maven approach is for module development. VASL for instance already uses Maven, but they manually put the Vengine.jar and all the library jars into their code. Since Vassal is an open source project, it would be possible to upload the engine to mavencentral, then VASL and possibly other modules could simply add a dependency to their pom.xml and start developing. I think it would also be possible to give module developers a template project which is already configured to pull the Vassal dependency and create the final module package, without them having to mess with compilers, compiler settings, source code and compiled class locations etc.

uckelman:
> Not for the regular compiler warnings, at least nothing that I know about, there is only the compiler args that can be passed to the compiler,

That's too bad, as the compiler prints a warning count. It's not like maven would even need to count the warnings itself.

yanlyub:
I checked again, and counting the warnings seems to not be a use case that is often needed. There are the arguments for printing out the warnings, the usages of deprecated methods/classes, there is also a -verbose argument which prints what the compiler is doing exactly, but none of those count warnings. One option is to redirect maven's output to a file, then count the lines that contain a warning.

BrentEaston:
Hi Yan,

I have cloned your Maven branch and imported it into Eclipse as a Maven project. It took about 5 minutes downloading a bunch of files related to the libraries and at then end, the project is non-functional with many errors.

All imports from import org.w3c.dom are generating the error 'The package org.w3c.dom is accessible from more than one module: <unnamed>, java.xml/. Same error for imports from javax.xml.

This results in Buildable not compiling, which cascades and generates errors in all components that implements Buildable.

Apart from that, it seems to have built a project that will work once we get the module issues sorted.

uckelman:
I suspect that problem is caused by the inclusion of `xml-apis=1.4.01.jar` from Batik, which Maven is pulling in because the Maven repository for Batik is telling it to. I'm not sure why this didn't cause a problem for me from the command line when I tried it. The `xml-apis` JAR is one that I've always intentionally _not_ had present in `libs` because it duplicates classes which have been in the JDK for ages now. I have never been able to find any explanation of why they still distribute this JAR with Batik.

I think what needs adding is the following:
```
<exclusions>
        <exclusion>
            <groupId>xml-apis</groupId>
            <artifactId>xml-apis</artifactId>
        </exclusion>
    </exclusions>
```
That would go inside one (or both?) of the `org.apache.xmlgraphics` dependency blocks in `pom.xml`.

uckelman:
I checked the dependency tree with `mvn dependency:tree` and it says that only `batik-dom` depends on `xml-apis`, so I think you need the exclusion only in that block.

pom.xml
```diff
@@ -0,0 +1,339 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.apache.xmlgraphics</groupId>
+			<artifactId>batik-dom</artifactId>
+			<version>${version.batik}</version>
+			<exclusions>
+				<exclusion>
+					<groupId>xalan</groupId>
+					<artifactId>xalan</artifactId>
+				</exclusion>
```
uckelman, line 86: Looks like we need to add this:
```
<exclusion>
   <groupId>xml-apis</groupId>
   <artifactId>xml-apis</artifactId>
</exclusion>
```

BrentEaston: I added that to my pom.xml and that fixed all of the XML relate errors, except now I am back to the original problem of

import sun.java2d.cmm.ProfileDeferralMgr;

not being resolved.

yanlyub: I have just spent a few hours trying to get this to run on Eclipse. Found the xml-apis issue as well and this same fix for it. But I also ran into another much bigger problem which has to do with Eclipse, Maven and JPMS (Java Platform Module System).

We use compiler arguments --add-exports, only one of them seems to be really needed for the ProfileDeferralMgr. But due to unfixed https://bugs.eclipse.org/bugs/show_bug.cgi?id=543631 the Eclipse Maven plugin (M2E) is not able to pass compiler arguments to the Eclipse project configuration. These arguments can be set manually in Eclipse, but they will be reset after each refresh by M2E.

Possible solutions that I see, in order of preference:
1. Check if this `ProfileDeferralMgr.deferred = false` workaround is still needed, there were some updates to ImageIO in the last years, with some luck this workaround and all the --add-exports arguments can be removed
2. Tell Eclipse users to switch to IntelliJ, the compiler argument has to be added there but adding them once is enough and they do not get overwritten by Maven
3. Tell Eclipse users to manually add the --add-exports after every update of the project configuration by M2E (not a good solution)
4. Do not switch to Maven, write an instruction for Eclipse and IntelliJ on how to set the compiler argument once (even worse than 3.)

uckelman: I wonder why that works for me from the command line. Hmm.

uckelman: What exactly is the error message?

BrentEaston: The import sun.java2d.cmm.ProfileDeferralMgr cannot be resolved.

BrentEaston: Also, I have no great objection to moving to Intellij. I only use Eclipse because it is what I started using many years ago.

uckelman: I happen to have the Java 13 source handy; `ProfileDeferralMgr` is still just as non-thread safe as it was when I first added the workaround. If we removed the workaround, we'd start getting  `ConcurrentModificationException` reports again as soon as we did.

There might be a different way of dealing with the problem if importing `ProfileDeferralMgr` is such a stumbling block now. I'll have a look at that tomorrow.

Aside: This is one of the most aggravating bugs I've encountered in the JDK. It throws an unchecked exception that violates every expectation of how image loaders should work. It was reported in 2010, and the fix would take 30 seconds to apply---all that's needed is a `synchronized` block inside `ProfileDeferralMgr.activateProfiles()`.

yanlyub: It works from the command line all right, the problem is getting the Eclipse IDE to show the code as compileable. Eclipse uses its own compile settings, these can be set manually when maven is not in use, this is how these compiler flags need to be set in Eclipse: http://www.vassalengine.org/forum/viewtopic.php?f=5&t=11991#p61234.

But, when maven is in use, the build configuration of Eclipse is set by maven. The problem is that maven does not pass the compiler arguments, the `--add-export`s in particular, into Eclipse's build configuration. Instead it resets these manual export settings to no exports. This leads to Eclipse not knowing how to compile ImageIOImageLoader and showing these two errors for that class:
```
The import sun.java2d.cmm cannot be resolved
ProfileDeferralMgr cannot be resolved to a variable
```

I just tried it again and it reproducible, I can import the maven project, Eclipse shows these two errors, then I manually add the --add-exports in this very complicated dialog, then the errors are gone. Then I do a "Update Maven Project" (alt+f5), and my manual export settings are gone and the two errors are back.

yanlyub: It seems 3 out of 4 compiler args were not needed anymore, it is only the `sun.java2d.cmm` one that is still needed for the `ProfileDeferralMgr.deferring = false` workaround.

yanlyub: My updated solutions to the `--add-exports` Eclipse problem, in order of preference:
1. Check if this `ProfileDeferralMgr.deferring = false` workaround is still needed, there were some updates to ImageIO in the last years, with some luck this workaround and the `--add-exports` compiler argument can be removed
2. Declare the Eclipse Java IDE as not supported for Vassal development. Suggest to use another IDE or editor. Write a step by step guide on how to get started with another code editor. I can help with IntelliJ.
3. Find a library for rendering images and rewrite the rendering code to use that library instead of ImageIO, thereby removing the need for the workaround and the `--add-exports` compiler arg
4. Tell Eclipse users to manually add the --add-exports after every update of the project configuration by M2E (not a good solution)
5. Do not switch to Maven, write an instruction for Eclipse and IntelliJ on how to set the compiler argument once (even worse than solution 4.)

yanlyub: Just now read your other comments here, so removing the workaround seems to be a non solution.

Can we drop support for Eclipse then? @BrentEaston is willing to try something else, @uckelman appears to not use an IDE at all and probably uses some command-line tool like `ed` or a brain-computer-interface that changes the bits on the disk drive, @Cattlesquat already got a bit jealous of the ease of use of IntelliJ.

uckelman:
I found a way to cajole `ProfileDeferralMgr.deferring` into being set to `false` via a call in a public class:
```
ICC_Profile.getInstance(ColorSpace.CS_sRGB).getPCSType();
```
`getPCSType()` calls `ProfileDeferralMgr.activateProfiles()`, which is the sole place that `ProfileDeferralMgr.deferring` is set other than where it's declared (and is also the function where the race we're heading off occurs).

So, we can use that instead. The function I'm using isn't guaranteed to do this by the docs, but given that the docs also don't say to expect the bug and the bug is a least a decade old now, I don't feel like they have the high ground here.

If you merge in master, you should be able to remove the last remaining `--add-exports`, and then I expect the problem in Eclipse will be gone.

uckelman:
BTW, the last run of Maven I did listed the warning count:
```
[INFO] 1138 warnings
```
I'm not sure if that was always there and I didn't see it, or if something changed to make that show up.

uckelman:
@BrentEaston Does this work in Eclipse for you now?

BrentEaston:
Perfect. I took a clean clone of Yan's maven branch, then imported it as an existing Maven project and it built cleanly. I've tested it by running a debug session from my custom code project after cleaning and rebuilding it and all worked as expected.

uckelman:
Great. I'm going to merge this. Because this moves all the source files, it's going to cause conflicts on basically every other PR when I do, but there's no avoiding that and the conflicts should be easy to resolve.

yanlyub:
> Great. I'm going to merge this. Because this moves all the source files, it's going to cause conflicts on basically every other PR when I do, but there's no avoiding that and the conflicts should be easy to resolve.

Perhaps a note to all contributors is in order, they should close their existing project and do a clean reimport as a Maven project. I wrote up a guide for this, will post it on the wiki now.

uckelman:
Ok, that's weird---now I'm not getting the warning count again, and no idea what changed.

BrentEaston:
> I wrote up a guide for this, will post it on the wiki now.

Great, looking forward to that because I have totally screwed up my installation. Time to check if my backups still run :)

PR #30, opened 2020-06-22
Bug 12551 Move Trails fixes (aka "baby's first bug fix")

Cattlesquat:
* Move trails no longer constantly re-initialized from what should only be starting parameters
* Move trails can be directly forced on/off/clear without need for indirection through "toggle", etc.


src/VASSAL/counters/Footprint.java
```diff
@@ -76,6 +79,9 @@
 
   // Type Variables (Configured in Ed)
   protected NamedKeyStroke trailKey;                // Control Key to invoke
+  protected NamedKeyStroke trailKeyOn;              //BR// Control Key to force trails on
```
yanlyub, line 82: I would give this field and the other new fields private visibility instead of protected - there are no classes that inherit from this in Vassal, and if there are subclasses of this in the modules they won't need these fields right away. If we make them protected now, we expose them as public API which we may have to support for a long time to keep backwards compatibility.

If at some point in future this class is inherited from and the subclass needs access to these fields, they can still be made protected, or stay private and get public getters/setters.

Cattlesquat: I don't have major objections if that's the consensus, although it does mean that some hypothetical overriding "custom class" wouldn't have access -- I'd used the protected/private levels for other similar items in the class (trailKey that already existed was protected, so went with that for e.g. trailKeyOn). So it's a little contrary to the normal Vassal architecture though I realize there are downsides in that. (What would anybody ever want to override this for? I guess to make it so those commands could have popup menu items attached to them, which I currently haven't provided in the interests of making a relatively minimal change).   But I certainly agree it wouldn't compromise the core value of the fix. Will await further discussion.

Cattlesquat:
I thought I was deleting .classpath out of the commit here, not out of the project.

Cattlesquat:
I think I've at least stopped it from wanting to delete .classpath from the project. No idea how to get it out of the PR which is what I wanted.

yanlyub:
> I think I've at least stopped it from wanting to delete .classpath from the project. No idea how to get it out of the PR which is what I wanted.

To get a file out of a PR it needs to be reset to the state it was in before the PR.

Cattlesquat:
Okay yeah I'd have to figure out how to tell git to show me the history of an individual file (.classpath), and figure out what the reference in it is that I can then set to get that version. I could literally just delete the whole PR and re-make it in probably a 100th the time it would take to figure that out, unless you're ready to create that particular chapter of your Git how-to :-)

yanlyub:
History of a single file, commit messages only: `git log path/to/file.txt`
History of a single file, full diff per commit: `git log -p path/to/file.txt`
Find commit hash for the commit you want to go back to, then:
Reset a single file to a certain commit: `git checkout <commit hash> -- path/to/file.txt`

Cattlesquat:
Well, maybe that worked, maybe it didn't. It did *something*. Can't tell.

Cattlesquat:
Oooh, "files changed" is 1 now -- I guess it DID work -- yay, thanks!

yanlyub:
@uckelman - have a look at https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits , it seems to be possible to squash several commits of a PR so it appears as a single commit in master, this would be good for this PR in particular, otherwise the several changes to .classpath will still end up in master.

src/VASSAL/counters/Footprint.java
uckelman: My inclination is to make new fields private unless and until someone asks for them to be protected.

Cattlesquat: Fixed & pushed.

Cattlesquat:
I made the new members private, and decluttered the comments. Should be more or less ready.

src/VASSAL/counters/Footprint.java
BrentEaston: But if we add private fields, can we please, please add public or at least protected getters and setters? The 'ask for them to be protected' process just wastes developers time and our time.

uckelman: Well, hmm. As this is a module-designer-facing class, if you want them protected, go for it.

Cattlesquat: Okay, done. It's a new PR since this one is merged, but does only that.

BrentEaston: Thanks Brian. The Traits in particular are the first classes most developers want to override. There should be no private members without at least protected getters/setters in any trait. Many years back, I went and pre-emptively changed any privates I found to protected (Yes, I know Yan, should have added getters/setters, cest la vie). There may be a few more hanging round that I missed or have been added since.

PR #31, opened 2020-06-22
Bug 13117 - mouse wheel now scrolls at a decent rate in both type of …

Cattlesquat:
…help window, and also HTMLChartWindow


uckelman:
So... I have a hard time believing that hard-coding the scroll rate is the right way to do this. Isn't there some way to get an appropriate scroll rate from the system?

yanlyub:
> So... I have a hard time believing that hard-coding the scroll rate is the right way to do this. Isn't there some way to get an appropriate scroll rate from the system?

Not in a platform-independent way. The slowness of the default scroll rate seems to be a common experience, a google search for "java desktop mouse scroll speed" turns up many results like "how to speed up scroll speed", "how to increase slow scrolling speed", "scroll speed too slow", and no results saying that it is too fast.

Hardcoding it to be a bit faster without getting too deep into platform dependent settings seems to make sense.

Why is there HelpWindow and DialogHelpWindow, two different classes which are almost exact copies of each other? That's the real code smell here.

Cattlesquat:
Yeah I agree with yanlyub -- it's awkward but is really just substituting one magic number (40 pixels) for the default one (1 pixel) which is manifestly terrible. And while windows and monitors can grow bigger, the normal pixel heights of fonts themselves grows much less slowly -- so little danger this will ever be "too much" in the lifetime of Vassal 3.X

Cattlesquat:
Oh and to answer yanlyub, I think one deals with the "plain text help file" and one deals with the "html help file", or something like that. And then the 3rd thing is for the HTMLChartWindow that's a way to display an HTML file from a Toolbar button rather than the Help menu. And we're not even covering all the ways to display HTML files in Vassal here -- only the ones that have scrollbars and aren't in external browsers!

uckelman:
> Why is there HelpWindow and DialogHelpWindow, two different classes which are almost exact copies of each other? That's the real code smell here.

It was probably done that way because one inherits from a `JFrame` and the other from `JDialog` and they were written before generics existed. This could be solved neatly by inheriting from a generic type, but isn't legal in Java so far as I can see:
```
public class HTMLWindow<T extends Window> extends T {
}
```
(Note that inheritance from a template type is perfectly legal C++...)

uckelman:
Give me a few minutes, I'm knocking together something to unify these...

yanlyub:
I've seen how they extend two different classes and cannot be unified because of this, my first idea was to unify their contents e.g. make this JScrollPane and its contents into one class e.g. VassalsCoolAndFastScrollingScrollPane and use that in both classes, and also to unify the algorithm that determines the width/height/xpos/ypos and reference it from both classes.

uckelman:
Have a look at #36.

Cattlesquat:
So it looks like:
(a) We'll still need the ChartWindowHTML part of this
(b) The new class from PR 36 should have its scrollbar set to scroll faster.

Basically Java (unfortunately) defaults all scrollbars to what usually turns out to be a super-slow speed for Mousewheels.

yanlyub:
> Have a look at #36.

Looks good! Inheritance from template type or not, there is no problem that cannot be solved by adding another level of indirection.

Cattlesquat:
@yanlyub It looks like I need to just revert these two conflicting files (since the relevant code moved elsewhere), but I don't know how to kick them out of my PR?

Cattlesquat:
This one should now be all fixed up. Instead of "40", uses 3 times the pixel of the current default font size it finds in the object. Makes a nice comfortable mousewheel scroll speed.

src/main/java/VASSAL/build/widget/HtmlChart.java
```diff
@@ -129,6 +131,10 @@ public Component getComponent() {
       scroller = new ScrollPane(htmlWin);
       scroller.getViewport().setPreferredSize(htmlWin.getPreferredSize());
       scroller.getViewport().setAlignmentY(0.0F);
+      
+      final Font f = new JLabel().getFont();
```
uckelman, line 135: I wonder if there's a way to do this without creating a `JLabel` which is never displayed. I don't want to hold up the PR for that, but it's something to think about.

uckelman:
I aim to give this a try this evening, and merge it if it checks out alright.

uckelman:
Observation: We already have a preference for scroll speed for maps (which is something I only just remembered). Would it make sense to use that for _everytying_?

Cattlesquat:
>>Observation: We already have a preference for scroll speed for maps (which is something I only just remembered). Would it make sense to use that for everything?
I think an appropriate *Map* scroll speed can be very different from a "lines of text" scroll speed. I think it would be appropriate (but probably unnecessary) to have a "scrollbar scroll speed" that applied to all these different windows, but I wouldn't combine it with the Map. The main thing here was getting away from the horrible Java default of "1 pixel" or whatever it is.

PR #33, opened 2020-06-22
Refactoring Deck

yanlyub:
Slightly changed the code in the Deck class, added a first batch of unit tests for the Deck, which are mostly converted from Stack's unit tests.

Also fixed some typos in StackTest.

### DIFF
--- a/src/VASSAL/counters/Deck.java
+++ b/src/VASSAL/counters/Deck.java
@@ -264,7 +264,7 @@ public class Deck extends Stack implements PlayerRoster.SideChangeListener {
   /**
   * Update map-level count property for a piece located at index
   */
-  private void updateCounts(int index, boolean increase) {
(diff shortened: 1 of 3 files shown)


src/VASSAL/counters/Deck.java
```diff
@@ -1361,12 +1329,10 @@ private File getSaveFileName() {
     File sf = fc.getSelectedFile();
     if (sf != null) {
       String name = sf.getPath();
-      if (name != null) {
```
uckelman, line 1364: Is `File.getPath()` guaranteed to return nonnull?

yanlyub: Yes it is guaranteed, a `File`'s path can only be set when constructing and `File` has null checks in all it's constructors and will throw an NPE if it encounters a null path.

PR #34, opened 2020-06-23
13129 - HTML Chatter w/ security & compatibility improvements

Cattlesquat:
I figured this would make faster progress if I
(a) actually did the rest of the work myself now that I'm "git-compatible" and
(b) addressed the issues Joel raised in the forum thread

See bug 13129 where I have uploaded a test module.

Notes:
* Chat messages have HTML tags auto-disabled. So players can't grief with them and CAN type < in the text without issues.
* "Standard" report messages have HTML processing disabled if the "compatibility" preference on the Chat pref tab is checked (it is checked by default). So old modules will function normally.


src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
 import VASSAL.configure.FontConfigurer;
+import VASSAL.configure.BooleanConfigurer;
 import VASSAL.i18n.Resources;
 import VASSAL.preferences.Prefs;
 import VASSAL.tools.ErrorDialog;
 import VASSAL.tools.KeyStrokeSource;
 import VASSAL.tools.ScrollPane;
-import VASSAL.tools.swing.SwingUtils;
+
+//BR// 23-June-2020 - significant Chatter update to provide full HTML support, including multiple colors/fonts and images, <b>bold</b> and <i>italic</i>
```
yanlyub, line 58: This is handled by the source code management / by git in this case, it keeps track of who changed what and when, and if the commit message contains a bug ID then it's also possible to look it up later on and find out why something was changed.

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
+  protected Color systemMsg, myChat, otherChat;
+  
+  protected boolean htmlCompatibility; //BR// Disable HTML parsing unless opted in, to maintain backwards compatibility
+	
   public static String getAnonymousUserName() {
     return Resources.getString("Chat.anonymous"); //$NON-NLS-1$
   }
 
-  public Chatter() {
+  public Chatter() {
```
yanlyub, line 97: Trailing whitespace, and also a piece of code that shows up in the diff even though nothing was changed here

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
+		add(input);
+	}
 
-  /** @deprecated use GlobalOptions.getPlayerId() */
-  @Deprecated public String getHandle() {
-    return GlobalOptions.getInstance().getPlayerId();
-  }
+	private String formatChat(String text) {
+    final String id = GlobalOptions.getInstance().getPlayerId();
+    return "&lt;" + (id.length() == 0 ? "(" + getAnonymousUserName() + ")" : id) + "&gt; - " + text; //$NON-NLS-1$ //$NON-NLS-2$ //BR//HTML-friendly angle brackets
```
yanlyub, line 149: `String.isEmpty()` is a slightly better readable way to express `String.length() == 0`.

Also, String.format() is a bit more readable in these cases, e.g.:
`
return String.format("&lt;%s&gt; - %s", id.length() == 0 ? "(" + getAnonymousUserName() + ")" : id, text);
`

or even:
`
final String idPart = id.isEmpty() ? "(" + getAnonymousUserName() + ")" : id;
return String.format("&lt;%s&gt; - %s", idPart, text);
`

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
-  @Override
+	public JTextField getInputField() {
+    return input;
+	}
+	
+	//BR// To make the player colors easy to override in a custom class 
+	//BR// (my modules have logic to assign individual player colors -- beyond the scope of the present effort but a perhaps a fun future addition)
+	protected String getChatStyle(String s)
+	{	  
+    if (s.startsWith(formatChat("").trim())) { //$NON-NLS-1$
```
yanlyub, line 160: Indentation looks a bit wild around this part, also there seems to be some trailing whitespace.

Cattlesquat: Do you have a good tool suggestion to clean up all the #$%^ tabs/whitespace. It has been an ongoing disaster :)

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
+	/**
+	 * Display a message in the text area
+	 */
+	public void show(String s) {
+		String style;
+		boolean html_allowed;
+
+		//BR// Choose an appropriate style to display this message in
+		s = s.trim();
+		if (s.length() > 0) {
```
yanlyub, line 176: A `if (!s.isEmpty())` might be better readable.

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
+				s = s.replaceFirst(tag, replace); // Swap in our new URL-laden tag for the old one.
+			} else {
+				break; // BR// If something went wrong in matching up the tag, don't loop forever
+			}
+		}
+		
+		//BR// Now we have to fix up any legacy angle brackets around the word <observer>
+    keystring = Resources.getString("PlayerRoster.observer");
+    replace   = keystring.replace("<","&lt;");
+    replace   = replace.replace(">","&gt;");
```
yanlyub, line 264: Can chain the calls here: `replace = keystring.replace(a, b).replace(c, d);`

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
+	//BR// Adds or updates a CSS stylesheet entry. Styles in the color, font type, and font size.
+	private void addStyle(String s, Font f, Color c, String font_weight, int size) {
+		if ((style == null) || (c == null)) return;
+		style.addRule(s + " {color:" + String.format("#%02x%02x%02x", c.getRed(), c.getGreen(), c.getBlue())
+				        + "; font-family:" + f.getFamily() + "; font-size:" + (size > 0 ? size : f.getSize()) + "; " + ((font_weight != "") ? "font-weight:" + font_weight + "; " : "") + "}");
+	}
+
+	//BR// Build ourselves a CSS stylesheet from our preference font/color settings.
+	private void makeStyleSheet(Font f) {
+		if (style != null) {
```
yanlyub, line 302: Could be changed to an "early return":
```
if (style == null) {
  return;
}

// proceed with non-null code without having to indent
```

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
+
+		globalPrefs.addOption(Resources.getString("Chatter.chat_window"), systemMsgColor);
+
+		systemMsg = (Color) globalPrefs.getValue(SYS_MSG_COLOR);
+
+
+		//
+		// my message color
+		//
+		final ColorConfigurer myChatColor = new ColorConfigurer( MY_CHAT_COLOR,
```
yanlyub, line 502: Can name the variable `myMessageColor` and remove the comment above

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
+		globalPrefs.addOption( Resources.getString("Chatter.chat_window"),
+		myChatColor );
+		  
+		myChat = (Color) globalPrefs.getValue(MY_CHAT_COLOR);		 
+
+		//
+		// other message color
+		//
+		
+		final ColorConfigurer otherChatColor = new ColorConfigurer( OTHER_CHAT_COLOR,
```
yanlyub, line 522: Can name the variable `otherMessageColor` and remove the comment above

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,477 +21,636 @@
...
+		} else {
+			return null;
+		}
+	}
+
+	/**
+	 * Displays the message, Also logs and sends to the server a {@link Command}
+	 * that displays this message
+	 */
+	public void send(String msg) {
```
yanlyub, line 578: No change in this method except indentation, probably better to revert it to the original so it won't show up in the diff.

Or change the condition in the next line to the slightly better readable `if (!StringUtils.isEmpty(msg))`.

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed (hopefully)

src/VASSAL/build/module/Chatter.java
Cattlesquat: Ran through a whitespace prettifier I had lying around. Hopefully better on whitespace and tab crud all around.

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed

src/VASSAL/build/module/Chatter.java
yanlyub: Code style settings in the IDE should be able to take care of that, in Eclipse under Window > Preferences > Java > Code Style > Formatter, once they are properly set they can be applied to the highlighted region or to a whole file with IIRC `ctrl+shift+f`. Took me a couple PRs to adjust my local settings to the projects standards, I'm hoping to get to a point where we have a standard code style that the developers can simply import into their IDE,  #26 is a step in that direction.

src/VASSAL/build/module/Chatter.java
```diff
@@ -21,92 +21,109 @@
...
+    doc = (HTMLDocument) conversation.getDocument();
+    kit = (HTMLEditorKit) conversation.getEditorKit();
+
+    style = kit.getStyleSheet();
+    myFont = new Font("SansSerif", Font.PLAIN, 12); //BR// Will be overridden by the font from Chat preferences
+
+    for (int i = 0; i < 15; ++i) {
+      try {
+        kit.insertHTML(doc, doc.getLength(), "<br>", 0, 0, null);
+      } catch (BadLocationException ble) {
```
yanlyub, line 111: Vassal's default code style is, else, catch, finally etc on a new line, i.e.

```
if (cond) {
  // foo
}
else {
}

try {
  // try
}
catch (Exception e) {
}
finally {
}

File: src/VASSAL/build/module/Chatter.java [hunk 12]
```diff
@@ -115,55 +132,214 @@ public void actionPerformed(ActionEvent e) {
...
+    } else {
+      style = "msg";
+      html_allowed = false;
+    }
+
+    //BR// Disable unwanted HTML tags in contexts where it shouldn't be allowed:
+    //BR// (1) Anything received from chat channel, for security reasons
+    //BR// (2) Legacy module "report" text when not explicitly opted in w/ first character or preference setting 
+    if (!html_allowed) {
+      s = s.replaceAll("<", "&lt;"); //BR// This prevents any unwanted tag from functioning
```
yanlyub, line 224: Can chain these and format like this, with the comments:
```
s = s.replaceAll(a, b) // comment 1
     .replaceAll(c, d) // comment 2
```

src/VASSAL/build/module/Chatter.java
```diff
@@ -115,55 +132,214 @@ public void actionPerformed(ActionEvent e) {
...
-  @Deprecated public String getHandle() {
+  @Deprecated
+  public String getHandle() {
     return GlobalOptions.getInstance().getPlayerId();
   }
 
+  //BR// Adds or updates a CSS stylesheet entry. Styles in the color, font type, and font size.
+  private void addStyle(String s, Font f, Color c, String font_weight, int size) {
+    if ((style == null) || (c == null)) return;
+    style.addRule(s + " {color:" + String.format("#%02x%02x%02x", c.getRed(), c.getGreen(), c.getBlue())
```
yanlyub, line 293: Looks a bit wild, some strategically placed newlines would make this easier readable, e.g.
```
s
+ " color:" + x
+ "; font-family:" + y
+ "; font-size:" + z
(etc)
```

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

src/VASSAL/build/module/Chatter.java
Cattlesquat: fixed for next push

Cattlesquat:
(1) Cleaned up a few "tab" and code-style things to satisfy checkstyle
(2) Now I know where i18n stuff goes, so fixed my strings to work with that.

uckelman:
You missed your chance to make your commit message `NOW I HAVE I18N HO HO HO`.

src/main/java/VASSAL/build/module/Chatter.java
```diff
@@ -115,55 +134,245 @@ public void actionPerformed(ActionEvent e) {
...
+  // To make the player colors easy to override in a custom class 
+  // (my modules have logic to assign individual player colors -- beyond the scope of the present effort but a perhaps a fun future addition)
+  protected String getChatStyle(String s) {   
+    if (s.startsWith(formatChat("").trim())) { //$NON-NLS-1$
+      return "mychat";      
+    } 
+    else {
+      return "other";
+    }   
+  }
```
uckelman, line 170: I'd write this as a ternary: `return s.startsWith(formatChat("").trim()) ? "mychat" : "other";`

src/main/java/VASSAL/build/module/map/KeyBufferer.java
```diff
@@ -100,8 +109,12 @@ public void mousePressed(MouseEvent e) {
     // Don't clear the buffer until we find the clicked-on piece
     // Because selecting a piece affects its visibility
     EventFilter filter = null;
+    BandSelectType bandSelect = BandSelectType.NONE;
     if (p != null) {
-      filter = (EventFilter) p.getProperty(Properties.SELECT_EVENT_FILTER);
+      filter = (EventFilter) p.getProperty(Properties.SELECT_EVENT_FILTER);      
+      if (!e.isPopupTrigger() && Boolean.TRUE.equals(p.getProperty(Properties.NON_MOVABLE))) {
```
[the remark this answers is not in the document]
Cattlesquat: Um, does this comment belong in this PR? I think this is just from when I re-based the PR on master. 

The answer to your question is: because if it's a right-click then it's not a band-select and therefore we proceed with normal clicking-on-things-and-putting-up-the-popup-menu.

src/main/java/VASSAL/build/module/Chatter.java
Cattlesquat: Fixed, checked in.

src/main/java/VASSAL/build/module/map/KeyBufferer.java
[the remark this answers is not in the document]
uckelman: I'm not sure why these lines show up as a change on _this_ PR. That said, I think this is wrong---`isPopupTrigger()` checks for something _waves hands_ which is not quite the same as a right click, and it got my attention because nothing in this file raises a context menu. I think we might want to check that the mouse button isn't the left here instead.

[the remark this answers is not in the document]
Cattlesquat: Oh right isPopupTrigger() means is THIS event going to open the context menu, and runs into the whole issue of some platforms do the trigger on "Pressed" and others do it on "Released" and when you're in the wrong method for your platform it doesn't tell you a damn thing. So this might produce hinky results on Linux which I *think* is the opposite whateverness from Windows. I will make a new PR and improve it.

Meanwhile no idea why THIS PR. The commands I issued to rebase this PR on master were:
git rebase --onto master HTMLChatter
git pull
git push -u origin HTMLChatter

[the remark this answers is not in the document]
Cattlesquat: I imagine that means when it comes time to merge this you'll want to do one of those Squash-N-Crunch kinds of merge.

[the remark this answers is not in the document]
uckelman: IIRC, the weird one for popup triggers is the Mac?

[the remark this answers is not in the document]
Cattlesquat: Upon reading up on it (because why not, "mouse is my thing"), it turns out that Windows is "one way" and Mac & Linux (both descendants of Unix) are "the other way".

uckelman:
Ok, I'm finally trying this out now. Your original description, from way back in April:

>  1. Report messages can use <b></b> and <i></i> tags to bold and italicize text.
>  2. Images from the VMOD images directory can likewise be dropped into the text with &lt;img src=…> tags (notice the dice rolling lines in the sample below! One of my favorite parts!)
> 3. It currently supports three extra game-message colors for Reports, activated by starting the message with a !, ?, or ~ symbol (which is trimmed out of the message)
> 4. If desired text can be additionally styled by dropping in things like e.g. <span style=”font-family:arial; font-size:20px;”>

I'm not having success with triggering any of the new behavior. (Yes, I did uncheck the "old behavior" box in the prefs.) What do I need to do?

Cattlesquat:
Ah! 
1. Go to bug 13129 (oops forgot I already had an issue open) and download the ChatterTestModule.vmod.
2. Run that module under this PR.
3. Every time you drag one of those pieces to a number-slot, it should generate a whole bunch of lines of text. 
4. They should all be different colors and have bold/italic and stuff
5. But the *first black line only* is the "Compatibility mode line", which tests "regular old-fashioned report messages". So IF you've got the compatibility box checked, notice that there's no bold or italic in that line, i.e. the HTML tags don't process. Or if you uncheck the box, note that that line produces different results. 
6. Meanwhile, also type some lines in Chat and notice they are colored.
7. Attempt to put HTML tags in your chat and notice that they are rejected
8. Merge the PR! :D

Cattlesquat:
Happy to also walk you through the "details of that". 

The Chatter also allows e.g. dice images to be displayed in report lines (e.g. for dice rolls). That's just through  tags in the Report messages. I didn't demonstrate that in the test module but happy to show you.

Cattlesquat:
If you want to try making your own "Report" trait messages then the theory is:
(1) Just putting plain old text in there goes with the "#1" text color. IF compatibility box not checked, you can also put html tabs in it e.g. <b></b> <i></i>  the latter of which will automatically find things in the images subdirectory.
(2) If the very first character on the line is | then it's an "opt-in" line of #1 color. All HTML tags will work regardless of checkbox.
(3) If very first character is "!" then it's color #2 and opted into HTML tags regardless of compatibility check box.
(4) First character "?" then color #3 
(5) First character ~ then color #4 
(6) First character ` then color #5

Cattlesquat:
Oh and of course in the "regular", not-opted in, report text (so with the checkbox check and none of the special first characters), then you can safely type things like "Is the first number <= the second number?" and it won't read the < as a tag.

Cattlesquat:
And yes in Report text, you can put < span.... > tags with special styling e.g. "invent your own 6th color! display in a different font!" stuff like that.

uckelman:
I've uploaded tests builds so people can try this change:

http://www.vassalengine.org/~uckelman/tmp/

They're named "VASSAL-3.3.1-265-g2e4bd1ee-pr34".

Cattlesquat:
Installed the windows one this morning, got:
image

I notice that happened to your Windows 32-bit test-case guy too (forum message). 

Anyway once we get that figured out I'll recruit some folks in the Module Design forum to give it some try-outs.

uckelman:
I messed up the symlinks when I added adding a 32-bit Windows JDK to my builds setup, which is why neither this build nor the 32-bit Windows test worked.

I've uploaded a new one, which worked in my Windows VM when I tried it. Download it again and try that.

Cattlesquat:
Okay, yeah, THAT one works. I will proceed with putting up a (better) test module for people to play with this and test with, and post something in (probably) Module Design forum.

I'll assume you will make a new version for the 32-bit Windows people and inform them.

uckelman:
> I'll assume you will make a new version for the 32-bit Windows people and inform them.

Already have.

Cattlesquat:
Happily, this (Chatter) is now getting some testing and positive reviews in forum thread: http://www.vassalengine.org/forum/viewtopic.php?f=6&t=12092

The suggestion of a "reset colors to defaults" button is an interesting one.

Cattlesquat:
@yanlyub Hey look "a Travis thing happened" and I fixed it all by myself like a Big Boy! :D

yanlyub:
Have a look at https://github.com/yanlyub/vassal/tree/HTMLChatter-fix and the one commit in it https://github.com/yanlyub/vassal/commit/b93a2a41642663809ba37094b4e690c851589488, if you like what you see, do a `git remote add yanlyub https://github.com/yanlyub/vassal.git` then check out that branch, push it to your repo and make a new PR from it.

Cattlesquat:
Closed in favor of #112

PR #39, opened 2020-06-24
Bug 13036 - can no longer accidentally band-select and drag entire contents of a deck

Cattlesquat:


uckelman:
I'm reading through this one presently. It amazes me that no one reported the problem previously.

src/main/java/VASSAL/build/module/map/KeyBufferer.java
```diff
@@ -243,6 +247,11 @@ else if (mapsel.contains(s.getPosition())) {
     // Does Not Stack units deselect normally once selected
     @Override
     public Object visitDefault(GamePiece p) {
+      Stack s = p.getParent();
+      if ((s != null) && (s instanceof Deck)) { // Clear any deck *members* out of the KeyBuffer. (yes, members of decks can be does-not-stack)
```
uckelman, line 251: You don't need any of the parentheses here save for the outermost ones.

Cattlesquat:
Actually I saw a bunch of "reports" on it, but it would always guys in the module design thread saying things like "well because I can't trust Vassal not to drag my whole deck onto the map" -- basically I don't think anybody ever "drilled down" to figure out what it took to "repro" the bug reliably. Because it didn't *look* like you could band select a deck--you'd band select it and then nothing would "look" selected, and if you then tried to drag on the deck you'd only get one card. Successful repro required you to band select the deck AND a counter, and then drag the counter. So people were having that happen "occasionally" but then never could figure out what had caused it to happen. I finally decided okay I bet I can figure out how to repro this, and once I repro'ed it the fix was obviously not hard.

PR #41, opened 2020-06-25
Added Maven Wrapper

yanlyub:
This adds the Maven Wrapper, described in https://github.com/takari/maven-wrapper

Key info:

Normally you instruct users to install a specific version of Apache Maven, put it on the PATH and then run the mvn command like the following:
``
mvn clean install
``

But now, with a Maven Wrapper setup, you can instruct users to run wrapper scripts:
``
./mvnw clean install
``

or on Windows
``
mvnw.cmd clean install
``


uckelman:
I tried this. It seems to work for me. What steps do people using IDEs need to take for this to work for them?

yanlyub:
> I tried this. It seems to work for me. What steps do people using IDEs need to take for this to work for them?

This just makes it easier to work on the command line, they won't need a separate installation of maven anymore, and no need having maven in their $PATH.

The IDEs will keep working with the `pom.xml` and use whatever maven is configured in the IDE.

PR #43, opened 2020-06-26
"Band Select" Improvements - immobile pieces don't "eat" band selects, and band-select gets Does-Not-Stack filter

Cattlesquat:
This PR (which is intentionally branched from the Bug 13036 don't-accidentally-drag-the-whole-deck PR) improves two issues:

(1) Presently if you have pieces marked "cannot be moved", then they will "eat" band-selects (attempting abortively to turn them into drags-across-the-map I suppose, even though of course an unmoveable piece cannot be dragged across the map). So anyway now a band-select can be started on top of an unmoveable piece (but this does not interfere with click-selecting said unmoveable piece, etc)
(2) Immobilized ("Does Not Stack") now includes an "Allow Band Select" combo box, which allows "normally", "Alt-Key", and "never" settings for whether pieces are picked up by the band-select "lasso". This allows (for example) immobile pieces to be excluded from group-selects without preventing them from being selected by a click (e.g. to activate a right-click menu). Modules which feature lots of immobile buttons/controls will benefit from having these filtered out of group select operations. Likewise module designers can choose to filter certain "less important pieces" out of band selects (I would immediately use this in Paths of Glory for e.g. trenches & control markers)

I wrote all the mouse logic for Rise of Nations (I mean who *else* is going to do it but the Lead Designer?) so I'm pretty confident that this is all "sound". I marked one spot for suggestion on a better approach than "heuristic magic number".


src/main/java/VASSAL/build/module/map/KeyBufferer.java
```diff
@@ -169,9 +191,21 @@ public void mouseReleased(MouseEvent evt) {
     PieceVisitorDispatcher d = createDragSelector(
       !evt.isControlDown(), evt.isAltDown(), map.componentToMap(selection)
     );
+    
+    //BR// If it was a legit band-select drag (not just a click), our special case only applies if piece is allowed to be band-selected
+    if (bandSelectPiece != null) {
+      Point finish = map.mapToComponent(evt.getPoint());
+      EventFilter bandFilter = (EventFilter) bandSelectPiece.getProperty(Properties.BAND_SELECT_EVENT_FILTER);
+      if ((bandFilter != null) && !(evt.isAltDown() && (bandFilter instanceof Immobilized.UseAlt))) {
```
yanlyub, line 199: For slightly better readability, this could be written like this (also, `Point finish` can be declared inside the next if):
```
    if (bandSelectPiece != null) {
      final EventFilter bandFilter = (EventFilter) bandSelectPiece.getProperty(Properties.BAND_SELECT_EVENT_FILTER);
      final boolean pieceAllowedToBeBandSelected = (bandFilter != null) && !(evt.isAltDown() && (bandFilter instanceof Immobilized.UseAlt));

      if (pieceAllowedToBeBandSelected) {
        final Point finish = map.mapToComponent(evt.getPoint());
        
        //BR// Open to suggestions about a better way to distinguish "click" from "lasso" (not that Vassal doesn't already suck immensely at click-vs-drag threshhold). FWIW, this "works".
        final boolean isLasso = finish.distance(anchor) >= 10;
        if (isLasso) { 
          bandSelectPiece = null;
        }
      }
    }
```

src/main/java/VASSAL/build/module/map/KeyBufferer.java
```diff
@@ -66,6 +68,11 @@
   protected Point anchor;
   protected Color color = Color.black;
   protected int thickness = 3;
+  protected GamePiece bandSelectPiece = null;
+  
+  private static final int NOT_BAND_SELECT     = 0;
```
yanlyub, line 73: Safer to use enums for this, else someone some day will get the idea to assign `bandSelect = 27`

And it's easy, define the enum at the bottom of this class:
```
  private enum BandSelectType {
    NOT,
    NORMAL,
    SPECIAL;
  }
```

Then use it just like you would use the int:
```
BandSelectType bandSelect = BandSelectType.NOT;
(...)
bandSelect = BandSelectType.SPECIAL;
(...)
if (bandSelect != BandSelectType.NOT) {
```

src/main/java/VASSAL/counters/Immobilized.java
```diff
@@ -361,7 +417,7 @@ else if (p.moveIfSelected) {
       b.add(movementOption);
       controls.add(b);
       controls.add(ignoreGridBox);
-    }
+}
```
yanlyub, line 420: This should be indented by 4 spaces

src/main/java/VASSAL/build/module/map/KeyBufferer.java
Cattlesquat: Great - added

src/main/java/VASSAL/build/module/map/KeyBufferer.java
Cattlesquat: added, thanks

src/main/java/VASSAL/counters/Immobilized.java
Cattlesquat: Whoopsie!

Cattlesquat:
@yanlyub could you have a look - I feel like I resolved the conflicts but git says they're still there. It looks like now the only "conflicts" are my actual changes but it won't let me mark them as resolved.

Cattlesquat:
@yanlyub (okay maybe I figured this one out, apparently I had to go play with the little editor in the conflict-resolver and tell it each of the things had been resolved)

uckelman:
Suggest a module for testing this.

Cattlesquat:
Sure, you want me to make you one real quick? Or just point you in the direction of one where you could then change the does-not-stack parameters in something and see it work?

src/main/java/VASSAL/build/module/map/KeyBufferer.java
```diff
@@ -147,8 +160,19 @@ else if (!s.isExpanded()) {
...
+      bandSelect = BandSelectType.NORMAL; //BR// Allowed to band-select  
+    }
+    
+    if (bandSelect != BandSelectType.NONE) {
+      bandSelectPiece = null;
       if (!e.isShiftDown() && !e.isControlDown()) { // No deselect if shift key down
         kbuf.clear();
+        
+        //BR// This section allows band-select to be attempted from non-moving pieces w/o preventing click-to-select from working 
+        if ((bandSelect == BandSelectType.SPECIAL) && (p != null) && !ignoreEvent) {
```
uckelman, line 172: Boolean operators have lower precedence than the comparison operators, so you don't need the extra parentheses here.

Cattlesquat: It's an old habit of mine. I personally find it more readable but am happy to "conform" if it's bothering you.

uckelman: > Sure, you want me to make you one real quick? Or just point you in the direction of one where you could then change the does-not-stack parameters in something and see it work?

No need to make one if there's one I can just download (or might already have handy).

Cattlesquat: If you pull down Paths of Glory 9.7, there are invisible pieces over every empty space. (See London or Berlin an example, many more).
1. First note that you can now start a band-select inside the boundaries of one of these without having the whole mouse stroke be "eaten" (you can try the same thing unpatched to see what I mean).
2. Then also notice that you nonetheless CAN band select these empty spaces (but why would you want to)
3. So then you can edit and go to the prototype "Empty Space" which is conveniently like the 3rd or 4th down the list in the prototypes. So you can edit the Does Not Stack for this, and use the new band select thing to make these things NOT band-selectable.
4. Now run the module and note that not only do the band selects not get "eaten" by the immobile piece being there, but that now these unwanted immobile pieces are filtered out of your resulting band selection (only other pieces remain in it).
5. And then if you want you can try the "Alt Key Only" version.

src/main/java/VASSAL/build/module/map/KeyBufferer.java
```diff
@@ -100,8 +109,12 @@ public void mousePressed(MouseEvent e) {
     // Don't clear the buffer until we find the clicked-on piece
     // Because selecting a piece affects its visibility
     EventFilter filter = null;
+    BandSelectType bandSelect = BandSelectType.NONE;
     if (p != null) {
-      filter = (EventFilter) p.getProperty(Properties.SELECT_EVENT_FILTER);
+      filter = (EventFilter) p.getProperty(Properties.SELECT_EVENT_FILTER);      
+      if (!e.isPopupTrigger() && (boolean)p.getProperty(Properties.NON_MOVABLE)) {
```
uckelman, line 115: I got an NPE on this. `getProperty()` returns `Object`s. In this case, the object you get will be a `Boolean`, which will be unboxed to a `boolean`... unless you get a `null`, which throws an NPE when you try casting it to a non-object type. The thing you need to do here is `Boolean.TRUE.equals(p.getProperty(Properties.NON_MOVABLE))`.

Cattlesquat: Ah! Thx. Fixed & checked in.

src/main/java/VASSAL/build/module/map/KeyBufferer.java
uckelman: Huh, I think I prefer the original map

Cattlesquat: Haha :) Good news, it's still there. One of my big projects in "rescuing" 9.0 was putting back the Classic Map and getting it working right too.

uckelman: This looks like it works. Thanks.

uckelman:
One last thing: Make an issue for these changes so it's got a number for the change log.

src/main/java/VASSAL/build/module/map/KeyBufferer.java
uckelman: I got PoG almost immediately after it was first published... The "Delux" map just looks wrong to me.

Cattlesquat:
Okay, made an issue (13149) and renamed this PR to include the number.

src/main/java/VASSAL/build/module/map/KeyBufferer.java
Cattlesquat: Yeah it was a major priority for the community to get both maps back working so that you could still play either the classic scenario or the "tournament" historical scenario still on the original map. I actually like both of the maps, but I didn't approve of only supporting one of them.

PR #44, opened 2020-06-26
Code cleanup, fixed compiler warnings

yanlyub:


yanlyub:
> Does adding generic parameters to raw types of non-private members or functions maintain binary compatibility?

It depends, if the signature stays the same after type erasure then the binary compatibility is preserved. Here is a good description: www.angelikalanger.com/GenericsFAQ/FAQSections/ProgrammingIdioms.html#FAQ104

I will recheck the changes in this PR as I may have changed some of the signatures accidentally.

src/main/java/VASSAL/build/module/gamepieceimage/SwatchComboBox.java
```diff
@@ -71,10 +71,10 @@ public SwatchRenderer() {
      * value and returns the label, set up to display the text and image.
      */
     @Override
-    public Component getListCellRendererComponent(JList list, Object value, int index, boolean isSelected,
+    public Component getListCellRendererComponent(JList<? extends String> list, String value, int index, boolean isSelected,
```
yanlyub, line 74: The wonders of type erasure, after this change this is how the compiler writes the bytecode:
```
    public Component getListCellRendererComponent(JList list, Object value, int index, boolean isSelected, boolean cellHasFocus) {
      ColorSwatch swatch = ColorManager.getColorManager().getColorSwatch((String)value);
```
Looks exactly like the source code before the change and should be binary compatible.

yanlyub:
> > Does adding generic parameters to raw types of non-private members or functions maintain binary compatibility?
> 
> It depends, if the signature stays the same after type erasure then the binary compatibility is preserved. Here is a good description: [www.angelikalanger.com/GenericsFAQ/FAQSections/ProgrammingIdioms.html#FAQ104](http://www.angelikalanger.com/GenericsFAQ/FAQSections/ProgrammingIdioms.html#FAQ104)
> 
> I will recheck the changes in this PR as I may have changed some of the signatures accidentally.

Double-checked all changes related to generics, after type erasure the compiled bytecode looks exactly like the source code before the changes. Should be binary compatible.

uckelman:
> > Does adding generic parameters to raw types of non-private members or functions maintain binary compatibility?
> 
> It depends, if the signature stays the same after type erasure then the binary compatibility is preserved. Here is a good description: [www.angelikalanger.com/GenericsFAQ/FAQSections/ProgrammingIdioms.html#FAQ104](http://www.angelikalanger.com/GenericsFAQ/FAQSections/ProgrammingIdioms.html#FAQ104)

Yep, I remember that article from when it was _new_.

PR #45, opened 2020-06-27
Bug 13137 - Drag Threshold - fixes bug where drag threshold was effectively 1 pixel. Also allows user to adjust w/ preference.

Cattlesquat:
Map's mouse listeners were translating the coordinates of mouse events from component coordinates to map coordinates before passing them on, but doing so in a way that actually overwrote the original event that awt DragGestureRecognizer was still using, resulting in an apples-to-oranges coordinate comparison that effectively reduced the dragThreshold to 1 at any zoom other than 100%. This is now fixed, and a user preference added to adjust the drag threshold (because 5 is still actually pretty damn tight)

### DIFF
--- a/src/main/java/VASSAL/build/GameModule.java
+++ b/src/main/java/VASSAL/build/GameModule.java
@@ -186,7 +186,7 @@ public abstract class GameModule extends AbstractConfigurable implements Command
   protected GpIdSupport gpidSupport = null;
   protected Long crc = null;
   
-  private static String oldDragThreshold; //BR//13137
+  private static String oldDragThreshold; //
 
   /**
    * @return the top-level frame of the controls window
@@ -882,7 +882,7 @@ public abstract class GameModule extends AbstractConfigurable implements Command
       theModule.checkGpIds();
     }
     
-    //BR//13137 - Save our old drag threshold
+    //Save our old drag threshold
     oldDragThreshold = System.getProperty("awt.dnd.drag.threshold");
     System.setProperty("awt.dnd.drag.threshold", Integer.toString(GlobalOptions.getDragThreshold()));
 
--- a/src/main/java/VASSAL/build/GameModule.java
+++ b/src/main/java/VASSAL/build/GameModule.java
@@ -185,6 +185,8 @@ public abstract class GameModule extends AbstractConfigurable implements Command
    */
   protected GpIdSupport gpidSupport = null;
   protected Long crc = null;
+  
+  private static String oldDragThreshold; //BR//13137
 
   /**
    * @return the top-level frame of the controls window
@@ -871,7 +873,7 @@ public abstract class GameModule extends AbstractConfigurable implements Command
         throw e;
       }
     }
-
+    
     /*
      *  If we are editing, check for duplicate, illegal or missing GamePiece Id's
      *  and update if necessary.
@@ -879,6 +881,10 @@ public abstract class GameModule extends AbstractConfigurable implements Command
     if (theModule.getDataArchive() instanceof ArchiveWriter) {
       theModule.checkGpIds();
     }
+    
+    //BR//13137 - Save our old drag threshold
+    oldDragThreshold = System.getProperty("awt.dnd.drag.threshold");
+    System.setProperty("awt.dnd.drag.threshold", Integer.toString(GlobalOptions.getDragThreshold()));
 
     /*
      * Tell any Plugin components that the build is complete so that they
@@ -893,6 +899,14 @@ public abstract class GameModule extends AbstractConfigurable implements Command
    * Unload the module
    */
   public static void unload() {
+    
+    //BR// Put our old drag threshold back, or if it wasn't set then return it to an unset state.
+    if (oldDragThreshold != null) {
+      System.setProperty("awt.dnd.drag.threshold", oldDragThreshold);      
+    } else {
+      System.setProperty("awt.dnd.drag.threshold", "");            
+    }
+    
     if (theModule != null) {
       if (theModule.shutDown()) {
         theModule = null;
--- a/src/main/java/VASSAL/build/GameModule.java
+++ b/src/main/java/VASSAL/build/GameModule.java
@@ -900,11 +900,11 @@ public abstract class GameModule extends AbstractConfigurable implements Command
    */
   public static void unload() {
     
-    //BR// Put our old drag threshold back, or if it wasn't set then return it to an unset state.
+    // Put our old drag threshold back, or if it wasn't set then return it to an unset state.
     if (oldDragThreshold != null) {
       System.setProperty("awt.dnd.drag.threshold", oldDragThreshold);      
     } else {
-      System.setProperty("awt.dnd.drag.threshold", "");            
+      System.clearProperty("awt.dnd.drag.threshold");            
     }
     
     if (theModule != null) {
--- a/src/main/java/VASSAL/build/GameModule.java
+++ b/src/main/java/VASSAL/build/GameModule.java
@@ -884,7 +884,7 @@ public abstract class GameModule extends AbstractConfigurable implements Command
     
     //Save our old drag threshold
     oldDragThreshold = System.getProperty("awt.dnd.drag.threshold");
-    System.setProperty("awt.dnd.drag.threshold", Integer.toString(GlobalOptions.getDragThreshold()));
+    System.setProperty("awt.dnd.drag.threshold", Integer.toString(GlobalOptions.getInstance().getDragThreshold()));
 
     /*
      * Tell any Plugin components that the build is complete so that they
--- a/src/main/java/VASSAL/build/module/GlobalOptions.java
+++ b/src/main/java/VASSAL/build/module/GlobalOptions.java
@@ -94,7 +94,7 @@ public class GlobalOptions extends AbstractConfigurable {
   private String autoReport = ALWAYS;
   private String markMoved = NEVER;
   
-  private static int dragThreshold = 10;
+  private int dragThreshold = 10;
 
   private Map<String,Object> properties = new HashMap<>();
   private static Map<String,Configurer> optionConfigurers = new LinkedHashMap<>();
@@ -477,10 +477,10 @@ public class GlobalOptions extends AbstractConfigurable {
     return isEnabled(markMoved, MARK_MOVED);
   }
   
-  public static int getDragThreshold() {
+  public int getDragThreshold() {
     return dragThreshold;
   }
-
+  
   public String getPlayerId() {
     playerIdFormat.setProperty(PLAYER_NAME, (String) GameModule.getGameModule().getPrefs().getValue(GameModule.REAL_NAME));
     playerIdFormat.setProperty(PLAYER_SIDE, PlayerRoster.getMyLocalizedSide());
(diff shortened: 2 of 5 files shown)


src/main/java/VASSAL/build/GameModule.java
```diff
@@ -893,6 +899,14 @@ public static void init(GameModule module) throws IOException {
...
    */
   public static void unload() {
+    
+    //BR// Put our old drag threshold back, or if it wasn't set then return it to an unset state.
+    if (oldDragThreshold != null) {
+      System.setProperty("awt.dnd.drag.threshold", oldDragThreshold);      
+    } else {
+      System.setProperty("awt.dnd.drag.threshold", "");            
+    }
+
```
uckelman, line 909: I think setting to `null` is what you want to unset the property, so you could do this unconditionally:

```System.setProperty("awt.dnd.drag.threshold", oldDragThreshold);```

File: src/main/java/VASSAL/build/module/Map.java [hunk 2]
```diff
@@ -1232,15 +1232,17 @@ public void mouseExited(MouseEvent e) {
...
+      mouseListenerStack.get(mouseListenerStack.size()-1).mouseClicked(mapEvent);
     }
     else if (multicaster != null) {
-      final Point p = componentToMap(e.getPoint());
-      e.translatePoint(p.x - e.getX(), p.y - e.getY());
-      multicaster.mouseClicked(e);
+      final Point p = componentToMap(mapEvent.getPoint());
+      mapEvent.translatePoint(p.x - mapEvent.getX(), p.y - mapEvent.getY());
+      multicaster.mouseClicked(mapEvent);
     }
```
uckelman, line 1246: You could eliminate quite a bit of repeated code here and below by adding a method which copies the event, does the translation, and returns the copied event.

Then you could do this:
```
public void mouseClicked(MouseEvent e) {
    if (!mouseListenerStack.isEmpty()) {
      mouseListenerStack.get(mouseListenerStack.size()-1).mouseClicked(translateEvent(e));
    }
    else if (multicaster != null) {
      multicaster.mouseClicked(translateEvent(e));
    }
  }
```
and a similar thing in `mousePressed()` and `mouseReleased()`.

src/main/java/VASSAL/build/module/GlobalOptions.java
```diff
@@ -181,6 +184,21 @@ public void propertyChange(PropertyChangeEvent e) {
...
+    );
+    dragThresholdConf.addPropertyChangeListener(new PropertyChangeListener() {
+      @Override
+      public void propertyChange(PropertyChangeEvent e) {
+        dragThreshold = dragThresholdConf.getIntValue(10);
+        System.setProperty("awt.dnd.drag.threshold", Integer.toString(dragThreshold));
+      }
+    });
+    prefs.addOption(dragThresholdConf);
+
```
uckelman, line 201: I'm a bit dubious about adding a preference for this. I can't come up with another program where the drag threshold is settable. How does everything else manage to get a sane value for a drag threshold without user intervention?

src/main/java/VASSAL/build/GameModule.java
```diff
@@ -185,6 +185,8 @@ public void propertyChange(PropertyChangeEvent evt) {
    */
   protected GpIdSupport gpidSupport = null;
   protected Long crc = null;
+  
+  private static String oldDragThreshold; //BR//13137
```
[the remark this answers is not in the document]
Cattlesquat: Haha no problem. I was modelling on stuff I'd seen in the code but that of course isn't necessarily a good reason.

src/main/java/VASSAL/build/GameModule.java
Cattlesquat: That was my first instinct, but was giving me null pointer errors.

src/main/java/VASSAL/build/module/GlobalOptions.java
Cattlesquat: Judgment call I guess. Of course we're a program that, at heart, lets our users design interfaces for other users, so that's a truly more "meta" situation than many programs. My ultimate profession has always been interface designer, so on the one hand no interface is ever good enough (I can begin to imagine the misery of going through the world with perfect pitch), but on the other hand no tool given to an interface designer ever goes to waste. (In a way, I would never have become involved in Vassal if the interface of the Paths of Glory module my friend talked me into playing last November didn't make me want to stab myself in the eye -- I spent the first two weeks telling myself "I am NOT fucking FIXING this. I am NOT fucking FIXING this." and then, you know, I decided to fix just this one thing, etc, etc, and the rest is history). So I have lots of professional history to say of course you want to give interface designers micro-control of interfaces, but in the end it's all an art form it's not like I've collected data that show it's better to have preferences. I just personally want them so I thought I'd, you know, be the change I want in the world :-)

src/main/java/VASSAL/build/GameModule.java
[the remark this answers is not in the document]
Cattlesquat: Cleaned that up.

src/main/java/VASSAL/build/module/Map.java
Cattlesquat: Fixed that up. Yan had mentioned that, but in a context where he was going to go do that in a second pass after the PR. But anyway now you've mentioned it and I've done it.

src/main/java/VASSAL/build/GameModule.java
Cattlesquat: And the documentation on awt.dnd.drag.threshold says that as long as it's not a positive integer it's TREATED as unset, so that's what pointed me toward just setting it empty string once using null didn't work.

Cattlesquat: Oh, found it: https://stackoverflow.com/questions/4323995/set-system-propery-to-null-in-java
Will do that instead.

Cattlesquat:
Fix checked in to clear the system property in the proper way. Hopefully you let me keep the preference :)

src/main/java/VASSAL/build/module/GlobalOptions.java
Cattlesquat: An example use-case would be a card game (e.g. Innovation) where you want the player to have to drag the card a more substantial percentage of the card's size (than 10 pixels) before it "engages".

src/main/java/VASSAL/build/GameModule.java
uckelman: Wow, that's an irritating API. I specifically checked that setting the value to `null` was ok before commenting, and apparently I missed it.

Cattlesquat: IKR?! Without stack overflow coming up I'd never have found that. So completely "not like the normal way to do things".

src/main/java/VASSAL/build/module/GlobalOptions.java
```diff
@@ -455,6 +476,10 @@ public boolean centerOnOpponentsMove() {
   public boolean isMarkMoveEnabled() {
     return isEnabled(markMoved, MARK_MOVED);
   }
+  
+  public static int getDragThreshold() {
```
uckelman, line 480: What's your reason for making this static? (The other members which are set from stored values have getters which aren't.)

src/main/java/VASSAL/build/module/GlobalOptions.java
```diff
@@ -92,6 +93,8 @@
   private String centerOnMoves = ALWAYS;
   private String autoReport = ALWAYS;
   private String markMoved = NEVER;
+  
+  private static int dragThreshold = 10;
```
uckelman, line 97: See static comment below.

uckelman:
This seems to work. I tried an outrageously large number (50) and I had to drag REAAAAAAALY far (probably 50 pixels!) before a drag started.

src/main/java/VASSAL/build/module/GlobalOptions.java
uckelman: Or above. In whichever direction the other comment is when GH shows it to you.

uckelman:
Observation: I tried dragging things (mainly images) in other applications (mainly Firefox) and it seems like the drag threshold there is 5-7, as near as I can tell.

uckelman:
Additional observation: Is this really something where users might want a different drag threshold per module? (I'm wondering if this and some other things, like scroll speed, might be more appropriately VASSAL preferences rather than module preferences...)

src/main/java/VASSAL/build/module/GlobalOptions.java
Cattlesquat: Oh right. I had to try to change it back to remember why -- it's because the place I need access to it in GameModule is static.

Cattlesquat:
I feel like it's more likely "per module" because what I'm thinking of is the different uses cases of "game with mainly huge-ass cards" versus "game with mainly tiny counters".

Cattlesquat:
>>Observation: I tried dragging things (mainly images) in other applications (mainly Firefox) and it seems like the drag threshold there is 5-7, as near as I can tell.

I'm fine with anything in the 5-10 range as our default. Because there are folks out there who I know feel "traumatized" by the 1-pixel threshold I chose to err on the high side -- you know, let them "definitely feel the effect". One of those customer service tricks... But if you felt really strong about 5 or something then sure.

uckelman:
I don't feel strongly about the default, was just noting that 10 is not far off what's used by something I could easily check.

Cattlesquat:
>>This seems to work. I tried an outrageously large number (50) and I had to drag REAAAAAAALY far (probably 50 pixels!) before a drag started.
Haha yes a feeling of AMAZING POWER! :D  I did test the crap out of this before checking it in. Because it sort of touches the entire universe of course.

Explained the static thing "above" just in case you didn't see it in all the notifs.

src/main/java/VASSAL/build/module/GlobalOptions.java
uckelman: Can't you make this non-static and get at it via `GlobalOptions.getInstance()`?

Cattlesquat: Uh, why apparently yes, he says sheepishly.  (A pattern that seems common here in java that I don't remember seeing much in my good ole C++ days)

Done, checked in.

uckelman: Singletons were common in terrible crappy old Java.

src/main/java/VASSAL/build/module/GlobalOptions.java
```diff
@@ -181,6 +184,21 @@ public void propertyChange(PropertyChangeEvent e) {
     classicMfd.addPropertyChangeListener( (evt) -> setUseClassicMoveFixedDistance(classicMfd.getValueBoolean()));
     prefs.addOption(classicMfd);
 
+    //BR// Drag Threshold
+    final IntConfigurer dragThresholdConf = new IntConfigurer(
+      DRAG_THRESHOLD,
+      Resources.getString("Mouse Drag Threshold"),  //$NON-NLS-1$
```
uckelman, line 190: Last thing, I think: This shouldn't be the literal string you want to appear in the prefs dialog, but instead the I18N key for that string. That is, you should have `"GlobalOptions.mouse_drag_threshold"` here, and then put the key-value pair for that into `src/main/resources/VASSAL/i18n/VASSAL.properties`.

Cattlesquat: So THAT'S where those things go! Done & checked in.

PR #46, opened 2020-06-27
Let's break the build - do not merge

yanlyub:
Had to try this.


uckelman:
Looks like it works.

yanlyub:
> Looks like it works.

Have to merge PR #47 though else it doesn't.

uckelman:
Looks like this worked, so that means we can close it without merging?

PR #49, opened 2020-06-28
Set git clone depth to 3 in travis build configuration

yanlyub:
A depth of 1 would probably be enough since we don't do anything with the git history during the build, but let's go with 3, just in case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ cache:
     - $HOME/.m2
 
 git:
-  depth: 3
+  depth: 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,6 @@ jdk: openjdk11
 cache:
   directories:
     - $HOME/.m2
+
+git:
+  depth: 3


uckelman:
What does this do?

yanlyub:
> What does this do?

Travis does a `git clone --depth=50` by default e.g. [here](https://travis-ci.com/github/vassalengine/vassal/builds/173285080#L165), which is too much for our "simple" build as we don't do anything with the git history during the build.

This reduces it to a depth of 3 which should be enough.

Here is the Travis build configuration manual, for further info: https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth

uckelman:
Oh, it's the clone depth. Is there any reason not to set it to 1?

yanlyub:
> Oh, it's the clone depth. Is there any reason not to set it to 1?

Looks good with depth 1: https://travis-ci.com/github/vassalengine/vassal/builds/173386972#L165

PR #52, opened 2020-06-30
Correct mistranslations in i18n VASSAL_nl.properties

jimu:
- Replace translation of WizardSupport.**SelectPlayMode**
  from ~~"Kies een spel module" ("Choose a game module")~~
  with "Kies speel modus" ("Select Play Mode")

- Correct capitalization of WizardSupport.WizardSupport.SelectPlayMode
  from ~~"kies speel modus"~~ to "Kies speel modus"


src/main/java/VASSAL/build/module/WizardSupport.java
```diff
@@ -252,6 +250,21 @@ public InitialWelcomeSteps createInitialWelcomeSteps() {
...
+   * 
+   * Test's whether GameModule.REAL_NAME is non-empty and not "newbie"
+   * 
+	 * @return <code>true</code> if user supplied a real name
+   */    
+  private boolean isRealName()
+  {
+	  final String name = (String)GameModule.getGameModule().getPrefs().getValue(GameModule.REAL_NAME);
+	  return name != null && !name.isBlank() && !name.equals(Resources.getString("Prefs.newbie"));
+  }
```
uckelman, line 265: Are you sure we want the `isBlank()` check? That's new.

jimu: I believe it's safe because:

1. The Oracle documentation describes [String isBlank()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#isBlank()) compatibility since v11

2. The VASSAL windows installer installs JRE 14:

```
"C:\Program Files\VASSAL-3.3.1\jre\bin\java" --version
openjdk 14.0.1 2020-04-14
OpenJDK Runtime Environment (build 14.0.1+7)
OpenJDK 64-Bit Server VM (build 14.0.1+7, mixed mode)
```

3. Maven enforces JavaVersion 11 in pom.xml:

```
<requireJavaVersion>
   <version>[11,)</version>

```

uckelman: That's not what I was asking about. I'm sure the function exists to be called.

What I'm asking about is whether introducing `isBlank()` is correct. It's not something we were checking for previously.

jimu: I misunderstood. In the future, I'll make sure to emphasize any changes in functionality in my commit message.

Currently users are told they need to enter their *real name* to proceed through the welcome/setup wizard. The `[finish]` and `[next]` buttons remain disabled until they do.

If they are clever or typo, they can enter a space for their name.   This seems like a bug or undocumented feature at best.  Users aren't allowed to be named "newbie" but they are allowed to be named " "?

I reasoned that `" "` and should be disallowed for the same reason `""` and `null` are disallowed.

uckelman: @BrentEaston Do you think this change will cause any problems?

BrentEaston: I think that is a reasonable change and I don't see any problems.

uckelman:
Merging. Thanks for the corrections.

Noted as Bug 13155.

PR #53, opened 2020-06-30
Maven multi module

yanlyub:
The next step towards publishing the Vassal jar on a public repository, for use by module developers.

The submodule "vasssal-app" contains the application itself, the result of building it is the Vassal jar that can be published on mavencentral and used as a dependency by module projects.

The submodule "release-prepare" copies the Vassal jar and its dependencies into a common directory for further processing by the makefile.

The two modules are aggregated under an umbrella pom.xml with the artifactId "vassal-parent".


yanlyub:
The CI proved its worth here, the original PR built on my machine but broke on Travis, this helped me find an error in the maven files and fix it.

yanlyub:
Closing this PR due to conflicts with PR #26, rebasing its branch on master and reopening as a new PR.

PR #58, opened 2020-07-03
Code formatting

yanlyub:
The first batch of code formatting commits aimed at reducing checkstyle warnings.


src/main/java/VASSAL/tools/image/GeneralFilter.java
```diff
@@ -106,7 +106,9 @@ public float apply(float t) {
     }
 
     @Override
-    public float getSamplingRadius() { return 1.0f; }
+    public float getSamplingRadius() {
```
uckelman, line 109: Very short function bodies should be able to be single-line.

src/main/java/VASSAL/tools/image/GeneralFilter.java
```diff
@@ -539,10 +556,10 @@ private static void apply_h(
...
         // Ugly, but fast.
         work[k] =
-         (s_a > 255 ? 255 : s_a < 0 ? 0 : (int)(s_a+0.5f)) << 24 |
-         (s_r > 255 ? 255 : s_r < 0 ? 0 : (int)(s_r+0.5f)) << 16 |
-         (s_g > 255 ? 255 : s_g < 0 ? 0 : (int)(s_g+0.5f)) <<  8 |
-         (s_b > 255 ? 255 : s_b < 0 ? 0 : (int)(s_b+0.5f));
+          (s_a > 255 ? 255 : s_a < 0 ? 0 : (int)(s_a+0.5f)) << 24 |
+            (s_r > 255 ? 255 : s_r < 0 ? 0 : (int)(s_r+0.5f)) << 16 |
+            (s_g > 255 ? 255 : s_g < 0 ? 0 : (int)(s_g+0.5f)) <<  8 |
+            (s_b > 255 ? 255 : s_b < 0 ? 0 : (int)(s_b+0.5f));
```
uckelman, line 562: The lines are misaligned with one another now.

src/main/java/VASSAL/chat/ui/RoomTreeRenderer.java
```diff
@@ -60,8 +60,8 @@ public Component getTreeCellRendererComponent(
     boolean expanded,
     boolean leaf,
     int row,
-    boolean hasFocus)
-  {
+    boolean hasFocus) {
```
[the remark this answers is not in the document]
yanlyub: This can't be enforced with checkstyle, it is either "LeftCurly on new line" or "LeftCurly on EOL".

Exceptions like a method definition with a long arguments list should generally not exist, otherwise this would be valid, and we would be back at C-style for an arbitrary amount of method definitions:
```
public void method(int arg1,
  int arg2)
{
```

The whole LeftCurly check could be disabled for method definitions but that would defeat its purpose.

src/main/java/VASSAL/launch/LaunchRequest.java
```diff
@@ -47,14 +47,46 @@
...
   enum Mode {
-    MANAGE    { public String toString() { return "manage"; } },
-    LOAD      { public String toString() { return "load";   } },
-    EDIT      { public String toString() { return "edit";   } },
-    IMPORT    { public String toString() { return "import"; } },
-    NEW       { public String toString() { return "new";    } },
-    EDIT_EXT  { public String toString() { return "edit-extension"; } },
-    NEW_EXT   { public String toString() { return "new-extension";  } },
-    TRANSLATE { public String toString() { return "translate"; } }
+    MANAGE    {
```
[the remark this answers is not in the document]
yanlyub: Far easier would be to define the enum this the right way.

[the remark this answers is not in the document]
uckelman: Which would be what?

src/main/java/VASSAL/tools/image/GeneralFilter.java
yanlyub: The general rule is to avoid having several statements on a single line. Adding exceptions to formatting rules makes the rules too complex, in this particular case it adds extra work to further edits of the code (what if I want to add a log statement during debugging?), makes setting breakpoints ambigious (is the breakpoint on the whole method or on the return statement?). Also, a good IDE will display the expanded version in a single line.

src/main/java/VASSAL/launch/LaunchRequest.java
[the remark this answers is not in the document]
yanlyub: Commit is coming up.

yanlyub:
To move things forward, and until we decide whether we want the left curly on the end of line or on a new line, I have changed the severity of the LeftCurly rule from warning to info.

yanlyub:
The first set of warnings are cleared with this PR, and the first rules are upgraded from warnings to errors.

Is this ok to be merged now or are there further change requests?

src/main/java/VASSAL/build/module/Inventory.java
```diff
@@ -1377,11 +1377,12 @@ protected String toSortKey() {
        * @return true if arguments looks processable, false else
        */
       protected boolean argsOK(Object arg0, Object arg1) {
-        return (arg0 != null &&
-                arg1 != null &&
-                arg0 instanceof CounterNode &&
-                arg1 instanceof CounterNode);
-        }
+        return (
```
uckelman, line 1380: This shouldn't have parentheses around it at all...

src/main/java/VASSAL/build/module/gamepieceimage/NewColorConfigurer.java
```diff
@@ -103,16 +103,15 @@ public void propertyChange(PropertyChangeEvent e) {
...
-                (java.awt.event.ActionEvent e) {
-              setValue(JColorChooser.showDialog
-                       (null, getName(), colorValue()));
-              csc.setValue(new ColorSwatch("", (Color) getValue())); //$NON-NLS-1$
-            }
-          });
+      b.addActionListener(new java.awt.event.ActionListener() {
+        @Override
+        public void actionPerformed
+          (java.awt.event.ActionEvent e) {
```
uckelman, line 109: This should be on the previous line.

src/main/java/VASSAL/counters/RestrictCommands.java
```diff
@@ -259,10 +259,11 @@ public String getState() {
     @Override
     public String getType() {
       SequenceEncoder se = new SequenceEncoder(';');
-      se.append(name.getValueString())
-      .append((actionOption.getSelectedIndex()==0) ? HIDE : DISABLE)
-      .append(propertyMatch.getValueString())
-      .append(watchKeys.getValueString());
+      se
```
uckelman, line 262: Why is this on it's own line now?

src/main/java/VASSAL/build/module/Inventory.java
yanlyub: Good point, and given enough time to analyze my IDE suggests some more changes in this area, I will add them to this PR.

src/main/java/VASSAL/build/module/properties/NumericPropertyPrompt.java
```diff
@@ -43,7 +43,8 @@ public String getNewValue(String oldValue) {
     String s = null;
     do {
       s = (String) JOptionPane.showInputDialog(dialogParent, promptText, null, JOptionPane.QUESTION_MESSAGE, null, null, oldValue);
-    } while (s != null && !isValidValue(s));
+    }
+    while (s != null && !isValidValue(s));
```
uckelman, line 47: Let's keep the `while` on the same line as the closing brace.

src/main/java/VASSAL/counters/RestrictCommands.java
yanlyub: Because it's only by chance that the length of the variable's name is equal to the indentation size, 2 characters. If the variable name changes, the indentation will stay properly indented.

src/main/java/VASSAL/build/module/properties/NumericPropertyPrompt.java
yanlyub: I reverted this, we will also activate the rule "RightCurly on same line" for "while" later on and catch the other occurrences if there are any.

yanlyub:
Caught the wrong rule but fixed it with the last commit, RightCurlyAlone is not enforced now for do-while loops. Enforcing RightCurly on same line for do-while is not active but we can consider activating it after this batch is done.

Also, was good to see in 745fd89 that the build breaks after only a couple seconds due to wrong formatting.

src/main/java/VASSAL/tools/version/VassalVersionTokenizer.java
```diff
@@ -150,27 +153,25 @@ public int next() throws VersionFormatException {
         break;
 
       case BUILD:
-        {
```
uckelman, line 153: This scope shouldn't be removed. It's there to prevent the local variables from being in the same scope as the bodies of other cases.

yanlyub: This sub-scope is not necessary. Accessing the local variables from another case body is a compile error. Declaring new variables with the same name is also a compile error.

Cattlesquat:
I will admit I'm sort of hoping some of my 3.3.2/3.3.3 bugfix PR's can be merged before an all-spanning code format checkin makes every line a conflict...

yanlyub:
> I will admit I'm sort of hoping some of my 3.3.2/3.3.3 bugfix PR's can be merged before an all-spanning code format checkin makes every line a conflict...

I'm ok with this being merged last, as long as this gets merged at all. I will resolve all upcoming conflicts.

src/main/java/VASSAL/tools/version/VassalVersionTokenizer.java
uckelman: Yes, the point of that scope was to prevent those exact compile errors...

src/main/java/VASSAL/tools/ipc/IPCMessageReceiver.java
```diff
@@ -38,7 +38,8 @@ public void run() {
 //System.err.println("received " + msg);
 
         lsup.notify(msg);
-      } while (!(msg instanceof Fin));
+      }
+      while (!(msg instanceof Fin));
```
uckelman, line 42: I thought you said you were reverting this change?

yanlyub: Missed that one, will revert it.

yanlyub:
Is this good to go or should I change anything else?

uckelman:
> Is this good to go or should I change anything else?

I'm not coming back to this until the PRs older than it are sorted, as otherwise it's going to cause them a load of conflicts.

PR #60, opened 2020-07-03
Bug 13146 - Fix class search to account for custom classes.

Cattlesquat:
This appears to fix the problem, from my tests.

### DIFF
--- a/src/main/java/VASSAL/configure/PropertiesWindow.java
+++ b/src/main/java/VASSAL/configure/PropertiesWindow.java
@@ -65,7 +65,7 @@ public class PropertiesWindow extends JDialog {
       if (Node.ELEMENT_NODE == child.getNodeType()) {
         // Cull Buildables from the state.
         try {
-          final Class<?> c = GameModule.getGameModule().getDataArchive().loadClass(((Element)child).getTagName(), false);
+          final Class<?> c = GameModule.getGameModule().getDataArchive().loadClass(((Element)child).getTagName());
           if (Buildable.class.isAssignableFrom(c)) {
             originalState.removeChild(child);
           }
--- a/src/main/java/VASSAL/configure/PropertiesWindow.java
+++ b/src/main/java/VASSAL/configure/PropertiesWindow.java
@@ -37,6 +37,7 @@ import org.w3c.dom.Node;
 import VASSAL.build.Buildable;
 import VASSAL.build.Builder;
 import VASSAL.build.Configurable;
+import VASSAL.build.GameModule;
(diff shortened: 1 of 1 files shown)


src/main/java/VASSAL/configure/PropertiesWindow.java
```diff
@@ -64,7 +65,7 @@ protected void initialize(final Configurable target, HelpWindow helpWindow) {
       if (Node.ELEMENT_NODE == child.getNodeType()) {
         // Cull Buildables from the state.
         try {
-          final Class<?> c = Class.forName(((Element)child).getTagName());
+          final Class<?> c = GameModule.getGameModule().getDataArchive().loadClass(((Element)child).getTagName(), false);
```
uckelman, line 68: Is there a reason we want `loadClass(String, boolean)` here instead of just `loadClass(String)`?

uckelman: (I suspect the answer is "no".)

Cattlesquat: I'm pretty sure the answer is actually yes. loadClass(String) calls straight through into the bowels of SecureClassLoader and its parents. Whereas loadClass(String, boolean) is the version that is overridden in DataArchive... and is also where it has a specific flow designed to deal with checking for classes-we-actually-already-loaded (e.g. custom classes). 

And we definitely want the "resolve = false" end of the boolean, because all we're really doing here in this loop is "checking to see if the class legitimately exists" -- because what we're calling in with is the name of a supposed child class of some main class that we're editing, and we're deciding whether to cull its node(s) from the internal XML tree.

yanlyub: Looks like loadClass(String) does simply call loadClass(String, boolean), which would call our overridden method. Still, I find calling the (String, boolean) version better, it shows explicitly what is called, if someone would jump into the method he would end up in our custom method right away and not have to navigate through internal Java methods. If for some reason the loadClass(String) would change the implementation and not call our method at all, we will not even notice that.

uckelman: My reason for wanting to avoid the boolean is that it's noise. Anyone modifying this code in the future will wonder what that `boolean` does, whether it should really be `false`, why this particular use of `loadClass` is not the same overload as the other ones in VASSAL... We could save future readers of the code all that trouble by using the overload without the `boolean`.

Cattlesquat: Fixed

PR #63, opened 2020-07-04
ZipArchive - replace File.renameTo() with FileUtils.forceDelete() and FileUtils.moveFile()

yanlyub:


src/main/java/VASSAL/tools/io/ZipArchive.java
```diff
@@ -453,7 +453,9 @@ private void writeToDisk() throws IOException {
     // Replace old archive with temp archive.
     if (!tmpFile.renameTo(archiveFile)) {
       try {
-        FileUtils.forceDelete(archiveFile);
+        if (archiveFile.exists()) {
```
uckelman, line 456: It's unexpected if `archiveFile` doesn't exist at this point, since we only just copied entries from it a few lines above.

yanlyub: Unexpected but not impossible, the documentation of File.renameTo() basically says "anything could happen", and we have a report with a stacktrace showing this exact situation - FileUtils.forceDelete() failing with a FileNotFoundException.

uckelman: I should state it more strongly, then: If `archiveFile` doesn't exist at this point, it's an error, and it's wrong to go on.

yanlyub: Is it necessarily an error? If it didn't exist when we read entries from it, there would have been exceptions before this part. If the file doesn't exist after the File.renameTo() it can only be due to renameTo() having failed but also having deleted the target file in the process. In that case the file is already deleted and we don't need to delete it again, just try the subsequent move.

uckelman: `File.renameTo()` failing but also deleting the target isn't the only way that the target could fail to exist when we call `FileUtils.forceDelete()`: It's also possible that something else deleted the target file between when we last read from it and when we tried to move the new file over it.

I had forgotten how bad `File.renameTo()` is. `File.renameTo()` is one those lovely functions where there's essentially no aspect of its behavior on which you can depend. (Python has loads of these as well, they're great!)

What about `Files.move()`? That didn't exist until Java 7. The documentation for that isn't as noncommittal about what it does. Perhaps we could do this

```Files.move(tmpFile.toPath(), archiveFile.toPath(), StandardCopyOption.REPLACE_EXISTING);```

instead? I have no sense of whether that would be more reliable.

I misunderstood the behavior of `FileUtils.forceDelete()` when the file doesn't exist. In the context of removing a file "force" usually means "do it, and don't complain if the file isn't  there" (cf. `rm -f foo`, where it's not an error if there is no `foo`). What I should have written to start with is this:
```
if (!tmpFile.renameTo(archiveFile)) {
  try {
    try {
      FileUtils.forceDelete(archiveFile);
    }
    catch (FileNotFoundException e) {
      // this is fine, the file's gone
    }
    FileUtils.moveFile(tmpFile, archiveFile);
  }
  catch (IOException e) {
    ...
  }
}
```

Comment (uckelman) [reply src/main/java/VASSAL/tools/io/ZipArchive.java:456]: Aside: If we could replace uses of `FileUtils` with `Files`, we wouldn't be far from being able to drop commons-io as a dependency.

Comment (yanlyub) [reply src/main/java/VASSAL/tools/io/ZipArchive.java:456]: I understand `File.renameTo()` simply calls the rename function/utility of the underlying filesystem, and am not surprised that the bug report we got is from Windows.

I would prefer my version to the one with try-catch, "using exceptions in ordinary situations" was one of the code smells in Bloch's Effective Java (http://www.thefinestartist.com/effective-java/57) and this `renameTo()` fail almost looks like an ordinary thing on Windows.

We could also reduce this multi-step renaming to a single call to `FileUtils.moveFile()`, it already does what we do here, tries `File.renameTo()` first and if that fails tries something better. Or we can try `Files.move()`, that looks good too.

Comment (uckelman) [reply src/main/java/VASSAL/tools/io/ZipArchive.java:456]: I see that `FileUtils.moveFile()` does do roughly what we do:

https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/FileUtils.java#L2011

It would take more digging to find out what exactly the `FileSystemProvider` used by `Files.move()` does on Windows.

Comment (uckelman) [reply src/main/java/VASSAL/tools/io/ZipArchive.java:456]: I think we should replace the whole thing with one call to `FileUtils.moveFile()`, given that it already does what we're doing.

Comment (yanlyub) [reply src/main/java/VASSAL/tools/io/ZipArchive.java:456]: Like this?
```
// Replace old archive with temp archive.
try {
  FileUtils.forceDelete(archiveFile);
  FileUtils.moveFile(tmpFile, archiveFile);
}
catch (...
```

uckelman: Yeah, I think so. That shouldn't behave any differently from what we have now, except maybe in the unusual case where `File.renameTo()` fails but it then becomes possible to remove the destination and move the source to it immediately after that... which is weird, but seemingly anything could happen on Windows with file handles.

PR #64, opened 2020-07-04
Try using the native look and feel on Windows once again

uckelman:


BrentEaston:
Tried this out, it's not nearly as nice as the default LaF.

BrentEaston:
Can we close this one? The windows native LaF is still truly awful compared to the one we use.

PR #71, opened 2020-07-08
13149 better right mouse detection in KeyBufferer

Cattlesquat:
Per our discussion in the other thread. This was Stack Overflow's top-voted answer for detecting pure right-button

### DIFF
--- a/src/main/java/VASSAL/build/module/map/KeyBufferer.java
+++ b/src/main/java/VASSAL/build/module/map/KeyBufferer.java
@@ -113,7 +113,7 @@ public class KeyBufferer extends MouseAdapter implements Buildable, MouseMotionL
     BandSelectType bandSelect = BandSelectType.NONE;
     if (p != null) {
       filter = (EventFilter) p.getProperty(Properties.SELECT_EVENT_FILTER);      
-      if (!SwingUtilities.isRightMouseButton(e) && Boolean.TRUE.equals(p.getProperty(Properties.NON_MOVABLE))) {  // Stack Overflow's top-rated way to detect right-button-down ;-) 
+      if (SwingUtilities.isLeftMouseButton(e) && Boolean.TRUE.equals(p.getProperty(Properties.NON_MOVABLE))) {   
         bandSelect = BandSelectType.SPECIAL; //BR// Don't "eat" band-selects if unit found is non-movable
       }
     }
--- a/src/main/java/VASSAL/build/module/map/KeyBufferer.java
+++ b/src/main/java/VASSAL/build/module/map/KeyBufferer.java
@@ -113,7 +113,7 @@ public class KeyBufferer extends MouseAdapter implements Buildable, MouseMotionL
     BandSelectType bandSelect = BandSelectType.NONE;
     if (p != null) {
       filter = (EventFilter) p.getProperty(Properties.SELECT_EVENT_FILTER);      
-      if (SwingUtilities.isLeftMouseButton(e) && Boolean.TRUE.equals(p.getProperty(Properties.NON_MOVABLE))) {   
+      if (!e.isPopupTrigger() && SwingUtilities.isLeftMouseButton(e) && Boolean.TRUE.equals(p.getProperty(Properties.NON_MOVABLE))) {   
         bandSelect = BandSelectType.SPECIAL; //BR// Don't "eat" band-selects if unit found is non-movable
       }
     }
--- a/src/main/java/VASSAL/build/module/map/KeyBufferer.java
+++ b/src/main/java/VASSAL/build/module/map/KeyBufferer.java
@@ -29,6 +29,7 @@ import java.awt.event.MouseEvent;
 import java.awt.event.MouseMotionListener;
 
 import javax.swing.JComponent;
+import javax.swing.SwingUtilities;
 
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
@@ -112,7 +113,7 @@ public class KeyBufferer extends MouseAdapter implements Buildable, MouseMotionL
     BandSelectType bandSelect = BandSelectType.NONE;
     if (p != null) {
       filter = (EventFilter) p.getProperty(Properties.SELECT_EVENT_FILTER);      
-      if (!e.isPopupTrigger() && Boolean.TRUE.equals(p.getProperty(Properties.NON_MOVABLE))) {
+      if (!SwingUtilities.isRightMouseButton(e) && Boolean.TRUE.equals(p.getProperty(Properties.NON_MOVABLE))) {  // Stack Overflow's top-rated way to detect right-button-down ;-) 
         bandSelect = BandSelectType.SPECIAL; //BR// Don't "eat" band-selects if unit found is non-movable
       }
     }


Cattlesquat:
https://stackoverflow.com/questions/4525733/java-mouse-event-right-click?rq=1

uckelman:
Checking for the right mouse button isn't correct here---shouldn't the test be that it's _not_ the left? There's also the middle button, and possibly several others if you have a fancy mouse.

uckelman:
Hold fire on this---I already have some code for dealing with detecting the right mouse button in the test build I'm waiting for any Mac touchpad user to try. I'll round that up this evening.

Cattlesquat:
Gotcha. The swing utilities have isLeftButton() and isMiddleButton() etc too, so it can certainly be worked how we want it. Think you're right since we're really doing band selects here that's ONLY with the left mouse button. 

I'll correct the line in the PR so it doesn't gnaw at me all day, but will await your test user etc.

Cattlesquat:
Okay further reading/thinking and I think I've got it here. I actually need to check BOTH in this situation. Inside the "if" block is what supports the "new elegant band-select flow", but higher priority is making sure that 100% of users can get a context menu (right click menu) on a piece. So:
* We check isPopupTrigger() on the admittedly extremely unlikely chance that somebody is using a one-button mouse where context menus get triggered by "left" mouse button plus a key-combo. Basically if this mouse event, after I return it unconsumed, will/might/could trigger a menu then I don't want to unselect the unit just because it is immobile.
* We check isLeftMouseButton() to filter out "fancy buttons" like middle, etc, that aren't part of doing drag-a-rubber-band-around-things functions.
* And on the off chance there's "something we didn't think of", like if an accessibility function flips the left & right mouse buttons but doesn't invert the SwingUtilities flags, then we end up getting the "safe flow" and the user doesn't get the elegant-new-band-select but retains 3.3.1 and prior level functionality i.e. they can at least get their context menus to pop.

uckelman:
Duelling SO answers: https://stackoverflow.com/questions/2972512/how-to-detect-right-click-event-for-mac-os

The first answer says `event.getButton() == MouseButton.BUTTON3` is wrong for Macs, because a Ctrl-click is _also_ a "right" mouse click. The second answer suggests `SwingUtilities.isRightMouseButton(event)`. The third answer points out that `SwingUtilities.isRightMouseButton(event)` incorrectly reports Cmd-click as a right click and incorrectly reports Ctrl-click as a left click. (Typical desktop Java dumpster fire...)

So... this is what I have in `VASSAL.tools.swing.SwingUtils` in the test build I'm waiting for anyone with a Mac laptop to try:
```
public static boolean isRightMouseButton(MouseEvent e) {
  if (e.getButton() == MouseEvent.BUTTON3) {
     return true;
  }

  if (!SystemUtils.IS_OS_MAC_OSX) {
    return false;
  }

  return e.getButton() == MouseEvent.BUTTON1 &&
    (e.getModifiersEx() & MouseEvent.CTRL_DOWN_MASK) != 0;
}
```
Basically, Button 3 is a right click no matter where you are, and on a Mac Ctrl + Button 1 is also right click.

Cattlesquat:
What context(s) are you calling your new isRightMouseButton in?

uckelman:
Look at the diff here: https://github.com/vassalengine/vassal/compare/master...uckelman:mac_right_mouse_button

uckelman:
Aside: This also makes me think that every single place we check the mouse button is wrong and has been practically forever, since we never check if a button is actually the left, just not the right.

Cattlesquat:
You're right on the aside for sure. I *also* think you almost certainly want:

`if (e.isPopupTrigger()) { return true; }`

At the top of that routine. And then STILL run the rest of the checks. Because essentially all the time we're "checking for right click" or "filtering out right click", we're wanting to make sure we don't divert the context menu events for something else. And context menu activation has all the multiplatform stuff. Notice I *don't* include "if (!e.isPopupTrigger()) return false;". Basically it's a "sufficient but not necessary" situation.

I'm also curious why you didn't use the version from that SO example:
`private static boolean isRightClick(MouseEvent e) {
    return (e.getButton()==MouseEvent.BUTTON3 ||
            (System.getProperty("os.name").contains("Mac OS X") &&
                    (e.getModifiers() & InputEvent.BUTTON1_MASK) != 0 &&
                    (e.getModifiers() & InputEvent.CTRL_MASK) != 0));
}` 

Your version is subtly different and you may have thought through it better, but it just caught my eye.

uckelman:
You're going to love how `MouseEvent.isPopupTrigger()` is implemented (this is from the JDK source):
``` 
public boolean isPopupTrigger() {
    return popupTrigger;
}
```
The `boolean` it returns is set from one of the ctor args. There's just no logic at all connecting it to right clicks. It could literally be _anything_. So, we'd best not rely on that for indicating a right click, just for indicating that we should show a context menu. (Everywhere else we're using it, there's a context menu lurking about which could be shown.)

Cattlesquat:
Right, but because it "could be anything", that means it should be aggressively checked when we want to make sure we're NOT interfering with a context menu. So in other words a lot of the times we want to say "if this isn't a right click" we also want to make sure "if this isn't for popping up context menu" -- and indeed we will often need to check both for this reason. 

Otherwise we end up in the exact situation that the Mac users were reporting "why can't I get context menus" -- and the answer is because some other flow was eating the event.

So in the flow I was working on there in KeyBufferer I'm (most importantly, 100% accuracy needed) not preventing the "Immobilized" piece from being able to receive context menus, while (secondary importance, 99% accuracy is acceptable) filtering things that obviously aren't drag-rubber-bands out.

uckelman:
> You're right on the aside for sure. I _also_ think you almost certainly want:
> 
> `if (e.isPopupTrigger()) { return true; }`
> 
> At the top of that routine.

It doesn't make sense to check `e.isPopupTrigger()` if we have no context menu to show. `KeyBufferer` doen't have one.

> I'm also curious why you didn't use the version from that SO example:

It's because `MouseEvent.getModifiers()` is deprecated.

Cattlesquat:
Imagine the following scenario from my Keybufferer flow: 
* Player presses the mouse on an immobilized piece (mousePressed fires)
* But whatever it is the player pressed IS a context menu event, and this IS a platform where those fire in mousePressed. 
* We don't have a context menu to display "right here", but we DO have a context to display later in the life of this same mouseEvent when it comes through to whatever calls menuDisplayer. (so see here is where I disagree with your statement "we have no context menu to show").
* We could choose to filter for solely right click, or for the somewhat-more-Mac-aware version of right click you have written. This would probably work 99.n% of the time. But it COULD fail if someone was on a platform where a different button from right click was used to pop context menus.
* Yet it would be 100% safe if we filtered out any  isPopupTrigger() events from this flow. We would still ALSO want to filter to try to get "only left mouse" as best we can. 

I mean, by instinct I'm probably more comfortable than you are "letting 0.5% of the users slide unless/until we get a problem reported". But my experience with Vassal so far is that there are a lot of Vassal people with various crazy/zany systems out there, and so I'm trying to give you the 100% safe solution rather than the 99.5% safe. 

If you want to run with fewer checks here, you're the boss, but I do think you're misunderstanding the value of isPopupTrigger() if you're assuming it should only be checked if "in this very method we might put up a context menu". I'm saying it's proper to check it if we want to know "in the lifetime of this _event_ through all our methods, does it trigger a context menu".

uckelman:
Oh ffs, it's as bad as I thought: I can drag pieces with mouse buttons 2, 4, and 5. That's completely fucked up.

uckelman:
WHY DID NO ONE TELL US ABOUT THIS BEFORE?!

uckelman:
I'm going to go through all of this and sort it out right now.

Cattlesquat:
Sounds good. I am happy to help with a "forensic review" if you want.

uckelman:
That's funny because I work in digital forensics. Nothing good ever happens in digital forensics.

uckelman:
@Cattlesquat Have a look at PR #79. That turned out to be rather more epic than I expected at the outset.

uckelman:
Fixed in #79 instead.

PR #75, opened 2020-07-09
Bug 13229: Save property map instead of piece clone as snapshot

yanlyub:
Walking on ice here with all the downcasting and relying on Vassal veterans' experience, is this roughly what was proposed in http://www.vassalengine.org/forum/viewtopic.php?f=5&t=11964#p61060?

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Inventory.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Inventory.java
@@ -95,6 +95,7 @@ import VASSAL.counters.PieceFilter;
 import VASSAL.counters.PieceIterator;
 import VASSAL.counters.Properties;
 import VASSAL.counters.PropertiesPieceFilter;
+import VASSAL.counters.PropertyExporter;
 import VASSAL.counters.Stack;
 import VASSAL.i18n.Resources;
 import VASSAL.i18n.TranslatableConfigurerFactory;
@@ -960,7 +961,7 @@ public class Inventory extends AbstractConfigurable
       Command comm = null;
       if (p != null) {
         // Save state first
-        p.setProperty(Properties.SNAPSHOT, PieceCloner.getInstance().clonePiece(p));
+        p.setProperty(Properties.SNAPSHOT, ((PropertyExporter) p).getProperties());
         if (tracker == null) {
           tracker = new BoundsTracker();
           tracker.addPiece(p);
--- a/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -89,6 +89,7 @@ import VASSAL.counters.PieceIterator;
 import VASSAL.counters.PieceSorter;
 import VASSAL.counters.PieceVisitorDispatcher;
 import VASSAL.counters.Properties;
+import VASSAL.counters.PropertyExporter;
 import VASSAL.counters.Stack;
 import VASSAL.tools.LaunchButton;
 import VASSAL.tools.image.ImageUtils;
@@ -680,8 +681,7 @@ public class PieceMover extends AbstractBuildable
     Command comm = new NullCommand();
     for (GamePiece piece : pieces) {
       if (piece.getProperty(Properties.SNAPSHOT) == null) {
-        piece.setProperty(Properties.SNAPSHOT,
-                          PieceCloner.getInstance().clonePiece(piece));
+        piece.setProperty(Properties.SNAPSHOT, ((PropertyExporter) piece).getProperties());
       }
       comm = comm.append(piece.keyEvent(key));
     }
--- a/vassal-app/src/main/java/VASSAL/counters/ActionButton.java
+++ b/vassal-app/src/main/java/VASSAL/counters/ActionButton.java
@@ -273,8 +273,7 @@ public class ActionButton extends Decorator implements EditablePiece, Loopable {
           ActionButton action = (ActionButton) piece;
           if (action.stroke != null && action.stroke.getKeyStroke() != null && action.bounds.contains(point)) {
             // Save state prior to command
-            p.setProperty(Properties.SNAPSHOT,
-              PieceCloner.getInstance().clonePiece(p));
+            p.setProperty(Properties.SNAPSHOT, ((PropertyExporter) p).getProperties());
             try {
               RecursionLimiter.startExecution(action);
               Command command = p.keyEvent(action.stroke.getKeyStroke());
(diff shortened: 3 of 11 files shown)


src/main/java/VASSAL/counters/ReportState.java
```diff
@@ -136,13 +136,13 @@ public Command myKeyEvent(KeyStroke stroke) {
...
-    GamePiece oldPiece = (GamePiece) getProperty(Properties.SNAPSHOT);
+    java.util.Map<String, Object> oldPiece = (java.util.Map<String, Object>) getProperty(Properties.SNAPSHOT);
 
-    boolean wasVisible = oldPiece != null && !Boolean.TRUE.equals(oldPiece.getProperty(Properties.INVISIBLE_TO_OTHERS));
+    boolean wasVisible = oldPiece != null && !Boolean.TRUE.equals(oldPiece.get(Properties.INVISIBLE_TO_OTHERS));
     boolean isVisible = !Boolean.TRUE.equals(outer.getProperty(Properties.INVISIBLE_TO_OTHERS));
 
     PieceAccess.GlobalAccess.hideAll();
-    String oldUnitName = oldPiece == null ? null : oldPiece.getLocalizedName();
+    String oldUnitName = oldPiece == null ? null : (String) oldPiece.get(BasicPiece.BASIC_NAME);
```
yanlyub, line 145: Not sure about this at all but it was the best replacement for `getLocalizedName()` I could find, BasicPiece also returns `getLocalizedName()` when asked for `getLocalizedProperty(BasicPiece.BASIC_NAME)`

src/main/java/VASSAL/counters/PropertyExporter.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import VASSAL.build.module.properties.PropertyNameSource;
+import VASSAL.build.module.properties.PropertySource;
+
+public interface PropertyExporter extends PropertyNameSource, PropertySource {
+
+  default Map<String, Object> getProperties() {
```
yanlyub, line 12: C++ devs should like this, it's almost like multiple inheritance.

Clean design would probably be unifying all three Property*-Interfaces into one.

src/main/java/VASSAL/counters/ReportState.java
BrentEaston: No, this is not going to produce the same result if a designer has provided translations for either Basic Piece Names or words in Text Label formats. getLocalizedName() is not a single property.

piece.getLocalizedName() starts at the outer and calls getLocalizedName() all the way down the trait stack and each trait can potentially add translated elements to the piece name. piece.get(BasicPiece.BASIC_NAME) just returns the raw untranslated base name from the BasicPiece trait. In reality, the only traits that modify the localizedName are Text Label and BasicPiece, but the potential is there.

The whole purpose of this exercise is to collect information for the ReportState trait, so I think the best way forward would be to create our own temporary property for the localized name.  Modify the PropertyExporter default implementation , add a new property key LOCALIZED_NAME="localizedName" and add a new property to your property map

result.put(LOCALIZED_NAME, getLocalizedName());

Then ReportState can use

String oldUnitName = oldPiece == null ? null : oldPiece.getProperty(PropertyExporter.LOCALIZED_NAME);

src/main/java/VASSAL/counters/PropertyExporter.java
```diff
@@ -0,0 +1,20 @@
...
+    final List<String> propertyNames = getPropertyNames();
+    java.util.Map<String, Object> result = new HashMap<>();
+    for (String propertyName : propertyNames) {
+      result.put(propertyName, getLocalizedProperty(propertyName));
+    }
+    return result;
```
BrentEaston, line 18: The list of propertyNames will never change once a GamePiece has been created. It is also relatively expensive to retrieve the list of names. Is it possibly to cache the propertyNames list in a static variable inside an interface default method implementation?

yanlyub: Could cache them in a `static final List<String>` but that would be a single List for all PropertyExporter objects in the JVM. Besides, remember what Donald Knuth said about premature optimization, once we optimize away the `PieceCloner.clone()` we can do some new measurements and if we see that `getPropertyNames()` is a major performance hog we can still see about caching the results. Some of its implementations are basically `return new ArrayList() { A, B, C }`, they look easy enough for the JVMs to optimize them during runtime if they are called often enough. Modern JVMs do wonders with small methods, they can turn them to native code on the fly.

BrentEaston: Sounds good to me.

uckelman:
Has anyone profiled this? (What profiler do you recommend these days?)

yanlyub:
> Has anyone profiled this? (What profiler do you recommend these days?)

We had a discussion about this here http://www.vassalengine.org/forum/viewtopic.php?f=5&t=11964

With profilers, the more they cost the better they are. JProfiler is pretty good but costs money. As an open-source project we could get a free YourKit license, it's supposed to be quite good.

BrentEaston:
We have had a Yourkit licence in the past, and renewed it once. They where quite happy to provide a license, we just need to provide a link to their website from ours. I think we used to have the link on the sourceforge page, I can't see it anymore. You've just got to write an email to them with the links to our project pages.

uckelman:
What I see in that discussion is that the existing code has been profiled, but not the replacement.

yanlyub:
It removes calls to `PieceCloner.clone()` which generally sounds like a good thing.

uckelman:
So we're replacing it because it's slow, but we don't know much time the replacement takes?

yanlyub:
We replaced it not only because it was slow but also because it was unnecessary to clone the whole piece only to save it's properties.

uckelman:
You _suspect_ that it's slow. If you've measured it, show me.

yanlyub:
> You _suspect_ that it's slow. If you've measured it, show me.

The results of the profiling and the rest of the discussion is here http://www.vassalengine.org/forum/viewtopic.php?f=5&t=11964

uckelman:
I was just looking at that. I see the "before" profiling. Where is the "after"?

yanlyub:
I used an evaluation version of JProfiler to profile this, I don't have it anymore. The free visualvm profiler spits out completely different results.

Have a look at @BrentEaston's comment here http://www.vassalengine.org/forum/viewtopic.php?f=5&t=11964#p61060 he explains very well why this change makes sense.

yanlyub:
@uckelman look here http://www.vassalengine.org/forum/viewtopic.php?f=6&t=11898&p=62819#p62814, try these two modules with master then with this PR's code. You will see the difference yourself, no profiling will be needed. Doing this change was a really good idea by @BrentEaston.

PR #77, opened 2020-07-09
13153 fix the start-a-log-file at beginning of game preference to work right, and give it a comprehensible name

Cattlesquat:
Possibly we don't want this to fire for an online game, but afaik it would have fired before (in the paths where it actually fired) -- but the implementation had grown so full of weeds that it's hard to know what was really intended. Anyway it would be easy to just ignore this in an online game.

### DIFF
--- a/src/main/java/VASSAL/build/module/GameState.java
+++ b/src/main/java/VASSAL/build/module/GameState.java
@@ -339,6 +339,18 @@ public class GameState implements CommandEncoder {
 
     if (gameStarted) {
       adjustSplitter();
+
+      if (gameStarting) {
+        SwingUtilities.invokeLater(new Runnable() {
+          @Override
+          public void run() {
+            Logger logger = GameModule.getGameModule().getLogger();
+            if (logger instanceof BasicLogger) {
+              ((BasicLogger)logger).queryNewLogFile(true);
+            }
+          }
+        });
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -339,6 +339,18 @@ public void setup(boolean gameStarting) {
...
+        SwingUtilities.invokeLater(new Runnable() {
+          @Override
+          public void run() {
+            Logger logger = GameModule.getGameModule().getLogger();
+            if (logger instanceof BasicLogger) {
+              ((BasicLogger)logger).queryNewLogFile(true);
+            }
+          }
+        });
+      }
```
uckelman, line 353: I'm curious, does this actually need to be run via `invokeLater`? The reason the splitter adjustment below this does is that it is Totally Broken, and I wasn't expecting showing the log file dialog to be Totally Broken.

uckelman: Or do you just want it to happen _after_ the splitter is adjusted?

Cattlesquat: It doesn't care about the splitter, it cares about all the other wonky game components already having been initialized

Cattlesquat: One example - StartupGlobalKeyCommands run during an invokeLater, so to have the logging begin after THEM, I definitely need invokeLater.

PR #80, opened 2020-07-10
Removed calls to deprecated IOUtils.closeQuietly

yanlyub:
Removes most of the leftover calls to closeQuietly(). In some cases this helped detangle the code and introduce detailed error messages where there was previously one catch-all error.

### DIFF
--- a/src/main/java/VASSAL/build/module/ModuleExtension.java
+++ b/src/main/java/VASSAL/build/module/ModuleExtension.java
@@ -22,6 +22,7 @@ import java.awt.event.ActionListener;
 import java.io.BufferedInputStream;
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
@@ -59,7 +60,6 @@ import VASSAL.configure.StringConfigurer;
 import VASSAL.i18n.Resources;
 import VASSAL.tools.ArchiveWriter;
 import VASSAL.tools.DataArchive;
-import VASSAL.tools.io.IOUtils;
 
 /**
  * An optional extension to a GameModule
@@ -122,30 +122,24 @@ public class ModuleExtension extends AbstractBuildable implements GameComponent,
     // Record that we are currently building this Extension
     GameModule.getGameModule().setGpIdSupport(this);
 
-    BufferedInputStream in = null;
-    try {
-      in = new BufferedInputStream(archive.getInputStream(fileName));
-    }
-// FIXME: should this be a FileNotFoundException?
-    catch (IOException e) {
-    }
-
-    if (in == null) {
-      build(null);
-    }
-    else {
+    try (BufferedInputStream in = new BufferedInputStream(archive.getInputStream(fileName))) {
       try {
         final Document doc = Builder.createDocument(in);
-        build(doc.getDocumentElement());
-        in.close();
+        if (doc != null) {
+          build(doc.getDocumentElement());
+        }
       }
-      // FIXME: review error message
       catch (IOException e) {
+        // FIXME: review error message
+        logger.error("Error while creating document from file {}", fileName, e);
         throw new ExtensionsLoader.LoadExtensionException(e);
       }
-      finally {
-        IOUtils.closeQuietly(in);
-      }
+    }
+    catch (FileNotFoundException e) {
+      logger.error("File {} not found in archive", fileName, e);
+    }
+    catch (IOException e) {
+      logger.error("Error while reading file {} from archive", fileName, e);
     }
 
     GameModule.getGameModule().add(this);
--- a/src/main/java/VASSAL/build/module/documentation/BrowserHelpFile.java
+++ b/src/main/java/VASSAL/build/module/documentation/BrowserHelpFile.java
@@ -24,8 +24,10 @@ import java.beans.PropertyChangeSupport;
 import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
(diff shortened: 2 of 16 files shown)


src/test/java/VASSAL/tools/ipc/IPCMessageDispatcherTest.java
```diff
@@ -45,7 +45,7 @@ public void testRun() throws IOException, InterruptedException {
         exactly(msg.length+1).of(out).flush();
 
         oneOf(out).writeObject(fin);
-        exactly(2).of(out).close();
+        exactly(1).of(out).close();
```
yanlyub, line 48: Reason for this is, the code originally did a `close()` inside a try-catch, and then a second `close()` in the finally block. And the test expected this behavior.

This is not a very clean way of closing streams, if the `close()` failed once, why would it work the second time around? And if we try it twice just in case, why twice, why not 4, 5 or 37 times, just to be on the safe side?

Cattlesquat: while (!isClosedYet()) close();

uckelman: The reason for checking that close is called twice is that under normal operation, `close()` really was called twice---once by calling `close()` directly and once by `IOUtils.closeQuietly()`. If it's called some other number of times, the test should fail.

After your change, once is the correct number of times; but twice wasn't wrong before.

src/main/java/VASSAL/build/module/ModuleExtension.java
```diff
@@ -122,30 +122,24 @@ public void build() {
...
-      in = new BufferedInputStream(archive.getInputStream(fileName));
-    }
-// FIXME: should this be a FileNotFoundException?
-    catch (IOException e) {
-    }
-
-    if (in == null) {
-      build(null);
-    }
-    else {
```
uckelman, line 136: This doesn't preserve behavior in the `in == null` case.

src/main/java/VASSAL/build/module/documentation/BrowserHelpFile.java
```diff
@@ -99,61 +101,64 @@ public void launch() {
...
-      try {
-        in = new ZipInputStream(new BufferedInputStream(
-              GameModule.getGameModule().getDataArchive()
-                .getInputStream("help/" + getContentsResource()))); //$NON-NLS-1$
-      }
-      catch (IOException e) {
-        // The help file was created with empty contents.
-        // Assume an absolute URL as the starting page.
-        url = new URL(startingPage);
-        return;
```
uckelman, line 121: How are you handling this case now?

src/main/java/VASSAL/launch/ModuleManager.java
```diff
@@ -413,14 +411,19 @@ public void run() {
             ErrorDialog.bug(e);
           }
           finally {
-            IOUtils.closeQuietly((Closeable) in);
-            IOUtils.closeQuietly(out);
             IOUtils.closeQuietly(clientSocket);
```
uckelman, line 414: You missed one.

src/main/java/VASSAL/build/module/ModuleExtension.java
yanlyub: The `in==null` case could only happen if `archive.getInputStream(fileName)` in pre-change line 127 threw an exception. This case is preserved and has better error messages now, it jumps from the outer `try` into either the `catch (FileNotFoundException e)` in line 138 or the `catch (IOException e) {` in line 141.

The `build(null)` did nothing, the `build(Element e)` method starts like this:
```
  public void build(Element e) {
    if (e == null) {
      return;
    }
```

src/main/java/VASSAL/build/module/documentation/BrowserHelpFile.java
yanlyub: Same way it was handled before, I moved this code to it's own method since it can be called from two catch blocks now:

```
  private void setFallbackUrl() {
    try {
      url = new URL(startingPage);
    }
    catch (MalformedURLException e) {
      logger.error("Malformed URL: {}", startingPage, e);
    }
  }
```

src/main/java/VASSAL/launch/ModuleManager.java
```diff
@@ -411,7 +411,14 @@ public void run() {
             ErrorDialog.bug(e);
           }
           finally {
-            IOUtils.closeQuietly(clientSocket);
```
uckelman, line 418: Is there a reason why we can't do `try (Socket clientSocket = serverSocket.accept()) {` at the head of the loop body and avoid the `finally` block completely?

yanlyub: Honestly, I am afraid of changing it that much due to these two lines in the middle of the while loop and the lack of unit tests:

```
clientSocket.close();
if (message == null || clientSocket.isClosed()) continue;
```

I have no idea what is going on there, why there's a need to check if the socket is closed right after calling `close()`, and when (if ever) the code after the `continue` is supposed to be executed, given normal circumstances where no exceptions are thrown.

uckelman: I added `clientSocket.isClosed()` there in `226c2deb`, noting:
```
Fixed Bug 2489325: SocketException in ModuleManager$SocketListener.run()
Check whether client is listening before trying to write reply.
```
I believe the problem that addressed was that the other end would sometimes hang up before a response could be written. I'm 50-50 on whether writing to the socket after this is a bug or intentional; Java's sockets implementation behaves unexpectedly in my experience, so I could certainly imagine `isClosed()` returning false here and the other end to still be waiting for a response.

Probably best to leave that alone for now unless you want to dig into testing it (noting that it's devilishly hard to test sockets code).

uckelman: "given normal circumstances where no exceptions are thrown" ---> That's the rub with sockets in Java. You do get exceptions in situations which are totally normal, such as the other end hanging up.

yanlyub: > Probably best to leave that alone for now unless you want to dig into testing it (noting that it's devilishly hard to test sockets code).

I think so too, for now this is a good first step, it removes the deprecated `closeQuietly()`, makes the second `close` inside the finally block more verbose, and maybe even highlights a previously unknown/unnoticed problem, while keeping the exact same functionality it had before.

PR #81, opened 2020-07-11
Add name of off-master branches to package filenames for clarity

uckelman:


yanlyub:
The git-commit-id-plugin still writes the output of `git describe --tags` into the property file which is pulled from the code into the about window.

uckelman:
> The git-commit-id-plugin still writes the output of `git describe --tags` into the property file which is pulled from the code into the about window.

I noticed that, but primarily wanted builds to have better names to avoid confusion when testing. Is it easy to make that plugin do the same?

yanlyub:
The plugin already writes 20+ different pieces of information from git into the properties, including the branch. I changed the code to include the branch: PR #82

PR #82, opened 2020-07-12
Added branch to version number

yanlyub:
Also changed the unit test to use the more readable and generally much better hamcrest asserts.

### DIFF
--- a/src/main/java/VASSAL/tools/version/GitProperties.java
+++ b/src/main/java/VASSAL/tools/version/GitProperties.java
@@ -21,12 +21,15 @@ public class GitProperties {
   private static final String KEY_GIT_CLOSEST_TAG_NAME = "git.closest.tag.name";
   private static final String KEY_GIT_CLOSEST_TAG_COMMIT_COUNT = "git.closest.tag.commit.count";
   private static final String KEY_GIT_COMMIT_ID = "git.commit.id";
+  private static final String KEY_GIT_BRANCH = "git.branch";
 
   private static final String DEFAULT_CLOSEST_TAG_NAME = "3.3.0";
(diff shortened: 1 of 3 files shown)


src/main/java/VASSAL/tools/version/GitProperties.java
```diff
@@ -59,12 +62,19 @@ private void readGitProperties() {
...
-    return String.format(
+    final String versionPrefix = String.format(
       "%s-%s-g%s",
       gitProperties.getProperty(KEY_GIT_CLOSEST_TAG_NAME, DEFAULT_CLOSEST_TAG_NAME),
       gitProperties.getProperty(KEY_GIT_CLOSEST_TAG_COMMIT_COUNT, DEFAULT_CLOSEST_TAG_COMMIT_COUNT),
       gitProperties.getProperty(KEY_GIT_COMMIT_ID, DEFAULT_COMMIT_ID).substring(0, GIT_COMMIT_ID_ABBREVIATED_LENGTH)
     );
+
+    final String versionSuffix = gitProperties.getProperty(KEY_GIT_BRANCH, DEFAULT_BRANCH);
+    if (!versionSuffix.equals(DEFAULT_BRANCH) && !versionSuffix.equals(SUPPRESS_BRANCH)) {
```
uckelman, line 73: I'm not following what this is intended to do w/r/t `DEFAULT_BRANCH`, as it's the empty string.

yanlyub: If I don't check for DEFAULT_BRANCH here it would return a version string with a trailing "-"

uckelman: Oh, I see. I think what's confusing is that the variable is called `DEFAULT_BRANCH` when the value doesn't name a branch.

PR #84, opened 2020-07-12
Fix JavaDoc search feature

jimu:
This commit fixes `make javadoc` so that it generates pages that return correct search results.

**PROBLEM DESCRIPTION**

`make javadoc` generates JavaDoc HTML documentation from the VASSAL source code. This documentation includes a SEARCH field in the upper-right corner of each page.  After filling in a search term in a browser and clicking on a search result, the browser follows a broken link. The page *does* exist, the link is just broken.

Example of the relative portion of an erroneous link:
```
   javadoc/undefined/VASSAL/command/AddPiece.html
```
Correct link:
```
   javadoc/VASSAL/command/AddPiece.html
```

**STEPS TO REPRODUCE PROBLEM**

1. From a command line, change to the root VASSAL project directory and run `make javadoc`
2. Use a browser to view `javadoc/index.html`
3. Enter `AddPiece` in the SEARCH field on the page
4. Click on the top result, `VASSAL.command.AddPiece`

The browser returns `ERR_FILE_NOT_FOUND` or something to that effect

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,15 @@ JDOC:=$(JAVAPATH)/javadoc
 JDEPS:=$(JAVAPATH)/jdeps
 JLINK:=$(JAVAPATH)/jlink
 
+# fixes javadoc 11 bug that generates invalid search urls
+JDOCVER:=$(shell javadoc --version 2>/dev/null | sed 's/^[^0-9]*\([0-9]\+\).*/\1/')
+ifeq ($(JDOCVER),11)
+  JDOCOPS:=--no-module-directories 
+else ifeq ($(JDOCVER),12)
+  JDOCOPS:=--no-module-directories 
+endif
+
+
 NSIS:=makensis
 #DMG:=dmg
 DMG:=../libdmg-hfsplus/dmg/dmg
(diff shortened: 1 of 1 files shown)


uckelman:
I wasn't able to reproduce the problem. It worked for me without any changes. What version of Java are you using? On what OS?

yanlyub:
For me the javadoc target of the makefile doesn't run at all.

The makefile variable `JDOCLINK:=file:///usr/share/javadoc/java/api` points nowhere, I don't have a `/usr/share/javadoc` at all.

The `make javadoc` starts, spits out 100 `package does not exist` and `error: cannot find symbol` errors, then stops.

We need to switch to https://maven.apache.org/plugins/maven-javadoc-plugin/, we will need to use it anyways if we want to publish the vassal jar on mavencentral.

uckelman:
> The `make javadoc` starts, spits out 100 `package does not exist` and `error: cannot find symbol` errors, then stops.

Have you checked that it didn't work? Mine does that also, but nonetheless produces the javadoc.

yanlyub:
> Have you checked that it didn't work? Mine does that also, but nonetheless produces the javadoc.

Yes, it does not even create the `JDOCDIR:=javadoc` directory.

jimu:
> What version of Java are you using? On what OS?

```
$ cat /etc/centos-release
CentOS Linux release 8.1.1911 (Core)

$ mvn --version
Apache Maven 3.5.4 (Red Hat 3.5.4-5)
Maven home: /usr/share/maven
Java version: 11.0.7, vendor: Oracle Corporation, runtime: /usr/lib/jvm/java-11-openjdk-11.0.7.10-1.el8_1.x86_64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.18.0-147.el8.x86_64", arch: "amd64", family: "unix"

$ javac --version
javac 11.0.7

$ javadoc --version
javadoc 11.0.7
```

uckelman:
I've tried this now, and when I run `make javadoc` from the PR, I get
```
javadoc: error - invalid flag: --no-module-directories
```
I don't find that flag in the man page for `javadoc`, which makes me wonder if it was removed between Java 11 and Java 14.

Suggestion so that it works in both cases?

yanlyub:
Indeed, this flag exists in my Java 11, but doesn't in my Java 13 and 14

uckelman:
This looks good. Thank you.

PR #85, opened 2020-07-12
3042 - Improved "Retire"/"Resign"/"Switch Sides" button flow.

Cattlesquat:
STEP 1 - Removed the pointless first dialog box from the "Retire" flow. This box added no actual additional choices, only added a needless extra step, especially for the common use case of "hotseat" players.

STEP 2 - Further improved the flow for Hotseat use case by offering the next "real" player slot (if one is available) as the default-selected option. Filtered out of the list are (a) player slots already occupied, and (b) player slots using any of the common names for "Solitaire" / "Solo" / "Referee".  If no available "real" slot is found, then the Observer option is selected as the default. 

Hotseat players can now click the button and hit Enter (or click Okay) to cycle to next player, which removes THREE extra clicks from their present flow. Meanwhile no other use cases should suffer negative effects.

Should correctly handle any number of sides.

Did a bunch of testing, but this obviously belongs in a "beta" build.

STEP 3 - "Retire" button gets an assignable hotkey field, just like Big Boy toolbar buttons get. 

COMMENT - I wish we would change the configurer name of this button to "Switch Sides" or "Switch Sides or Retire" something more intuitive to what the most common use case of this button is, but I'll leave that aside unless folks get interested in it. It's literally just a few text changes in the .properties files.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
@@ -471,46 +471,23 @@ public class PlayerRoster extends AbstractConfigurable implements CommandEncoder
     // Thus hotseat players can easily cycle through the player positions as they will appear successively as the default.
     // Common names for Solitaire players (Solitaire, Solo, Referee) do not count as "real" player sides, and will be skipped.
     // If we have no "next" side available to offer, we stay with the observer side as our default offering.
-    boolean found = false;       // Haven't found our own side in the list.
+    boolean found = false;       // If we find a usable side
     String mySide = getMySide(); // Get our own side, so we can find the "next" one
-    String nextChoice = translatedObserver; // For now default to observer
-    if (mySide != null) {                   // If I'm presently an observer, then my next side is just going to be the first real one we find later
-      for (String s : sides) {
-        if (mySide.equals(s)) {             // We're looking for "our" side
-          found = true;                     // We found it, so next "real" one is the one we want
-          continue;            
-        }
-        if (alreadyTaken.contains(s)) {     // Don't want anything that's already taken
-          continue;
-        }
-        if (Resources.getString("PlayerRoster.solitaire").equals(s) ||   // Don't want a solitaire side
-            Resources.getString("PlayerRoster.solo").equals(s) ||
-            Resources.getString("PlayerRoster.referee").equals(s)) {
-          continue;
-        }
-        if (!found) {                       // Don't actually want ANYTHING until we've found "our" current side in the list
-          continue;
-        }
-        nextChoice = s;                     // Aha! We've found a next side after ours, AND it's not taken. THAT's the default we want. 
+    int myidx = (mySide != null) ? sides.indexOf(mySide) : -1; // See if we have a current non-observe side.
+    int i = (myidx >= 0) ? ((myidx + 1) % sides.size()) : 0;   // If we do, start looking in the "next" slot, otherwise start at beginning.
+    for (int tries = 0; i != myidx && tries < sides.size(); i = (i+1) % sides.size(), tries++) { // Wrap-around search of sides
+      String s = sides.get(i);
+      if (!alreadyTaken.contains(s) &&  
+          !Resources.getString("PlayerRoster.solitaire").equals(s) &&
+          !Resources.getString("PlayerRoster.solo").equals(s) &&
+          !Resources.getString("PlayerRoster.referee").equals(s)) {
+        found = true; // Found an available slot that's not our current one and not a "solo" slot.
         break;
-      }      
+      }
     }
     
-    if (translatedObserver.equals(nextChoice)) {
-      // If we get in here, that either means "our" side was the last available one we found in the list,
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -483,32 +466,69 @@ protected String promptForSide() {
     }
 
     availableSides.removeAll(alreadyTaken);
+
+    // If a "real" player side is available, we want to offer "the next one" as the default, rather than observer.
+    // Thus hotseat players can easily cycle through the player positions as they will appear successively as the default.
+    // Common names for Solitaire players (Solitaire, Solo, Referee) do not count as "real" player sides, and will be skipped.
+    // If we have no "next" side available to offer, we stay with the observer side as our default offering.
+    boolean found = false;       // Haven't found our own side in the list.
```
uckelman, line 474: What if the only available sides are _before_ the side we have now?

Cattlesquat: See line 499 down below the if block.

uckelman: Ok.

So the rule is: Select the first available non-solo side after the current one, wrapping around to the head of the list if necessary?

Cattlesquat: Correct! A "modulo" next-available-non-solo-side, if you will. So imagine you're in the hotseat use case, then whenever you click the "resign" button it comes up with the next logical side for you, so you can basically just click resign button and hit Enter and you'll be on the next side (or hit the Resign Hotkey and then Enter). In a two-player game it will just swap you back and forth between the two sides. Obviously at any point you can still manually pick any of the other sides, or "Observer".

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -483,32 +466,69 @@ protected String promptForSide() {
...
+        if (Resources.getString("PlayerRoster.solitaire").equals(s) ||
+            Resources.getString("PlayerRoster.solo").equals(s) ||
+            Resources.getString("PlayerRoster.referee").equals(s)) {
+          continue;
+        }
+        nextChoice = s;
+        break;
+      }                  
+    }
+
```
uckelman, line 513: I think you could to this whole thing more simply like this:
```
  found = false;
  int myidx = sides.indexOf(mySide);
  if (myidx != -1) {
    // we have a side

    int i = (myidx + 1) % sides.size();
    for ( ; i != myidx && !found; i = (i+1) % sides.size()) {
      String s = sides[i];

      if (!alreadyTaken.contains(s) && 
          !Resources.getString("PlayerRoster.solitaire").equals(s) &&
          !Resources.getString("PlayerRoster.solo").equals(s) &&
          !Resources.getString("PlayerRoster.referee").equals(s)) {
        found = true;
      }
    }
  }

  if (found) {
    nextChoice = sides[i];
  }
```

Circling back to the beginning of the list gets folded in as a non-exceptional case this way, and finding the index of the side we already have gets moved outside the loop.

Cattlesquat: We want a "break" in there after the "found = true", right?

Cattlesquat: Oh I see the !found hiding in the for controls

uckelman: You could `break` instead. It's 6 of one, a half dozen of the other for that.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
uckelman: Makes sense. That's a good rule.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
uckelman: Try this out, of course---I typed it up but didn't actually compile or try it.

uckelman: Anyhow, make this less convoluted and I'm ready to merge it.

Cattlesquat: k, yeah turns out we need to break because otherwise "i" gets bollixed on to the next spot in the for-loop logic. Couple other minor complexities I'm ironing out, but it's gonna be a real compact-ass thing compared to before. Will have something up shortly.

uckelman: Yes, breaking is the right thing, and take the `!found` test out of the loop condition. I did say I didn't test this. :)

Cattlesquat: Okay all done & checked in. Runs correctly.

PR #88, opened 2020-07-13
13172 - Undo button can now be assigned a hotkey by module designer

Cattlesquat:
Still left it no key by default, but now it can at least be assigned one.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
@@ -130,10 +130,13 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
     button = mod.getToolBar().add(stepAction);
     button.setToolTipText(Resources.getString("BasicLogger.step_forward_tooltip"));  //$NON-NLS-1$
     button.setAlignmentY((float) 0.0);
+    
+    final NamedKeyStrokeListener undoKeyListener = new NamedKeyStrokeListener(undoAction, null);
+    mod.addKeyStrokeListener(undoKeyListener);
 
     final NamedKeyStrokeListener stepKeyListener = new NamedKeyStrokeListener(stepAction, NamedKeyStroke.getNamedKeyStroke(KeyEvent.VK_PAGE_DOWN, 0));
     mod.addKeyStrokeListener(stepKeyListener);
-
+    
     final KeyStrokeListener newLogKeyListener = new KeyStrokeListener(newLogAction, KeyStroke.getKeyStroke(KeyEvent.VK_W, InputEvent.ALT_DOWN_MASK));
     mod.addKeyStrokeListener(newLogKeyListener);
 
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -147,6 +150,22 @@ public void propertyChange(PropertyChangeEvent evt) {
       }
     });
     stepIconConfig.fireUpdate();
+    
+    final NamedHotKeyConfigurer stepKeyConfig = new NamedHotKeyConfigurer("stepHotKey", Resources.getString("BasicLogger.step_forward_hotkey"), stepKeyListener.getNamedKeyStroke());  //$NON-NLS-1$ //$NON-NLS-2$
+    GlobalOptions.getInstance().addOption(stepKeyConfig);
+    stepKeyConfig.addPropertyChangeListener(new PropertyChangeListener() {
```
uckelman, line 156: Could make this a lambda.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -159,16 +178,21 @@ public void propertyChange(PropertyChangeEvent evt) {
     });
     undoIconConfig.fireUpdate();
 
-    final NamedHotKeyConfigurer stepKeyConfig = new NamedHotKeyConfigurer("stepHotKey", Resources.getString("BasicLogger.step_forward_hotkey"), stepKeyListener.getNamedKeyStroke());  //$NON-NLS-1$ //$NON-NLS-2$
-    GlobalOptions.getInstance().addOption(stepKeyConfig);
-    stepKeyConfig.addPropertyChangeListener(new PropertyChangeListener() {
+    final NamedHotKeyConfigurer undoKeyConfig = new NamedHotKeyConfigurer("undoHotKey", Resources.getString("BasicLogger.undo_hotkey"), undoKeyListener.getNamedKeyStroke()); //$NON-NLS-1$ //$NON-NLS-2$
+    GlobalOptions.getInstance().addOption(undoKeyConfig);
+    undoKeyConfig.addPropertyChangeListener(new PropertyChangeListener() {
```
uckelman, line 183: And this one.

uckelman:
This is fine.

That said, we really screwed up way back at the start by not assigning the standard system hotkeys to standard things like Undo and Redo. That ship has sailed for V3---we can't now be sure that the standard hotkeys don't conflict with what's in some module----but it's a thing I intend to do right in V4.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
Cattlesquat: Lambdas up in latest push.

Cattlesquat:
Agree on "claiming" some of the core default standard keys for Vassal. Although I also think that V4 should, like an IDE, have a near-religious commitment to allowing the module designer to redefine pretty much any key. If some jackwagon wants Ctrl+Z to mean "Fire Forward Phaser Banks" then we should let him.

uckelman:
Agreed, the capability to rebind keys is a good thing, so are unsurprising defaults.

PR #92, opened 2020-07-14
Rewrote tests not to use org.jmock.lib.legacy.ClassImposteriser

uckelman:
This lets us drop our dependency on jmock-legacy.


uckelman:
Also, I feel strongly that "Imposteriser" should be spelled with a Z, so this way I don't have to look at spelled with an S anymore.

PR #95, opened 2020-07-15
13174 - Pieces in Piece Palettes (i.e. PieceSlots) can be given a scale factor

Cattlesquat:
Hello Vassal fans! Tired of having things in your piece palettes show up either "ginormous" or "too small", subject to the arbitrary whims of what constitutes 100% size in your art set? **WELL LOOK NO FURTHER!** This Amazing New Patch allows YOU the module designer to apply a scale ratio to any of the widgets in your piece palettes, and indeed each different widget can have different scales, so you can scale a "card" widget differently from a "counters" widget, etc, etc, etc. 

image


yanlyub:
No CI build here, none here https://travis-ci.com/github/vassalengine/vassal/pull_requests, what happened there..

Cattlesquat:
Uh... weird? This is from a VERY recent branch of master -- I branched it yesterday or the day before.

uckelman:
Huh, that's puzzling.

yanlyub:
Builds can be skipped but the commit message needs to contain certain keywords for that, none of which are in the last commit here.. https://docs.travis-ci.com/user/customizing-the-build#skipping-a-build

I'll try pushing a change into my formatting PR, it's probably due for a merge-update anyways.

uckelman:
That's really weird, as Travis did build the PR Brent made just now.

Cattlesquat:
This isn't even "a very complicated PR".

yanlyub:
Seems to work again, it built Brent's PR and also my latest push.

Maybe it just doesn't like Brian anymore after that broken build from yesterday :grinning:

Cattlesquat:
Anything I can push on this that would restore the behavior?

Cattlesquat:
I just scrolled back through my "git command line" history in the shell and it was all the completely normal standard stuff.

yanlyub:
Travis probably just had a short outage, it seems to work again.

> Anything I can push on this that would restore the behavior?

Any change should trigger a build, anything. Add a javadoc somewhere or add additional info to an existing javadoc, or change the code formatting a little e.g. add a space somewhere where one should be, there's a lot of code like `1+i+n` which should rather be `1 + i + n`.

Cattlesquat:
Yep, that woke it up.

doc/ReferenceManual/PieceWindow.htm
```diff
@@ -52,6 +52,13 @@ <h2>Game Piece Palette<br>
 Button Icon:&nbsp; </span>The icon image on the toolbar button.<br>
             <span style="font-weight: bold;">Hotkey:&nbsp; </span>A
 keyboard shortcut for the toolbar button.<br>
+            <span style="font-weight: bold;">Image scale for all child 
+            members:&nbsp; </span> An scale factor to be used to draw
```
uckelman, line 56: "A scale"

vassal-app/src/main/java/VASSAL/build/widget/PieceSlot.java
```diff
@@ -101,6 +102,20 @@ public PieceSlot(PieceSlot piece) {
   public PieceSlot(CardSlot card) {
     this((PieceSlot) card);
   }
+  
+  
+  // If we're a child of a piece widget that allows scale control, get our scale from that. Otherwise default to 1.0 
+  public double getScale() {
+    Widget w = this;
+    while (w.getParent() != null) {
```
uckelman, line 110: `while ((w = w.getParent()) != null) {` saves you a call to `w.getParent()`.

Cattlesquat: Done. At some point in my C++ past I got trained never to do assignments in comparisons :)

uckelman: There are two circumstances in which I think they're a good idea: when you're assigning to what is basically an iterator in the condition of a while loop, and when you're checking the return value of a system call for an error code.

Cattlesquat: yeah makes sense here because it actually saves an extra call.

uckelman:
Do you think "child members" instead of "children" makes us sound like aliens?

vassal-app/src/main/java/VASSAL/build/widget/ListWidget.java
```diff
@@ -54,16 +55,42 @@
...
+  @Override
+  public double getScale() {
+    return scale;
+  }  
+  
+  // True if we're part of main (docked) Piece Palette
+  protected boolean isMainPiecePalette() {
+    Widget w = this;
+    while (w.getParent() != null) {
+      w = w.getParent();
```
uckelman, line 86: Here's one more I assignment in `while` condition you could do that I missed the first time.

Cattlesquat:
I am confident that it does not :D

vassal-app/src/main/java/VASSAL/build/widget/ListWidget.java
Cattlesquat: Taken care of

PR #96, opened 2020-07-15
Bug 13178.

BrentEaston:
1. Add Place Marker traits in Prototype definitions to the list of
pieceSlot definitions used to refresh module.
2. Fix bug in recording of Place Marker traits that could cause pieces
to be refreshed to their creating piece, not their own updated
definition.
3. Minor formatting fixes in affected classes.
4. Report GPId updates by checker in Chat Window when updating modules in Editor. [Will normally never be seen except for first edit after naive editing of the buildfile, but many modules may get some reports after this fix as Place Markers in old Prototypes are assessed for the first time.

### DIFF
--- a/src/main/java/VASSAL/build/GameModule.java
+++ b/src/main/java/VASSAL/build/GameModule.java
@@ -65,6 +65,7 @@ import VASSAL.build.module.PlayerRoster;
 import VASSAL.build.module.Plugin;
 import VASSAL.build.module.PredefinedSetup;
 import VASSAL.build.module.PrivateMap;
+import VASSAL.build.module.PrototypeDefinition;
 import VASSAL.build.module.PrototypesContainer;
 import VASSAL.build.module.RandomTextButton;
 import VASSAL.build.module.ServerConnection;
@@ -952,6 +953,12 @@ public abstract class GameModule extends AbstractConfigurable implements Command
     for (PieceSlot pieceSlot : theModule.getAllDescendantComponentsOf(PieceSlot.class)) {
       checker.add(pieceSlot);
     }
+    // Add any PieceSlots in Prototype Definitions
+    for (PrototypesContainer pc : theModule.getComponentsOf(PrototypesContainer.class)) {
+      for (PrototypeDefinition pd : pc.getDefinitions()) {
+        checker.add(pd);
+      }
+    }  
     checker.fixErrors();
   }
 
--- a/src/main/java/VASSAL/build/module/GameRefresher.java
+++ b/src/main/java/VASSAL/build/module/GameRefresher.java
@@ -34,8 +34,6 @@ import javax.swing.JPanel;
 import javax.swing.JTextArea;
 import javax.swing.WindowConstants;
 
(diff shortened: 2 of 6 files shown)


src/main/java/VASSAL/build/GameModule.java
```diff
@@ -952,6 +953,12 @@ protected void checkGpIds() {
     for (PieceSlot pieceSlot : theModule.getAllDescendantComponentsOf(PieceSlot.class)) {
       checker.add(pieceSlot);
     }
+    // Add any PieceSlots in Prototype Definitions
+    for (PrototypesContainer pc : theModule.getComponentsOf(PrototypesContainer.class)) {
+      for (PrototypeDefinition pd : pc.getDefinitions()) {
+        checker.add(pd);
+      }
```
uckelman, line 960: I think you could do `pc.getDefinitions.forEach(checker::add);` here for the inner loop?

src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -122,6 +121,14 @@ public void execute(boolean testMode, boolean useName) {
     for (PieceSlot slot : theModule.getAllDescendantComponentsOf(PieceSlot.class)) {
       gpIdChecker.add(slot);
     }
+    
+    // Add any PieceSlots in Prototype Definitions
+    for (PrototypesContainer pc : theModule.getComponentsOf(PrototypesContainer.class)) {
+      for (PrototypeDefinition pd : pc.getDefinitions()) {
+        gpIdChecker.add(pd);
+      }
```
uckelman, line 129: Similarly here for the inner loop.

src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -320,21 +325,24 @@ public void windowClosing(WindowEvent we) {
         @Override
         public void actionPerformed(ActionEvent e) {
```
uckelman, line 326: Since you're already making changes right here---these `ActionListeners` could all become lambdas. E.g.,
```
runButton.addActionListener(e -> run());
```

src/main/java/VASSAL/build/module/PrototypesContainer.java
```diff
@@ -43,6 +44,10 @@
   private Map<String,PrototypeDefinition> definitions =
     new HashMap<>();
 
+  public Collection<PrototypeDefinition> getDefinitions() {
+    return definitions.values();
+  }
```
uckelman, line 49: Do we want the prototype defs mapping to be modifiable here? (It would be if we return `values()` instead of wrapping that in a `Collections.unmodifiableCollection()` before returning it.) I'm not saying either way, as I'm not familiar with this code.

uckelman:
A few comments/questions, and then some conflicts to resolve.

BrentEaston:
Agree with all comments.  I am going to create a new branch and re-apply the changes rather than try and fight with the conflicts. Will get it done tonight.

BrentEaston:
Closing. Replace with updated version based on latest source

PR #98, opened 2020-07-16
13184 CounterDefailtViewer can now display overlapping counters

Cattlesquat:
MouseoverStackViewer has an IMHO annoying behavior that it insists on only finding things "at the exact pixel location" of the first piece it finds. And this is even a specific induced behavior requiring special code! And yet in all the games that I'm used to, there are multiple "layers" of pieces (so incapable of formally stacking) and so of course everything sticks out overlappingly in a space, and usually when I'm rolling over something I'm much MORE curious about the stuff that's half-obscured underneath (regardless of whether it's "on the exact pixel") than I am about the thing that's on top of the stack obscuring everything else.

Anyway, since this was an induced behavior, the "business end" of the fix for it is basically a one-liner, and everything else is just me adding actual user-documentation and stuff because yay me.

image


Cattlesquat:
image

Cattlesquat:
I guess I should mention that I fixed some wrong-indentation things that were yellow-highlighted as I was going through here. Which shows up as "more changes".

src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -652,8 +656,11 @@ protected void apply(GamePiece p) {
           lastLayer = layer;
         }
 
-        if (foundPieceAt == null) {
-          foundPieceAt = p.getPosition();
+        // Our show-overlap setting controls whether we're allowed to show all pieces overlapping this spot, or only pieces at the exact location of the first piece we find.
+        if (!showingOverlap) {
+          if (foundPieceAt == null) {
+            foundPieceAt = p.getPosition();
+          }
```
uckelman, line 663: Test the conditions together since the only line which does anything is enclosed in both blocks.

`if (!showingOverlap && foundPieceAt == null) {`

Cattlesquat: Done

Cattlesquat:
This one re-based and ready to go.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -589,14 +591,16 @@ else if (displayWhat.equals(EXC_LAYERS)) {
     protected int lastLayer = -1;
     protected int insertPos = 0;
     protected Point foundPieceAt;
+    protected boolean showingOverlap;
 
-    public Visitor(Filter filter, Map map, Point pt) {
+    public Visitor(Filter filter, Map map, Point pt, boolean showOverlap) {
```
uckelman, line 596: This could break anything which extends it, or CounterDetailViewer. What you should do here to prevent that is add the following:
```
@Deprecated
public Visitor(Filter filter, Map map, Point pt) {
  this(filter, map, point, false);
}
```
so that the old ctor still exists for anything which calls it.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -815,7 +820,8 @@ else if (drawSingleDeprecated) {
       SHOW_MOVE_SELECTED,
       SHOW_NON_MOVABLE,
       UNROTATE_PIECES,
-      SHOW_DECK
+      SHOW_DECK,
+      SHOW_OVERLAP,
```
uckelman, line 824: Trailing comma.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -847,6 +853,7 @@ else if (drawSingleDeprecated) {
        Resources.getString("Editor.MouseOverStackViewer.non_moveable"), //$NON-NLS-1$
        Resources.getString("Editor.MouseOverStackViewer.unrotated_state"), //$NON-NLS-1$
        Resources.getString("Editor.MouseOverStackViewer.top_deck"), //$NON-NLS-1$
+       Resources.getString("Editor.MouseOverStackViewer.show_overlap"), //$NON-NLS-1$
```
uckelman, line 856: Trailing comma.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -877,7 +884,8 @@ else if (drawSingleDeprecated) {
       Boolean.class,
       Boolean.class,
       Boolean.class,
-      Boolean.class
+      Boolean.class,
+      Boolean.class,
```
uckelman, line 888: Trailing comma.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
yanlyub: This is legitimate Java code and makes further work on this code easier by allowing to copy+paste the last line and to adjust it according to needs.

Cattlesquat: Yeah I actually "found these" with trailing commas in them and simply extended the pattern.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
Cattlesquat: Fixed.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
Cattlesquat: I recommend we leave the commas, but let me know if you want me to take them out.

Otherwise I addressed your other thing and it's perhaps ready to go.

PR #99, opened 2020-07-17
13186 - Fix deck drawing that allowed drawn card to be seen prematurely

Cattlesquat:
Per email thread. Fixes a bug introduced by PR#6 but apparently not in 3.3.2.

### DIFF
--- a/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -1093,8 +1093,8 @@ public class PieceMover extends AbstractBuildable
 
         final GamePiece piece = dragContents.nextPiece();
         final Point pos = relativePositions.get(index++);
-        final Map map = piece.getMap();
-
+        final Map map = piece.getMap();        
+        
         if (piece instanceof Stack){
           stackCount = 0;
           piece.draw(g, EXTRA_BORDER - boundingBox.x + pos.x,
@@ -1112,10 +1112,19 @@ public class PieceMover extends AbstractBuildable
           else {
             stackCount = 0;
           }
-
+          
           final int x = EXTRA_BORDER - boundingBox.x + pos.x + offset.x;
           final int y = EXTRA_BORDER - boundingBox.y + pos.y - offset.y;
+          
+          String owner = "";
+          if (piece.getParent() instanceof Deck) {
+            owner = (String)piece.getProperty(Properties.OBSCURED_BY);
+            piece.setProperty(Properties.OBSCURED_BY, ((Deck) piece.getParent()).isFaceDown() ? Deck.NO_USER : null);
+          }
           piece.draw(g, x, y, map == null ? target : map.getView(), zoom);
+          if (piece.getParent() instanceof Deck) {
+            piece.setProperty(Properties.OBSCURED_BY, owner);
+          }
 
           final Highlighter highlighter = map == null ?
             BasicPiece.getHighlighter() : map.getHighlighter();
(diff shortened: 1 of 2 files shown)


Cattlesquat:
It was this commit: https://github.com/vassalengine/vassal/pull/6/commits/02cccb7f0228888d4a842311386114da10e8002b

src/main/java/VASSAL/build/module/map/PieceMover.java
```diff
@@ -1112,10 +1109,19 @@ private void drawDragImage(BufferedImage image, Component target,
...
+          
+          String owner = "";
+          if (piece.getParent() instanceof Deck) {
+            owner = (String)piece.getProperty(Properties.OBSCURED_BY);
+            piece.setProperty(Properties.OBSCURED_BY, ((Deck) piece.getParent()).isFaceDown() ? Deck.NO_USER : null);
+          }
           piece.draw(g, x, y, map == null ? target : map.getView(), zoom);
+          if (piece.getParent() instanceof Deck) {
+            piece.setProperty(Properties.OBSCURED_BY, owner);
+          }
```
uckelman, line 1124: Doing this in a draw method is truly horrible, but I don't see a good alternative right now without some _major_ refactoring.

Cattlesquat: Yeah it's really less the Deck's implementation that is bad here than it is the Obscured implementation. We can look at a deeper fix if we want, but this should "hold the water back" for now.

Cattlesquat: If you want to REALLY laugh and laugh, look at the draw method for **Deck** -- 

It does this:
      Object owner = top.getProperty(Properties.OBSCURED_BY);
      top.setProperty(Properties.OBSCURED_BY, faceDown ? NO_USER : null);

And this time it DOESN'T restore the state. So I will revise my previous statement to "It's really less Deck's appalling implementation that's affecting us here..." :O

Cattlesquat: That probably causes subtle occasional funky behavior by cards in decks, by the way. The whole reason this ended up getting brought to Brent's attention at all was because I went looking for the source of some 3.2.17 funky deck behavior, and found Brent's patch had "partially" resolved it.

uckelman:
NB: I'm not adding this to the change log because the bug never appeared in a release.

src/main/java/VASSAL/build/module/map/PieceMover.java
uckelman: Geez, maybe fix that, too?

Cattlesquat: Okay I think a downstream fix might be along these lines:
#1 - Obscurable.java (only) gets a special draw method with an extra parameter to allow overriding the obscured state
#2 - Deck.java and PieceMover.java utilize that parameter instead of changing the game state.

uckelman: (It's funny how GH makes all your number signs into issue links. I was wonder what these had to do with the first two PRs.)

Cattlesquat: Yeah I'll pull together a little mini-refactor patch and submit it for you to have a look at.

Cattlesquat: (Now that this got less urgent and became more of an interesting/educational housecleaning project)

Cattlesquat: PR #100 (TIL how to make PR autolink, haha)

BrentEaston: The Drag image is only created once per drag, so the overhead isn't significant. The one in Deck.draw() is worse. It does restore the state though at line 993. At a minimum, the Deck image should be cached, it does not change that often compared to the number of times the Deck is redrawn. It only changes when the top card changes.

I think Yan already has a Deck rewrite in his sights, but I think this will be difficult to achieve without introducing significant bugs and subtle behaviour changes.

PR #100, opened 2020-07-17
Clean up Deck & PieceMover to not write into game state during draw method

Cattlesquat:
Well this was fun, "thinky", and actually not such a deep refactor after all.


Cattlesquat:
AND I nabbed PR #100 😎

Cattlesquat:
@BrentEaston Check out PR #100 I think it cleans this up a lot better.

Cattlesquat:
Oh this IS #100 sorry haha

Cattlesquat:
On your comment - keep in mind this is only replacing the two flows where we were manually "Obscuring to Deck.NO_USER" before the call. I *think* there is no path there where the peek-y tops of obscurables could be displayed, right? Aren't those only displayed when deck is only "obscured to others", not "obscured to me"? I think with the owner (even temporarily) being Deck.NO_USER, we always get "obscured to me"?

Cattlesquat:
Oh now I understand what you meant. Right. I'll have a look for a way to fix.

Cattlesquat:
Okay I think THIS actually cleans it up. More trouble than it's worth? @BrentEaston @uckelman 

I rebased on master so only the last commit in the list has any meaningful changes. 

Someone can still go back and have Deck cache its image.

src/main/java/VASSAL/counters/GamePiece.java
```diff
@@ -50,7 +59,7 @@
    * @param obs the Component on which this piece is being drawn
    * @param zoom the scaling factor.
    */
-  public void draw(Graphics g, int x, int y, Component obs, double zoom);
+  public void draw(Graphics g, int x, int y, Component obs, double zoom, EnumSet<DrawFlags> flags);
```
uckelman, line 62: This will break every single custom class which implements `GamePiece`, by changing the signature of a function that they all have to implement. This has to be done a different way. (Too late tonight anymore for me to comment further.)

Cattlesquat: True. Dang. Well... maybe the band-aid we put on it is good enough then?

The alternative I can think of off the top of my head is going back to my "find the obscurable" route and then manually walking the rest of the "above" Decorators and drawing them. Bleh?

BrentEaston: I stumbled across Interface default methods the other day which I think is designed to work around this. 

We would need to keep the original function signature, but add the new signature to the interface and provide a default implementation that is then used by every implementer of the Interface that doesn't specifically override that function. 

It's still dangerous, because some trait's don't follow the normal pattern in draw (draw inner, draw me).

BrentEaston: Honestly, I don't think it is worth this much potential for damage. What are we actually trying to achieve? The band-aid fixes the problem and we are no worse off, with a change to 1 source file as opposed to 185. 

The Drag shadow generation is not really an issue as it only get's done once at the start of a drag.

The Deck drawing the top card would be nice to fix, but for what actual gain? Efficiency improvement? Marginal in this case I think. 

The principle of 'its a bad thing to play with state in a draw'. Sure, it is worth investigating a better way, for the day in the far distant future when drawing is happening on a different thread.

Cattlesquat: Yeah I'll just close this and we can always come back to the issue later. Many better fish to fry.

PR #101, opened 2020-07-17
Chatter depreciated calls

Stew-rt:
Found Depreciated swing calls in Chatter.java 

I'm not 100% on one or two things here:
1) I couldn't find these methods (over-ridden or otherwise) being called from anywhere else in the code, so I've assumed they are intended to be public, and should be depreciated, so I've
  1.1) Marked them as depreciated
  1.2) Made them call the new over-ridden methods
2) the new methods require floats, not ints, casting ints to floats is all good, but I've cast floats to int for the return from the depreciated call.. which is blasphemous.

Thoughts would be appreciated, so I can know what to do in this project in future.

### DIFF
--- a/src/main/java/VASSAL/build/module/Chatter.java
+++ b/src/main/java/VASSAL/build/module/Chatter.java
@@ -375,7 +375,7 @@ public class Chatter extends JPanel implements CommandEncoder, Buildable {
     }
   }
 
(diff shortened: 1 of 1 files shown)


uckelman:
As `PView` and `WrappedView` here are both private inner classes, there's no need to deprecate anything---just fix them, because nothing outside of `Chatter` can be depending on them.

PR #104, opened 2020-07-18
Refactored PlayerWindow and its SplitPane handling

yanlyub:
This cleans up the code in PlayerWindow and its splitting functionality.


src/main/java/VASSAL/build/module/PieceWindow.java
```diff
@@ -70,7 +70,7 @@
   public static final String HIDDEN = "hidden"; //$NON-NLS-1$
   protected static UniqueIdManager idMgr = new UniqueIdManager("PieceWindow"); //$NON-NLS-1$
   protected JComponent root;
-  protected ComponentSplitter.SplitPane mainWindowDock;
+  protected SplitPane mainWindowDock;
```
uckelman, line 73: This is going to break every module which touches the splitter. It will cause a problem with exactly the same modules as the change I had to revert in 3.3.1.

yanlyub: I see I have changed too much, I will propose a version that retains API compatibility.

yanlyub:
This is now API compatible with the previous state.

I propose to deprecate all `protected ComponentSplitter.SplitPane` fields, find out which modules use them and how they use them, and offer them a replacement.

vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -152,6 +152,11 @@ public void propertyChange(PropertyChangeEvent evt) {
   };
 
   protected PlayerWindow frame = new PlayerWindow();
+
+  /**
+   * @deprecated deprecated without replacement, modify/subclass {@link PlayerWindow} instead.
+   */
```
uckelman, line 158: This is misleading. Nobody should be subclassing `PlayerWindow`.

uckelman:
I've tried this now with several of the OCS modules, which are some of those which use ComponentSplitter.SplitPane directly. They seem to still work.

yanlyub:
Wow that went fast, I just wanted to push a minor improvement..

PR #105, opened 2020-07-18
13191 - SEARCH function added to EDITOR (holy crap I can't wait to get to actually use this)

Cattlesquat:
Reasonably decent search function for Editor. Currently can match "name" or "class name" of any entry in the tree.
Offers toggles for whether names and/or class names are matched.
Offers toggle for "match case"

Open to any suggestions for improving interface & functionality. 

I'm thinking about ways to be able to match the internals of Traits (less about the mechanics of getting stuff out of the traits and more about what the UI flow would be like)

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
+++ b/vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
@@ -131,10 +131,10 @@ public class ConfigureTree extends JTree implements PropertyChangeListener, Mous
   protected Action helpAction;
   
   // Search parameters
-  protected String searchString;
-  protected boolean matchCase;
-  protected boolean matchNames;
-  protected boolean matchTypes;
+  protected String searchString;     // Current search string
+  protected boolean matchCase;       // True if case-sensitive
+  protected boolean matchNames;      // True if match configurable names
+  protected boolean matchTypes;      // True if match class names
 
   public static Font POPUP_MENU_FONT = new Font("Dialog", 0, 11);
   protected static List<AdditionalComponent> additionalComponents = new ArrayList<>();
@@ -404,7 +404,22 @@ public class ConfigureTree extends JTree implements PropertyChangeListener, Mous
     }
     
     return foundNode;
-  }   
+  }  
+  
+  
+  // Finds out how many total nodes match the search string
+  public final int getNumMatches(String searchString) {    
+    List<DefaultMutableTreeNode> searchNodes = getSearchNodes((DefaultMutableTreeNode)getModel().getRoot());
+    int hits = 0;
+
+    for (int index = 0; index < searchNodes.size(); index++) {
+      if (checkNode(searchNodes.get(index), searchString)) {
+        hits++;
+      }
+    }
+    
+    return hits;    
+  }
   
     
   // Search action - runs search dialog box, then searches
@@ -441,6 +456,11 @@ public class ConfigureTree extends JTree implements PropertyChangeListener, Mous
         find.addActionListener(new ActionListener() {
           @Override
           public void actionPerformed(ActionEvent e) {
+            boolean anyChanges = (!searchString.equals(search.getText())) || 
+                                 (matchCase != sensitive.isSelected()) ||
+                                 (matchNames != names.isSelected()) ||
+                                 (matchTypes != types.isSelected());
+            
             searchString = search.getText();
             matchCase    = sensitive.isSelected();
             matchNames   = names.isSelected();
@@ -452,7 +472,12 @@ public class ConfigureTree extends JTree implements PropertyChangeListener, Mous
               chat (Resources.getString("Editor.search_all_off"));                            
             }
             
-            if (!searchString.isEmpty()) {              
+            if (!searchString.isEmpty()) {
+              if (anyChanges) {
+                int matches = getNumMatches(searchString);
+                chat (String.valueOf(matches) + " " + Resources.getString("Editor.search_count") + searchString);
+              }
+              
               DefaultMutableTreeNode node = findNode(searchString);                
               if (node != null) {
                 TreePath path = new TreePath(node.getPath());
--- a/vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
+++ b/vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
@@ -42,12 +42,14 @@ import javax.swing.Action;
 import javax.swing.Box;
 import javax.swing.BoxLayout;
 import javax.swing.JButton;
+import javax.swing.JCheckBox;
 import javax.swing.JComboBox;
 import javax.swing.JDialog;
 import javax.swing.JFrame;
 import javax.swing.JLabel;
 import javax.swing.JOptionPane;
 import javax.swing.JPopupMenu;
+import javax.swing.JTextField;
 import javax.swing.JTree;
 import javax.swing.KeyStroke;
 import javax.swing.SwingUtilities;
@@ -98,6 +100,7 @@ public class ConfigureTree extends JTree implements PropertyChangeListener, Mous
   protected EditorWindow editorWindow;
   protected Configurable selected;
   protected int selectedRow;
+  protected String searchCmd;
   protected String moveCmd;
   protected String deleteCmd;
   protected String pasteCmd;
@@ -111,6 +114,7 @@ public class ConfigureTree extends JTree implements PropertyChangeListener, Mous
   protected KeyStroke pasteKey;
   protected KeyStroke deleteKey;
   protected KeyStroke moveKey;
+  protected KeyStroke searchKey;
   protected KeyStroke helpKey;
   protected KeyStroke propertiesKey;
   protected KeyStroke translateKey;
@@ -119,9 +123,14 @@ public class ConfigureTree extends JTree implements PropertyChangeListener, Mous
   protected Action pasteAction;
   protected Action deleteAction;
   protected Action moveAction;
+  protected Action searchAction;
   protected Action propertiesAction;
   protected Action translateAction;
   protected Action helpAction;
+  
+  // Search parameters
+  protected String searchString;
+  protected boolean matchCase;
 
   public static Font POPUP_MENU_FONT = new Font("Dialog", 0, 11);
   protected static List<AdditionalComponent> additionalComponents = new ArrayList<>();
@@ -141,6 +150,7 @@ public class ConfigureTree extends JTree implements PropertyChangeListener, Mous
     addMouseListener(this);
     addMouseMotionListener(this);
     addTreeSelectionListener(this);
+    searchCmd = Resources.getString("Editor.search"); //$NON-NLS-1$
     moveCmd = Resources.getString("Editor.move"); //$NON-NLS-1$
     deleteCmd = Resources.getString("Editor.delete"); //$NON-NLS-1$
     pasteCmd = Resources.getString("Editor.paste"); //$NON-NLS-1$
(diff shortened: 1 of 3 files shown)


Cattlesquat:
These screen shots seem like a good idea, for whatever reason.

image

src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -284,6 +321,205 @@ protected JPopupMenu buildPopupMenu(final Configurable target) {
     addAction(popup, buildImportAction(target));
     return popup;
   }
+  
+  
+  // Enumerates our configure tree in preparation for searching it
+  private final List<DefaultMutableTreeNode> getSearchNodes(DefaultMutableTreeNode root) {
```
yanlyub, line 327: `final` modifier has no effect in private methods

src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -284,6 +321,205 @@ protected JPopupMenu buildPopupMenu(final Configurable target) {
...
+    List<DefaultMutableTreeNode> searchNodes = new ArrayList<DefaultMutableTreeNode>();
```
yanlyub, line 328: Can use the "diamond operator" `<>` here, the compiler is smart enough to figure out what type we need:
```
List<SomeType> myList = new ArrayList<>();
```

src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -284,6 +321,205 @@ protected JPopupMenu buildPopupMenu(final Configurable target) {
     addAction(popup, buildImportAction(target));
     return popup;
   }
+  
+  
+  // Enumerates our configure tree in preparation for searching it
```
yanlyub, line 326: Could make this a proper Javadoc, especially on the public methods down below, the syntax for these comments to be detected as Javadoc is `/** */` e.g.:
```
/**
 * A nice comment describing the method
 */
public void aMethod() {
```

src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -284,6 +321,205 @@ protected JPopupMenu buildPopupMenu(final Configurable target) {
...
+    }
+    
+    return foundNode;
+  }  
+  
+  
+  // Finds out how many total nodes match the search string
+  public final int getNumMatches(String searchString) {    
+    List<DefaultMutableTreeNode> searchNodes = getSearchNodes((DefaultMutableTreeNode)getModel().getRoot());
+    int hits = 0;
```
yanlyub, line 413: Can be rewritten using Java 8's stream API, the readability is slightly better then:
```
    return (int) searchNodes
      .stream() // entry point of the Stream API
      .filter(node -> checkNode(node, searchString)) // leaves only the elements in the stream that match the predicate
      .count(); // counts elements in the stream, returns a long, that's why it's casted to int a few lines above
```

src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -284,6 +321,205 @@ protected JPopupMenu buildPopupMenu(final Configurable target) {
...
+            if (!matchNames && !matchTypes) {              
+              matchNames = true;
+              names.setSelected(true);
+              chat (Resources.getString("Editor.search_all_off"));                            
+            }
+            
+            if (!searchString.isEmpty()) {
+              if (anyChanges) {
+                int matches = getNumMatches(searchString);
+                chat (String.valueOf(matches) + " " + Resources.getString("Editor.search_count") + searchString);
```
yanlyub, line 478: Can omit the `String.valueOf()` here and just start the concatenation with the int

src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -284,6 +321,205 @@ protected JPopupMenu buildPopupMenu(final Configurable target) {
...
+              else {
+                chat (Resources.getString("Editor.search_none_found") + searchString);
+              }
+            }            
+          }
+        });
+        
+        
+        JButton cancel = new JButton(Resources.getString(Resources.CANCEL));
+        cancel.addActionListener(new ActionListener() {
```
yanlyub, line 496: Shorter version using Java 8 lambdas:
```
cancel.addActionListener(e1 -> d.dispose());
```

src/main/java/VASSAL/configure/ConfigureTree.java
Cattlesquat: @yanlyub Can you point me in the direction of where I'd get the stream api figured out?

src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -284,6 +321,205 @@ protected JPopupMenu buildPopupMenu(final Configurable target) {
...
+      }
+    }
+    
+    return hits;    
+  }
+  
+    
+  // Search action - runs search dialog box, then searches
+  protected Action buildSearchAction(final Configurable target) {
+    final Action a = new AbstractAction(searchCmd) {
```
yanlyub, line 427: I would consider extracting this anonymous AbstractAction into it's own class since it's quite long and the ConfigureTree is already a few hundred lines long. Eclipse has decent refactoring support, extracting this as an inner class or even moving into it's own file should be quite easy.

src/main/java/VASSAL/configure/ConfigureTree.java
yanlyub: This guide looks good but you don't have to get too deep into that topic, the most common methods you will need are `stream()`, `filter()`, `map()`, `count()`, `findFirst()`, `forEach()`, if you roughly understand how they work you can look up the rest later: https://stackify.com/streams-guide-java-8/

Or just learn it as you go, in this particular example we first `filter()` the elements we want (same thing you did with the `if` inside the for loop), then we `count()` what is left (like you did with the `hits++`).

Cattlesquat: Well that certainly made it shorter!

src/main/java/VASSAL/configure/ConfigureTree.java
Cattlesquat: Probably above my skill level with high level Java architecture. I tried to get the Eclipse refactor menu to "do anything" but I could only convince it to move a bunch of minor variables into some other class, couldn't figure out how to get it to extract even a single method let alone a group of them. And discovered when poking around with moving the search stuff to another class "by hand" that they call back into the main class for getting names and stuff. I could probably spend an hour or two and make something that "would run" but it would also almost certainly be worse that what's there now. 

Since the real problem isn't the search itself, but the bulkiness of the overall class, I'd suggest it as a refactoring project for a java pro :-)

yanlyub: Ok I'll look into making a PR into this PR.

yanlyub: Have a look at https://github.com/Cattlesquat/vassal/pull/1

Cattlesquat:
@yanlyub Your PR-into-PR added a merge conflict in ExtensionTree -- you'll probably know best how to resolve it.

yanlyub:
Just some code formatting changes, I made a new PR https://github.com/Cattlesquat/vassal/pull/2

yanlyub:
Whoops, what have we done..

yanlyub:
14 checkstyle violations in ExtensionTree, I got and old version of it into the PR

Cattlesquat:
Heh - I'll pull the next version after dinner. Almost got my new Mac virtual box running Vassal now too :)

yanlyub:
I think I've overdone it, now git doesn't know that ExpressionTree was moved, and the "new" deprecated ExpressionTree is a completely new file. I think it's time to rewrite history, then I'll do the "light" version of the refactoring and keep the old files where they are:

1. make sure all pending changes are committed
2. checkout your local Bugfix13191 branch
3. create a new branch off of it, just in case
4. checkout Bugfix13191 branch again
5. rewind it the "hard" way back to b1ace63: `git reset --hard b1ace63`
6. force-push it with `git push --force`

Then the branch and this whole PR will be back at b1ace63 and I'll propose another PR.

Cattlesquat:
k, did it!

yanlyub:
Wrote a new PR for this: https://github.com/Cattlesquat/vassal/pull/3

Not sure if you get notifications for these PR-into-PRs, I didn't get any when someone did this with one of my PRs, neither did @uckelman when I made a PR into one of his'.

Cattlesquat:
I don't get a notif when you make one, only when you comment.

Cattlesquat:
Woot!

Cattlesquat:
Okay this one's now properly rebased and ready to go.

Cattlesquat:
Oooh, @yanlyub I like what you did with the refactor. That made "saving the search parameters" from session to session super-tidy to implement.

vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -1118,4 +1174,336 @@ public AdditionalComponent(Class<? extends Buildable> p, Class<? extends Buildab
...
+      this.matchTypes = matchTypes;
+      writePrefs();
+    }
+
+    public void setFrom(final SearchParameters searchParameters) {
+      this.searchString = searchParameters.getSearchString();
+      this.matchCase = searchParameters.isMatchCase();
+      this.matchNames = searchParameters.isMatchNames();
+      this.matchTypes = searchParameters.isMatchTypes();
+      writePrefs();
```
uckelman, line 1271: You don't need any of the `this`es here.

vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -1118,4 +1174,336 @@ public AdditionalComponent(Class<? extends Buildable> p, Class<? extends Buildab
...
+      
+      prefs.addOption(null, new StringConfigurer(SearchParameters.SEARCH_STRING, null, ""));
+      prefs.addOption(null, new BooleanConfigurer(SearchParameters.MATCH_CASE,   null, false));
+      prefs.addOption(null, new BooleanConfigurer(SearchParameters.MATCH_NAMES,  null, true));
+      prefs.addOption(null, new BooleanConfigurer(SearchParameters.MATCH_TYPES,  null, true));
+      
+      this.searchString = (String) prefs.getValue(SearchParameters.SEARCH_STRING);
+      this.matchCase    = (Boolean)prefs.getValue(SearchParameters.MATCH_CASE);
+      this.matchNames   = (Boolean)prefs.getValue(SearchParameters.MATCH_NAMES);
+      this.matchTypes   = (Boolean)prefs.getValue(SearchParameters.MATCH_TYPES);
```
uckelman, line 1217: You don't need any of the `this`es here.

uckelman:
Ok, so... problems: The first problem I have when I try this is that Ctrl-F doesn't bring up the dialog, nor does selecting the entry from the Edit menu.

uckelman:
o i c

The dialog doesn't display when nothing is selected in the editor, as happens when it starts. So... should the dialog display anyhow? Should the root start selected in the editor? Should the dialog menu entry be disabled when nothing is selected?

vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
Cattlesquat: done

Cattlesquat:
Selecting the root by default sounds best. I'll just have to go figure out a quick command to do that and a place to put it.

Cattlesquat:
There - selects root at start.

uckelman:
I opened a PR against your branch which does two things (mainly it's dialog layout). The layouts which come with Swing are just awful.

PR #107, opened 2020-07-19
13197 - Mac interface cleanup and Legacy preference

Cattlesquat:
Stage 1 - This should clean up the rest of the Mac Mouse bullshit and add the legacy preference. Keyboard shortcuts will be more of a bear.

Rather than have methods named e.g. "isRightMouse" but they lie to us and screw with our heads when we deal with Macs (wait, does "Control" mean "Control" right now, or does "Control" mean "Command"?), I have created names that hopefully are both understandable when read by a non-initiate but also "different enough" to remind us they have special meaning (and let us keep things straight in our heads when dealing with Macs).
**SwingUtils.isVanillaLeftButtonDown()** -- is left mouse button down ("except on Mac when it's pretending to be right mouse")
**SwingUtils.isContextMouseButtonDown()** -- is right button down (OR on Mac is left button pretending to be right button)
**SwingUtils.isSelectionToggle()** -- normally "Ctrl+LeftClick" on non-Macs. Other things on Macs, depending on "stuff".
**SwingUtils.isShortcutKeyDown()** -- ... in development. Name may slightly evolve if my needs do. Normally "Ctrl". On Mac usually "Command" but when legacy pref "Control"

This was the easy part, partially because of all the work done on this for 3.3.2. 

Beating the keyboard shortcuts into line is going to be more "uncharted territory" so a good thing for me to work on ... tomorrow!

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,6 @@
 3.3.3-beta1 - ???
 
+* 13196: Ctrl+Click toggles rather than only deselects
 * 13171: SetGlobalProperty makes menuCommands the wrong type of array
 * 13164: Upgrade Apache Commons libraries: commons-io 2.7, commons-codec 1.14,
     commons-lang3 3.10


yanlyub:
This is good.

There's a lot of missing Override annotations in that area, perhaps you could add them, they make navigating the class/interface structure easier for humans and development tools:
```
interface AnInterface {
  void methodA();
  void methodB();
}

class AClass implements AnInterface {
  @Override
  void methodA() {}

  @Override
  void methodB() {}
}
```

Cattlesquat:
Will do

Cattlesquat:
STAGE 2 - Module KeyStroke's translated back and forth from "Command" to "Control" at the appropriate places. 
**SwingUtils.getKeySystemToVassal()** - Translate's Mac's platform-specific Command key into "Ctrl" for storing-in-module purposes
**SwingUtils.getKeyVassalToSystem()** - Translate's Vassal-stored keystrokes from "Ctrl" into Command key if we're on a Mac for checking-for-actual-keystroke purposes.
* if **macLegacy** preference is in effect, then this is all just a pass-through.

Here is a secondary detail that I need advice on: I need to do ONE of the following:
(a) We could make macLegacy a "requires restart" preference. In this case I'd need @yanlyub's advice on how to fix my little configurer path so that SwingUtils.setMacLegacy() *does* get called at Vassal startup, so that SwingUtils knows what mode we're in, but that it *does not* get called every time the user checks/unchecks the preference (but locally GlobalOptions.setPrefsMacLegacy() still gets called, or in some way the preference gets properly written out)
(b) OR, someone could suggest a mechanism/algorithm to have every VASSAL.tools.KeyStrokeListener know to "unregister" and then re-"register" its keystrokes.

Cattlesquat:
Meanwhile, @uckelman we are probably ready for some test builds so that I can test this on Mac. The thought of installing a dev environment on my wife's Macbook Pro sends cold chills up my spine...

Cattlesquat:
Or if there's some way that "through the magic of maven" I can cause a mac build to get made here on my Windows machine...

yanlyub:
It is still a "make" project on Linux, maven builds the Java part but the make system takes care of packaging the final distribution.

If you can install git and have a Java installation, you could clone the repo, use the maven wrapper to build the jar, unzip the Vassal's "other" distribution somewhere, replace it's jar files with the ones you built, and start it using the VASSAL.sh script.

Cattlesquat:
I'd have to figure out how to do all of that on a Mac (sort of like having my right arm tied behind my back and several fingers of my left hand removed), and during the brief windows I can borrow my wife's laptop! 

Now if there's a way to build a Mac version on my _Windows_ machine...

Cattlesquat:
@yanlyub Investigating installing a Ubuntu virtual machine on my Windows. We'll see how far I get :)

Cattlesquat:
@yanlyub Okay, so let's say I've got a little Ubuntu virtual machine running, and it has "git" and "java 11 sdk" installed for at least some value of "installed". Like I can type "java" at a $ prompt and get a spew of java-y stuff. And I can type "git" and I've told it I exist (user.name and stuff) and it recognizes me. 

I "shared" my Windows folder of vassal over, but when I type "git status" in that folder from the Ubuntu side of the world, I get an immense spew of "everything in the world has changed", so in other words, git isn't "really" playing well with that folder (on Windows side typing "git status" in same folder says I'm all clean). 

So maybe I'd better make a new folder in my Ubuntu drive and just pull Vassal down into THAT. 

But meanwhile, what OTHER stuff do I need to get to be able to build the makefile? I tried "sudo apt-get install make" and it "made something install", but when I typed make in the vassal directory it was "sad".

uckelman:
Changes to everything is due likely to line ending differences between Windows and Linux.

What you need for making Mac builds is genisoimage (for which there will be an Ubuntu package) and dmg (for which there is probably not an Ubuntu package). I was already intending to see about updating dmg and providing a simple way to build it; I may run that down in the coming week.

Cattlesquat:
I've gotten "surprisingly far". What does one type to tell the makefile to build a Mac version (as opposed to just compiling the build)? Also... is "dmg" the name of a specific program? I thought it was more of a file format?
image

yanlyub:
Yes git on linux handles line endings differently from windows, just clone the repo into a new folder inside your VM.

You could install several more JDKs, they come in handy when you want to try different Java and Vassal versions, Vassal 3.2 will need JDK 8. You could install these three for instance: a JDK 8 for running Vassal 3.2, JDK 11 as it's the long-term release and the minimum required JDK for building Vassal 3.3, and JDK 14 as it's the most modern one and it's VM should run Vassal in the most performant way.

For the makefile to build a release, ultimately you will have to ask @uckelman as I never tried setting up my machine to do the full build cycle, I only know that you can install nsis and genisoimage from the package manager.

To install all these packages:
```
sudo apt install openjdk-14 openjdk-11 openjdk-8 nsis genisoimage
```
You can also type `sudo apt install openj` or `sudo apt install geniso` then keep hitting tab, apt has tab-completion nowadays.

One of the Javas will be set as the default, you can choose which one like this: https://askubuntu.com/a/740782

But even then you can still choose which one you use by using the absolute path, e.g.:
```
$ /usr/lib/jvm/java-11-openjdk-amd64/bin/java -version
$ /usr/lib/jvm/java-8-openjdk-amd64/bin/java -version
```
(They all are in `/usr/lib/jvm/<jdkversion>/bin/java` and you can use tab completion after typing `/usr/lib/jvm/`)

In Eclipse / IntelliJ you can add any available JDK in the compiler settings (IntelliJ will automatically find them anyways) and select which one is used for the project.

yanlyub:
> I've gotten "surprisingly far". What does one type to tell the makefile to build a Mac version (as opposed to just compiling the build)?

It's `make release-macosx`, and you see all available make targets in the last line of the Makefile, behind the `.PHONY:`. And make has tab completion too, you can just type `make ` then hit tab a couple times, or for instance `make r` + `tab` + `tab` + `-m` + `tab` + `tab`.

Cattlesquat:
Here's how far I get on "make release-macosx" so far. 
image

uckelman:
> Here's how far I get on "make release-macosx" so far.
> image

Huh, so wget _didn't_ create the `jdks` directory for you?

Cattlesquat:
Well THAT was easily remedied with mkdir. 
Next up:
image

uckelman:
Are you current with master? I suspect you're getting a stray newline in `tmp/module_deps`, and there's a commit on master to fix that.

Cattlesquat:
Seems to think everything is up to date

Cattlesquat:
I'd pulled the version down for the very first time a couple hours ago.

uckelman:
It looks like `tmp/module_deps` wasn't written out correctly, then. Whats in it?

Cattlesquat:
jdk.crypto.ec,Error: unknown option: --ignore-missing-depsUsage: jdeps <options> <path...>]use --help for a list of possible options

lol

Cattlesquat:
Maybe it doesn't like my Java 11 jdk. Still trying to figure out where to get 14. Yan's command earlier didn't find it.

yanlyub:
Which ubuntu version did you install?

Cattlesquat:
how do I tell?

Cattlesquat:
By the way I did manage to get a jdk14 installed, and got a lot farther. Am now missing, as Joel predicted, libdmg-hfsplus

Cattlesquat:
Apparently I am about to go on a "Voyage of Discovery" of downloading THAT repository and compiling it :)

yanlyub:
Try `sudo apt install build-essential` first, this will install the whole C compiler toolset. Then you might need to install single libraries like libcrypto in their development versions, you can search the ubuntu repository with `apt search libcrypto`, and the package you need usually has the `lib-` prefix and `-dev` suffix e.g. `libssl-dev`, install that one with `sudo apt install`.

Cattlesquat:
Here is where currently stuck (this is in trying to build libdmg-hfsplus)
image

Cattlesquat:
I have installed all of the things you described above. Of course I'm probably missing something *else*. I googled an article about how to install/build libdmg-hfsplus and followed all of those steps, but it doesn't remedy this.

yanlyub:
I just tried it myself, getting the same compile error in filevault.c or filevault.h or however this build log is supposed to be read.

yanlyub:
And my computer is setup for projects like this and can successfully compile other C/C++ projects with cmake, regular make, I regularly compile working versions of crispy-doom, openttd, gzdoom, emacs. With custom CFLAGS and all. I think the code of this libdmg-hfsplus is just broken. Maybe @uckelman can add his working dmg binary to the Vassal codebase.

Cattlesquat:
https://github.com/planetbeing/libdmg-hfsplus/issues/14

Cattlesquat:
Found that saying that the new libssl-dev doesn't build it, but the command they describe to find the older version doesn't produce it for me.

yanlyub:
Yes I can't get libssl1.0-dev for my ubuntu 20.04 either, they don't keep all the old versions in their repos. Ugh. Unmaintained C/C++ projects and their dependency management. Care to find out how to manually build the old libssl 1.0 lib? :rofl:

Cattlesquat:
This article, down under "still facing this in 2020" section... https://github.com/rvm/rvm/issues/4764

Cattlesquat:
BUILT that little fucker!

Cattlesquat:
By cracky I think I just built a mac version too. Now to go build the mac version of the branch I want (haha forgot to have that pulled down)

Cattlesquat:
This is like having a weird little alien computer inside my computer! What an odd and intriguing toy! I guess next I can make a "fake Mac" to test on...

Cattlesquat:
Seriously, I can't believe I was actually alive and working during the time before "hey google, how do I solve this weird completely esoteric technical problem?"  "Now how do I solve the NEXT one?" etc...

yanlyub:
Wait, you don't have Dosbox installed which puts a little MS-DOS computer inside your modern machine and allows you to play all the good old DOS games?

Cattlesquat:
Haha I've had various versions installed in the past, including the "TRS-80" simulator so I could play my stuff from the early 80's. But my last versions self-obsoleted when I went to 64-bit Windows a few years ago and I haven't re-invested yet. :)

Cattlesquat:
Aaaaaaand... it runs on the Mac!  (For certain values of "runs" ... meaning I still have some debugging of the actual code I came here to debug, but hey!)

Can anybody diagnose why this code doesn't remove the bit flags I think I'm and-ing out? 

    public KeyStroke getKeySystemToVassal (KeyStroke k) {
      if (macLegacy) {
        return k;
      }
      int modifiers = k.getModifiers();
      if ((modifiers & InputEvent.META_DOWN_MASK) != 0) {
        modifiers &= ~InputEvent.META_DOWN_MASK;
        modifiers |= InputEvent.CTRL_DOWN_MASK;
        return KeyStroke.getKeyStroke(k.getKeyCode(), modifiers);
      } 
      else {
        return k; 
      }      
    }
    
    @Override
    public KeyStroke getKeyVassalToSystem (KeyStroke k) {
      if (macLegacy) {
        return k;
      }
      int modifiers = k.getModifiers();
      if ((modifiers & InputEvent.CTRL_DOWN_MASK) != 0) {
        modifiers &= ~InputEvent.CTRL_DOWN_MASK;
        modifiers |= InputEvent.META_DOWN_MASK;
        return KeyStroke.getKeyStroke(k.getKeyCode(), modifiers);
      } 
      else {
        return k; 
      }      
    }

Cattlesquat:
Can anybody diagnose why the "KeyStroke.getKeyStroke(k.getKeyCode(), modifiers)" in those two calls end up creating a keystroke that has BOTH masks in it (it has the one I got rid of AND the new one). I've confirmed I'm passing in the right modifiers I want, but what comes out has both modifiers!

Cattlesquat:
Oh it's probably the "we're all writing to the same fucking instance of the event" problem. Fuck me.

Cattlesquat:
Or something. It's "like" that but I can't figure out what it is. When one of you guys wakes up, see if you can eyeball why those two routines return things with BOTH of the masks in them. It's successfully and-ing the old mask out of the "modifiers" variable, but then what I get back as a return from the whole routine has BOTH masks in it -- the one I or-red in AND the one I supposedly and-ed out. 

And the line: `final int mods = SwingUtils.getKeyVassalToSystem(k).getModifiers();` therefore gives me a "mods" that has both masks in it.

yanlyub:
It seems this has to do with the "old" modifiers still being in the KeyStrokes, they need to be and-ed out as well, in getKeySystemToVassal:
```
      modifiers &= ~InputEvent.META_DOWN_MASK;
      modifiers &= ~InputEvent.META_MASK;
```

and likewise in getKeyVassalToSystem:
```
      modifiers &= ~InputEvent.CTRL_DOWN_MASK;
      modifiers &= ~InputEvent.CTRL_MASK;
```

Both old modifiers are deprecated so it's probably better to add a comment to the code saying that they are needed.

uckelman:
A thing I just remembered: You don't actually need `dmg` for building a test DMG. `dmg` is what _compresses_ the DMG output by `genisoimage`. We want a compressed DMG for releases, but if all you're doing is testing, you can comment that out an use the uncompressed one.

uckelman:
> Yes I can't get libssl1.0-dev for my ubuntu 20.04 either, they don't keep all the old versions in their repos. Ugh. Unmaintained C/C++ projects and their dependency management. Care to find out how to manually build the old libssl 1.0 lib? rofl

This is all unnecessary if you guys just give me a chance to deal with `dmg`. There's a much newer version than the one I'm using. I probably can't compile the one I'm using now either---I compiled it a long time ago and never bothered to update it because I was the only one using it and it works.

yanlyub:
An automated build would be great, we could give it to any CI server and you wouldn't be the only guy on this planet anymore who can produce all the necessary releases.

Cattlesquat:
@uckelman Totally great to deal with dmg but in the meantime I actually got it to compile (got the old libssl through some hanky panky from one of those articles). So you're absolutely right that's probably not the thread we want everything to hang from but I was "on a tear" and got it running. Need a Mac build? ;)

@yanlyub Oh you're probably right it's probably still got old-style modifier in the mask and then adding the new one back in because it sees that. Whee!

uckelman:
If you're asking Java to do something with key modifiers, there's a good shot that if it sees one of the `*_MASK_DOWN` modifiers set that it will also set the corresponding `*_MASK` modifier, and vice versa, for backwards compatibility. You might be seeing that. (There's a `SequenceEncoder` test which looks for exactly that thing.)

Cattlesquat:
Yep that was it. This should go quickly now :-)

Cattlesquat:
Okay, next little mystery, @uckelman and @yanlyub. I'm less mystified by "why this doesn't work now" and more mystified by "how did this ever work". The following is from ForwardToKeyBuffer.  Note that we're sending all three of "keyPressed", "keyTyped", and "keyReleased" in to "process". And there's special code explicitly designed to consume & ignore the "keyTyped", so yay for that. But then the keyReleased arrives and gets processed, and there's nothing here designed to consume/ignore it. 

So the behavior I'm seeing is... my "translated" command gets translated correctly, and it gets recognized correctly... and it activates TWICE! And yet this is the exact logic that was in here before, apart from me changing the getKeyStrokeForEvent() to filter through my special method.

Any ideas?

`  @Override
  public void keyPressed(KeyEvent e) {
    process(e);
  }

  @Override
  public void keyReleased(KeyEvent e) {
    process(e);
  }

  @Override
  public void keyTyped(KeyEvent e) {
    process(e);
  }

  protected void process(KeyEvent e) {
    // If we've consumed a KeyPressed event,
    // then automatically consume any following KeyTyped event
    // resulting from the same keypress
    // This prevents echoing characters to the Chat area if they're keycommand for selected pieces
    if (lastConsumedEvent != null
        && lastConsumedEvent.getWhen() == e.getWhen()) {
      e.consume();
    }
    else {
      lastConsumedEvent = null;
    }
    final int c = e.getKeyCode();
    // Don't pass modifier keys alone to counters
    if (!e.isConsumed() && c != KeyEvent.VK_SHIFT && c != KeyEvent.VK_CONTROL && c != KeyEvent.VK_ALT && c != KeyEvent.VK_META) {
      Command comm = KeyBuffer.getBuffer().keyCommand
          (SwingUtils.getKeyStrokeForEvent(e));
      if (comm != null && !comm.isNull()) {
        GameModule.getGameModule().sendAndLog(comm);
        e.consume();
        lastConsumedEvent = e;
      }
    }
  }
`

yanlyub:
I can only approach this from the theoretical perspective, having almost no practical experience with these keyboard/mouse parts of Java.

Isn't "keyTyped" same as "keyPressed" and "keyReleased" combined into one event? And if "keyPressed" is processed and the "keyTyped" that belongs to that same "keyPressed" is ignored, shouldn't the "keyReleased" that also belongs to the "keyPressed" be ignored as well?

Cattlesquat:
Well that's the logic I'd be coming at it with too, and yet "somehow this always worked". So I'm wondering if there's "secret Java voodoo" at work (like the kind where the KeyStroke regrew its CTRL_DOWN_MASK because of the deprecated flag still being on)

Cattlesquat:
Okay I think I've found it. There's an extra piece of KeyStroke I need to make sure I put into the translated version, because we're apparently using it.

Cattlesquat:
(And yes, after I got my first Mac build done last night, I realized - DUH - that I could debug 95% of this by just making a 'Windows' version of the Input Classifier to swap some keys)

Cattlesquat:
DONE! And works on wife's actual Mac. Will deserve a beta, but that's what 3.3.3 is planned to have anyway.

Cattlesquat:
@uckelman @yanlyub This is my "Merge Candidate 0" for this Mac Panacea Patch, so go ahead and review when you have a chance.

Cattlesquat:
Meanwhile, I think I've created a monster...
image

src/main/java/VASSAL/build/module/map/ForwardToKeyBuffer.java
```diff
@@ -87,10 +88,10 @@ protected void process(KeyEvent e) {
       lastConsumedEvent = null;
     }
     final int c = e.getKeyCode();
-    // Don't pass SHIFT or CONTROL only to counters
-    if (!e.isConsumed() && c != KeyEvent.VK_SHIFT && c != KeyEvent.VK_CONTROL) {
+    // Don't pass modifier keys alone to counters
+    if (!e.isConsumed() && c != KeyEvent.VK_SHIFT && c != KeyEvent.VK_CONTROL && c != KeyEvent.VK_ALT && c != KeyEvent.VK_META) {
```
yanlyub, line 92: For such complex conditions, I like to create a boolean variable and use the text I would otherwise put in the comment for this variable's name, e.g.:
```
final boolean onlyModifierKeys = (c == KeyEvent.VK_SHIFT || c == KeyEvent.VK_CONTROL || c == KeyEvent.VK_ALT || c == KeyEvent.VK_META);
if (!e.isConsumed() && !onlyModifierKeys) {
...
```

This makes the `if` easier to read by making the comment part of the code.

src/main/java/VASSAL/tools/swing/SwingUtils.java
```diff
@@ -53,115 +72,300 @@ public static AffineTransform descaleTransform(AffineTransform t) {
...
+    KeyStroke getKeySystemToVassal (KeyStroke k);
+    
+    /*
+     * @returns translation of keystroke from Vassal to local system (to handle Mac platform support)
+     */
+    KeyStroke getKeyVassalToSystem (KeyStroke k);    
   }
 
+  
+  // Pass through -- presently handles all systems except Macs
```
yanlyub, line 104: You can turn this into javadoc, then the IDE's will show neat popup menus with this text when hovering over the class name.

src/main/java/VASSAL/tools/swing/SwingUtils.java
```diff
@@ -53,115 +72,300 @@ public static AffineTransform descaleTransform(AffineTransform t) {
...
+      return k;
+    }
+    
+    @Override
+    public KeyStroke getKeyVassalToSystem (KeyStroke k) {
+      return k;
+    }
   }
 
+  // All the special Mac-platform-specific mouse and keyboard code lives here
```
yanlyub, line 147: Javadoc

src/main/java/VASSAL/tools/swing/SwingUtils.java
```diff
@@ -53,115 +72,300 @@ public static AffineTransform descaleTransform(AffineTransform t) {
...
+  } 
 
   private static final InputClassifier INPUT_CLASSIFIER =
     SystemUtils.IS_OS_MAC_OSX ?
       new MacInputClassifier() : new DefaultInputClassifier();
 
-  /*
-   * @return whether the event is effectively for the left button
+  /**
+   * @return whether the event is effectively for the left button. Deprecated in favor of isVanillaLeftButtonDown()
```
yanlyub, line 279: Can write it like this, then it's going to be a proper reference to the method and all the tools will easily find it and allow navigating from this javadoc right into the method, and also rename this reference if the method is renamed:

  /**
   * @return whether the event is effectively for the left button.
   * 
   * @deprecated Deprecated in favor of {@link #isVanillaLeftButtonDown(MouseEvent)}
   */

yanlyub:
Looks good, I left a few minor suggestions, and if you could change the base class for the unit tests `VASSAL.tools.swing.AbstractSwingUtilsTestIsMouseButton` in lines 46 and 61, make it test the new methods instead of the deprecated ones, then I won't have to do that later on :grin:

uckelman:
@Cattlesquat Have a look at PR #109. It has bootstrap.sh, which you can run to download and the JDKs needed for packaging, and it also downloads and compiles a current version of dmg.

src/main/java/VASSAL/build/module/map/ForwardToKeyBuffer.java
Cattlesquat: By the way this is one reason why I personally prefer to see all the (c != KeyEvent.VK_SHIFT) things individually in parentheses. I know they aren't formally necessary but boy do they make the if statements more readable. 

Anyway, did what you suggested.

BrentEaston: I agree, that is my thought as well. Just because you can write the code in fewer characters doesn't mean you should. It makes no difference at all to a modern compiler, but makes the visual parsing of long if statements FAR easier.

uckelman: That line is just too long for easy reading. Splitting the conditions across two or more lines would help.

Cattlesquat: Already done and checked in

Cattlesquat:
This latest attempt to merge appears to have "deleted the entire project" in this branch. Need to figure out how to rewind.

Cattlesquat:
@yanlyub uh.... I most definitely did not mean to "close this". I was trying to get rid of the part where somehow merging deleted the whole project. Any hints?

uckelman:
Hmm. Let me mess with this, as I was hoping to look at this next.

Cattlesquat:
Yeah it all seemed fine until I tried to merge 13196 into it after you merged that.

Cattlesquat:
And then I tried to rewind it but apparently not the right way.

Cattlesquat:
I still have "the source code" from it on my Linux virtual machine, and I was able to cram it into the Bugfix13197 branch on github "Cattlesquat" fork. But that didn't update the PR the way it normally does. I should probably just stop touching anything related to this until you guys figure it out.

Cattlesquat:
PR #110 has a raw push of what I had on this in the other branch, right as I finally finished and built it the other day. 

It's obviously probably got everything "in the wrong directories" and what not. But it's there "just in case".

PR #112, opened 2020-07-22
13129 - New HTML Chatter with colors, fonts/bold/italic, image tags, etc

Cattlesquat:
image

Here is a whole new version of the PR where I just copied final version of the "only two files that ever changed" over into a new branch and committed it. That seems to accomplish in 10 minutes what will apparently take several of us several hours/days to untangle.


Cattlesquat:
Original PR #34 

Here's the link to the forum thread where people have been doing some testing: http://www.vassalengine.org/forum/viewtopic.php?f=6&t=12092

Cattlesquat:
Did a late-breaking push on this for (a) javadoc and (b) embarrassingly, to make it not break the custom-class version of it I gave to people. Also the copyright notice had 2003 so I updated it to 2020 and that made it seem appropriate to add my name. But if there's some different protocol for that just let me know.

uckelman:
So, I have one more comment, which might annoy you: There are no docs. How is anybody going to know how to use this or that it even exists?

Cattlesquat:
Haha, go pull down PR #117 and open GamePiece.htm to see my idea of decent documentation :-)

I was already planning to add a Chat log section to the docs, just waiting for it to get merged.

Cattlesquat:
I'm guessing I currently have about 50 "man" hours into this documentation pass :-)

Cattlesquat:
@uckelman Hey Joel could you check if I've done the "Disable HTML" preference right in this? The  _goal_ is that anybody who loads up a module that's naive to this new chatter gets the preference checked by default, so that old modules don't suddenly have < sign interpreted as a tag and stuff. But if a module designer says "okay this module uses the new chatter" by turning off the preference it should remember that in the module, so that when some _player_ starts a new game with it they get the module using the HTML by default. Does that mean I need to put in one of those funky GlobalOptions "Always, Never, Use User Preference" kinds of thing? I'd thought I could just put a module preference into GlobalOptions but I want to make sure that isn't going to have every single _player_ needing to explicitly uncheck the box. 

And of course I discovered this... while starting to write up some documentation for this, haha :D   Not the first thing either -- Brent and I "learned some interesting new things" about Trait Order while I was documenting that. 

So anyway I'm getting this all docced up but planning to do it in the big Doc PR. Also Brent has all that Expressions stuff coming in PR 146 that I'm guessing we'll want to get in 3.3.3 beta (perhaps that's the "cut off PR"?). 

Give me a couple days notice when you're thinking of actually starting legit 3.3.3 betas, so I can get the docs PR to a viable place and have the beta start w/ the right documentation in it? I'm also thinking of a very small "code PR" to take the ancient PDF off the Editor's Help menu and substitute a live HTML "landing page" instead.  

Lmk thoughts.

Brian

Cattlesquat:
Try out the latest version of PR #117 -- and pull up ChatLog.htm   It gets linked from GameModule.htm which I currently haven't finished "prettying up" but I'm coming to it soon (see GamePiece.htm and Map.htm for examples of what "prettied up" is starting to look like). 

You will see I've left myself a couple sentences I need to fill in once we get the final preference mode set per previous message.

Cattlesquat:
I have concluded that this DOES need a GlobalOptions so that the Module Designer is the primary controller of when module has HTML Support.

vassal-app/src/main/java/VASSAL/build/module/Chatter.java
```diff
@@ -57,57 +52,84 @@
...
+
+    for (int i = 0; i < 15; ++i) {
+      try {
+        kit.insertHTML(doc, doc.getLength(), "<br>", 0, 0, null);
+      } 
+      catch (BadLocationException ble) {
+        ErrorDialog.bug(ble);
+      } 
+      catch (IOException ex) {
+        ErrorDialog.bug(ex);
```
uckelman, line 120: If one of these exceptions happens, does this mean it's then going to happen another 14 times? (I.e., should the try/catch be outside the loop?)

Cattlesquat: changed & checked in

PR #114, opened 2020-07-22
Release vassal jar, source-jar and javadoc-jar into maven repository

yanlyub:
First version of the new Maven release process.

Things to try:
- build with `mvn clean package`, observe how the jar, source-jar and javadoc-jar end up in /vassal-app/target/
- build with `mvn clean deploy`, observe how the /snapshot/ directory is created, containing a maven repository with maven meta files and the snapshot release (jar, source-jar and javadoc-jar)
- set new version with `mvn versions:set -DnewVersion=3.3.3`, build with `mvn clean deploy`, observe how the /releases/ directory is created, containing the final release, and how these files appear in git ready for a push

Things to note:
- the maven-javadoc-plugin is configured to run with -Xdoclint:none else it would break the build due to hundreds of errors
- the `--no-module-directories` issue is fixed differently, it's in the configuration of the maven-javadoc-plugin in the `<bottom>` and the `<additionalJOption>` elements, this should make it work regardless of the JDK version

Things left to decide & fix:
- handling of version numbers, setting the version number is now done manually, this could be added to the makefile to make it call `mvn versions:set` with a version number that is maintained in the makefile
- what to do with the git-commit-id plugin and the version number generated by git
- link this whole new release process with the makefile, some makefile targets probably need adjusting to the new `mvn deploy` target


yanlyub:
The last commit is supposed to make the Makefile the single source of truth for the maven version number:

- variable `MAVEN_VERSION` contains the version number for maven, this will need to be set manually
- goal `$(LIBDIR)/Vengine.jar` sets the maven version to the value of `MAVEN_VERSION` prior to building the jar and deploying a possible release under /releases
- new goal `post-release` sets the maven version to the value of `MAVEN_VERSION` and does nothing else

I am afraid of unifying the various version number variables, there is now `VNUM`, `MAVEN_VERSION` and `VERSION`. They can probably be reduced to two or even one.

The release workflow will look roughly like this:
- set `MAVEN_VERSION` to `3.3.3-SNAPSHOT`
- work work work
- build intermediary release
- work work work
- set `MAVEN_VERSION` to `3.3.3-beta1`
- build, tag, push
- set `MAVEN_VERSION` to `3.3.3-SNAPSHOT`
- `make post-release`, push
- work work work
- set `MAVEN_VERSION` to `3.3.3-beta2`
- build, tag, push
- set `MAVEN_VERSION` to `3.3.3-SNAPSHOT`
- `make post-release`, push
- work work work
- set `MAVEN_VERSION` to `3.3.3`
- build, tag, push
- set `MAVEN_VERSION` to `3.3.4-SNAPSHOT`
- `make post-release`, push

We could also try several alpha releases without making full linux/win/mac distributions to see this cycle in action and get some first jar/src-jar/javadoc-jar artifacts into the maven repository (into /releases).

vassal-app/src/main/java/VASSAL/tools/version/Version.java
```diff
@@ -1,96 +0,0 @@
-/*
```
uckelman, line 1: It's not among the classes in the the dependency report I ran last... This is _probably_ safe. Probably.

uckelman:
One thing we need to work out yet is how to handle these:
```
VNUM:=3.3.2

GITBRANCH:=$(shell git rev-parse --abbrev-ref HEAD)
GITCOMMIT:=$(shell git rev-parse --short HEAD)
GITDESC:=$(shell git describe --tags)

ifeq ($(GITBRANCH), master)
VERSION:=$(GITDESC)
else
VERSION:=$(GITDESC)-$(GITBRANCH)
endif
```
So...

`VNUM` has to be just the dotted-decimal part; we need that for substitution into files for the Mac and Windows packages. Probably what we want then is:
```
MAVEN_VERSION:=$(VNUM)-SNAPSHOT
```
so that we don't repeat the numeric part. This also means we'll want to get rid of `GITDESC`, but  then we need to do... something else... for forming `VERSION`. It's handy to have the branch and commit in `VERSION`, because it's part of the package names and install directories. It's also handy having that version be reported by VASSAL itself, in the errorLog and in the About window. I'm not sure how that latter thing will happen now.

yanlyub:
> MAVEN_VERSION:=$(VNUM)-SNAPSHOT

Sounds good, then only the suffix of `MAVEN_VERSION` will have to be manually adjusted:
```
MAVEN_VERSION:=$(VNUM)
MAVEN_VERSION:=$(VNUM)-SNAPSHOT
MAVEN_VERSION:=$(VNUM)-alpha17
MAVEN_VERSION:=$(VNUM)-beta42
```

> part of the package names and install directories

If we want different package names and directories for what will possibly be several different releases with the same Maven snapshot version, there is probably no way around having this version number set by git, an option would be to use the Maven version and add a suffix from git to it e.g. `3.4.5-SNAPSHOT-gittag-23'

> It's also handy having that version be reported by VASSAL itself, in the errorLog and in the About window. I'm not sure how that latter thing will happen now.

That one is still coming from `Info.getVersion()` which is coming from the git.properties file written by the git-id-maven-plugin. Does this need to change?

uckelman:
I think I want this:
```
ifeq ($(shell git describe ---tags), $(VNUM))
  # we are at a release tag
  VERSION:=$(MAVEN_VERSION)
else ifeq ($(GITBRANCH), master)
  # we are somewhere else on master
  VERSION:=$(MAVEN_VERSION)-$(GITCOMMIT)
else
  # we are on some branch
  VERSION:=$(MAVEN_VERSION)-$(GITBRANCH)-$(GITCOMMIT)
endif
```
Or maybe we don't even need the special case for master?

uckelman:
And then it would be good if the version reported internally matched that.

yanlyub:
> And then it would be good if the version reported internally matched that, because it's confusing when they don't match. Do you think this makes sense?

Yes this makes sense. I think it would be best to remove the maven-git-id plugin entirely and have the makefile pass this git info to maven, e.g. `mvn deploy -Dgitversion=$(VERSION)`. I'll prepare a commit with this.

yanlyub:
Travis doesn't like the last commits.. "GitHub payload is missing a merge commit (mergeable_state: "unknown", merged: false)"

I'll rebase this on top of master, will be cleaner anyways.

uckelman:
>     * the `--no-module-directories` issue is fixed differently, it's in the configuration of the maven-javadoc-plugin in the `<bottom>` and the `<additionalJOption>` elements, this should make it work regardless of the JDK version

Was this forgotten? I still see the part for adjusting this in the Makefile.

yanlyub:
> Was this forgotten? I still see the part for adjusting this in the Makefile.

What do we want to do with `make javadoc`?

uckelman:
Have I misunderstood? I thought that there was a bit added to one of the pom.xml files for handling the javadoc?

yanlyub:
> Have I misunderstood? I thought that there was a bit added to one of the pom.xml files for handling the javadoc?

Yes and it works too, I'm just afraid of removing the `make javadoc` target entirely. The javadoc generated with maven is for the overall maven release of jar, src-jar and javadoc-jar, and I don't know what `make javadoc` is there for, I never tried it and don't know what it produces. Didn't want to break anything.

uckelman:
`make javadoc` is there so I can produce javadoc I can browse.

yanlyub:
Is the new javadoc.jar good enough for this? You would need to fish it out of vassal-app/target or from releases/ or snapshots/ and unzip it. Then we can remove the `make javadoc` goal entirely.

uckelman:
It seems to produce the same thing. I'll adjust the Makefile.

What's the procedure for publishing the JARs?

yanlyub:
Release JARs will end up in a directory structure under /releases, snapshot JARs under /snapshots. /snapshots is in .gitignore so they will stay on your local machine, /releases will be pushed to github.

Try to build a snapshot release, look what happens in /snapshots. Try a full release, remove the `-SNAPSHOT` from the maven version number and build, and look what happens in /releases and how git suggest adding and pushing it.

uckelman:
Wait, pushed, as in committed first? That's kind of gross, committing JARs to the repo, isn't it?

yanlyub:
Well, this will "misuse" github as our maven repository, and allow modules to add our .jar as a dependency. I thought that would be the easiest option. We can also look into publishing them to some external site like bintray or maven central but that would mean more work.

uckelman:
What about a second github repo?

yanlyub:
Would be possible too, we would need to make this second repo known to the makefile and have it move the release packages there, and then commit&push in that second repo.

uckelman:
What does the bit that, say, VASL, would put into their pom.xml look like when pulling from GitHub?

yanlyub:
They would need to define the repository in addition to the dependency itself:
```
<!-- Add repository -->
<repositories>
    <repository>
    <id>vassal-repo</id>
    <url>https://github.com/vassalengine/vassal/raw/master/releases</url>
    </repository>
</repositories>

<!-- Add dependency -->
<dependency>
    <groupId>org.vassalengine</groupId>
    <artifactId>vassal-app</artifactId>
    <version>3.3.3</version>
</dependency>
```

Just like we define an alternative repository (alternative to maven central) for mockito and org.netbeans:wizard here: https://github.com/vassalengine/vassal/blob/d6f416037148ea8e022959ad35507127aefa2953/vassal-app/pom.xml#L166

uckelman:
So the repository can be anything accessible by HTTP(S), right? We could put this on our own server?

yanlyub:
> So the repository can be anything accessible by HTTP(S), right? We could put this on our own server?

Yes I think so.

uckelman:
There's a Maven repo at http://www.vassalengine.org/maven now which contains a build marked 3.3.3-alpha1.

yanlyub:
`vassal-parent` needs to be there as well, `vassal-app` is a submodule and depends on it. The pom.xml of `vassal-app` inherits everything from the pom.xml of `vassal-parent`.

uckelman:
It's there now.

uckelman:
When I tried using the repo from VASL's `pom.xml` without `vassal-parent` present, it did appear to work, however.

yanlyub:
Have to be careful when trying it on the same machine where you built the vassal release, the `maven install` target installs the release artifacts into the local Maven repo. Can be that VASL's Maven got it from there and didn't have to get it from vassalengine.org/maven.

And it looks good now, my local Maven is able to import the dependency including the `vassal-parent`. I will prepare a module template and probably put it into it's own GitHub repo, it can then be copied into a new repo under github/vassalengine.

uckelman:
The VASL PR is here, if you're interested: https://github.com/vasl-developers/vasl/pull/723

PR #115, opened 2020-07-23
Fix checkstyle violations, various other code improvements

yanlyub:


vassal-app/src/main/java/VASSAL/tools/image/GeneralFilter.java
```diff
@@ -690,9 +715,9 @@ private static void apply_v(
 
         dst[dx + i*dw] =
           a << 24 |
-          (r > a ? a : r) << 16 |
-          (g > a ? a : g) <<  8 |
-          (b > a ? a : b);
+          (Math.min(r, a)) << 16 |
+          (Math.min(g, a)) <<  8 |
+          (Math.min(b, a));
```
uckelman, line 720: Don't make code changes to `GeneralFilter` without benchmarking it. This is the most heavily-optimized code we have, and small changes like  this can have a surprisingly significant impact on tiling speed.

vassal-app/src/main/java/VASSAL/counters/Embellishment.java
```diff
@@ -1295,10 +1237,6 @@ else if (suffix.isSelected()) {
           isPrefix.set(index, NEITHER);
         }
       }
-      else if (index == 0) {
```
uckelman, line 1298: Oh boy. I wonder if this was supposed to be an `index < 0` case.

vassal-app/src/main/java/VASSAL/build/module/map/KeyBufferer.java
```diff
@@ -172,7 +169,7 @@ else if (!s.isExpanded()) {
         kbuf.clear();
         
         //BR// This section allows band-select to be attempted from non-moving pieces w/o preventing click-to-select from working 
-        if (bandSelect == BandSelectType.SPECIAL && p != null && !ignoreEvent) {
```
uckelman, line 175: How do you know you don't need to check these here?

yanlyub: `bandSelect` can only have the value `SPECIAL` if:

1. the `p != null` check in line 112 resolved to `true` and `bandSelect` was set to `SPECIAL` in line 116
2. the `if (p != null && !ignoreEvent)` in line 121 resolved to `true` otherwise `bandSelect` would have been set to `NORMAL` in the else-block in line 163

Also, this whole method is way too long at 85 lines.

vassal-app/src/main/java/VASSAL/tools/image/GeneralFilter.java
yanlyub: Reverted

vassal-app/src/main/java/VASSAL/build/module/map/KeyBufferer.java
uckelman: I would like to leave this alone for context until the whole thing can be refactored.

vassal-app/src/main/java/VASSAL/counters/Embellishment.java
uckelman: Hmm. This was an `else` until 9c43d480df8.

uckelman: @BrentEaston Do you have any memory of what you were trying to do in that commit? The code in this block won't have run since then, as the test a bit above is `index >= 0`.

BrentEaston: The offending commit comment says

Fixed bug 2442887: Exception if no Layer levels and type new level name

All I managed to do is stop the else from ever being executed, which was presumably the actual correct solution - remove the else altogether.  The actual problem being the if nothing has been selected yet, index will be -1 which will blow up the code in the else.

PR #120, opened 2020-07-25
Fix checkstyle warnings, various code improvements

yanlyub:
Second and last batch of fixes for the current checkstyle warnings.

The only warning left after this is the one `finalize()` method, but that is more tricky and we can fix it later.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/image/svg/SVGRenderer.java
+++ b/vassal-app/src/main/java/VASSAL/tools/image/svg/SVGRenderer.java
@@ -162,7 +162,7 @@ public class SVGRenderer {
     px = new AffineTransform(px);
     px.scale(scale, scale);
 
-    final Rectangle2D rect = new Rectangle2D.Float(0, 0, defaultW, defaultH);
+    new Rectangle2D.Float(0, 0, defaultW, defaultH);
 
     r.addTranscodingHint(Rasterizer.KEY_WIDTH, (float) aoi.getWidth());
     r.addTranscodingHint(Rasterizer.KEY_HEIGHT, (float) aoi.getHeight());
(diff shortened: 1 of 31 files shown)


vassal-app/src/main/java/VASSAL/tools/imageop/AbstractOpImpl.java
```diff
@@ -32,7 +33,7 @@
 /**
  * An abstract representation of an operation which may be applied to an
  * {@link Image}. <code>ImageOp</code> is the base class for all such
- * operations. The results of all operations are memoized (using a
+ * operations. The results of all operations are memorized (using a
```
uckelman, line 36: `memoized` is correct.

https://en.wikipedia.org/wiki/Memoization

vassal-app/src/main/java/VASSAL/tools/image/svg/SVGRenderer.java
```diff
@@ -162,7 +162,7 @@ public BufferedImage render(double angle, double scale, Rectangle2D aoi) {
     px = new AffineTransform(px);
     px.scale(scale, scale);
 
-    final Rectangle2D rect = new Rectangle2D.Float(0, 0, defaultW, defaultH);
+    new Rectangle2D.Float(0, 0, defaultW, defaultH);
```
uckelman, line 165: This line is cruft which should have been removed when the line which used `rect` was removed. It can be deleted.

vassal-app/src/main/java/VASSAL/tools/imageop/AbstractOpImpl.java
yanlyub: Never heard of this before, we simpleton practical developers/programmers just call this "caching". My IntelliJ complained about this too. It's a good sign though, IntelliJ only starts complaining about spelling once all errors and warnings from are gone from a class.

PR #123, opened 2020-07-26
Bug 13124 remove unused dependencies

yanlyub:
This is up to `VASSAL.tools.BackgroundTask.start()` in unused-dependencies.txt

### DIFF
--- a/vassal-app/src/main/java/VASSAL/chat/ServerAddressBook.java
+++ b/vassal-app/src/main/java/VASSAL/chat/ServerAddressBook.java
@@ -77,8 +77,12 @@ public class ServerAddressBook {
   protected static final String P2P_TYPE = P2PClientFactory.P2P_TYPE;
   protected static final String P2P_MODE_KEY = P2PClientFactory.P2P_MODE_KEY;
   protected static final String P2P_SERVER_MODE = P2PClientFactory.P2P_SERVER_MODE;
+
+  /**
+   * @deprecated will be removed without replacement
+   */
+  @Deprecated
   protected static final String P2P_CLIENT_MODE = P2PClientFactory.P2P_CLIENT_MODE;
-  // protected static final String PRIVATE_TYPE = PrivateClientFactory.PRIVATE_TYPE;
   protected static final String TYPE_KEY = ChatServerFactory.TYPE_KEY;
   protected static final String DESCRIPTION_KEY = "description"; //$NON-NLS-1$
   protected final int LEAF_ICON_SIZE = IconFamily.SMALL;
@@ -91,7 +95,7 @@ public class ServerAddressBook {
   private DefaultListModel<AddressBookEntry> addressBook;
   private AddressBookEntry currentEntry;
   private boolean enabled = true;
-  private PropertyChangeSupport changeSupport = new PropertyChangeSupport(this);
+  private final PropertyChangeSupport changeSupport = new PropertyChangeSupport(this);
   private static ServerAddressBook instance;
   private static String localIPAddress;
   private static String externalIPAddress;
@@ -143,7 +147,7 @@ public class ServerAddressBook {
   }
 
   private static String discoverMyIpAddressFromRemote() throws IOException {
-    String theIp = null;
+    String theIp;
     HttpRequestWrapper r = new HttpRequestWrapper("http://www.vassalengine.org/util/getMyAddress"); //$NON-NLS-1$
     List<String> l = r.doGet(null);
     if (!l.isEmpty()) {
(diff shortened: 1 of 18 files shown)


vassal-app/src/main/java/VASSAL/chat/ServerAddressBook.java
```diff
@@ -1057,73 +1005,6 @@ protected void setAdditionalProperties(Properties props) {
...
-//      props.setProperty(TYPE_KEY, PRIVATE_TYPE);
-//    }
-//
-//    protected void setAdditionalProperties(Properties props) {
-//      serverIp.setText(props.getProperty(PrivateClientFactory.URL));
-//      serverPort.setText(props.getProperty(PrivateClientFactory.PORT));
-//    }
-//
-//  }
-
```
uckelman, line 1126: Don't remove this commented out bit. We might need this shortly.

vassal-app/src/main/java/VASSAL/chat/ServerAddressBook.java
```diff
@@ -77,8 +77,6 @@
   protected static final String P2P_TYPE = P2PClientFactory.P2P_TYPE;
   protected static final String P2P_MODE_KEY = P2PClientFactory.P2P_MODE_KEY;
   protected static final String P2P_SERVER_MODE = P2PClientFactory.P2P_SERVER_MODE;
-  protected static final String P2P_CLIENT_MODE = P2PClientFactory.P2P_CLIENT_MODE;
```
uckelman, line 80: Why is this being removed?

vassal-app/src/main/java/VASSAL/tools/BackgroundTask.java
```diff
@@ -17,45 +17,12 @@
...
 
 /**
- * Utility task for starting a thread that performs one task,
- * {@link #doFirst}, then queues another another
- * task, {@link #doLater}, for the Event Handler thread to execute
- * This is basically a simple version of Sun's SwingWorker class.
- *
  * @deprecated Use {@link SwingWorker} now that we ship the JAR for it.
  */
 @Deprecated
```
uckelman, line 25: Why delete all the insides of this class but not the whole class? (Is something depending on the class, but none of it's members?)

vassal-app/src/main/java/VASSAL/chat/ServerAddressBook.java
yanlyub: It is not used anywhere

vassal-app/src/main/java/VASSAL/tools/BackgroundTask.java
yanlyub: It's weird but yes it looks like it, else the class itself would appear in the unused-dependencies list, like other classes which don't have any dependents anymore.

vassal-app/src/main/java/VASSAL/chat/ServerAddressBook.java
yanlyub: I put it back with a deprecated marker, if modules extend this class they can define that field themselves.

vassal-app/src/main/java/VASSAL/tools/BackgroundTask.java
uckelman: `BackgroundTask` is used by `CASL.VASL.VASLThread` in the VSQL module:

https://sourceforge.net/p/vassalengine/svn/HEAD/tree/VSQL-src/trunk/CASL/VASL/VASLThread.java

I think that's current? (@BrentEaston would know.) I don't see how that can possibly compile after gutting `BackgroundTask`, since it won't have a `start()` to call.

yanlyub: We have a detected dependency on the class itself and it's constructor by VASLThread:
https://github.com/vassalengine/vassal/blob/6f8c7f08dc3801669b07c07195e6713a6c9762a1/vassal-app/src/test/resources/depreport/depreport.txt#L155

But not on the methods inside it, that's why they appear among unused dependencies: https://github.com/vassalengine/vassal/blob/6f8c7f08dc3801669b07c07195e6713a6c9762a1/vassal-app/src/test/resources/depreport/unused-dependencies.txt#L126

Either the module we did the dependency analysis for is a different version from the SVN trunk, or it's an issue with the analysis tool, maybe it cannot detect method calls on anonymous types?

yanlyub: Either way, I reverted `BackgroundTask` to how it was before.

BrentEaston:
This one is out of date now due to my work on the graceful deprecation stuff. Leave this one here for now. I will apply the changes to my deprecation branch so these will get included when you get to it tomorrow Joel. Don't bother resolving the conflicts Yan, I'll do that in my branch.

BrentEaston:
I have included these changes in PR#166, this one is good to be closed now.

PR #127, opened 2020-07-26
Refactored WizardSupport

yanlyub:
Minor refactoring.

`PredefinedSetup.isMenu()` is already checked in `PredefinedSetup.addPredefinedSetup()`, no need to check it again in the renderer.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
@@ -46,6 +46,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Vector;
 
 import javax.swing.Action;
 import javax.swing.Box;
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
```diff
@@ -533,18 +535,15 @@ public void actionPerformed(ActionEvent e) {
...
-            c.setText(value == null ? "" : value.toString());
-          }
+        public Component getListCellRendererComponent(JList<?> list, Object value, int index, boolean isSelected, boolean cellHasFocus) {
+          final JLabel c = (JLabel) super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus);
+          final String labelText =
+            value instanceof PredefinedSetup
+              ? ((PredefinedSetup) value).getConfigureName()
+              : value == null
+                ? ""
+                : value.toString();
```
uckelman, line 545: Easier to read:
```
value instanceof PredefinedSetup ?
  ((PredefinedSetup) value).getConfigureName() :
  (value == null ? "" : value.toString());
```

yanlyub: Arguable, but the ternary operator looks this weird in the rest of the code so I'll change it to at least make it consistently weird.

PR #128, opened 2020-07-27
Code formatting

yanlyub:
Fixes a bunch of checkstyle warnings.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Inventory.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Inventory.java
@@ -341,8 +341,8 @@ public class Inventory extends AbstractConfigurable
 
                 piece.draw(
                   g,
-                  (int)(-r.x * os_scale),
-                  (int)(-r.y * os_scale),
+                  (int) (-r.x * os_scale),
+                  (int) (-r.y * os_scale),
                   c,
                   pieceZoom * os_scale
                 );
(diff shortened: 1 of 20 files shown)


vassal-app/src/main/java/VASSAL/build/module/Inventory.java
```diff
@@ -341,8 +341,8 @@ public void paintIcon(Component c, Graphics g, int x, int y) {
 
                 piece.draw(
                   g,
-                  (int)(-r.x * os_scale),
-                  (int)(-r.y * os_scale),
+                  (int) (-r.x * os_scale),
+                  (int) (-r.y * os_scale),
```
uckelman, line 345: I hadn't realized we'd enabled this one. Could we turn this back off?

vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
```diff
@@ -231,12 +231,12 @@ public void setup(boolean gameStarting) {
         p.translate(r.x, r.y);
       }
       if (placeNonStackingSeparately()) {
-        for (int i=0;i<s.getPieceCount();++i) {
+        for (int i=0; i<s.getPieceCount(); ++i) {
```
uckelman, line 234: Spacing around `=`, `<` here.

uckelman:
Otherwise looks ok.

vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
yanlyub: Will be checked by rule "WhitespaceAround" that we will enable later on. For now we have enabled "WhitespaceAfter".

PR #133, opened 2020-07-27
13218 - Inform a piece "that wants to know" where the context menu click was

Cattlesquat:
http://www.vassalengine.org/forum/viewtopic.php?f=2&t=12170

This was getting mentioned again in the forums, and I have a really lightweight solution I'd been using in a custom class. It seems like something that many module designers might like to do.

What it actually does: when something gets executed from a right-click menu of a piece, then IF the piece has dynamic properties for "ClickedX" and "ClickedY" they are updated with the map coordinates of the most recent click.

Use Case:
(1) I have a "big" piece (e.g. a picture of a province, or a battleship or whatever)
(2) I right click on it, and a context menu comes up
(3) Maybe what I pick from the menu is "Put a tank here"
(4) Piece that's processing the click can retrieve the map coordinates where the right-click took place, and use that to allow "placing the tank" to be at the location of the click rather than always at one exact spot. 

Why I did it that way: lightweight and doesn't create any "Command" traffic _unless_ the target piece actually has those properties.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
@@ -37,7 +37,10 @@ import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 
 import VASSAL.build.Buildable;
+import VASSAL.build.GameModule;
 import VASSAL.build.module.Map;
+import VASSAL.command.ChangeTracker;
+import VASSAL.command.Command;
 import VASSAL.counters.Deck;
 import VASSAL.counters.EventFilter;
 import VASSAL.counters.GamePiece;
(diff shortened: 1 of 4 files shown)


yanlyub:
This removes the "Search" feature that we added in another PR, looks like something got messed up in git

Cattlesquat:
Whoa, weird.

Cattlesquat:
@yanlyub Somehow my copy of master on github seems to be fucked up -- we had some funky PR merging/reverting last night. My normal commands don't seem to be restoring it.

Cattlesquat:
@yanlyub Take a look at https://github.com/Cattlesquat/vassal and see if you can figure out what I can do to make all that "4 commits ahead of master" go away and be back to just pointing at master again?

Cattlesquat:
I don't know how to reset an upstream repo

Cattlesquat:
Got it. StackOverflow to the rescue :)

Closing this one, reopening a clean one.

PR #134, opened 2020-07-27
13218 Allow a piece to find out its context menu click location

Cattlesquat:
http://www.vassalengine.org/forum/viewtopic.php?f=2&t=12170

This was getting mentioned again in the forums, and I have a really lightweight solution I'd been using in a custom class. It seems like something that many module designers might like to do.

What it actually does: when something gets executed from a right-click menu of a piece, then IF the piece has dynamic properties for "ClickedX" and "ClickedY" they are updated with the map coordinates of the most recent click.

Use Case:
(1) I have a "big" piece (e.g. a picture of a province, or a battleship or whatever)
(2) I right click on it, and a context menu comes up
(3) Maybe what I pick from the menu is "Put a tank here"
(4) Piece that's processing the click can retrieve the map coordinates where the right-click took place, and use that to allow "placing the tank" to be at the location of the click rather than always at one exact spot.

Why I did it that way: lightweight and doesn't create any "Command" traffic unless the target piece actually has those properties.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
@@ -37,7 +37,10 @@ import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 
 import VASSAL.build.Buildable;
+import VASSAL.build.GameModule;
 import VASSAL.build.module.Map;
+import VASSAL.command.ChangeTracker;
+import VASSAL.command.Command;
 import VASSAL.counters.Deck;
 import VASSAL.counters.EventFilter;
 import VASSAL.counters.GamePiece;
--- a/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
@@ -267,6 +267,16 @@ public class MenuDisplayer extends MouseAdapter implements Buildable {
 
     // It is possible for the map to close before the menu is displayed
     if (map.getView().isShowing()) {
+
+      // Inform the piece where player clicked, if it wants to know.  
+      final ChangeTracker ch = new ChangeTracker(p);
+      p.setProperty("ClickedX", e.getPoint().x);
+      p.setProperty("ClickedY", e.getPoint().y);
+      Command comm = ch.getChangeCommand();
+      if ((comm != null) && !comm.isNull()) {
+        GameModule.getGameModule().sendAndLog(comm);
+      }      
+
       popup.show(map.getView(), pt.x, pt.y);
     }
 


uckelman:
It is eluding me how this works. Where is the code for handling this on the side of the piece?

Cattlesquat:
No code needed on the piece side. Piece just defines a Dynamic Property trait for ClickedX and ClickedY and it will automatically be filled in with the click location whenever there is a click on the piece. Part of what makes this lightweight is that since it only happens if the piece defines these "mutable properties", there is no "Command" generated, no game-state change, if the feature is not presently in use (unlike a perma-expose-this-property-always type of situation).

vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
```diff
@@ -267,6 +270,16 @@ public void popupMenuWillBecomeVisible(PopupMenuEvent evt) {
...
+
+      // Inform the piece where player clicked, if it wants to know.  
+      final ChangeTracker ch = new ChangeTracker(p);
+      p.setProperty("ClickedX", e.getPoint().x);
+      p.setProperty("ClickedY", e.getPoint().y);
+      Command comm = ch.getChangeCommand();
+      if ((comm != null) && !comm.isNull()) {
+        GameModule.getGameModule().sendAndLog(comm);
+      }      
+
```
BrentEaston, line 282: Hi Brian,

Sorry, I am only getting to review this now. This change as implemented has a few problems.

1. If the designer does not create Dynamic Properties ClickedX and ClickedY, the change still appears to work because ClickedX and ClickedY will end up stored in the scratch-pad property HashMap in BasicPiece. However, you end up with all the same problems as the backXXX variables used to have:- Not sent to clients or log or save files, undo does not work.

Because it appears to work on a single client, the designer may 'forget' or decide not to create the Dynamic Properties.

As an alternative, you could make ClickedX and ClickedY Persistent properties using the interface I created to fix SendToLocation by doing something like this:

Command comm  = new NullCommand();
comm = setPersistentProperty(BasicPiece,CLICKED_X, String.valueOf(e.getPoint().x));
comm.append (setPersistentProperty(BasicPiece,CLICKED_Y, String.valueOf(e.getPoint().y));
if ((comm != null) && !comm.isNull()) {

2. However, there is a deeper problem in that whether using user supplied Dynamic Properties or persistent properties, this solution generates a Command each time the player right-clicks on a piece, regardless of whether or not they actually select a menu option. 

When you replay the log, there will appear to be empty steps to click through. 

Even if you do select a Command, this will generate 2 steps, an apparently empty one when the Clicked props are set, followed by the actual menu selection execution. Currently, a single menu selection only ever generate a single log step. Now it will generate 2.

The Command generated by the KeyCommand selected on the menu needs to be appended to the Command to set the properties,but only after the menu item is selected.
 
This one is tricky because the individual KeyCommands in the popup handle their own activation directly through KeyCommand.actionPerformed().  

You may be able to mediate this through the KeyBuffer. Set the ClickedX and ClickedY into the KeyBuffer in MenuDisplayer, then use them in KeyCommand.actionPerformed() which already accesses the KeyBuffer to get the list of currently selected units.

There are cases where actionPerformed is call directly on KeyCommands not as a result of the MenuDisplayer, so you will also need to determine that the actionPerformed is actually a result of a MenuDisplayer selection. Tricky, but I suspect the KeyBuffer will only have units in it at MenuDisplayer time.

Cattlesquat: Ooooo. Didn't know about the scratchpad property. Is there a way to check definitively for the existence of the DynamicProperty in advance of sending the SetProperty?

BrentEaston: Not really, because the same Command may be applied to a whole Stack by KeyBuffer. On an individual Counter, you would need to run down the trait stack and check each one. Just change them to Persistent properties, that solves the problems and is simpler for the designer. No need to add 2 more traits to the stack just to access ClickedX/Y.

BrentEaston: I would definitely approach it by 
1. adding clickPoint to KeyBuffer and have menuDisplayer set it.
2. Modify KeyBuffer.keyCommand to set the Persistent properties from clickPoint, then append p.keyEvent(stroke) for each piece. The handles the case where a menu command is applied to a selection of units.

I think that's all you need. All Keycommands issued from MenuDisplayer (and only KeyCommand issued from the Menudisplayer) have global set to true, so you can do all the work in KeyBuffer

Cattlesquat: Please check #156 -- it seems to be running correctly but give it a once-over.

PR #136, opened 2020-07-27
Fixed compile warnings, minor refactoring

yanlyub:


vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -175,13 +177,7 @@ public void setPiece(GamePiece piece) {
...
     }
 
-    if (piece == null) {
-      inUseModel.insertElementAt(new BasicPiece(), 0);
-    }
-    else {
-      inUseModel.insertElementAt(piece, 0);
-    }
-
+    inUseModel.insertElementAt(Objects.requireNonNullElseGet(piece, BasicPiece::new), 0);
```
uckelman, line 180: Isn't `piece == null ? new BasicPiece() : piece` both shorter and easier to read?

vassal-app/src/main/java/VASSAL/chat/DynamicClientFactory.java
```diff
@@ -31,11 +31,9 @@ public ChatServerConnection buildServer(Properties param) {
     String url = param.getProperty(URL);
     DynamicClient client = url == null ? new DynamicClient() : new DynamicClient(url);
     if (param.containsKey(DYNAMIC_TYPE)) {
-      //param = new Properties(param);
       // Force the type of the dynamically created server proxy
       param.setProperty(ChatServerFactory.TYPE_KEY, param.getProperty(DYNAMIC_TYPE));
     }
-    client.setOverrides(param);
```
uckelman, line 38: Why remove this?

yanlyub: The method doesn't do anything and we marked it as deprecated: https://travis-ci.com/github/vassalengine/vassal/builds/177365837#L701

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
yanlyub: My IDE suggests this "modern" way of doing this but either way works for me. This is probably meant to be use with a static import of the method, then it would just be `model.insertElementAt(requireNonNullElseGet(piece, BasicPiece::new), 0)`

PR #137, opened 2020-07-28
13219 & 12946 - (a) KeyNamer.java referenced in buildFile and translates key names  (b) Restore any missing fundamental classes to buildFile

Cattlesquat:
"While we're at it" I also made it restore wrongly-deleted classes to the buildFile (e.g. PlayerRoster).

This PR is branched from #110 

@BrentEaston When you come online could you give this a quick eyeball and see if it looks like I tagged all the right bases?


Cattlesquat:
@BrentEaston It's just the last two commits in the list that have any meaning. Everything else is just because I branched this from the 13197 PR

yanlyub:
If you turn the static method upside down, you could build the final string from left to right in a `StringBuffer` then at the end create the final string and return it. That would be a tiny bit more optimal than allocating all the intermediary string objects and feeding them to the GC.

Design-wise, for the sake of unit tests the static method should be a regular method otherwise it cannot be mocked, also it has nothing to do with the rest of the class and should be in it's own class, but I get why it has to be that way :disappointed:

Cattlesquat:
Hi Yan,

Take a look -- I switched it over to StringBuilder. Please check if I got all the nuances of when to replace/insert/capitalize right. 

Also if you can think of a good way to make the method non-static, please do stick up a PR-into-PR?

Brian

yanlyub:
Right, `StringBuilder` has the `insert()` method, rarely used but that works too, and no need to turn the method upside down.

The static method, I saw this comment "Exists as a buildable (buildFile element) so that it can be overridden by custom classes", that qualifies the method as a necessary part of the class, maybe a compromise would be to just make it a regular non-static method, the callers could then do a `String s = new KeyNamer().getKeyString(aKeyStroke);`, modules could still override the method, unit tests would be able to mock it, and design wise it wouldn't look like an "alien element" inside that class anymore.

Cattlesquat:
So in this case, the method is actually currently called from "exactly one place" (the place that is getting requests funneled in from all over):

`public static String getString(KeyStroke k)`  in `VASSAL.configure.HotKeyConfigurer`

So that method is itself static, but then _that_ method is getting from about 8 places, I think all non-static.

Does this suggest any good methods for correcting/improving to you?

yanlyub:
Well yes, but `HotKeyConfigurer.getString()` is already there and is static and who knows how many modules use it in a static way, so it cannot be changed without a lengthy process of deprecating it, analyzing whether any modules use it, then replacing it with a non-static version. But the new code can be made non-static, which will improve maintainability in the long run.

vassal-app/src/main/java/VASSAL/configure/HotKeyConfigurer.java
```diff
@@ -27,7 +27,12 @@
 import javax.swing.JTextField;
 import javax.swing.KeyStroke;
 
+import org.apache.commons.lang3.StringUtils;
```
uckelman, line 30: Unused?

vassal-app/src/main/java/VASSAL/build/module/Chatter.java
```diff
@@ -50,6 +50,8 @@
 import javax.swing.text.View;
 import javax.swing.text.WrappedPlainView;
 
+import org.apache.commons.lang3.StringUtils;
```
uckelman, line 53: Unused?

vassal-app/src/main/java/VASSAL/build/module/KeyNamer.java
```diff
@@ -0,0 +1,111 @@
+/*
```
uckelman, line 1: Missing copyright line.

vassal-app/src/main/java/VASSAL/build/module/KeyNamer.java
```diff
@@ -0,0 +1,111 @@
...
+      break;
+    default:
+      break;
+    }
+    if (sb.toString().isEmpty() || sb.toString().contains("Keys.")) { //$NON-NLS-1$
+      sb.append(KeyEvent.getKeyText(code)); 
+    }
+
+    final int mods = SwingUtils.genericToSystem(k).getModifiers();
+    if ((mods & KeyEvent.SHIFT_DOWN_MASK) > 0) {
```
uckelman, line 93: Technically you could have a mask on the high bit and get a result less than 0 indicating that the masked bit was on, because Java foolishly lacks unsigned integral types... So these should be `!= 0`.

uckelman (changes requested):

uckelman (approved):

PR #140, opened 2020-07-29
Remove unnecessary modifiers from interfaces

yanlyub:
There are still lots of unnecessary `public` modifiers in almost all interfaces, even though everything is `public` by default in an interface, anything else would make no sense, an interface is a contract with the outside world and private/protected/package-visibility members don't belong in an interface.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/AutoConfigurable.java
+++ b/vassal-app/src/main/java/VASSAL/build/AutoConfigurable.java
@@ -73,7 +73,7 @@ public interface AutoConfigurable extends Configurable {
    */
   public VisibilityCondition getAttributeVisibility(String name);
 
-  public static class Util {
+  public class Util {
     public static void buildAttributes(Element e, AutoConfigurable parent) {
       if (e != null) {
         NamedNodeMap n = e.getAttributes();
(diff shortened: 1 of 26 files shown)


vassal-app/src/main/java/VASSAL/build/AutoConfigurable.java
```diff
@@ -73,7 +73,7 @@
    */
   public VisibilityCondition getAttributeVisibility(String name);
 
-  public static class Util {
+  public class Util {
```
uckelman, line 76: Won't this change make the inner class depend on the outer class when it doesn't need to? Is that binary compatible?

yanlyub: There is no outer class, it's an interface, completely different thing. Classes inside interfaces are by default static, they cannot be non-static. This is binary compatible, this is not even a change, just removal of redundant information.

uckelman: Ah. I didn't have enough context above this to see that the containing block was an interface.

uckelman: Looks like all the others are also that way.

yanlyub: Yes I'm too missing a feature that lets me open the whole file from the PR view, I need it often and it's very inconvenient to open the PR and the changed file locally or navigate to it's version on GitHub in another browser tab.

PR #141, opened 2020-07-29
Bug 13178 - GameRefresher fixes - Version 2

BrentEaston:
Updated with suggestions, based on latest source


yanlyub:
"modifier out of order with the JLS suggestions", JLS suggests `private final` or `private static final`, else checkstyle breaks the build. Visibility/access modifier is the most important and comes first, `static` is 2nd important, and `final` almost unimportant and comes last.

BrentEaston:
Issues fixed. Forgot to test Maven build locally first.

PR #142, opened 2020-07-30
Bug 13014: Module editor frequently stuck thinking module isn't writeable

uckelman:
@Cattlesquat: Does this fix any of the problems you noted?

(Please do not use this for anything you can't lose. I have added unit tests for ZipArchive, which both the old and new ZipArchive pass, and I have tried various things successfully---but I could imagine there still being some problems.)

### DIFF
--- a/vassal-app/pom.xml
+++ b/vassal-app/pom.xml
@@ -235,6 +235,7 @@
 				<configuration>
 					<enableAssertions>true</enableAssertions>
 					<testFailureIgnore>false</testFailureIgnore>
+          <trimStackTrace>false</trimStackTrace>
 					<excludes>
 						<exclude>**/AllTests.java</exclude>
 					</excludes>
--- a/vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
+++ b/vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
@@ -1,6 +1,5 @@
 /*
- *
- * Copyright (c) 2000-2008 by Rodney Kinney, Joel Uckelman
+ * Copyright (c) 2000-2020 by Rodney Kinney, Joel Uckelman
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Library General Public
@@ -252,7 +251,6 @@ public class ArchiveWriter extends DataArchive {
 
       archive = new ZipArchive(tmp, filename);
       archiveName = filename;
-      archive.flush();
 
       tmp.revert();
       tmp.close();
--- a/vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
+++ b/vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
@@ -452,7 +452,9 @@ public class ZipArchive implements FileArchive {
 
     // Replace old archive with temp archive.
     try {
-      FileUtils.forceDelete(archiveFile);
+      if (archiveFile.exists()) {
+        FileUtils.forceDelete(archiveFile);
+      }
       FileUtils.moveFile(tmpFile, archiveFile);
     }
     catch (IOException e) {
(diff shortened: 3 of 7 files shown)


yanlyub:
Hamcrest asserts are generally recommended for new test code.

Cattlesquat:
It seems to at least restore "3.2.17" functionality:
(1) I was able to save a new module w/o error
(2) Trying to re-save it when I had the vmod open in 7-zip produced the "correct" error w/o e.g. asking me to send an error report, etc.
(3) Was able to load a "big module" and then save it out with a different name successfully.

So then I went for the "bonus test" and tried having it save For the People 3.2 after I'd already started the module in the player. This still failed (even though I theoretically "should" be able to save the module), but it failed in the same way 3.2.17 and gave the "correct" error.

So we haven't fixed the deeper bug of "why is the module sometimes being held 'open' when it shouldn't be" in all cases, but at least this is back to what appears to be a 3.2.17 level of mostly-okay.

Happy to do more testing, or more specific testing, if you have any in mind.

BrentEaston:
Just an aside, It can be dangerous to save a module from the Editor while a game is open if you use custom code that plays around with component states.

Cattlesquat:
Ooh interesting - what's the flavor of failure in that case? Component states just not being what we expect? Or actual garbage into the module? 

My own custom code tends to be pretty light touch and uses "standard" methods (I.e. Commands) to change properties, but it's good to know where the spiky pits are...

*Sent from the wild frontier of mobile games :-)

BrentEaston:
My GTS modules have a set of Chart Maps for holding division reinforcement cards, 1 for each Division. The big ones might have 20+ of these, each with it's own tab. Smaller scenarios might only need a couple of these, so I hide the other 18. If I save the module at this point, the 18 'hidden' chart maps don't get written to the buildfile and disappear from the module. I really must go back and work out some better way of doing that, I probably did something dodgy :)

uckelman:
> Hamcrest asserts are generally recommended for new test code.

How would those look with those? Maybe make a PR?

uckelman:
@Cattlesquat I didn't specifically set out to fix any of the problems you mentioned, though I though there was a chance that swapping out the old `ZipFile` implementation could fix one or more of them. I'll try to run down the problems you're seeing next.

yanlyub:
> How would those look with those?

```
import static org.hamcrest.MatcherAssert.assertThat; 
import static org.hamcrest.Matchers.*;

assertThat(actual, is(expected));
assertThat(actual, equalTo(expected));
assertThat(aBoolean, is(true));
assertThat(actualByteArray, is(expectedByteArray));
assertThat(actual, is(anyOf(expected1, expected2)));
assertThat(collection, hasItem(item));
assertThat(collection, containsInAnyOrder(elem1, elem2, elem3));
```

http://hamcrest.org/JavaHamcrest/tutorial

Cattlesquat:
Sad but VERY prudent

uckelman:
Here's one example: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8229887

yanlyub:
> Here's one example: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8229887

This one is closed/fixed, it does not count :smile:

yanlyub:
Fixed in 11.0.5, we're at 11.0.8 by now.

Cattlesquat:
Well yeah but jackwagons will be jackwagons. My game developer spidey sense says there will be plenty of never-bothered-to-upgrade people. I guess we don't have the problem for Windows/Mac, so I guess it's a question of "How Many Linux Jackwagons Does It Take To Ruin Our Saturdays"?

Cattlesquat:
This would be one of those times where if we'd been collecting telemetry all along we'd be able to answer that question.

Cattlesquat:
Not that I'm volunteering to be in charge of VassalEngine.Org's new "Privacy Policy" etc, etc, etc. NOT IT! (he says, finger already on nose)

PR #146, opened 2020-07-30
Expression Builder/Beanshell improvements

BrentEaston:
- Add ? to Expression Builder
 - Add existing Java Math functions to Expression Builder
 - Add existing Java String functions to Expression Builder
 - Create random and isRandom functions and add to Expression Builder
 - Create Sum and Count functions and add to Expression Builder
 - Add additional 'hints' that can be displayed below function input in Expression Builder
 - Clean up PieceDefiner to Java 11 standards

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/BeanShellFunctionMenu.java
+++ b/vassal-app/src/main/java/VASSAL/configure/BeanShellFunctionMenu.java
@@ -174,7 +174,7 @@ public class BeanShellFunctionMenu extends JPopupMenu {
     }
     else {
       final StringConfigurer result = new StringConfigurer(null, "", "");
-      new FunctionBuilder(result, (JDialog) configurer.getTopLevelAncestor(), op, desc, parmDesc, target, hints, options).setVisible(true);
+      new FunctionBuilder(result, (JDialog) configurer.getTopLevelAncestor(), op, desc, parmDesc, target, hints, options, configurer.getSelectedText()).setVisible(true);
       if (result.getValue() != null && result.getValueString().length() > 0) {
         configurer.insertName(result.getValueString());
       }
--- a/vassal-app/src/main/java/VASSAL/configure/BeanShellFunctionMenu.java
+++ b/vassal-app/src/main/java/VASSAL/configure/BeanShellFunctionMenu.java
@@ -1,5 +1,7 @@
 package VASSAL.configure;
 
+import static VASSAL.configure.BeanShellExpressionConfigurer.Option;
+
 import VASSAL.build.AbstractBuildable;
 import VASSAL.build.AbstractConfigurable;
 import VASSAL.build.Buildable;
@@ -29,7 +31,6 @@ import javax.swing.JPopupMenu;
 public class BeanShellFunctionMenu extends JPopupMenu {
 
   protected static final String[] SUM_COUNT_HINTS = new String[]{"WARNING - This function needs to scan many pieces and is relatively slow. Use sparingly", "$property$ variables may be used in the Match Expression and will be evaluated on the source piece.", "See the Reference Manual for detailed usage."};
-  protected static final String[] STRING_HINTS = new String[]{"This function can only be used on a String or Function return value.", "Use GetProperty(\"propertyName\").function() to apply this function to a Property value."};
 
   protected static int maxScrollItems = 0;
   protected BeanShellExpressionConfigurer configurer;
@@ -116,15 +117,15 @@ public class BeanShellFunctionMenu extends JPopupMenu {
 
     final JMenu stringMenu = new JMenu("String");
     addFunction(stringMenu, ".length", "Return the length of a string", new String[] {"Target String"}, "()");
-    addFunction(stringMenu, ".contains", "Return true if string contains the specified search string", new String[] { "Target String", "Search String" }, "(string)", STRING_HINTS);
-    addFunction(stringMenu, ".startsWith", "Return true if string starts with the specified search string", new String[] { "Target String", "Search String" }, "(string)", STRING_HINTS);
-    addFunction(stringMenu, ".endsWith", "Return true if string ends with the specified search string", new String[] { "Target String", "String String" }, "(string)", STRING_HINTS);
-    addFunction(stringMenu, ".matches", "Return true if string matches the specified Regular Expression", new String[] { "Target String", "Regular Expression" }, "(regExpr)", STRING_HINTS);
-    addFunction(stringMenu, ".indexOf", "Return the position of the specified search string", new String[] { "Target String", "Search String" }, "(string)", STRING_HINTS);
-    addFunction(stringMenu, ".lastIndexOf", "Return the last position of the specified search string", new String[] { "Target String", "Search String" }, "(string)", STRING_HINTS);
-    addFunction(stringMenu, ".substring", "Return substring starting at a given position, up to the end of the string", new String[] { "Target String", "Starting position" }, "(start)", STRING_HINTS);
-    addFunction(stringMenu, ".substring", "Return substring between positions", new String[] { "Target String", "Starting position", "End Position" }, "(start, end)", STRING_HINTS);
-    addFunction(stringMenu, ".replace", "Replace one substring with another", new String[] { "Target String", "String to find", "String to replace" }, "(old, new)", STRING_HINTS);
+    addFunction(stringMenu, ".contains", "Return true if string contains the specified search string", new String[] { "Target String", "Search String" }, "(string)");
+    addFunction(stringMenu, ".startsWith", "Return true if string starts with the specified search string", new String[] { "Target String", "Search String" }, "(string)");
+    addFunction(stringMenu, ".endsWith", "Return true if string ends with the specified search string", new String[] { "Target String", "String String" }, "(string)");
+    addFunction(stringMenu, ".matches", "Return true if string matches the specified Regular Expression", new String[] { "Target String", "Regular Expression" }, "(regExpr)");
+    addFunction(stringMenu, ".indexOf", "Return the position of the specified search string", new String[] { "Target String", "Search String" }, "(string)");
+    addFunction(stringMenu, ".lastIndexOf", "Return the last position of the specified search string", new String[] { "Target String", "Search String" }, "(string)");
+    addFunction(stringMenu, ".substring", "Return substring starting at a given position, up to the end of the string", new String[] { "Target String", "Starting position" }, "(start)");
+    addFunction(stringMenu, ".substring", "Return substring between positions", new String[] { "Target String", "Starting position", "End Position" }, "(start, end)");
+    addFunction(stringMenu, ".replace", "Replace one substring with another", new String[] { "Target String", "String to find", "String to replace" }, "(old, new)");
 
     final JMenu randomMenu = new JMenu("Random");
     addFunction(randomMenu, "Random", "Return a random whole number between 1 and a number (inclusive)", new String[] { "Highest number" }, "(x)");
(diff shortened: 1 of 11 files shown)


Cattlesquat:
I put a bunch of these through their paces with one of my little test modules -- looking REAL good!
* Suggest more consistency in guidance from the builder dialog boxes. For example the ? ternary let me put in all 3 fields (good!) but the string functions just inserted a .contains() asking for the parameter but not the original string (confusing).
* I couldn't figure out what Compare() is actually supposed to do (how is it different from ==)? I tried to "compare" two strings like it said I could and it complained of "bad data". 
* Love how the "Math" functions don't require the GetProperty wrapper around property names. MUCH easier to use than the String stuff. Any way we could get the String stuff working like that too? Because StateList.contains(CurrentZone) is a helluva lot easier to read (and type, and understand) than GetProperty("StateList").contains(GetProperty("CurrentZone"))
* Did a smoke test on the random number generator, nothing massive, but it seemed good.

I'll put "getting this all fully documented in the new Expression.htm docs" on my to-do-soon list!

BrentEaston:
Thanks Brian.

Hmm. I will have a think about the consistency issue.  Probably what we want is that the String functions do have an input field for the target string as you suggest. But it should be pre-populated with any part of the expression that you have selected before hitting the Insert button and the result will replace the selection on return.

I think the answer for Compare is to just remove it. I think I actually put it in very early on before I modified the Beanshell source to support == directly on Strings. Also the fact that it doesn't actually work :)

The problem with the String functions is because the function is being applied directly to the target String, rather than taking the String as an argument like the Math functions. You can use property names directly as the arguments of the string function (e.g. GetProperty("Army").contains(PlayerSide)). I will delve in deeper and try and work out what is going on. Just a thought, I wonder if (propname).contains("x") works? 

I use the inbuilt RNG and the recommended library calls to correctly generate integers in a range, so should be no dramas there.

Cattlesquat:
Well I will be damned! 

`(States).contains(State) ? "Yes!" : "No"`

Was able to determine that Arkansas is a state and that Poop is not!

Cattlesquat:
So if your Stringmotizer asked me both ends, and then automatically added parentheses-that's-probably-brackets-to-you around the first argument if it didn't see a GetProperty wrapping, that might be huge. 

One other suggestion is the popup boxes (which already contain some nice guidance) could each give one example of a "properly formed sample specimen" of that time?

BrentEaston:
`(States).contains(State) ? "Yes!" : "No"`

its got something to do with the way I treat the property name is Beanshell. They are basically undefined Java variables that I then set to the value of the corresponding property. But States.contains(State) is valid Java, so it shouldn't choke. I should be able to work that out. The fallback plan will be to add () of the target string is a simple property name.

Could you give me an example of the 'guidance' you mean? I'm not sure at which level you mean? because what you type into the Function argument boxes is not necessarily what gets added into the previous step. For example, quotes in Strings and property match expressions get escaped and the Property Match Expression gets Beanshell braces wrapped around it. [Note the Sum and Count functions will (should!) work with Classic syntax as well, but no need to advertise that.]

Cattlesquat:
1. The most effective part of 'guidance' I think is asking me for all the parameters on the Strings (and I think you already do on the Random/Math functions). And you ask for all the parameters for the ternary, which is good. Obviously when I wrote the first one I was still wondering what Compare was for and how it worked, so that's off the list anyway.
2. I think a specimen or two of a properly formed use of the function would be hugely helpful to learners:
    `(States).contains("Arkansas")`
    `Math.max(ThisNumber,ThatNumber)`
    `(Wins > 5) ? "Wow!" : "Dang."`

BrentEaston:
But which popup box do you display that on? Because you aren't actually typing that stuff in, it is being generated for you, so it is not obvious what you should type into the individual argument boxes to get that result.

BrentEaston:
How about, instead of examples, I have a 'Visualiser' below the argument boxes which shows the generated result of what you are typing in. And this could even have the Tick/cross validator display showing if you are generating valid Beanshell or not?

Would we still need examples?
That would then make sense to have example

Cattlesquat:
image

Forgive the hacked-together nature of this and the fact that my sample expressions are neither very good nor even wholly accurate, but this gives you the idea. 

The idea of a visualizer has potential, but I think the Examples are more important -- see some examples of how the thing can actually be put to work. Because I could easily be typing stuff in that's not really right, so getting to see it visualized (as not actually right) doesn't help Lead Me Down The Path Of Righteousness. And the green-tick-marks are absolutely useful and fantastic, but if I don't have some examples then I'm kind of stabbing-in-the-dark as a learner. Whereas seeing something that somebody has actually done that definitely works starts to serve as a mental template for Ways I Can Bend This Thing To My Will.

Cattlesquat:
Probably hard to tell in that shot, but I subtly "greyed" the examples so they read as a slightly different type of text than the main black "officially the UI" text.

BrentEaston:
I don't think the examples under the individual parameter fields are necessary. They are just repeats of what you have below in the full examples, and really they could be anything. From your examples, it looks like you are saying you can only use a number or a String.

The examples at the bottom by themselves then have no context because they contain information that is not present on the page (in this case the ? and the :)

However, if we throw the visualiser in there, then you can see the full expression developing as you type in the parameter boxes and you can try and make it look like one of the examples. 

I agree, the examples are important, because it is not necessarily obvious what should go in each box. What was a Property Match Expression again? They will at least get you started.

I will have a play with those ideas for v2.

BrentEaston:
Brian,

I think I have cracked the States.contains(State) problem.

States.contains(State) was being interpreted as an unknown method States,contains() because the Parser didn't know what 'States' was, so assumes it is a class qualifier for the contains method.

To fix this, I had to re-organise the Beanshell expression evaluator to pass the values of property names in as function arguments. This forces the parser to see them as variables, not as part of a method signature.

However, achieving this adds a whole new layer of overhead to each call to evaluate a Beanshell expression. The arguments need types, and since I don't know what types the Beanshell expression is expecting, I have to examine the data in the property names used in each call and generate a suitable matching function signature and potentially re-parse the expression with the new argument list. All properties are stored as Strings, so if I see a "1", I don't know if the receiving property is going to be used as an int or a String, or if next time, that same argument will be "1A", so I need to regenerate the argument list each call and let auto-boxing sort out any discrepancies. 

It's all a bit of a rabbit hole and the fix is starting to look worse than the problem.

If we can live with the syntax of (States).contains(State) to perform String functions on Property Names, then I can auto-detect when this is needed in the Expression Builder and automagically convert States.contains(State) to (States).contains(State). I should also be able to detect any usage of States.stringFunction() in a user typed entry and flag this as an error in the Validator and display a meaningful error as a tooltip on the red X. Should even be able to auto-correct it if you click on the red X!

What do you think? My feeling is it is a bad idea to make Beanshell processing even slower. It would be an extra overhead on every Beanshell evaluation, not a one time set up.

Cattlesquat:
I agree. Not worth a big perf hit, especially when the (States) syntax is pretty good.

Cattlesquat:
One other quasi-related thing to look at if you have time is: http://www.vassalengine.org/tracker/show_bug.cgi?id=13224 

In my push to de-emphasize Old Style notation in the documentation examples etc, I've come across one I can't get rid of yet because it only accepts Old Style :) -- Deck count-matching-expressions.

Cattlesquat:
Oh and one more thing from the new Expression Builder that I'd forgotten because it was one of the very first things I tested:

I made a "Count" item and into the blank I typed `MyColor=="Red"` but what I got back was:

`Count("{MyColor==\"Red\"}")`

Which seems like a bit of a mess (but it works). But it would certainly be better and more readable if Count() and its ilk would at least accept:

`Count({MyColor=="Red"})`

Or even better of course:

`Count(MyColor=="Red")`

Since I'm not sure I think the "new stuff" should have to support Old Style expressions. Especially if the price is `Count("{MyColor==\"Red\"}")` :D 

Brian

BrentEaston:
I will fix the Deck Count expressions as part of this.

Count("{MyColor==\"Red\"}")  - Yep. Unfortunately, that's the only way it can be done.
Count({MyColor=="Red"}) - Nope, that's not java syntax, won't parse.
Count(MyColor=="Red") - That will count the property named "true" or the property named "false".

Don't forget, Beanshell is just straight Java syntax. so the only way to pass a comparison expression as the argument of a function is as a String. And any quotes in the String have to be escaped. Hence you get Count("{MyColor==\"Red\"}").

The best I can do is have the Expression Builder take care of that for you, if you use it.

This is regardless of supporting old-style expressions, that just came for free. Classic expressions don't need quotes, so end up looking a lot better.

 Count("MyColor=Red").

The most I could do would be to remove classic expression support altogether and the basic call would then be

 Count("MyColor==\"Red\"").

BrentEaston:
My comment did not come through properly, everything got unescaped. This is the correct version:

`Count("{MyColor==\"Red\"}") - Yep. Unfortunately, that's the only way it can be done.`

The inner quotes have to be escaped. Ugly, but I can't see anyway around it.

BrentEaston:
Hi Brian,
Try this version. It has the new support for calling String functions directly on Property Names. I did some small test, then fired up POG to some stress testing and I think I found a bug in POG v9.8. You may have fixed this already. When you remove the Hungary Neutral marker, you get a data error 

- Bad Data in Module: Expression evaluation error Expression={LocationName!=AP Forces}, Error=

The source is a Beanshell Fragment {LocationName!=$LocationName$} that is missing the quotes. Should be {LocationName!="$LocationName$"}

Cattlesquat:
Oh my, how did that ever work? Must have been getting masked by something in 3.2.17. I’d noticed little occasional “Bad Data” errors starting to pop up in PoG in 3.3.2, but the message makes it look sort of like “Old Style” expressions, and so I was sort of vaguely looking for Old Style expressions as I was “keeping my eye out for it”. But with that little hint I found it immediately.

 

Meanhile -- I’ll check out the new version!

 

Brian

Cattlesquat:
o i c ... It looks like a common pattern I inherited in PoG was if statements that said  LocationName!=$OldLocationName$  which is of course acceptable Old Style notation. But somehow, at some point,  at least one of those snuck its way *inside* a {  }. Could have been a poorly judged cut-and-paste , or some other form of self-trickery. The dangers of "old code bases", just the Module Version. I swear I was never seeing any data errors from it in 3.2.17 so I wonder if something was masking the reporting of it (even if the code wasn't "really working")?

Cattlesquat:
Hi Brent -- one more "Expression Builder" bug I filed a while back that might be worth hitting while you've got all that "paged into your brain" as Joel says: http://www.vassalengine.org/tracker/show_bug.cgi?id=12980

Cattlesquat:
Hi Brent -- I pulled this down to give it a try but wasn't able to get it to build. Checkstyle was mad about "import java.awt.*" so I tried taking that out thinking I would then just explicitly import whatever it then wanted, but then it wanted "Component" and when I did an "import java.awt.Component;" that didn't clear it up for reasons I couldn't immediately decipher so I decided to give up and wait for you. I'll give it a pull & a try as soon as you have it building again.

yanlyub:
You can skip the checkstyle check locally, add a `<skip>true</skip>` to the plugin configuration, here in the `<configuration>` block: https://github.com/vassalengine/vassal/blob/2d301d84f9b1793e901c43b7788c2a7055e95e10/vassal-app/pom.xml#L247

Alternatively a user property should work as well, add a `<checkstyle.skip>true</checkstyle.skip>` to the `<properties>` block here: https://github.com/vassalengine/vassal/blob/2d301d84f9b1793e901c43b7788c2a7055e95e10/vassal-app/pom.xml#L17

Here is the documentation of the `skip`: https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html#skip

I even think that most or even all maven plugins have this skip option which can be added to their configuration. Just don't commit it, it's for local use only :smile:

BrentEaston:
The problem is that my Intellij keeps sneaking in import java.awt.* and javax.swing.* into the import list. I needed to go to 

File > Settings > Project Settings > Code Style > Java > Imports > General > Class count to use import with '*

and remove those two from the list and bump the class counts to 99.

Locally in BeanShellExpressionConfigurer, you can just replace the * import with

import java.awt.Component;
import java.awt.Window;
import java.awt.event.KeyAdapter;
import java.awt.event.KeyEvent;
import java.awt.image.BufferedImage;

BrentEaston:
I'mhalfway through another set of changes so can;t check in a fixed version just yet.

BrentEaston:
> Hi Brent -- one more "Expression Builder" bug I filed a while back that might be worth hitting while you've got all that "paged into your brain" as Joel says: http://www.vassalengine.org/tracker/show_bug.cgi?id=12980

Brian, I think I have already fixed part 1 of this for the Set Global Property case in the version you have (once you fix the awt import). There is still an issue with the second part, will look into it.

BrentEaston:
Hi Brian, try this version. It fixes the import issue and adds the Visualiser I talked about. Do we still need examples as well or can these be one click away in the Reference manual? On my list left I have:
1. Bug 12980
2. Neaten up the Function Builder
3. Try and improve the error reporting on broken Beanshell expressions.

BrentEaston:
Wow, can view and add comments to PR's directly from IntelliJ. Big Thumbs up!
This version should resolve bug 12980 and tidies things up.  Can you give it a good run through? It is close to done I think.

Cattlesquat:
pulling it down now

Cattlesquat:
Works great, Brent! The "visualizer" I agree should be sufficient coupled with some documentation. 

I was able to "fool" it with things like Random(5, "String") into giving me a spurious green check (valid Java syntax but manifestly failing type), but that's what green check has always done before. It would be a plus if it rejected obviously-wrong types but that would be a whole side project I'm sure.

BrentEaston:
Yes, it is only checking for syntax. It would be possible, but a lot of work to check for that, but it will only pick up the obvious errors as I can only tell the type of the argument at this time if it is a constant. I can't check what any function calls or expressions would return. I think that is probably best covered in the docs - Green Tick only means valid syntax, it doesn't check for people doing something stoopid.

BrentEaston:
Joel, can we get this PR into 3.3.3 as well? It is basically a general bug fix and usability update to Beanshell. While it will not improve the performance of existing modules, the new Sum and Count functions have the potential to make a big improvement to modules that are currently using GKC's and Global Variables to do the same thing.

This PR fixes issues/bugs12980, 12999 and 13100

Cattlesquat:
Time to mark it no longer "in progress" :)

vassal-app/src/main/java/VASSAL/configure/BeanShellFunctionMenu.java
```diff
@@ -0,0 +1,327 @@
...
+    addOperator(comparisonMenu, "=~", "Matches Regular Expression");
+    addOperator(comparisonMenu, "!~", "Does not match Regular Expression");
+    add(comparisonMenu);
+
+    final JMenu logicalMenu = new JMenu("Logical");
+    addOperator(logicalMenu, "&&", "And");
+    addOperator(logicalMenu, "||", "Or");
+    addOperator(logicalMenu, "(", "Left parenthesis");
+    addOperator(logicalMenu, ")", "Right parenthesis");
+    add(logicalMenu);
```
uckelman, line 105: Isn't there a logical not?

BrentEaston:
Oops, update coming soon. Took a Logic professor to see that one :)

vassal-app/src/main/java/bsh/BeanShellExpressionValidator.java
```diff
@@ -1,187 +1,213 @@
...
+   * 
+   * @param s Expression
+   * @return stripped expression
+   */
+  public static String stripBraces(String s) {
+    String expr = s;
+    if (s.trim().startsWith("{") && s.trim().endsWith("}")) {
+      final int start = s.indexOf("{");
+      final int end = s.lastIndexOf("}");
+      StringBuilder buffer = new StringBuilder(s.length());
```
uckelman, line 146: I don't get this. If `s` starts with `{`, don't you already know that the index of `{` is 0, and if it ends with `}` that the index of the last `}` is `s.length()-1`?

BrentEaston: No, s.trim() starts with "{", not s. There may be extra leading or trailing spaces that I need to take into account, but it is not stripBraces() job to remove them. The purpose is to replace the outermost enclosing {}, if it exists with space characters, but not otherwise disrupt the the string.

Could probably do this better, but it is Module Editor only and limited use, I'm not too fussed about efficiency

uckelman: I misread. Thanks.

PR #154, opened 2020-08-04
"13218 Part Deux" - Resolve ClickedX and ClickedY to work The Better Brent Way

Cattlesquat:
Now Featuring Persistent Properties to avoid the issue where a module designer could neglect to implement part of the feature and get the "illusion of it working".

See commentary in #134


Cattlesquat:
Good god why are there documentation commits in here, sigh.

Cattlesquat:
Upstream master was screwed up

PR #155, opened 2020-08-04
13218 "Part Deux" - Better method for getting ClickedX and ClickedY passed through

Cattlesquat:
Now done The Better Brent Way (tm). To prevent module designer from failing to implement part of the feature and therefore experiencing the false illusion of success.

See commentary in #134


Cattlesquat:
My upstream master was screwed up.

PR #161, opened 2020-08-05
Bug 13280: Null protection in key translators

Cattlesquat:
Fixes issue from forums today.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/swing/SwingUtils.java
+++ b/vassal-app/src/main/java/VASSAL/tools/swing/SwingUtils.java
@@ -201,7 +201,7 @@ public class SwingUtils {
     @Override
     @SuppressWarnings("deprecation")
     public KeyStroke systemToGeneric (KeyStroke k) {
-      if (macLegacy || (k == null)) {
+      if (macLegacy) {
         return k;
       }
       int modifiers = k.getModifiers();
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/tools/swing/SwingUtils.java
```diff
@@ -201,7 +201,7 @@ public boolean isSelectionToggle(MouseEvent e) {
     @Override
     @SuppressWarnings("deprecation")
     public KeyStroke systemToGeneric (KeyStroke k) {
-      if (macLegacy) {
+      if (macLegacy || (k == null)) {
```
uckelman, line 204: I had a look at where these are used. Nothing should be passing nulls to these. The original call in `KeyStrokeListener` had an implicit null check because we were calling `stroke.equals(key)` and `stroke` had already been verified to be nonnull. We can omit the null checks here.

Cattlesquat: I had made a similar tour and concluded that nonetheless some "future-proofing" was reasonable, but I've removed those two extra checks in the new push.

yanlyub:
The "defensive" way of writing code would be to check for null at the beginning of every single method in this call chain. It bloats up the code but makes it very stable and bugs very easy to find if they appear.

The more modern way is to use Optional<> instead of passing regular types, this forces an explicit null-check by whoever wants to read the variable.

PR #164, opened 2020-08-06
Help Menu -- Trying to get help menu to have more useful entries.

Cattlesquat:


Cattlesquat:
@uckelman Here is my little Help Menu thing. In EditorWindow.java, look for three groups of `//BR//` comments.

Cattlesquat:
Meanwhile see ~line 250 and thereabouts in EditorWindow.java in "master" branch -- and notice that it's "supposed to" put an entry that says "Reference Manual" but instead puts an entry that says "Component Help".

Cattlesquat:
And I have absolutely no idea why that block is coded from the block two items up (that puts the Readme.html). 

And of course when I try to "copy the idea" of either block, I can't get it to work.

BrentEaston:
Something is going wacko. This is from my standard branch:

image

Same module with your branch:

image

BrentEaston:
Ah, much better. The 'Using the Editor' option seems to be missing it's manual page though. And that Table of Contents page is just a place holder yes? I'm guessing this branch doesn't have the new manual pages yet?

Cattlesquat:
Yep, just finally got it figured out. Two main things:
(1) Whoever came before me here was using the form 'ShowHelpAction(url,null)' which just happened to work on item #1 ("Help") because it wanted to say Help anyway, and Help is apparently what you get 100% of the time. Meanwhile in item #3 which in master as you see "Component Help" even though it is being told to say "Reference Manual", was using a different weird-o form of the code, and obviously only partially working. So the **right** thing to do was `ShowHelpAction("TheStringYouWantToSee", url, null)`
(2) Meanwhile down on the bottom option I had my own bug where there was already a "Properties" option elsewhere for a different kind of "Properties" and it was colliding and having a little pooch-up-party.

Cattlesquat:
Correct, this branch is "anticipating" the docs branch I believe it's PR#117 being merged in.

Cattlesquat:
And I believe THIS branch is now ready to merge in behind it. I fell into a "Classic Vassal Trap" of assuming whoever had written the previous Help Menu code had done it correctly :D  ... and then had the bad luck to compound it with the Properties / Properties collision.

Cattlesquat:
I went back and looked and it was actually broken even in 3.2.17 (saying Component Help instead of Reference Manual, and just getting lucky on the rest)

uckelman:
What am I supposed to be seeing in the Editor's Help menu now?

PR #166, opened 2020-08-06
Issue 13243 - Graceful Deprecation and removal of unused class members

BrentEaston:
This is the first part of a review of all classes with Deprecated members and marking them up with the new graceful deprecation scheme to generate warning dialogs to users of modules with custom code that calls Deprecated class methods, saying they have until such and such a date to get their act together (1 year from now). This is a Draft PR, but please feel free to comment. 

As part of this PR, I plan to examine each Class containing Deprecated members and
a) Retrieve any recently removed Deprecated methods
b) Tag each Deprecated Field and Method with the annotation @Deprecated(since = "2020-08-06", forRemoval = true)
c) Add a call to ProblemDialog.showDeprecated ("2020-08-05"); to each Deprecated method
d) Fix all JavaDoc, whitespace, Java 11 issues and as many compiler and IntelliJ code warnings as possible.

This commit contains the core changes to display the new Dialog and also updates the first handful of classes (by chance, the ones that my modules use :)). I have also added a similar 'Out of Data and Inefficient' dialog if any module tries set the BasicPiece.SNAPSHOT property the old way with a GamePiece generated by the PieceCloner, and converts the GamePiece to a Property Map instead of breaking the module.

Yan is building a custom Annotation processor that will report in our Builds when Deprecated members are nearing or pass their expiry (Issue 13244).

This is a model we can use going forward. My plan for 3.3.3 is to process all Class members currently marked as Deprecated.


yanlyub:
The "forRemoval=true" is meant to express that it will be removed in the next release and the compiler is supposed to print a stronger warning when it's still in use, are we sure we want this for code that will stay another year?

BrentEaston:
I think so. We want any module developers to fix their modules asap as if it was being removed tomorrow. One year is not that long in the greater scheme of things. It is tomorrow in the development cycle sense.

I have tried to add @@upressed("deprecation") warnings to deprecated methods that are calling other deprecated methods to keep their warnings out of our build.

The only one that was a but of an issue was HelpWindowExtension which you are trying to get rid of and still being used.

I think

vassal-app/src/main/java/VASSAL/build/module/map/BoardPicker.java
```diff
@@ -172,9 +175,7 @@ public void addTo(Buildable b) {
     for (Board board : possibleBoards) {
       board.setMap(map);
     }
-    if (b instanceof Translatable) {
```
uckelman, line 175: Ditto here---isn't this going to cause us a problem if `b` isn't a `Translatable`?

vassal-app/src/main/java/VASSAL/build/module/dice/DieServer.java
```diff
@@ -237,17 +252,17 @@ public void reportResult(RollSet mroll, FormattedString format) {
...
         }
       }
 
       if (reportTotal)
-        val += total;
+        val.append(total);
 
-      val = formatResult(roll.getDescription(), val, format);
-      GameModule.getGameModule().getChatter().send(val);
+      val = new StringBuilder(formatResult(roll.getDescription(), val.toString(), format));
```
uckelman, line 264: Why are we making a second `StringBuilder` here?

vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
```diff
@@ -353,27 +338,55 @@ public void run() {
...
+  public static Future<?> showDeprecated(String date) {
+    final StackWalker.StackFrame frame =
+      StackWalker.getInstance(Set.of(StackWalker.Option.RETAIN_CLASS_REFERENCE), 2)
+        .walk(f -> f.skip(1).limit(1).collect(Collectors.toList())).get(0);
+    final String method = frame.getClassName() + "." + frame.getMethodName();
+    String expiry;
+    try {
+      expiry = LocalDate.parse(date, DateTimeFormatter.ofPattern("uuuu-M-d")).plusYears(1).format(DateTimeFormatter.ofPattern("d-MMM-uuuu", Resources.getLocale()));
+    }
+    catch (Exception ignored) {
```
uckelman, line 369: What's the actual exception type which could come out of the try block?

vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
```diff
@@ -353,27 +338,55 @@ public void run() {
...
   }
+
+  public static Future<?> showDeprecated(String date) {
+    final StackWalker.StackFrame frame =
+      StackWalker.getInstance(Set.of(StackWalker.Option.RETAIN_CLASS_REFERENCE), 2)
+        .walk(f -> f.skip(1).limit(1).collect(Collectors.toList())).get(0);
+    final String method = frame.getClassName() + "." + frame.getMethodName();
+    String expiry;
+    try {
+      expiry = LocalDate.parse(date, DateTimeFormatter.ofPattern("uuuu-M-d")).plusYears(1).format(DateTimeFormatter.ofPattern("d-MMM-uuuu", Resources.getLocale()));
```
uckelman, line 367: Is there a reason we're not using the `ISO_LOCAL_DATE` formatter in both places here?

vassal-app/src/main/java/VASSAL/counters/Obscurable.java
```diff
@@ -86,7 +87,7 @@
 
   public Obscurable() {
     this(ID + "M;", null);
-  }
+  } //$NON-NLS-1$//
```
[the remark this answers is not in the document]
BrentEaston: Ah Yes. Intellij collapses one line methods and Constructors and it is easy to misplace the comment.

vassal-app/src/main/java/VASSAL/counters/Embellishment.java
```diff
@@ -80,13 +81,13 @@
...
+  public static final String ID = "emb2;"; // New type encoding //$NON-NLS-1$//
 
-  public static final String IMAGE = "_Image";
-  public static final String NAME = "_Name";
-  public static final String LEVEL = "_Level";
-  public static final String ACTIVE = "_Active";
+  public static final String IMAGE = Embellishment0.IMAGE;
+  public static final String NAME = Embellishment0.NAME;
+  public static final String LEVEL = Embellishment0.LEVEL;
+  public static final String ACTIVE = Embellishment0.ACTIVE;
```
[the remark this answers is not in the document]
BrentEaston: True, since we may potentially remove Embellishment0 at some point. Would be interesting to know if there are still modules that use it. Updated

vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
```diff
@@ -199,7 +196,7 @@ public void writeMapAsImage() {
     // something with the minimum size or font metrics
     final int l = "Saving map image as ".length() + file.getName().length() + 6;
     final StringBuilder b = new StringBuilder();
-    for (int i = 0; i < l; i++) b.append("N");
+    b.append("N".repeat(Math.max(0, l)));
```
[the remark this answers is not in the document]
BrentEaston: :) I'd like to take the credit, but it was a clever suggestion by Intellij.

vassal-app/src/main/java/VASSAL/build/module/map/GlobalMap.java
```diff
@@ -132,7 +127,7 @@ public void addTo(Buildable b) {
     map.addDrawComponent(this);
     map.getToolBar().add(launch);
 
-    if (b instanceof Translatable) {
+    if (b != null) {
```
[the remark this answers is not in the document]
BrentEaston: Fixed.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -1093,7 +1128,7 @@ else if (FONT_SIZE.equals(name)) {
       }
       if (value != null) {
         fontSize = (Integer) value;
-        font = font.deriveFont(fontSize);
+        font = font.deriveFont((float) fontSize);
```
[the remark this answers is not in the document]
BrentEaston: This one was a bit weird.

font.deriveFont(int) derives a font with a new Style
font.deriveFont(float) derives a font with a new Size

So the original code trying to derive a size 12 font, was actually deriving a Style 12 font, which does not exist so has no effect. The cast forces the size version to be called.

Presumably, no-one has ever been able to change the size of the Counter Detail View text.

vassal-app/src/main/java/VASSAL/build/module/map/BoardPicker.java
BrentEaston: Same again, fixed.

vassal-app/src/main/java/VASSAL/build/widget/PieceSlot.java
```diff
@@ -125,7 +123,7 @@ protected void copyFrom(PieceSlot piece) {
 
   public class Panel extends JPanel {
     private static final long serialVersionUID = 1L;
-    protected PieceSlot pieceSlot;
+    protected final PieceSlot pieceSlot;
```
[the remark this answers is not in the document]
BrentEaston: Yeeesss, but in the context of this piece, there is no possible reason for ever setting this to anything other than the enclosing slot. I think this should have been final in the first place and any over-riding code is making a grave mistake by changing it and should be alerted.

vassal-app/src/main/java/VASSAL/build/module/dice/DieServer.java
BrentEaston: Replaced that and the next line with

   GameModule.getGameModule().getChatter().send(formatResult(roll.getDescription(), val.toString(), format));

vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
BrentEaston: Ah, I think that was left over from an earlier version. No longer required, I have removed the try/catch block.

vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
BrentEaston: On Input, ISO_LOCAL_DATE is inflexible on the number of digits used for year, month or day. "uuuu-M-D" will happily parse "20-8-4" or "2020-08-04".

On output, I find the '3 Aug 2020' form far friendlier for multi-locale usage. No change of confusion. That's just my opinion though.

BrentEaston:
Resolved some issues with changes pushed. Commented on other issues.

vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
uckelman: That's exactly why I like ISO-8601 dates. :)

I think we should use a yyyy-mm-dd format for the deprecation annotation. Showing month names in messages is ok.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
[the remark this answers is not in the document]
uckelman: Oh wow. That is not a great API design! I'm now going to check every single place we call `deriveFont()`.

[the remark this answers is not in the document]
uckelman: Fortunately that was the only one where we used an int as the font size. I'm amazed!

PR #169, opened 2020-08-08
13236 - guard infinite loop in $..$ strings

Cattlesquat:


Cattlesquat:
This is also 4769

Cattlesquat:
We probably need both RecursionLimiter.infiniteLoopError(e) and RecursionLimiter.infiniteLoopWarning(e)

Depending on the situation where we're limiting recursion? Or is the RecursionLimiter *always* for module-designer-induced badness?

BrentEaston:
If you look at where the RecursionLimiter is implemented, it is only in places that can catch infinite loops caused by modules. We don't guard against it in Vassal itself, because if we see a Stack Overflow that is not caused by a module, we just fix the underlying bug and it never happens again.

So yes, RecursionLimiter  issues are always module errors.

Cattlesquat:
Okay cool. Once #184 gets merged so I have the new BadData stuff, I'll do another pass at this.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -231,70 +246,80 @@ public String evaluate(PropertySource ps, boolean localized) throws ExpressionEx
...
+        else if ("true".equals(value)) {
+          setVar(var, true);
+        }
+        else if ("false".equals(value)) {
+          setVar(var, false);
+        }
+        else {
           try {
-            setVar(var, Float.valueOf(value).floatValue());
+            setVar(var, Integer.valueOf(value).intValue());
```
yanlyub, line 279: Boxing/unboxing to parse a primitive VASSAL.script.ExpressionInterpreter.evaluate(PropertySource, boolean) [VASSAL.script.ExpressionInterpreter] At ExpressionInterpreter.java:[line 279] DM_BOXED_PRIMITIVE_FOR_PARSING (https://travis-ci.com/github/vassalengine/vassal/builds/179361630#L544, http://findbugs.sourceforge.net/bugDescriptions.html#DM_BOXED_PRIMITIVE_FOR_PARSING)

vassal-app/src/main/java/VASSAL/tools/FormattedString.java
```diff
@@ -133,19 +146,28 @@ public String getLocalizedText(PropertySource ps) {
...
+        }
+        else {
+          ErrorDialog.dataWarning(new BadDataReport(Resources.getString("Error.expression_error"), format.getExpression(), e));
+        }
+        return "";
       }
-      return "";
+    }
+    catch (RecursionLimitException e) {
+      return "&lt;infiniteloop&gt;";
```
uckelman, line 167: Strictly speaking, we'd have to have a general solution to the halting problem to know that there's an infinite loop. All we actually know is that you have a very long execution consistent with being in an infinite loop.

("possible infinite loop" is both accurate and brief.)

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
uckelman: `Integer.valueOf(value).intValue()` can just be `Integer.parseInt(value)` (and similarly, `Float.parseFloat(value)`).

vassal-app/src/main/java/VASSAL/tools/FormattedString.java
Cattlesquat: Oh right I think Brent wants me to return empty string but print Bad Module Data here. I was waiting for the BMD PR but now I can fix this up since it merged.

PR #170, opened 2020-08-08
10314 - NPE Fix - GameRefresher - dialog can be null if the dialog box is closed quickly while test is running

Cattlesquat:
see start() method for a picture of how that happens


uckelman:
Is this Bug 10314? http://www.vassalengine.org/tracker/show_bug.cgi?id=10314

Cattlesquat:
Bwahahaha yes it is

PR #171, opened 2020-08-08
2705 - NPE Fix - FreeRotator - protect from NPE if piece got moved off map by e.g. GKC.

Cattlesquat:
All other paths in this trait have explicit NPE protect for "the piece might no longer be on a map", but this path had no protection, so it will NPE any time the piece gets GKC'ed off a map during its interactive free rotating thinger.


uckelman:
Is this Bug 2705? http://www.vassalengine.org/tracker/show_bug.cgi?id=2705

Cattlesquat:
Why yes. Yes it IS bug 2705 :D

PR #172, opened 2020-08-08
Minor optimizations

yanlyub:


vassal-app/src/main/java/VASSAL/tools/SequenceEncoder.java
```diff
@@ -178,7 +178,7 @@ public String nextToken() {
         int end = i;
         while (begin < end) {
           if (val.charAt(end - 1) == '\\') {
-            buffer.append(val.substring(begin, end - 1));
+            buffer.append(val, begin, end - 1);
```
uckelman, line 181: I know why it was like this: Originally this used a `StringBuffer`, and that overload of `append` didn't exist in `StringBuffer` until Java 5.

vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
```diff
@@ -42,13 +44,13 @@ public DeobfuscatingInputStream(InputStream in) throws IOException {
     super(null);
 
     final byte[] header = new byte[ObfuscatingOutputStream.HEADER.length()];
-    readFully(in, header, 0, header.length);
+    readFully(in, header, HEADER_LENGTH);
```
uckelman, line 47: I wouldn't do it this way. The thing that `readFully` needs is the actual length of the array, which `header.length` will always be.

PR #173, opened 2020-08-08
13217 - can't paste a cut ancestor to its own child

Cattlesquat:


uckelman:
What's the reason for the Chatter changes?

Cattlesquat:
It's an avenue for the Editor to splort text straight into the chatter for "warnings" (already use more or less this for the Search feature when you e.g. get no hits, but made it wider available to the rest of the editor, since I was planning on doing more of this). Doesn't need the full normal GameModule version of this because we're in the Editor and we don't need to do Commands and messages to other players and stuff. At least that was my take.

PR #175, opened 2020-08-08
2377 NPE in PlaceMarker because not checking if parent is null, which…it can totally legitimately be

Cattlesquat:
Gets reported about once a year for the last 10 years :D

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
@@ -205,7 +205,7 @@ public class PlaceMarker extends Decorator implements TranslatablePiece {
         target = parent == null ? outer : parent;
       }
       c = m.getStackMetrics().merge(target, marker);
-      if (index >= 0) {
+      if ((parent != null) && (index >= 0)) { // Parent can come down to us as null
         final ChangeTracker ct = new ChangeTracker(parent);
         parent.insert(marker, index);
         c = c.append(ct.getChangeCommand());
--- a/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
@@ -190,19 +190,23 @@ public class PlaceMarker extends Decorator implements TranslatablePiece {
         m.getPieceCollection().canMerge(outer, marker)) {
       final Stack parent = getParent();
       GamePiece target = outer;
-      int index = -1;
+      int index = -1;      
       switch (placement) {
       case ABOVE:
-        target = outer;
         break;
       case BELOW:
-        index = parent == null ? 0 : parent.indexOf(outer);
+        if (parent != null) {
+          index = parent.indexOf(outer);
+        }
         break;
       case STACK_BOTTOM:
         index = 0;
         break;
       case STACK_TOP:
-        target = parent == null ? outer : parent;
+        if (parent != null) {
+          target = parent;
+        }
+        break;
       }
       c = m.getStackMetrics().merge(target, marker);
       if ((parent != null) && (index >= 0)) { // Parent can come down to us as null


Cattlesquat:
Hi @uckelman I saw your chats about this when I got back online.

I'm not sure the BELOW case is actually wrong. Also in any event, parent could be null in STACK_BOTTOM too, and that would also crash it. 

Given that no one has been complaining about "Place Marker isn't putting my counter in the right place in the stack", but we have 10 years of NPE reports, I'd say top priority is to get the NPE fixed? 

I always thought it was sort of odd PlaceMarker had all these choices anyway, when it's so much easier to use the apply-after keystroke to fix things how you want, plus you usually have to do that anyway.

```java
if (m.getStackMetrics().isStackingEnabled() &&
        !Boolean.TRUE.equals(marker.getProperty(Properties.NO_STACK)) &&
        !Boolean.TRUE.equals(outer.getProperty(Properties.NO_STACK)) &&
        m.getPieceCollection().canMerge(outer, marker)) {
      final Stack parent = getParent();
      GamePiece target = outer;
      int index = -1;
      switch (placement) {
      case ABOVE:
        target = outer;
        break;
      case BELOW:
        index = parent == null ? 0 : parent.indexOf(outer);
        break;
      case STACK_BOTTOM:
        index = 0;
        break;
      case STACK_TOP:
        target = parent == null ? outer : parent;
      }
      c = m.getStackMetrics().merge(target, marker);
      if ((parent != null) && (index >= 0)) { // Parent can come down to us as null
        final ChangeTracker ct = new ChangeTracker(parent);
        parent.insert(marker, index);
        c = c.append(ct.getChangeCommand());
      }
    }
```

BrentEaston:
Those options pre-date the whole 'apply key after xxxx' concept.

uckelman:
What I'm saying about the `BELOW` case is that I don't understand why someone bothered to _set_ `index` at al when `parent` is null, because you can't use `index` for anything in that case. Shouldn't it stay `-1`?

uckelman:
Also the `ABOVE` case is weird. `target` is already `outer`, so why assign `outer` to it again?

uckelman:
This does the same thing, but doesn't leave me wondering what's going on:
```
      switch (placement) {
      case ABOVE:
        break;
      case BELOW:
        if (parent != null) {
          index = parent.indexOf(outer);
        }
        break;
      case STACK_BOTTOM:
        index = 0;
        break;
      case STACK_TOP:
        if (parent != null) {
          target = parent;
        }
        break;
      }
```
It's probably not that important, but setting `index` to `0` in that one case is confusing when it's not going to be used.

Cattlesquat:
Updated

uckelman:
@BrentEaston Are you satisfied that this fixes the bug correctly? [Your comment from 2012](http://www.vassalengine.org/tracker/show_bug.cgi?id=2377#c11) on the ticket makes me wonder.

BrentEaston:
This is one of the most annoying bugs I ever looked at. No-one was every able to supply a way to reproduce it.

It is likely that if parent is null, we don't want to be Placing Markers anyway. For example, if you where to right-click and select place marker on a counter in a PieceSlot. However, Place Marker traits are supposed to be disabled in this case. 

Note the

command.setEnabled(getMap() != null && markerSpec != null);

in myGetKeyCommands.

So yes, my 2012 comment still stands, there is some deeper issue and this 'fix' will just mask it and cause Place Marker to quietly and inexplicably fail now and then. Some of the comments seem to indicate that the players where just doing normal game 'things' like 'flippin a token'.

One thing all of these stack traces seem to have in common is that the Place Marker trait is activated by clicking on a button

at javax.swing.AbstractButton.fireActionPerformed(Unknown Source) ~[na:1.8.0_241]
	at javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source) ~[na:1.8.0_241]
	at javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source) ~[na:1.8.0_241]
	at javax.swing.DefaultButtonModel.setPressed(Unknown Source) ~[na:1.8.0_241]
	at javax.swing.AbstractButton.doClick(Unknown Source) ~[na:1.8.0_241]

It never seems to occur as a result of a manual interaction with the counter and there are no Vassal classes about that button. I'd sure like to know what that button is. 

To fix this properly, we really need to dump some more detailed error information into the log. In this case, walk the stack up to that AbstactButton and print it's name and any other info we can grab from it. Without a repro that's about all we can do.

Cattlesquat:
I don't disagree with any of that, however this has been the case for over 10 years now, and we haven't gotten anywhere. I don't think that leaving it there to NPE a few times a year indefinitely is a good idea either. 

I would suggest putting this fix into place for now & making a note/ticket that someone build an error dumper or something that would wait for this problem to arise and try to get the user to send us the walked information on it (I don't know how to do that part).

uckelman:
We found the cause and the solution. Closing in favor of #267

PR #177, opened 2020-08-08
13200 - mainWindowDock could be null here

Cattlesquat:
It's a module we don't have in our library and marked 1.0 so quite likely some brand new module he's making. It looks like it might be some thing where he doesn't actually *have* a map in the module yet or something like that. 

Anyway, the thinger that came up null was being protected from null elsewhere so it probably should be here.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -1949,8 +1949,8 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
         return false;
       }
     }
-    // should be impossible
-    return true;
+    // should be impossible (and yet is not, if there are no maps in the module at all -- BR)
+    return false;
   }
 
   /**
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -1965,12 +1965,14 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
       final GameModule g = GameModule.getGameModule();
 
       if (shouldDockIntoMainWindow()) {
-        mainWindowDock.showComponent();
-        final int height = (Integer)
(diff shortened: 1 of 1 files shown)


Cattlesquat:
This is also 10102

uckelman:
The only way I can see that this could happen is if `Map.setup()` ran but `Map.addTo()` didn't run first, since the latter is the sole place that `mainWindowDock` is assigned to.

Cattlesquat:
Or (dun dun DUNH) if the module has no Main Map?

This had several dupes out there, so it wasn't a one-off freak bug.

uckelman:
How can `shouldDockIntoMainWindow()` return true if there's no main map?

Cattlesquat:
Check the amazing code of `shouldDockIntoMainWindow()` I believe we are hitting the flow covered by the (lol) "Should be impossible" comment. I think more accurately it means the module has no Map *at all*.

Cattlesquat:
I mean, I'm not saying these guys aren't jackwagons... :D

uckelman:
Ah, so! Maybe that should return `false` in the case that nothing is found!

uckelman:
BTW: I tried rewriting `shouldDockIntoMainWindow()` shortly before you arrived, and concluded that it's actually quite hard to adjust to be any better because the inner call to `shouldDockIntoMainWindow()` could be calling an overrriden method in a subclass.

Cattlesquat:
`shouldDockIntoMainWindow()` now returns false in the "impossible" case.

uckelman:
Wait, wait---`shouldDockIntoMainWindow()` is a non-static member of `Map`. So if you're calling it, you definitely have a `Map`. This means that the `Map` you have is not one returned by `GameModule.getGameModule().getComponentsOf(Map.class)`. So you somehow have a `Map` that the module doesn't know about?!

Cattlesquat:
I guess one other possibility is that `shouldDockIntoMainWindow()` changes values in between the addTo() and the setup()?

Could that happen for example if someone changed configuration and then restarted the module without restarting editor?

PR #178, opened 2020-08-08
12986 - Stack could be getting created not-on-a-map

Cattlesquat:
There is evidence elsewhere in StackMetrics.java (see line 271) that Stacks sometimes exist "not on a map". And if one is *created* not-on-a map, as tried to happen here, this line will blow up. So I put in null protection for the map.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/StackMetrics.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/StackMetrics.java
@@ -653,8 +653,7 @@ public class StackMetrics extends AbstractConfigurable {
           }
           else {
             fixedParent = createStack(fixed, true);
-            fixedParent.setMap(fixed.getMap());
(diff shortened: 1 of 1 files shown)


Cattlesquat:
Also, this is 2714, ho ho ho

Cattlesquat:
And 3449 hahaha

uckelman:
The fix for this one worries me, because it makes a case where no command gets created. @BrentEaston had a comment on one of the issues in which this occurs where he said that doing this would mask the actual bug.

Cattlesquat:
Various ideas:
* We could just not create a stack in this case (where we aren't on a map). Just put the other piece in the same "location". 
* We could print some warning into the chat log (sort of like the "bad data" messages for expressions)

Cattlesquat:
This bug is alive and well as recently as 3.3.2 (e.g. bug report 13337)

uckelman (approved):

PR #184, opened 2020-08-09
12648 - guard SetupStack from unbuilt pieces

Cattlesquat:
This is one of those places where one could optionally print a "failed to build piece" warning to the chatter line, if we like the idea of going in that direction. 

Module designer had manifestly screwed up his piece definitions. But we can obviously handle better than "NPE".

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -531,18 +531,16 @@ public abstract class GameModule extends AbstractConfigurable implements Command
    * Display the given text in the control window's status line.
    * Save the messages for later if the Chatter has not been initialised yet
    */
-  public static void warn(String s) {
-    GameModule module = getGameModule();
-    if (module != null) {
-      if (module.chat == null) {
-        module.deferredChat.add(s);
-      }
-      else {
-        module.chat.show(" - " + s); //$NON-NLS-1$
-      }
+  public void warn(String s) {
+    if (chat == null) {
+      deferredChat.add(s);
+    }
+    else {
+      chat.show(" - " + s); //$NON-NLS-1$
     }
   }
-
+  
+  
   /**
    * @return a single Random number generator that all objects may share
    */
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -191,7 +191,7 @@ public abstract class GameModule extends AbstractConfigurable implements Command
    */
   protected GpIdSupport gpidSupport = null;
   protected Long crc = null;
-  
+
   private static String oldDragThreshold; //
 
   /**
@@ -401,15 +401,15 @@ public abstract class GameModule extends AbstractConfigurable implements Command
       l.addKeyStrokeSource(s);
     }
   }
-  
-  
+
+
   /**
-   * If our keyboard mapping paradigm changes (example: Mac Legacy preference checked/unchecked), we need to reregister all of our KeyStrokeListeners 
+   * If our keyboard mapping paradigm changes (example: Mac Legacy preference checked/unchecked), we need to reregister all of our KeyStrokeListeners
    */
   public void refreshKeyStrokeListeners() {
     keyStrokeListeners.forEach(l -> l.setKeyStroke(l.getKeyStroke()));
   }
-  
+
 
   @Deprecated public void fireKeyStroke(KeyStroke stroke) {
     if (stroke != null) {
@@ -767,7 +767,7 @@ public abstract class GameModule extends AbstractConfigurable implements Command
    */
   public void sendAndLog(Command c) {
     if (c != null && !c.isNull()) {
-      synchronized(loggingLock) {
+      synchronized (loggingLock) {
         if (loggingPaused) {
           if (pausedCommands == null) {
             pausedCommands = c;
@@ -794,7 +794,7 @@ public abstract class GameModule extends AbstractConfigurable implements Command
    * @return
    */
   public boolean pauseLogging() {
-    synchronized(loggingLock) {
+    synchronized (loggingLock) {
       if (loggingPaused) {
         return false;
       }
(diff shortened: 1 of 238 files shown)


BrentEaston:
Yes, This fixes the original problem, the NPE, at the expense of making the failure invisible.

In cases like this, I think we need to do a 'BadBulldFileReport', or something like that. Maybe a 'BadDesignerReport; :)

So 

if (p == null) {
  show a bad something report in chatter
else 
  do the normal stuff
end

We would need to provide enough information to allow the Designer to track down the problem.

In general, If we can't be sure it is a Design problem, or we can't determine enough information to direct the Designer to the fix, then we should probably raise a Bug Report.

In this particular case, we can direct them to the name of the SetupStack that is failing to build. We can't tell then the name of the piece, because the piece build is failing. But we could tell them, the 3rd piece failed to build.

Cattlesquat:
Yeah especially after the discussion today when we were fixing up the threading issue with chat "warnings", this seems like a really good way to go.

Cattlesquat:
@BrentEaston @uckelman See what you guys think of this and if so once it's merged I'll use this in other similar situations.

vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
```diff
@@ -414,9 +415,14 @@ protected Stack initializeContents() {
...
-        GameModule.getGameModule().getGameState().addPiece(p);
-        s.add(p);
+        if (p != null) { // In case slot fails to "build the piece", which is a possibility.
+          p = PieceCloner.getInstance().clonePiece(p);
+          GameModule.getGameModule().getGameState().addPiece(p);
+          s.add(p);
+        } 
+        else {
+          Chatter.warning("- Bad Piece Data - Failed to create piece in At-Start Stack: " + getConfigureName());
+        }
```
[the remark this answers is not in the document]
BrentEaston: Rather than run this directly into Chatter.warning, it might be better to use the existing structures and do this:

ErrorDialog.dataError(new BadDataReport(this, message, data));

This particular issue is a bad PieceDefinition in a PieceSlot, so you could add a new Constructor to BadDataReport(PieceSlot slot, String message, String data)  and then just do

ErrorDialog.dataError(new BadDataReport(slot, "Piece Definition failed to build", "Attempting to build piece "+num+" in At-Start stack"+getConfigureName());

The BadDataReport would handle the actual formatting, This could probably be used anywhere the PieceCloner fails to build a piece.

Have a Play, might be better to sub-class BadDataReport into BadPieceDefintionReport, BadTraitReport, BadOtherThingReport. etc. That would allow you to tailor the Constructors better. The Trait stuff in BadDataReport should really be in a subclass BadTraitReport. 

This will also give us a central place to change how we handle these.

Cattlesquat:
Okay, main files to have a look at are:
* SetupStack.java & PieceSlot.java for the generation of the reports
* ErrorDialog - I changed "dataError" to "dataWarning" for reasons of clarity (this generated the majority of file changes in this PR)
* BadDataReport - I added a new PieceSlot version per Brent's suggestion. So far not feeling the need for subclasses - the type matching on the parameters works fine, and there's less looking-up-the-right-name. 

There were some entertaining "bootstrapping" adventures with PieceSlot itself -- I originally put a Resources.getString() in there, but in my test case I'd malformed my "bad piece" definition so badly that it failed during the module builder which meant "failed to get resource bundle" and some other fun stuff. So we'll have to be careful that we don't put Resource.getString() in places where it could fail before the Resources are loaded.

The At-Start stack itself is fine with that. 

Pic of my test-run:
image

vassal-app/src/main/java/VASSAL/tools/ErrorDialog.java
```diff
@@ -332,20 +332,18 @@ public static void bug(final Throwable thrown) {
     );
   }
 
-  private static final Set<String> reportedDataErrors =
+  private static final Set<String> reporteddataWarnings =
```
uckelman, line 335: `reportedDataWarnings`

Cattlesquat: oops

Cattlesquat: a global-search-and-replace gone slightly astray. Fortunately not MUCH astray :)

vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -531,12 +531,15 @@ public GamePiece createPiece(String type, GamePiece inner) {
    * Display the given text in the control window's status line.
    * Save the messages for later if the Chatter has not been initialised yet
    */
-  public void warn(String s) {
-    if (chat == null) {
-      deferredChat.add(s);
-    }
-    else {
-      chat.show(" - " + s); //$NON-NLS-1$
+  public static void warn(String s) {
```
uckelman, line 534: This change won't be binary compatible.

Cattlesquat: fixed

Cattlesquat:
(1) GameModule.warn() returned to non-static
(2) ErrorDialog.dataError() "deprecated version" left as a deprecation-warning-to-the-unwary-subclasser
(3) In testing all sorts of different garbage I could feed this, I found a few more failure paths and locked them up. 

I'm realizing that a number of our Trait and Component classes are going to need some "hardening" against BAD MODULE DATA. Will just take that on a class by class basis starting with some of the obvious worst-offenders. But I think it's a good principle to assume that module designers will feed in some fairly stinky garbage and it's better if it looks like Bad Module Data(tm) than a Vassal Train Wreck.

BrentEaston:
I agree. If we are going to make it easier for designers to add custom code, then we have to be really tight with how we handle potential errors and make sure it is clearly a MODULE bug and not a VASSAL bug. The bug bonus is these issues will stop feeding bugs into bugzilla.

Cattlesquat:
I'm not very good at unit tests, but I bet there's substantial unit-testing that could be done of traits/components being forced to eat crud and say yum.

uckelman (approved):

PR #187, opened 2020-08-09
11827 - Send-to-Location w/ gridLocation option when board doesn't have any grid assigned

Cattlesquat:


BrentEaston:
Definite need for a BadDataReport. This is the classic case where the SendToLocation just doesn't work for no apparent reason and with out an error, the designers and our time is wasted trying to work it out.

These sorts of errors will nearly always be picked up by the module designer during development if we can get good error messages out

ErrorDialog.dataError(new BadDataReport(this, 'Attempt to Send to GridLocation, but destination has no grid defined', 'Destination board/gridname info'); // With appropriate translation!

Cattlesquat:
Okay this is fixed up with "Bad Module Data"

PR #189, opened 2020-08-10
i18n preparation work

BrentEaston:
This is basically just a refactor of the core i18n routines, plus a bunch of comments and FIXME's in preparation for a sweep through the code to bring the i18n up to date, ensure consistency and complete the i18n for the Editor.


BrentEaston:
Resolve Conflicts

yanlyub:
```
[ERROR] /home/travis/build/vassalengine/vassal/vassal-app/src/main/java/VASSAL/i18n/Resources.java:363:20: WhitespaceAround: '=' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3) [WhitespaceAround]
```

And more in `Resources.java`, checkstyle got stricter.

uckelman:
Is this ready for review?

BrentEaston:
Yes please.

vassal-app/src/main/java/VASSAL/configure/StringEnum.java
```diff
@@ -1,6 +1,4 @@
 /*
- *
- * Copyright (c) 2000-2003 by Rodney Kinney
```
uckelman, line 3: Probably shouldn't remove the copyright notices.

vassal-app/src/main/java/VASSAL/configure/StringEnumConfigurer.java
```diff
@@ -1,7 +1,4 @@
 /*
- *
- * Copyright (c) 2000-2009 by Rodney Kinney, Brent Easton
- *
```
uckelman, line 4: Ditto.

vassal-app/src/main/java/VASSAL/i18n/Resources.java
```diff
@@ -327,28 +369,21 @@ public static String getString(ResourceBundle bundle, String id) {
...
-  protected static final String EN_BUNDLE = "VASSAL_en.properties";
-
   /**
    * Custom Class Loader for loading VASSAL property files.
    * Check first for files in the VASSAL home directory.
    *
    * @author Brent Easton
    */
-  public class VassalPropertyClassLoader extends ClassLoader {
+  public static class VassalPropertyClassLoader extends ClassLoader {
```
uckelman, line 378: This would cause an exception if anyone subclassed it but didn't recompile. (Odds are that no one will have, though.)

vassal-app/src/main/java/VASSAL/i18n/Resources.java
```diff
@@ -205,7 +217,7 @@ private void addLocale(Locale l) {
    *
    * Resources.getString(Resources.VASSAL)
    */
-  public static final String VASSAL = "General.VASSAL"; //$NON-NLS-1$
```
uckelman, line 208: Deprecate?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/configure/StringEnum.java
BrentEaston: True

vassal-app/src/main/java/VASSAL/i18n/Resources.java
BrentEaston: Ah, that was a bug actually. I had replaced an existing in-use  key which is a no-no. Reverted, but VASSAL key moved to top of file.

vassal-app/src/main/java/VASSAL/i18n/Resources.java
BrentEaston: True, best to leave it. Reverted.

BrentEaston:
That should resolve the issues.

uckelman (approved):

PR #190, opened 2020-08-10
13188 - "Gentle SVG format failures" - Follow same pattern as SVGImageUtils to catch non-px format

Cattlesquat:
image

Much cleaner fail, and now provides a bit of guidance.

We can come back and try to support other formats later, if we want, but in any event this should help with the bug reports.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/image/svg/SVGImageUtils.java
+++ b/vassal-app/src/main/java/VASSAL/tools/image/svg/SVGImageUtils.java
@@ -86,10 +86,10 @@ public class SVGImageUtils {
       synchronized (factory) {
         doc = factory.createDocument(null, in);
       }
-    } 
+    }
     catch (FileNotFoundException e) {
       throw new ImageNotFoundException(name, e);
-    } 
+    }
     catch (DOMException | IOException e) {
       throw new ImageIOException(name, e);
     }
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/tools/image/svg/SVGImageUtils.java
```diff
@@ -78,60 +79,55 @@ public static Dimension getImageSize(InputStream in) throws IOException {
...
-        root.getAttributeNS(null, "height").replaceFirst("px", ""))+0.5);
+      final int width = (int) (Float.parseFloat(root.getAttributeNS(null, "width").replaceFirst("px", "")) + 0.5);
+      final int height = (int) (Float.parseFloat(root.getAttributeNS(null, "height").replaceFirst("px", "")) + 0.5);
 
       return new Dimension(width, height);
-    }
+    } 
     catch (NumberFormatException e) {
-      throw new ImageIOException(name, e);
+      throw new ImageIOException(name + " (Only 'px' units supported for height and width in SVG files)", e);
```
BrentEaston, line 106: Much better overall, but what is that extra trailing whitespace on line 104? There seems to be a few of them added to this file.

uckelman:
Closing in favor of #196, which handles all absolute units.

PR #193, opened 2020-08-11
Increase checkstyle strictness

yanlyub:


yanlyub:
#228 needs to be merged before this.

PR #194, opened 2020-08-11
Remove finalize() method

yanlyub:


vassal-app/src/main/java/VASSAL/tools/menu/MenuScroller.java
```diff
@@ -433,29 +429,6 @@ public void keepVisible(int index) {
     keepVisibleIndex = index;
   }
 
-  /**
-   * Removes this MenuScroller from the associated menu and restores the
-   * default behavior of the menu.
-   */
-  public void dispose() {
```
uckelman, line 440: Doesn't this need to be deprecated instead of removed?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/tools/menu/MenuScroller.java
yanlyub: Theoretically, yes. Practically, is anyone likely to call this?

uckelman: No, but I would prefer not getting bug reports about it.

uckelman (approved):

PR #195, opened 2020-08-12
Remove possible and actual bugs, some code cleanup

yanlyub:


vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/LayoutConfigurer.java
```diff
@@ -400,6 +400,8 @@ public void setColumns(int cols) {
     }
 
     public int getIntValue() {
+      // TODO is this supposed to call enclosing object's VASSAL.build.module.gamepieceimage.LayoutConfigurer.getValue
+      // or VASSAL.configure.Configurer.getValue from superclass?
```
uckelman, line 404: It should be IntConfigurer.getValue(), from the superclass. getValue() from the outer class returns something which is completely the wrong type.

PR #196, opened 2020-08-12
Bug 10414: SVG with non-px dimensions throws ImageIOException

uckelman:
CSS 2.1 defined 96px = 1in, so we can now fix dimensions for all SVG which doesn't use font-relative sizes. This should prevent all reasonable SVG from throwing exceptions due to dimensions.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/image/svg/SVGImageUtils.java
+++ b/vassal-app/src/main/java/VASSAL/tools/image/svg/SVGImageUtils.java
@@ -110,7 +110,7 @@ public class SVGImageUtils {
     final String vbs = root.getAttributeNS(null, SVGConstants.SVG_VIEW_BOX_ATTRIBUTE);
 
     final BridgeContext bctx = new BridgeContext(new UserAgentAdapter());
-    final UnitProcessor.Context uctx = UnitProcessor.createContext(bctx, root); 
+    final UnitProcessor.Context uctx = UnitProcessor.createContext(bctx, root);
 
     float w = -1.0f;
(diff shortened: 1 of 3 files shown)


uckelman:
NB: Image dimensions in em and ex will do... something, but not something reliable. It's not a good idea to use them in this context, as one can't really be sure what the default font size _is_, but they won't throw now.

vassal-app/src/main/java/VASSAL/tools/image/svg/SVGImageUtils.java
```diff
@@ -78,35 +91,108 @@ public static Dimension getImageSize(InputStream in) throws IOException {
...
 
-      return new Dimension(width, height);
+    // try to parse the height
+    if (!hs.isEmpty()) {
+      try {
+        h = UnitProcessor.svgVerticalLengthToUserSpace(hs, SVGConstants.SVG_HEIGHT_ATTRIBUTE, uctx);
+      }
+      catch (BridgeException e) {
+        // the height was invalid
+        throw new IOException(e);
```
Cattlesquat, line 137: Note that the former "ImageIOException" I believe goes to a non-stack-dump "warning format", Does IOException? I would certainly recommend image-file loading not lead to stack-dump.

uckelman: If you look outward from where this particular overload of `getImageSize()` is called, you'll find the call is inside a catch which rethrows `IOException` wrapped in an `ImageIOException`. See lines 94-104 (new).

PR #197, opened 2020-08-12
Close stream when saving the game

yanlyub:


vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -746,14 +746,15 @@ else if (END_SAVE.equals(theCommand)) {
...
 // FIXME: Extremely inefficient! Write directly to ZipArchive OutputStream
     final String save = saveString();
-    final FastByteArrayOutputStream ba = new FastByteArrayOutputStream();
-    try (OutputStream out = new ObfuscatingOutputStream(ba)) {
-      out.write(save.getBytes(StandardCharsets.UTF_8));
-    }
+    try (FastByteArrayOutputStream ba = new FastByteArrayOutputStream()) {
+      try (OutputStream out = new ObfuscatingOutputStream(ba)) {
+        out.write(save.getBytes(StandardCharsets.UTF_8));
+      }
```
uckelman, line 752: Shouldn't the scope of the `FBAOS` end here?

uckelman:
The comment has me wondering why we don't do this:
```
    try (FileArchive archive = new ZipArchive(f)) {
      try (OutputStream zout = archive.getOutputStream(SAVEFILE_ZIP_ENTRY)) {
        try (OutputStream oout = new ObfuscatingOutputStream(zout)) {
          oout.write(save.getBytes(StandardCharsets.UTF_8));
        }
      }
      (new SaveMetaData()).save(archive);
    }
```

yanlyub:
Is the `FastByteArrayOutputStream` extremely inefficient, or writing the save to a String first, then write it into the stream?

uckelman:
`FastByteArrayOutputStream` itself is quite efficient itself, but I'm having a hard time seeing why we should be using it essentially as a very large buffer in this case, to collect all the bytes before writing them out again, instead of writing the bytes directly.

Writing the save state to a `String` first is _also_ inefficient, but I don't see an obvious way to fix that from within this method.

It would be worth saving something very large to see if there's much difference between using the `FBAOS` and writing directly.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -746,15 +746,11 @@ else if (END_SAVE.equals(theCommand)) {
   public void saveGame(File f) throws IOException {
 // FIXME: Extremely inefficient! Write directly to ZipArchive OutputStream
     final String save = saveString();
-    try (FastByteArrayOutputStream ba = new FastByteArrayOutputStream()) {
-      try (OutputStream out = new ObfuscatingOutputStream(ba)) {
+    try (FileArchive archive = new ZipArchive(f)) {
+      try (OutputStream out = new ObfuscatingOutputStream(archive.getOutputStream(SAVEFILE_ZIP_ENTRY))) {
```
uckelman, line 750: Won't we fail to close the `OutputStream` from the archive if the `OOS` ctor throws?

Shouldn't it be this instead?

```
try (OutputStream zout = archive.getOutputStream(SAVEFILE_ZIP_ENTRY);
     OutputStream out = new ObfuscatingOutputStream(zout)) {
```

yanlyub: Both should work but the one with explicit declaration of the stream variables looks safer.

uckelman (approved):

PR #198, opened 2020-08-12
Fix floating point comparison

yanlyub:


vassal-app/src/main/java/VASSAL/build/module/map/boardPicker/board/HexGrid.java
```diff
@@ -1062,4 +1062,10 @@ public int getSnapScale() {
   public void setSnapScale(int snapScale) {
     this.snapScale = snapScale;
   }
+
+  private static boolean isClose(double x, double y) {
+    final double bounds = 0.01;
```
yanlyub, line 1067: Is 0.01 a good value here?

Cattlesquat: If there's gonna be an isClose(), I recommend it go in some kind of MathUtils and the tolerance be passed as a parameter, because different situations will want different tolerances.

yanlyub: It looks like this is the only case of floating point comparison in all of the code.

vassal-app/src/main/java/VASSAL/build/module/map/boardPicker/board/HexGrid.java
```diff
@@ -387,7 +387,7 @@ else if (DY.equals(key)) {
         val = Double.valueOf((String) val);
       }
       dy = (Double) val;
-      if (dx == sqrt3_2 * 64.0) {
+      if (isClose(dx, sqrt3_2 * 64.0)) {
         dx = sqrt3_2 * dy;
       }
```
BrentEaston, line 392: What this code is actually trying to do is to check if the hex width (dx) still has its default value of 64.0. If so, then update the width to to a nice value based on the hex height (dy), otherwise leave it alone.  Note the other 64.0 in the constructor.

This code was working before, probably only because 64.0 does have an exact binary representation (2^6E0). While this change is more correct, it will have no effect.

BrentEaston: Forget the bit about 64.0, I misread the code.

sqrt3_2 * 64.0 is what dx is set to when a new HexGrid is initialised, so this == does work because it is comparing (sqrt3_2 * 64.0) to (sqrt3_2 * 64.0). It is essentially checking if dx still in it's default state.  Changing this code in this way actually changes the behavior and is technically incorrect, though it is extremely unlikely that it will ever be noticed.

PR #204, opened 2020-08-13
Remove call to deprecated method

yanlyub:


vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -438,9 +438,16 @@ public void fireKeyStroke(KeyStroke stroke) {
   }
 
   public void fireKeyStroke(NamedKeyStroke stroke) {
-    if (stroke != null && !stroke.isNull()) {
-      fireKeyStroke(stroke.getKeyStroke());
+    if (stroke == null || stroke.isNull() || stroke.getKeyStroke() == null) {
+      return;
     }
```
uckelman, line 443: Aren't you calling `stroke.getKeyStroke()` and checking it's nonnull twice in the case where it's nonnull?

yanlyub: Whoops, of course.

Fixed, rebased, squashed and force-pushed.

PR #205, opened 2020-08-14
Performance improvement of SequenceEncoder

yanlyub:
A first try at improving performance of SequenceEncoder.

One open question is, how to detect whether the primitives need escaping due to the escape character being in `[0-9truefalse.-]`.

Possible solution (pseudocode):
```
class SequenceEncoder {
  boolean dangerousDelimiter = false;

  SequenceEncoder(String delimiter) {
    if (delimiter.matches("[0-9truefalse.-]")) dangerousDelimiter = true;
  }

  void append(int i) {
    if (dangerousDelimiter)
      possiblyEscapeAndAppend(i);
    else
      fastAppend(i);
  }
}
```

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/StringArrayConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/configure/StringArrayConfigurer.java
@@ -25,7 +25,9 @@ import java.awt.event.ActionListener;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.swing.Box;
 import javax.swing.DefaultListModel;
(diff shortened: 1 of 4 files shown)


vassal-app/src/main/java/VASSAL/configure/StringArrayConfigurer.java
```diff
@@ -184,6 +186,29 @@ public static String arrayToString(String[] s) {
     if (s == null || s.length == 0) {
       return "";
     }
+
```
yanlyub, line 189: This is faster, if none (or few) commas happen to be in the input. And slower if the input contains many commas.

uckelman:
This:
```
  private static final String UGLY = "-.0123456789aeflrstu";
  private final boolean uglyDelim;

  public SequenceEncoder(char delimiter) {
    delimit = delimiter;
    uglyDelim = UGLY.indexOf(delimit) != -1;
  }

  public SequenceEncoder append(int i) {
    if (uglyDelim) {
      return append(String.valueOf(i));  
    }
    startBufferOrAddDelimiter();
    buffer.append(i);
    return this;
  }
```
and then the same for `long`, `double`, and `boolean`.

uckelman:
Ah, `double` might need a few more "ugly" characters, becuase of `Infinity` and `NaN`, and I think for scientific notation the `e` is actually capitalized, so:

```
  private static final String UGLY = "-.0123456789NaeEfiIlnrstuy";
```

uckelman:
Aside: I don't like checking `uglyDelim` on each call to `append()` as it's a constant. It's too bad we can't set a function pointer to the approriate `append()` in the ctor and never branch after that, or have a `SequenceEncoder` factory that returns ugly or non-ugly subclasses as appropriate.

uckelman:
You can do this for `char`:

```
  public SequenceEncoder append(char c) {
    if (c == '\\' || c == '\'') {
      buffer.append('\'');
      if (c == delim) {
        buffer.append('\\');
      }
      buffer.append('\'');
    }
    else {
      if (c == delim) {
        buffer.append('\\');
      }
      buffer.append(c);
    }
    return this;
  }
```

uckelman:
Also, changing the existing `buffer.append("'")` to `buffer.append('\'')` should create less garbage.

PR #209, opened 2020-08-14
Refactored command encoders

yanlyub:


vassal-app/src/main/java/VASSAL/build/module/noteswindow/SecretNotesController.java
```diff
@@ -81,11 +81,21 @@
   public static final int COL_NAME = 2;
   public static final int COL_REVEALED = 3;
 
-  // Date formatter to save and restore date/times in the save file
+  private static final String INTERNAL_DATETIME_FORMAT = "M/d/yy h:mm a";
```
uckelman, line 84: Why change the date format?

vassal-app/src/main/java/VASSAL/launch/BasicModule.java
```diff
@@ -242,20 +242,23 @@ public String encode(Command c) {
...
       return null;
     }
-    String s = encodeSubCommand(c);
+    StringBuilder sb = new StringBuilder();
+    sb.append(encodeSubCommand(c));
     String s2;
     Command[] sub = c.getSubCommands();
     if (sub.length > 0) {
-      SequenceEncoder se = new SequenceEncoder(s, COMMAND_SEPARATOR);
+      SequenceEncoder se = new SequenceEncoder(COMMAND_SEPARATOR);
```
uckelman, line 250: This isn't right. If `s` contains the separator, it won't get escaped properly now.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/noteswindow/SecretNotesController.java
yanlyub: This is a fix, it was `m/d/yy h:mm a` before i.e. `minutes/day/year hours:minutes`. Now it is `month/day/year hours:minutes`.

uckelman: How is this going to affect saved games?

yanlyub: If they were saved with minutes > 12 then it will probably break them. Otherwise the minutes will become months.

uckelman: The date's just for display purposes, so fixing the format shouldn't break anything. This ought to be tested with a log from before the change, to check that it's still loadable.

uckelman (commented):
The date's just for display purposes, so fixing the format shouldn't break anything. This ought to be tested with a log from before the change, to check that it's still loadable.

uckelman:
The ticket for the date format bug is 13336.

uckelman (approved):

PR #210, opened 2020-08-14
13275 - Ctrl+MouseWheel will zoom in and out (Command + Mousewheel on Mac, of course)

Cattlesquat:
This is actually pretty sweet, works a LOT better than I was expecting. And of course module designer has control of exactly what "zoom stops" are provided.

I don't think there's any need to provide a preference to e.g. turn this on and off, because realistically who does "ctrl + mousewheel" when they don't "mean it" anyway.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
@@ -671,7 +671,7 @@ public class Zoomer extends AbstractConfigurable implements GameComponent {
     listener = new MouseWheelListener() {
       @Override
       public void mouseWheelMoved(MouseWheelEvent e) {
-        if ((e.getScrollAmount() == 0) || !map.getComponent().hasFocus()) {
+        if (e.getScrollAmount() == 0) {
           return;
         }
         
@@ -686,6 +686,8 @@ public class Zoomer extends AbstractConfigurable implements GameComponent {
             zoomOut();
           }
         }
+        
+        map.getComponent().getParent().dispatchEvent(e); // So that the scrollbars can still find our event.
       }
     };
     
--- a/vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
@@ -689,10 +689,8 @@ public class Zoomer extends AbstractConfigurable implements GameComponent {
         map.getComponent().getParent().dispatchEvent(e); // So that the scrollbars can still find our event.
       }
     };
-    
-    if (listener != null) {
-      map.getComponent().addMouseWheelListener(listener);
-    }
+        
+    map.getComponent().addMouseWheelListener(listener);
   }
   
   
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
```diff
@@ -660,8 +666,36 @@ public void addTo(Buildable b) {
...
+          else if ((units > 0) && state.hasLowerLevel()) {
+            zoomOut();
+          }
+        }
+        
+        map.getComponent().getParent().dispatchEvent(e); // So that the scrollbars can still find our event.
+      }
+    };
+    
+    if (listener != null) {
```
uckelman, line 693: How could `listener` be null here?

vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
```diff
@@ -660,8 +666,36 @@ public void addTo(Buildable b) {
     map.getToolBar().add(zoomInButton);
     map.getToolBar().add(zoomPickButton);
     map.getToolBar().add(zoomOutButton);
-  }
 
+    // Ctrl+Mousewheel to zoom in/out.
+    listener = new MouseWheelListener() {
```
uckelman, line 671: This could be a lambda, as `MouseWheelListener` has only one method.

vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
Cattlesquat: I'm happy to turn it into one if you supply the syntax. I will confess to having googled for 10 minutes and can't figure it out. Lambdas for me make the code take up fewer lines at the expense of rendering it incomprehensible. I do need "listener" to be set for later, if that matters.

uckelman: It would look like this:
```
      listener = e -> {
        if (e.getScrollAmount() == 0) {
          return;
        }

        if ((e.getScrollType() == MouseWheelEvent.WHEEL_UNIT_SCROLL) && SwingUtils.isSelectionToggle(e)) {
          int units = e.getUnitsToScroll();

          if ((units < 0) && state.hasHigherLevel()) {
            zoomIn();
          }
          else if ((units > 0) && state.hasLowerLevel()) {
            zoomOut();
          }
        }

        map.getComponent().getParent().dispatchEvent(e); // So that the scrollbars can still find our event.
      };

Cattlesquat: done

Cattlesquat:
Should be all set (at least from previous review)

uckelman:
I am shocked at how well this works. I had thought performance would not be great, but the graphics subsystem seems able to keep up.

Cattlesquat:
I will admit that although I was more optimistic than you were, that I too was surprised. 

Similar for my "ping" feature (or "flare" or whatever I called it) -- have a look kind of cool.

PR #212, opened 2020-08-14
SequenceEncoder optimizations

uckelman:
This improves SequenceEncoder performance by around 5% when creating save game strings.


uckelman:
SequenceEncoder.Decoder is now about 12% faster.

PR #217, opened 2020-08-17
Cleanup

uckelman:


uckelman:
Closing. Going to do this PR in two parts instead.

PR #219, opened 2020-08-18
Bug 13291 fix -$$ variables in Count & Sum

BrentEaston:
Fix for Sum/Count issue. Still Investigating related issue.


BrentEaston:
Looking at a more comprehensive fix that will also fix the backward compatibility issue reported by jrwatts.

PR #220, opened 2020-08-18
13296 - Flare feature for map (call opponent's attention to a location)

Cattlesquat:
image

VASL has had a feature like this for a long time, and I actually had a random person come along and "insert it" into PoG as a custom class. Neither version is as good as This Awesome Feature (of course). It has also come up recently in the forums w/ various people asking for it. 

This version adds:
* Default of Alt+Click to send the ping. Ctrl+Click (an inherently terrible idea of course) is included because it was the VASL (and derivative custom class) keystroke, and so the usual bring-the-old-guys-along reasons. Possibly I could include things like other mouse buttons entirely (e.g. mousebuttons 3, 4, 5), though I want to think about it a bit since this is currently a module designer setting and not everybody would have those buttons (or want to be forced to use them). 
* Can configure the color & size of the circle
* Can choose an unanimated circle (this is the only option for the old VASL etc classes)
* But can also do a cool animated pingy-thing, w/ control of speed of animation, etc.
* Yay features!!!!!!!!!!!!!!!!!!!!!!1!1111!!!!!!!!!!!

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -495,9 +495,9 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
   @Override
   public void build(Element e) {
     ActionListener al = e1 -> {
-        if (mainWindowDock == null && launchButton.isEnabled() && theMap.getTopLevelAncestor() != null) {
-          theMap.getTopLevelAncestor().setVisible(!theMap.getTopLevelAncestor().isVisible());
-        }
+      if (mainWindowDock == null && launchButton.isEnabled() && theMap.getTopLevelAncestor() != null) {
+        theMap.getTopLevelAncestor().setVisible(!theMap.getTopLevelAncestor().isVisible());
+      }
     };
     launchButton = new LaunchButton(Resources.getString("Editor.Map.map"), TOOLTIP, BUTTON_NAME, HOTKEY, ICON, al);
     launchButton.setEnabled(false);
@@ -667,11 +667,11 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
       new MandatoryComponent(this, StackMetrics.class)).append(idMgr);
 
     final DragGestureListener dgl = dge -> {
-        if (dragGestureListener != null &&
-            mouseListenerStack.isEmpty() &&
-            SwingUtils.isDragTrigger(dge)) {
-          dragGestureListener.dragGestureRecognized(dge);
-        }
+      if (dragGestureListener != null &&
+          mouseListenerStack.isEmpty() &&
+          SwingUtils.isDragTrigger(dge)) {
+        dragGestureListener.dragGestureRecognized(dge);
+      }
     };
 
     DragSource.getDefaultDragSource().createDefaultDragGestureRecognizer(
@@ -2270,9 +2270,9 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
         if (r.x + r.width > d.width) r.x = d.width - r.width;
         if (r.y + r.height > d.height) r.y = d.height - r.height;
 
-      theMap.scrollRectToVisible(r);
+        theMap.scrollRectToVisible(r);
+      }
     }
-  }
   }
 
   /**
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -520,7 +520,7 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
       addChild(new KeyBufferer());
       addChild(new ImageSaver());
       addChild(new CounterDetailViewer());
-      setMapName("Main Map"); //FIXME shouldn't this be localized?
+      setMapName(Resources.getString("Map.main_map"));
     }
     if (getComponentsOf(GlobalProperties.class).isEmpty()) {
       addChild(new GlobalProperties());
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -520,6 +520,7 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
       addChild(new KeyBufferer());
       addChild(new ImageSaver());
       addChild(new CounterDetailViewer());
+      addChild(new Flare());
       setMapName(Resources.getString("Map.main_map"));
     }
     if (getComponentsOf(GlobalProperties.class).isEmpty()) {
@@ -530,6 +531,9 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
     }
     if (getComponentsOf(HighlightLastMoved.class).isEmpty()) {
       addChild(new HighlightLastMoved());
+    }
+    if (getComponentsOf(Flare.class).isEmpty()) {
+      addChild(new Flare());
     }
     setup(false);
   }
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -92,6 +92,7 @@ import VASSAL.build.module.map.CounterDetailViewer;
 import VASSAL.build.module.map.DefaultPieceCollection;
 import VASSAL.build.module.map.DrawPile;
 import VASSAL.build.module.map.Drawable;
+import VASSAL.build.module.map.Flare;
 import VASSAL.build.module.map.ForwardToChatter;
 import VASSAL.build.module.map.ForwardToKeyBuffer;
 import VASSAL.build.module.map.GlobalMap;
@@ -179,7 +180,7 @@ import VASSAL.tools.swing.SwingUtils;
  * contained in the <code>VASSAL.build.module.map</code> package
  */
 public class Map extends AbstractConfigurable implements GameComponent, MouseListener, MouseMotionListener, DropTargetListener, Configurable,
-  UniqueIdManager.Identifyable, ToolBarComponent, MutablePropertiesContainer, PropertySource, PlayerRoster.SideChangeListener {
+    UniqueIdManager.Identifyable, ToolBarComponent, MutablePropertiesContainer, PropertySource, PlayerRoster.SideChangeListener {
   protected static boolean changeReportingEnabled = true;
   protected String mapID = ""; //$NON-NLS-1$
   protected String mapName = ""; //$NON-NLS-1$
@@ -663,8 +664,8 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
 
     final DragGestureListener dgl = dge -> {
       if (dragGestureListener != null &&
-        mouseListenerStack.isEmpty() &&
-        SwingUtils.isDragTrigger(dge)) {
+          mouseListenerStack.isEmpty() &&
+          SwingUtils.isDragTrigger(dge)) {
         dragGestureListener.dragGestureRecognized(dge);
       }
     };
@@ -1521,10 +1522,10 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
         final Rectangle vrect = scroll.getViewport().getViewRect();
 
         if ((sx == -1 && vrect.x == 0) ||
-          (sx ==  1 && vrect.x + vrect.width >= theMap.getWidth())) sx = 0;
+            (sx ==  1 && vrect.x + vrect.width >= theMap.getWidth())) sx = 0;
 
         if ((sy == -1 && vrect.y == 0) ||
-          (sy ==  1 && vrect.y + vrect.height >= theMap.getHeight())) sy = 0;
+            (sy ==  1 && vrect.y + vrect.height >= theMap.getHeight())) sy = 0;
 
         // Stop if the scroll vector is zero
         if (sx == 0 && sy == 0) scroller.stop();
@@ -2024,7 +2025,7 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
       if (mainWindowDock != null) {
         if (mainWindowDock.getHideableComponent().isShowing()) {
           Prefs.getGlobalPrefs().getOption(MAIN_WINDOW_HEIGHT)
-            .setValue(mainWindowDock.getTopLevelAncestor().getHeight());
+               .setValue(mainWindowDock.getTopLevelAncestor().getHeight());
         }
         mainWindowDock.hideComponent();
         toolBar.setVisible(false);
@@ -2143,7 +2144,7 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
       // If no piece at destination and this is a stacking piece, create
       // a new Stack containing the piece
       if (!(p instanceof Stack) &&
-        !Boolean.TRUE.equals(p.getProperty(Properties.NO_STACK))) {
+          !Boolean.TRUE.equals(p.getProperty(Properties.NO_STACK))) {
         final Stack parent = getStackMetrics().createStack(p);
         if (parent != null) {
           c = c.append(placeAt(parent, pt));
@@ -2265,9 +2266,9 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
         if (r.x + r.width > d.width) r.x = d.width - r.width;
         if (r.y + r.height > d.height) r.y = d.height - r.height;
 
-        theMap.scrollRectToVisible(r);
-      }
+      theMap.scrollRectToVisible(r);
     }
+  }
   }
 
   /**
@@ -2494,7 +2495,7 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
   public Class<?>[] getAllowableConfigureComponents() {
     return new Class<?>[]{ GlobalMap.class, LOS_Thread.class, ToolbarMenu.class, MultiActionButton.class, HidePiecesButton.class, Zoomer.class,
       CounterDetailViewer.class, HighlightLastMoved.class, LayeredPieceCollection.class, ImageSaver.class, TextSaver.class, DrawPile.class, SetupStack.class,
-      MassKeyCommand.class, MapShader.class, PieceRecenterer.class };
+      MassKeyCommand.class, MapShader.class, PieceRecenterer.class, Flare.class };
   }
 
   @Override
@@ -2699,7 +2700,7 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
     @Override
     public Object visitStack(Stack s) {
       if (s.getPosition().equals(pt) && map.getStackMetrics().isStackingEnabled() && !Boolean.TRUE.equals(p.getProperty(Properties.NO_STACK))
-        && s.topPiece() != null && map.getPieceCollection().canMerge(s, p)) {
+          && s.topPiece() != null && map.getPieceCollection().canMerge(s, p)) {
         return map.getStackMetrics().merge(s, p);
       }
       else {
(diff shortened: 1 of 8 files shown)


Cattlesquat:
[Count_them_Pieces.ZIP](https://github.com/vassalengine/vassal/files/5091411/Count_them_Pieces.ZIP)
This module (rename to VMOD, github required me to name it ZIP) will let you give it a quick test. Just start the module and Alt+Click around the map. You can edit the little "Map Flare" entry on the map if you want to play with configuring it.

vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -507,7 +508,7 @@ public void build(Element e) {
       addChild(new KeyBufferer());
       addChild(new ImageSaver());
       addChild(new CounterDetailViewer());
-      setMapName("Main Map");
+      setMapName("Main Map"); //FIXME shouldn't this be localized?
```
uckelman, line 511: Yes.

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,343 @@
...
+import VASSAL.build.module.GameComponent;
+import VASSAL.build.module.GlobalOptions;
+import VASSAL.command.CommandEncoder;
+import VASSAL.build.AbstractConfigurable;
+import VASSAL.build.AutoConfigurable;
+import VASSAL.command.FlareCommand;
+import VASSAL.configure.ColorConfigurer;
+import VASSAL.configure.StringEnum;
+import VASSAL.i18n.Resources;
+import VASSAL.tools.swing.SwingUtils;
```
uckelman, line 47: Sort the JDK imports into one group, ours into another.

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,343 @@
...
+  public String getDescription() {
+    return Resources.getString("Map Flare");
+  }
+  
+  public static String getConfigureTypeName() {
+    return Resources.getString("Map Flare");
+  }
+
+  public Class<?>[] getAttributeTypes() {
+    return (Class<?>[]) new Class[] { FlareKeyConfig.class, Integer.class, Color.class, Integer.class, Integer.class };
```
uckelman, line 100: No need to cast, just do `new Class<?>[]`.

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,343 @@
...
+    }
+    else if (CIRCLE_COLOR.equals(key)) {
+      return new StringBuilder().append(ColorConfigurer.colorToString(this.color)).toString();
+    }
+    else if (PULSES.equals(key)) {
+      return new StringBuilder().append(this.pulses).toString();
+    }
+    else if (PULSES_PER_SEC.equals(key)) {
+      return new StringBuilder().append(this.pulsesPerSec).toString();
+    }
```
uckelman, line 130: For all of these, you're not doing any concatenation, so you can just return `String.valueOf()` for the numbers, and `ColorConfigurer.colorToString(this.color)` directly.

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,343 @@
...
+  
+
+
+  public Flare() {
+    this.circleSize = 100;
+    this.color = Color.RED;
+    this.active = false;
+    this.flareKey = FLARE_ALT;
+    this.pulses       = 6;
+    this.pulsesPerSec = 3;
```
uckelman, line 88: I think every `this` in the whole file is superfluous. (You generally only need it in places where there's potential for ambiguity.)

vassal-app/src/main/java/VASSAL/command/FlareCommand.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package VASSAL.command;
+
+import java.awt.Point;
+
+import VASSAL.build.module.map.Flare;
+
+public class FlareCommand extends Command {
+  private Flare finder;
+  private Point clickPoint;
```
uckelman, line 26: These could both be final, as they're set only in the ctor.

vassal-app/src/main/java/VASSAL/command/FlareCommand.java
```diff
@@ -0,0 +1,49 @@
...
+  protected Command myUndoCommand() {
+    return null;
+  }
+    
+  public int getValue() {
+    return 0;
+  }
+    
+  public Point getClickPoint() {
+    return this.clickPoint;
```
uckelman, line 47: You don't need any of the this'es in this file either.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,343 @@
...
+    if (this.animate) {
+      this.framesPerPulse = 30 / this.pulsesPerSec;
+      if (this.framesPerPulse < 1) this.framesPerPulse = 1;
+      this.frames = this.pulses * framesPerPulse;
+    } 
+    else {
+      this.frames = DEFAULT_FRAMES;
+    }
+    
+    this.map.getView().repaint();
```
uckelman, line 222: You could limit the repaint just to the dirty area.

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,343 @@
...
+      if (animate) {
+        diameter = diameter * (framesPerPulse - (frame % framesPerPulse)) / framesPerPulse;
+      }
+      if (diameter <= 0) {
+        return;
+      }
+
+      // translate the piece center for current zoom
+      final Point p = map.mapToDrawing(clickPoint, os_scale);                             
+
```
uckelman, line 240: Setting an antialiasing hint here improves rendering quality quite a bit:

```
g2d.setRenderingHint(RenderingHints.KEY_ANTIALIASING,
                     RenderingHints.VALUE_ANTIALIAS_ON);
```

uckelman (changes requested):

uckelman:
I can't get Alt+Click to work on Linux. (Ctrl+Click works fine.)

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
Cattlesquat: I am interested in doing that - can you point me to some sample/related code?

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
uckelman: There's more than one way you can do it. If you have view coordinates handy, you can call the four-argument version of `repaint(x, y, w, h)`,

If you have map coordinates handy, which you do, you can do this:
```
map.repaint(new Rectangle(
  clickPoint.x - circleSize / 2 - 3,
  clickPoint.y - circleSize / 2 - 3,
  circleSize + 6,
  circleSize + 6
));
```
That includes a bit of padding for the stroke width. (It _shouldn't_ extend 3px outward in all directions, given that it's only got a width of 3.)

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,346 @@
...
+      GameModule.getGameModule().removeCommandEncoder(this);
+    }
+  }
+
+
+  public void startAnimation(final boolean isLocal) {
+    if (!isLocal) {
+      map.centerAt(this.clickPoint);
+    }
+    active = true;
```
BrentEaston, line 209: Should the centring respect the 'Centre on Opponents Moves' preference? If a player has turn off move centering for whatever reason, this could potentially be very annoying. My feeling is it should respect the preference. You can use 

if (GlobalOptions.getInstance().centerOnOpponentsMove()) {

What do think about adding a Message Format for reporting Flares? 

<Bob(Germany)> dropped a map flare at U01.03

This would allow the person who has map re-centering turned off to at least know they happened.

BrentEaston: At the risk of adding features no-one has asked for, it would be neat to have an option so that a text input box pops up when you request a flare and you can type text in and this text appears in the message format, and hovers over the Flare. Of course, players can just type that text into the chat box. Wait for someone to ask for it :)

Cattlesquat:
@BrentEaston Please have a look at latest ReportFormat code - it's "working" but I'm sure there are better ways to do some things.

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,481 @@
...
+  private boolean circleScale;
+  private int pulses;
+  private int pulsesPerSec;
+  private boolean animate;
+  private String flareKey;
+  private Color color;
+  private Point clickPoint;
+  private FormattedString reportFormat = new FormattedString();
+  private volatile boolean active;
+  private PropertySource propertySource;
```
BrentEaston, line 75: Remove, not needed.

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,481 @@
...
+    reportFormat.setProperty(FLARE_LOCATION, map.locationName(clickPoint));
+    final Zone z = map.findZone(clickPoint);
+    reportFormat.setProperty(FLARE_ZONE, (z != null) ? z.getName() : "");
+    reportFormat.setProperty(FLARE_MAP, map.getMapName());
+    reportFormat.setProperty(GlobalOptions.PLAYER_NAME, (String) mod.getProperty(GlobalOptions.PLAYER_NAME));
+    reportFormat.setProperty(GlobalOptions.PLAYER_NAME_ALT, (String) mod.getProperty(GlobalOptions.PLAYER_NAME_ALT));
+    reportFormat.setProperty(GlobalOptions.PLAYER_SIDE, (String) mod.getProperty(GlobalOptions.PLAYER_SIDE));
+    reportFormat.setProperty(GlobalOptions.PLAYER_SIDE_ALT, (String) mod.getProperty(GlobalOptions.PLAYER_SIDE_ALT));
+    reportFormat.setProperty(GlobalOptions.PLAYER_ID,   (String) mod.getProperty(GlobalOptions.PLAYER_ID));
+    String reportText = reportFormat.getLocalizedText(propertySource);
```
BrentEaston, line 425: This is sufficient

    reportFormat.setProperty(FLARE_LOCATION, map.locationName(clickPoint));
    final Zone z = map.findZone(clickPoint);
    reportFormat.setProperty(FLARE_ZONE, (z != null) ? z.getName() : "");
    reportFormat.setProperty(FLARE_MAP, map.getMapName());
    String reportText = reportFormat.getLocalizedText(map);

Passing the map to getLocalizedText will satisify all of the Player Name/Id/Side options as map.getProperty() drops through to GameModule.getProperty() which satisifies the GlobalOptions properties.

BrentEaston (changes requested):

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
BrentEaston: Plus you get the bonus advantage that the designer can specify any Map or Module level properties in the Message Format (using $$ notation) and these will also be satisfied by passing the map as the PropertySource.

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
BrentEaston: Just the PropertySource.

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,583 @@
...
+    animator.start();
+  }
+
+  /**
+   * Draw the Flare at current animation frame
+   * @param g - Graphics
+   * @param map - Map component
+   */
+  public void draw(final Graphics g, final Map map) {
+    if (active && clickPoint != null) {
```
yanlyub, line 395: I'd invert this condition and do an early return, then the whole method doesn't need to be indented an extra level:
```
if (!active || clickPoint == null) {
  return;
}

// rest of the code
```

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,583 @@
...
+  }
+
+  /**
+   * Encodes Flare commands into string form.
+   * @param c Command
+   * @return String version of Flare Command, or null if not a Flare Command.
+   */
+  public String encode(final Command c) {
+    if (c instanceof FlareCommand) {
+      return COMMAND_PREFIX + ((FlareCommand) c).getClickPoint().x + "," + ((FlareCommand) c).getClickPoint().y + "," + ((FlareCommand) c).getMapId();
```
yanlyub, line 438: Would make the comma a constant and use it here and in `decode()`:
```
private static final String DELIMITER = ",";
String encode() {
  return field1 + DELIMITER + field2 + DELIMITER + field3;
}
Command decode(String s) {
  Decoder sd = new Decoder(s, DELIMITER);
}
```

vassal-app/src/main/java/VASSAL/tools/SequenceEncoder.java
```diff
@@ -223,6 +223,15 @@ public Decoder(String value, char delimiter) {
       stop = val != null ? val.length() : 0;
     }
 
+    /**
+     * Decoder constructor that accepts a string for the delimiter (so that a single constant value can be defined for a module, e.g. private static final String DELIMITER = "\t";)
```
uckelman, line 227: I'm not grasping what this is for. What prevents defining a `private static final char DELIMITER = '\t';`?

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,674 @@
...
+    // translate the click location for current zoom
+    final Point p = map.mapToDrawing(clickPoint, os_scale);
+    g2d.setRenderingHint(RenderingHints.KEY_ANTIALIASING,
+            RenderingHints.VALUE_ANTIALIAS_ON);
+
+    // draw a circle around the selected point
+    g2d.setColor(color);
+    g2d.setStroke(new BasicStroke((float)(STROKE * os_scale)));
+    g2d.setRenderingHint(RenderingHints.KEY_ANTIALIASING,
+            RenderingHints.VALUE_ANTIALIAS_ON);
```
uckelman, line 448: No need to set the hints twice. Kill the earlier one?

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,674 @@
...
+    }
+
+    public String[] getI18nKeys(AutoConfigurable target) {
+      return new String[] {
+        FLARE_ALT_LOCAL,
+        SystemUtils.IS_OS_MAC_OSX && !GlobalOptions.getInstance().getPrefMacLegacy() ? FLARE_COMMAND_LOCAL : FLARE_CTRL_LOCAL,
+        FLARE_ALT_SHIFT_LOCAL,
+        SystemUtils.IS_OS_MAC_OSX && !GlobalOptions.getInstance().getPrefMacLegacy() ? FLARE_SHIFT_COMMAND_LOCAL : FLARE_CTRL_SHIFT_LOCAL,
+        SystemUtils.IS_OS_MAC_OSX && !GlobalOptions.getInstance().getPrefMacLegacy() ? FLARE_ALT_COMMAND_LOCAL : FLARE_CTRL_ALT_LOCAL,
+        SystemUtils.IS_OS_MAC_OSX && !GlobalOptions.getInstance().getPrefMacLegacy() ? FLARE_ALT_SHIFT_COMMAND_LOCAL : FLARE_CTRL_ALT_SHIFT_LOCAL,
```
uckelman, line 659: Could get the pref once and store it the value.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/tools/SequenceEncoder.java
Cattlesquat: So the problem is that some (all?) of the other places the delimiter is used, you actually want/need _String_ not _char_. Meaning that for your trouble of trying to do a good job defining your delimiter rather than just hardcoding it everywhere, you end up getting punished by either having to define two different names (e.g. DELIMITER_STRING and DELIMITER_CHAR) -- which is both bad and inconvenient, or you have to manually convert the String to a char when it is used, which is also kind of bad. So really the encoder should just accept the same String version of the delimiter that is most useful for all other purposes where these are coming up. (see the various uses of DELIMITER in this PR for a tour).

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
Cattlesquat: oh that's funny

uckelman (approved):

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
```diff
@@ -0,0 +1,673 @@
...
+  }
+
+  /**
+   * Serializes Flare commands into string form.
+   * @param c Command
+   * @return String version of Flare Command, or null if not a Flare Command.
+   */
+  public String encode(final Command c) {
+    if (c instanceof FlareCommand) {
+      return COMMAND_PREFIX + ((FlareCommand) c).getId() + DELIMITER + ((FlareCommand) c).getClickPoint().x + DELIMITER + ((FlareCommand) c).getClickPoint().y;
```
uckelman, line 470: These are the places you say you need `DELIMITER` to be a `String`? I think it would work with `DELIMITER` as a `char` as well.

uckelman: Also, it's dangerous not to use `SequenceEncoder` if any of the items you're encoding are `Strings` which you don't control. E.g., what happens if `getId()` has `DELIMITER` in it?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/map/Flare.java
Cattlesquat: I am certainly not an expert on "why we use the delimiters that we use" in the various SequenceEncoders. I started with a fairly traditional/naive delimiter (e.g. "," or something), and Brent suggested I change that to "\t" or something "for reasons", and then Yan suggested I define as DELIMITER so that I had it all in one place rather than hard-coding it. This seemed like a plausible idea, but when I attempted to do that I discovered that while "usually" the delimiters appear being +'ed into strings e.g. "Thing" + "\t" + "OtherThing", that SequenceDecoder (only) really wanted a char '\t' or whatever. Which then made the whole idea of "making the code less confusing and/or more maintainable" by defining DELIMITER seem farcical unless the same thing could be used in all places, and the easiest candidate for change was letting SequenceDecoder accept a string delimiter. I suppose we could throw an exception if it was other than exactly 1 character long. 

Anyway if you don't like the simplifications I'm happy to go back to hard-coding "\t" or whatever "like most classes do". Or if you do like them, just let me know which way you want it although trying to append `char`'s into the middle of Strings is surprisingly inconvenient in Java so I'm not a fan of that.

Cattlesquat: Okay the char version of DELIMITER "just works". God knows what I was doing. Probably forgot to declare DELIMITER to be char and didn't notice that was the only problem.

Pulled out the special bonus version of SequenceEncoder, which earned me the opportunity to "mergetool" it about 6 consecutive times during the rebase :D

Cattlesquat:
Did a "safe" SequenceEncoder. (Now that I know what one is)

Cattlesquat:
map.findZone(clickPoint) was MURDEROUSLY writing bullshit into MY point!

uckelman (approved):

PR #235, opened 2020-08-21
13355: changes to EnsureVisible method map.java

claudiociardelli:
Modified method EnsureVisible. Chenged behaviour, instead of showing a rectangle with the moved piece on one of the edges, it will show a rectangle so that the piece moved is in the center.

In addition the centering on opponent move will be triggered by a unit that is already visible by the opponent but that moves "too close" to the border of the visible area. 
For the moment "Too close" is defined as a band around the edges of 5% of the height or width of the visible rectangle.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/GlobalOptions.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GlobalOptions.java
@@ -68,6 +68,7 @@ public class GlobalOptions extends AbstractConfigurable {
   public static final String NON_OWNER_UNMASKABLE = "nonOwnerUnmaskable"; //$NON-NLS-1$
   public static final String PROMPT_STRING = "promptString"; //$NON-NLS-1$
   public static final String CENTER_ON_MOVE = "centerOnMove"; //$NON-NLS-1$
+  public static final String CENTER_ON_MOVE_SENSITIVITY = "centerOnMoveSensitivity"; //$NON-NLS-1$
   public static final String MARK_MOVED = "markMoved"; //$NON-NLS-1$
   public static final String AUTO_REPORT = "autoReport"; //$NON-NLS-1$
   public static final String CHATTER_HTML_SUPPORT = "chatterHTMLSupport"; //$NON-NLS-1$
@@ -214,6 +215,11 @@ public class GlobalOptions extends AbstractConfigurable {
     
     BooleanConfigurer config = new BooleanConfigurer(CENTER_ON_MOVE, Resources.getString("GlobalOptions.center_on_move"), Boolean.TRUE); //$NON-NLS-1$
     prefs.addOption(config);
+
+    //CC// center_on_move_border_proximity_pct (is the pct of distance from border to center of window that triggers a recenter
+    final IntConfigurer pctRecenterOn = new IntConfigurer(CENTER_ON_MOVE_SENSITIVITY,
+      Resources.getString("GlobalOptions.center_on_move_sensitivity"), 10 ); //$NON-NLS-1$
+    prefs.addOption(pctRecenterOn);
        
     validator = new SingleChildInstance(gm, getClass());
   }
--- a/vassal-app/src/main/java/VASSAL/build/module/GlobalOptions.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GlobalOptions.java
@@ -506,13 +506,11 @@ public class GlobalOptions extends AbstractConfigurable {
   }
 
   public double centerOnOpponentsMoveSensitivity() {
-    Integer intSensitivity = (int) GameModule.getGameModule().getPrefs().getValue(CENTER_ON_MOVE_SENSITIVITY);
-    double sensitivity;
-    sensitivity = intSensitivity.doubleValue();
+    int sensitivity = (Integer) GameModule.getGameModule().getPrefs().getValue(CENTER_ON_MOVE_SENSITIVITY);
     if (sensitivity > 100) {
       sensitivity = 100;
     }
-    if (sensitivity < 0) {
+    else if (sensitivity < 0) {
       sensitivity = 0;
     }
     return sensitivity;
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2226,10 +2227,38 @@ public void centerAt(Point p, int dx, int dy) {
...
+      // If r is already visible decide if unit is close enough to
+      // border to justify a recenter
+
+      // if r is within a band of  n%width/height of border, trigger recenter
+      r_norecenter.width = (int) round(r_current.width * no_recenter_pct);
+      r_norecenter.height = (int) round(r_current.height * no_recenter_pct);
+      r_norecenter.x = r_current.x + (int) round(r_current.width - r_norecenter.width) / 2;
+      r_norecenter.y = r_current.y + (int) round(r_current.height - r_norecenter.height) / 2;
+
+      if (p.x < r_norecenter.x) bTriggerRecenter = true;
```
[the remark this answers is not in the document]
claudiociardelli: Agreed, but leaving it in lines makes it more readable. I personally prefer that.

vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2226,10 +2227,38 @@ public void centerAt(Point p, int dx, int dy) {
   /** Ensure that the given region (in map coordinates) is visible */
   public void ensureVisible(Rectangle r) {
     if (scroll != null) {
+      final double no_recenter_pct = 0.9;
```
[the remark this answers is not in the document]
claudiociardelli: Thx for pointing that out. I Refactored all the variables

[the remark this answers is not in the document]
Cattlesquat: Cool. Be sure to "push" your new changes after you commit them on your local branch, so that they will appear here.

vassal-app/src/main/java/VASSAL/build/module/GlobalOptions.java
```diff
@@ -498,7 +504,20 @@ public boolean autoReportEnabled() {
   public boolean centerOnOpponentsMove() {
     return Boolean.TRUE.equals(GameModule.getGameModule().getPrefs().getValue(CENTER_ON_MOVE));
   }
-  
+
+  public double centerOnOpponentsMoveSensitivity() {
+    Integer intSensitivity = (int) GameModule.getGameModule().getPrefs().getValue(CENTER_ON_MOVE_SENSITIVITY);
+    double sensitivity;
```
uckelman, line 510: You don't need to turn this into a double yourself; you can just return the int and it will be upcast.

vassal-app/src/main/java/VASSAL/build/module/GlobalOptions.java
```diff
@@ -498,7 +504,20 @@ public boolean autoReportEnabled() {
...
+    sensitivity = intSensitivity.doubleValue();
+    if (sensitivity > 100) {
+      sensitivity = 100;
+    }
+    if (sensitivity < 0) {
```
uckelman, line 515: Should be an `else if`.

vassal-app/src/main/java/VASSAL/build/module/GlobalOptions.java
uckelman: I think what you want here is:
```
int sensitivity = (Integer) GameModule.getGameModule().getPrefs().getValue(CENTER_ON_MOVE_SENSITIVITY);
```

vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2226,10 +2227,43 @@ public void centerAt(Point p, int dx, int dy) {
...
+      Rectangle rNorecenter = new Rectangle(0, 0);
+
+      // If r is already visible decide if unit is close enough to
+      // border to justify a recenter
+
+      // Close enough means a strip of the window along the edges whose
+      // width is a % of the edge to center of the window
+
+      // The % is defined in GlobalOptions.CENTER_ON_MOVE_SENSITIVITY
+      final double noRecenterPct = (100 - GlobalOptions.getInstance().centerOnOpponentsMoveSensitivity()) / 100;
```
uckelman, line 2242: If you want floating point division, you need to ensure one of the operands is a floating point type, say by using `100.0` instead of `100`.

vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2226,10 +2227,43 @@ public void centerAt(Point p, int dx, int dy) {
...
+      rNorecenter.width = (int) round(rCurrent.width * noRecenterPct);
+      rNorecenter.height = (int) round(rCurrent.height * noRecenterPct);
+      rNorecenter.x = rCurrent.x + (int) round(rCurrent.width - rNorecenter.width) / 2;
+      rNorecenter.y = rCurrent.y + (int) round(rCurrent.height - rNorecenter.height) / 2;
+
+      if (p.x < rNorecenter.x) bTriggerRecenter = true;
+      if (p.x > (rNorecenter.x + rNorecenter.width)) bTriggerRecenter = true;
+      if (p.y < rNorecenter.y) bTriggerRecenter = true;
+      if (p.y > (rNorecenter.y + rNorecenter.height)) bTriggerRecenter = true;
+
```
uckelman, line 2254: This unnecessarily evaluates conditions after you've already set `bTriggerRecenter` to true.

Better would be:
```
bTriggerRecenter = p.x < rNorecenter.x || p.x > (rNorecenter.x + rNorecenter.width) ||
                   p.y < rNorecenter.y || p.y > (rNorecenter.y + rNorecenter.height);
```

vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2226,10 +2227,43 @@ public void centerAt(Point p, int dx, int dy) {
...
+      if (bTriggerRecenter == true) {
```
uckelman, line 2255: You never need to check booleans for equality. Just write `if (bTriggerRecenter) {`.

uckelman (changes requested):

uckelman (approved):

PR #237, opened 2020-08-22
Toggle password visibility

yanlyub:


uckelman:
The height of the whole row is bad: 

image

uckelman:
The best way is to use a MigLayout for these.

yanlyub:
> The height of the whole row is bad:
> image

I can't reproduce that, I increased the row height to 40 now, hope it's enough.

uckelman:
Setting fixed sizes for those components isn't the best. The last commit here (https://github.com/uckelman/vassal/commits/pw_layout) uses different layout.

uckelman:
Closing in favor of PR 343, which also contains the commit from this PR.

PR #238, opened 2020-08-22
3.5 - 1807: ChessClocks - These BadBoys might be Ready To Rip!!!

Cattlesquat:
image
image image

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
@@ -20,6 +20,7 @@ package VASSAL.build.module.chessclockcontrol;
 import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.Font;
+import java.awt.Insets;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.ArrayList;
@@ -506,7 +507,7 @@ public class ChessClock extends AbstractConfigurable implements CommandEncoder, 
             baseline = String.format("%d0:00.0", minutes/10); //NON-NLS
           }
           else {
-            baseline = String.format("0:00.0"); //NON-NLS
+            baseline = "0:00.0"; //NON-NLS
           }
         }
         else {
@@ -515,7 +516,7 @@ public class ChessClock extends AbstractConfigurable implements CommandEncoder, 
             baseline = String.format("%d0:00", minutes/10); //NON-NLS
           }
           else {
-            baseline = String.format("0:00"); //NON-NLS
+            baseline = "0:00"; //NON-NLS
           }
         }
       }
@@ -523,7 +524,7 @@ public class ChessClock extends AbstractConfigurable implements CommandEncoder, 
 
     // adjust size to prevent button wobbling
     final Insets ins = timerButton.getInsets();
-    final int width = ins.left + timerButton.getFontMetrics(timerButton.getFont()).stringWidth(getFormattedButtonText() + baseline + ins.right;
+    final int width = ins.left + timerButton.getFontMetrics(timerButton.getFont()).stringWidth(getFormattedButtonText() + baseline + ins.right);
     timerButton.setPreferredSize(new Dimension(width, timerButton.getPreferredSize().height));
 
     return tocking != oldTocking;


vassal-app/src/main/java/VASSAL/build/module/Map.java.orig
```diff
@@ -0,0 +1,3499 @@
+/*
```
uckelman, line 1: This file shouldn't have been committed.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -455,6 +465,18 @@ protected boolean allSidesAllocated() {
...
+  /**
+   * @param side Name of a side to see if it's a "solo side"
+   * @return True if the side is "Solitaire", "Solo", "Moderator", or "Referee"
+   */
+  public static boolean isSoloSide(String side) {
+    return SOLITAIRE.equals(side) ||
+           SOLO.equals(side)      ||
+           MODERATOR.equals(side) ||
+           REFEREE.equals(side);
+  }
```
uckelman, line 477: This is going to be fun when someone is playing a Start Wars game... Han may need to shoot first...

Is there a better way we can do this than string matching?

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -475,10 +497,7 @@ protected String promptForSide() {
     int i = (myidx >= 0) ? ((myidx + 1) % sides.size()) : 0;   // If we do, start looking in the "next" slot, otherwise start at beginning.
     for (int tries = 0; i != myidx && tries < sides.size(); i = (i + 1) % sides.size(), tries++) { // Wrap-around search of sides
       String s = sides.get(i);
-      if (!alreadyTaken.contains(s) &&
-          !Resources.getString("PlayerRoster.solitaire").equals(s) &&
-          !Resources.getString("PlayerRoster.solo").equals(s) &&
-          !Resources.getString("PlayerRoster.referee").equals(s)) {
```
uckelman, line 481: It looks like we were translating these? How's that handled now?

Cattlesquat: Oh I should change that to call the isSoloSide thinger. Didn't spot that one.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
Cattlesquat: That's definitely something in the back of my mind. When I pulled this out to be a separate method I was definitely thinking about making it a single-centralized-method for handling this kind of check, and that we'd probably eventually want to "do a better job". I think the "right thing to do" is to extend the PlayerRoster to let Solo/Moderator player sides be explicitly named by the module designer w/ some kind of checkbox or something. Which seems more like a future PR?

vassal-app/src/main/java/VASSAL/build/module/Map.java.orig
Cattlesquat: Oh I see it was THAT day.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
Cattlesquat: Oh right, this is stuff I took out. "I" was thinking these were translated here but we weren't actually calling the translate-the-side method it is just using straight side names.

Which, now that you mention it, I think this has just "been wrong" for a long time. I will have a look.

Cattlesquat: So confusing. In "PlayerRoster", "sides" means translated sides. But in "PlayerInfo", "side" means untranslated side. Anyway I have now set it to be used consistent with the way this stuff is apparently done.

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+   * @return Our command for restoring from a saved game (or adding an online player)
+   */
+  public Command getRestoreCommand() {
+    return new UpdateClockControlCommand(chessClocksVisible, onlineGame);
+  }
+
+  /**
+   * Our "command" format for passing information about the master clock control between computers (or to/from save and log files)
+   */
+  public class UpdateClockControlCommand extends Command {
```
uckelman, line 691: Does this need to be public?

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+      }
+      else if (command.contains(CHESSMENU_HIDE)) {
+        pressControlButton();
+      }
+    }
+
+    /**
+     * Build our context menu
+     */
+    void buildPopup() {
```
uckelman, line 772: private?

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+      item = new JMenuItem(CHESSMENU_NEXT + "  " + NamedHotKeyConfigurer.getString(nextListener.getNamedKeyStroke()));
+      item.addActionListener(this);
+      popup.add(item);
+    }
+
+    /**
+     * Build and display our context menu
+     * @param p Coordinates for popup
+     */
+    void doPopup(Point p) {
```
uckelman, line 814: private?

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+  public String getShowDays() {
+    return showDays;
+  }
+
+  /**
+   * @return the number of chess clocks currently ticking/active
+   */
+  public int getClocksTicking() {
+    int running = 0;
+    if (instanceIsActive && (getInstance() != null)) {
```
uckelman, line 449: Why do we need `getInstance()` in a non-static context?

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -0,0 +1,826 @@
...
+  public ChessClock(String side) {
+    this();
+    this.side       = side;
+    this.buttonText = side;
+    setConfigureName(side);
+    setTimerButton();
+  }
+
+
+  public ChessClock() {
```
uckelman, line 115: Could the `ChessClockControl` be passed in to this ctor instead of being fetched using static methods?

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -0,0 +1,826 @@
...
+
+    image        = "";
+    elapsedTime  = 0;
+    verifiedTime = 0;
+    clockTicking = false;
+
+    // Timer is static so it gets created exactly once ever
+    if (timer == null) {
+      timer = new Timer(100, this);
+      timer.stop();
```
uckelman, line 155: Is this necessary? I thought timers didn't start on creation.

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -0,0 +1,826 @@
...
+  public static final String GENERIC   = "Player";
+
+  private static Color defaultColor;       // Stores default look-and-feel colors for buttons
+  private static Color defaultFontColor;
+
+  protected Color tickingBackgroundColor = new Color(255, 255, 0);  // Starting (default) colors
+  protected Color tickingFontColor       = new Color(51, 51, 51);
+  protected Color tockingFontColor       = new Color(0, 0, 0);
+
+  protected static Timer timer;    // Millisecond timer (but only actually updated every ~100ms)
```
uckelman, line 90: If you have only one timer, should it be non-static and in the `ChessClockControl` instead of static and here?

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -0,0 +1,826 @@
...
+  protected void startTimer() {
+    if (!timer.isRunning()) {
+      timer.start();
+    }
+  }
+
+  /**
+   *  Defines the Command by which information about this clock is communicated between machines (and to/from save and log files)
+   */
+  public class UpdateTimerCommand extends Command {
```
uckelman, line 234: Does it need to be public?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
yanlyub: Can it be static? Inner classes should be static whenever they can.

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+
+  /**
+   * When we need to find a ChessClockControl in a static context
+   * @return First ChessClockControl in our module
+   */
+  public static ChessClockControl getInstance() {
+    GameModule gm = GameModule.getGameModule();
+    if (gm.getAllDescendantComponentsOf(ChessClockControl.class).size() > 0) {
+      return gm.getAllDescendantComponentsOf(ChessClockControl.class).get(0);
+    }
```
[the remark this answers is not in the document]
yanlyub: `!Collection.isEmpty()` is more readable than `Collection.size() > 0`

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+/**
+ * CHESS CLOCK CONTROLLER for VASSAL.
+ *
+ * This Module-level component manages a set of one or more {@link ChessClock} clocks. It creates a master "Chess Clocks" button on the
+ * Module's toolbar. Clicking the button shows/hides the clocks themselves, when no clocks are running, or stops all clocks if one
+ * is running. The Chess Clock Control component also contains the configurable elements that specify how the clocks are to be displayed.
+ */
+public class ChessClockControl extends AbstractConfigurable
+        implements CommandEncoder, GameComponent, UniqueIdManager.Identifyable {
+  protected static UniqueIdManager idMgr = new UniqueIdManager("ChessClockControl"); //$NON-NLS-1$
```
yanlyub, line 65: Should be `final`, with an UPPERCASE_NAME

yanlyub: And if the class is designed to be subclassable, also `private`, subclasses will need their own IdManager.

If the class is not designed to be subclassable, make the class `final` and replace all `protected` with `private`.

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+
+  /**
+   * Builds the clock control's component hierarchy from a given XML element, or a null one is given initializes
+   * a brand new default "new clock control" hierarchy.
+   * @param e XML element to build from, or null to build the default hierarchy
+   */
+  @Override
+  public void build(Element e) {
+    super.build(e);
+    if (e == null) {
```
yanlyub, line 162: To reduce indentation levels, consider an early return:
```
if (e != null) {
  return;
}
// rest here
```

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+  }
+
+  /**
+   * Sets the value of one of this component's XML attributes
+   * @param key the name of the attribute. Will be one of those listed in {@link #getAttributeNames}
+   * @param value New value for the attribute. Can be either String version or the actual Object.
+   */
+  @Override
+  public void setAttribute(String key, Object value) {
+    if (NAME.equals(key)) {
```
yanlyub, line 311: A `switch` here?

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+  }
+
+  /**
+   * Gets the value of one of this component's XML attributes
+   * @param key the name of the attribute. Will be one of those listed in {@link #getAttributeNames}
+   * @return String value of the attribute.
+   */
+  @Override
+  public String getAttributeValueString(String key) {
+    if (NAME.equals(key)) {
```
yanlyub, line 363: And possibly a `switch` here

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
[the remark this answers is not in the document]
yanlyub: Maybe even like this:
```
return gm.getAllDescendantComponentsOf().stream().findFirst().orElse(null);
```

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+    }
+  }
+
+  /**
+   * Deserializes our command from a string version, if the command belongs to us.
+   * @param command Serialized string command
+   * @return An {@link UpdateClockControlCommand}
+   */
+  public Command decode(final String command) {
+    if (command.startsWith(COMMAND_PREFIX)) {
```
yanlyub, line 651: Possibly an early return here

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+    }
+  }
+
+  /**
+   * Serializes our command into a string, if it belongs to us
+   * @param c Command to serialize. Only serialized if it's an UpdateClockControlCommand.
+   * @return Serialized command, or null if command passed wasn't an UpdateClockControlCommand.
+   */
+  public String encode(final Command c) {
+    if (c instanceof UpdateClockControlCommand) {
```
yanlyub, line 669: And early return here

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+    public boolean isLoggable() {
+      return false;
+    }
+  }
+
+
+  /**
+   * Processes right-clicks on our button to put up a context menu
+   */
+  class ChessMouseListener implements MouseListener, ActionListener {
```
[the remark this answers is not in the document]
yanlyub: And static if possible

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -0,0 +1,826 @@
...
+ * Clicking the button starts and stops the clock, which we attempt to keep (roughly) in sync across multiple
+ * players' computers. A pair (or more, or I suppose only one) of them are then added to {@link ChessClockControl}.
+ *
+ * @author Brian Reynolds and Michael Kiefte
+ *
+ * Class originally created by Michael Kiefte for Twilight Struggle, including handshaking for verification of time totals between online
+ * machines. Made more generically configurable, ChessClockControl added, and integrated into standard VASSAL, by Brian Reynolds.
+ *
+ */
+public class ChessClock extends AbstractConfigurable implements CommandEncoder, GameComponent, ActionListener {
```
yanlyub, line 65: A heretical question, but why don't we split the Model from the View here?

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -0,0 +1,826 @@
...
+  private String getFormattedButtonText() {
+    return (StringUtils.isEmpty(buttonText) ? "" : buttonText + " ");
+  }
+
+  /**
+   * Updates clock text on the button, based on our configured style
+   * @return half second tick-tock pulse on/off
+   */
+  public boolean setTimerButton() {
+    long time = elapsedTime;
```
yanlyub, line 414: Can't we use a library here to do these calculations e.g Java 8's Date/Time API?

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -0,0 +1,826 @@
...
+    return new Class[0];
+  }
+
+  /**
+   * Deserializes our command from a string version, if the command belongs to us.
+   * @param command Serialized string command
+   * @return An {@link UpdateTimerCommand}
+   */
+  public Command decode(final String command) {
+    if (command.startsWith(COMMAND_PREFIX)) {
```
yanlyub, line 672: Early return

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -0,0 +1,826 @@
...
+    }
+  }
+
+  /**
+   * Serializes our command into a string, if it belongs to us
+   * @param c Command to serialize. Only serialized if it's an UpdateTimerCommand.
+   * @return Serialized command, or null if command passed wasn't an UpdateTimerCommand.
+   */
+  public String encode(final Command c) {
+    if (c instanceof UpdateTimerCommand) {
```
yanlyub, line 696: Early return

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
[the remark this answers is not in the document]
Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
Cattlesquat: I think I "inherited that". I'll take it out we can hope it works :)

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
Cattlesquat: apparently it can be private, so it is. Changing it to static made lots of things not compile.

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+      command.execute();
+      GameModule.getGameModule().sendAndLog(command);
+    });
+    GameModule.getGameModule().addKeyStrokeListener(pauseListener);
+  }
+
+  /**
+   * @return Our list of chessclocks
+   */
+  public Set<ChessClock> getChessClocks() {
```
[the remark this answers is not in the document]
uckelman: So, there's no order guarantee on a `Set`. I figured you might care about iteration order?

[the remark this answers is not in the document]
Cattlesquat: I do, actually - what's the right thing instead? A List?

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -0,0 +1,843 @@
...
+
+    if (chessclocks.isEmpty()) { // If we don't have any clocks, can't start one.
+      return command;
+    }
+
+    Iterator<ChessClock> it = chessclocks.iterator();
+    ChessClock first = null;
+    boolean found = false;
+    while (it.hasNext()) {
+      ChessClock clock = it.next();
```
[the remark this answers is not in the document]
Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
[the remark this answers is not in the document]
uckelman: Yes, `List`.

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
yanlyub: The `idMgr`, should be `final`, and since it then would be a constant it should have an uppercase name, `ID_MGR`.

Like what we did here, but since this is new code it's not too late to change the field name to the proper convention for final static fields: https://github.com/vassalengine/vassal/pull/352/files

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -310,6 +316,10 @@ protected static String getMySide(boolean localized) {
     return players.toArray(new PlayerInfo[0]);
   }
 
+  public List<String> getSides() {
+    return (ArrayList)sides;
```
uckelman, line 320: You don't need that cast.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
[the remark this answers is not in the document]
Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
Cattlesquat: A thing I liked about the static version was that, in the admittedly oddball case where someone had *multiple* ChessClock controllers, they'd all still share a single timer instance.

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
Cattlesquat: done

uckelman:
Fix yer conflict, yo.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -310,6 +316,10 @@ protected static String getMySide(boolean localized) {
     return players.toArray(new PlayerInfo[0]);
   }
 
+  public List<String> getSides() {
+    return sides;
+  }
```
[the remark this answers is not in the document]
Cattlesquat: fixed

uckelman:
So, I'm finally trying this now (!!!). When I start a clock with tenths of seconds, the width of the display keeps changing which causes the right end of it to wobble.

uckelman:
Do you want to provide that chess clock icon as the default?

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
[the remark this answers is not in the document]
uckelman: I was going to say that you could return an unmodifiable list instead of a new one? But I guess this is fine.

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -447,43 +448,80 @@ public boolean setTimerButton() {
...
+    String baseline;
 
     if (doDays) {
       if (doTenths) {
         timerButton.setText(String.format("%s%d:%02d:%02d:%02d.%d", getFormattedButtonText(), days, hours, minutes, seconds, tenths)); //NON-NLS
+        baseline = "0:00:00:00.0"; //NON-NLS
       }
       else if (doSeconds) {
         timerButton.setText(String.format("%s%d:%02d:%02d:%02d", getFormattedButtonText(), days, hours, minutes, seconds)); //NON-NLS
+        baseline = "0:00:00:00.0"; //NON-NLS
```
uckelman, line 460: Is having the tenths in the string here a copy-paste mistake?

Cattlesquat: Yes :)

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -515,15 +516,15 @@ else if (doSeconds) {
             baseline = String.format("%d0:00", minutes/10); //NON-NLS
           }
           else {
-            baseline = String.format("0:00"); //NON-NLS
```
uckelman: :)

PR #239, opened 2020-08-23
Fix MarkTB's Text Label problem -- apparently was using same style sheet as BasicHTML

Cattlesquat:
Always nice to fix something by *removing* two lines!


uckelman:
Cherry-picked to PR 240.

PR #242, opened 2020-08-23
Add private server type, remove Jabber server type

uckelman:


BrentEaston (commented):
Looks good. Just a couple of comments.

I don't think you need (or want) to commit any changes to the _xx.properties files. My understanding is that if you remove a key or change a translation in the main properties files, Crowdin will take care of removing the translation from the _xx files or marking that key as needing a re-translate.

Second issue is with the server stats button (Connect with red question mark). I have selected a private server (which I can't connect to), but the server stats button is still showing the module stats from the main Vassal server. Should it be trying to get those from the Private server? I'm not sure about the message posting stuff either, is that being directed to the correct server?

uckelman:
> I have selected a private server (which I can't connect to), but the server stats button is still showing the module stats from the main Vassal server. Should it be trying to get those from the Private server? I'm not sure about the message posting stuff either, is that being directed to the correct server?

This stuff depends on our web server and shouldn't be displayed at all for a private server. I'm going to do a bit of refactoring so it's not displayed for private servers.

uckelman:
I've reverted the changes to the translations.

BrentEaston (approved):

PR #243, opened 2020-08-24
13353: Use native file chooser on linux

yanlyub:


vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
```diff
@@ -70,49 +65,156 @@ public static FileChooser createFileChooser(Component parent, DirectoryConfigure
...
    */
   public static FileChooser createFileChooser(Component parent, DirectoryConfigurer prefs, int mode) {
-    FileChooser fc;
-    if (SystemUtils.IS_OS_MAC_OSX) {
-      // Mac has a good native file dialog
-      fc = new NativeFileChooser(parent, prefs, mode);
-    }
-    else if (mode == FILES_ONLY && SystemUtils.IS_OS_WINDOWS) {
-      // Window has a good native file dialog, but it doesn't support selecting directories
-      fc = new NativeFileChooser(parent, prefs, mode);
```
uckelman, line 80: Have you checked that the file choosers this switches us to support selecting directories?

yanlyub: Checked and changed this PR to use the native dialog only for file load/save on Linux, while keeping the Swing dialog for selecting directories.

uckelman (approved):

PR #244, opened 2020-08-24
13323 - Player switching sides now reported in Chat Log

Cattlesquat:
image

Previously players could switch sides w/ little transparency. This isn't foolproof of course but it's good transparency.

Also cleaned up various little code nubbins.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
@@ -18,13 +18,12 @@
 package VASSAL.build.module;
 
 import java.awt.Component;
-import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
-import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -245,27 +239,33 @@ protected void launch() {
...
+    Add a = new Add(this, me.playerId, me.playerName, me.side);
     a.execute();
-    GameModule.getGameModule().getServer().sendToOthers(a);
+
+    c = c.append(a);
+    gm.getServer().sendToOthers(c);
 
     newSide = getMySide();
-    fireSideChange(oldSide, newSide);
+    fireSideChange(mySide, newSide);
```
uckelman, line 268: How does this work? Why can we get rid of `oldSide`?

Cattlesquat: Since I discovered the code for that method looked like this: :)
```java
protected void launch() {
    final String mySide = getMySide();
    if (mySide == null && allSidesAllocated()) {
      return;
    }

    final String oldSide = getMySide();
```

uckelman: Derp.

PR #249, opened 2020-08-27
3.5 - 13333 - Better handling of logfile/savefile comments

Cattlesquat:
image
image
image

image

(1) Clarified text that this properly appears in "Save" files as well as "Log" files.
(2) When loading a file with metadata comments, display the comments more vividly.
(3) When saving a file, ALSO display the metadata comments if any (better feedback to comment maker that comments are being accepted, plus a clue to where they might be viewed)
(4) Clearer explanation of what is being prompted for, at time of prompt.
(5) Display filename being loaded when loading (I constantly wish for this, so now whee!)
(6) Moved the prompt for comments to before the (potentially long) save-file-write process.


uckelman:
Is that an actual screenshot from that dialog? Why doesn't the massively long text wrap?

Cattlesquat:
image
Eek, no! It looks like this now. When I make things better I should remember to update early shitty screenshots :)

BrentEaston:
Hi Brian,

A couple of things.

1. Why are there extensive Chatter changes in this PR? Are they supposed to be there?

2. If I create a new log file and save it as test2.vlog, then close, create a new game and the create a new log file to overwrite the test2.vlog I just created, I get this dialog
image

Otherwise looking good.

BrentEaston (commented):
Query about the Chatter changes being in this PR and the bug with the version checking when overwriting a log file

Cattlesquat:
So #1 is -- the path I'm using to send things to the chatter sends them as "system messages" (you know the ones that begin with the "-", as opposed to game message that begin with a "*"). And up until this PR, unfortunately only the GAME messages could use the color-styling codes. Which it turns out was an oversight. So I just added the color styling code to the "system message" path in the Chatter so that I could correctly send them as system messages, but have them come out as the nice green messages you see in the screenshot. This will, not incidentally, now allow those styling to work in the future for any future system messages that need to be emitted through the chatter.

And #2 is -- that's some message that Yan & Joel have somehow conjured into existence. It has nothing to do with me or this PR. I think part of why you're seeing it is the weird maven numbering thing about "3.5-SNAPSHOT" being "before 3.5" but it sort of thinks its 3.5 or something like that. So basically whenever you're saving and loading stuff in "debug mode" in IntelliJ you get that as essentially a garbage message. I sort of suspect that we're actually getting it too many times even not in debug mode (among other things I saw Yan & Joel discussing recently how people were getting the message and it was too frightening and so Joel rewrote it a bit to make it less friendly), but anyway it may be getting shown a bit overkill amount of times but they seem to know about it, understand it, etc, and I haven't really gotten a brain dump about it. So anyway -- not related to this PR even though this PR is about log files  (go load and save some logfiles in master in IntelliJ and it will do the same garbage I'm pretty sure)

BrentEaston (approved):

PR #256, opened 2020-08-28
13326 - Saving a module will default to .VMOD, an extension to .VEXT

Cattlesquat:
Up until now, if you make a new module, tell it to "save" or "save as", and type in "Name", you get a module named literally "Name", with no extension. 

This fix corrects that, adding ".vmod" to module saves if no fileext has been added, and ".vext" to a module extension missing a fileext.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
+++ b/vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
@@ -34,13 +34,18 @@ import VASSAL.tools.imageop.Op;
 import VASSAL.tools.io.FileArchive;
 import VASSAL.tools.io.IOUtils;
 import VASSAL.tools.io.ZipArchive;
+import org.apache.commons.lang3.StringUtils;
+
+import javax.swing.JOptionPane;
 
 /**
  * An ArchiveWriter is a writeable DataArchive. New files may be added
- * with the {@link #addFile} and {@link #addImage} methods.
+ * with the {@link #addFile} and {@link #addImage} methods. {@link #save()} and {@link #saveAs()} will
+ * cause the archive to be written out, with FileChooser invoked if appropriate.
  */
 public class ArchiveWriter extends DataArchive {
   private String archiveName;
+  private String defaultExtension;
   private boolean isTempArchive = false;
 
   /**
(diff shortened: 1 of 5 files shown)


vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
```diff
@@ -217,19 +242,38 @@ public void addFile(String fileName, byte[] content) {
     }
   }
 
+  /**
+   * Saves the archive, prompting for a name only if none has ever been provide.
```
uckelman, line 246: d

vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
```diff
@@ -217,19 +242,38 @@ public void addFile(String fileName, byte[] content) {
...
+   * @throws IOException IOException
+   */
   public void save() throws IOException {
     save(false);
   }
 
+  /**
+   * Saves the archive, prompting for a name only if none has ever been provide.
```
uckelman, line 254: d

vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
```diff
@@ -240,13 +284,31 @@ protected void write(FileArchive fa, boolean notifyModuleManager)
...
-    final String filename = fc.getSelectedFile().getPath();
+    String filename = fc.getSelectedFile().getPath();
+
+    if (!StringUtils.isEmpty(defaultExtension) && (filename.lastIndexOf('.') < 0)) {
+      if (defaultExtension.indexOf('.') >= 0) {
+        filename = filename + defaultExtension;
+      }
+      else {
+        filename = filename + "." + defaultExtension;
+      }
```
uckelman, line 307: Would it be simpler to expect the default extension to supply its own `.`, and then not check for it here?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
Cattlesquat: Fair enough. I'm probably overengineering given the small number of places/times this needs to get called.

PR #259, opened 2020-08-29
Add JsonSaver feature

yanlyub:


BrentEaston (changes requested):
Hi Yan,

There are several things I think we need to look at.

1. As it stands, this can't released. You are exporting properties that can not normally be seen by an opponent. I did a test and created and Obscured Counter, then changed sides, On the board, I can only see the mask back (a '?' marker) and the Text saver shows me:

K2: ?

The Concealed counter just shows its Concealed name and there is no clue to what it could be, as you would expect.

However, this is what the Json Saver shows me:

  "Map" : {
    "K2" : [ {
      "_name" : "?",
      "Nation" : "German",
      "_Active" : true,
      "OldLocationName" : "",
      "Stype" : "Squad",
      "CurrentY" : 296,
      "Berserk_Active" : false,
      "OldBoard" : "",
      "PieceName" : "?",
      "LocationName" : "K2",
      "Side" : "Axis",
      "OldZone" : "",
      "ClickedX" : 770,
      "BasicName" : "8-3-8 Assault Engineer",
      "ClickedY" : 298,
      "OldX" : 149,
      "_Name" : "",
      "OldY" : 42,
      "Berserk_Name" : "",
      "_Level" : 1,
      "TextLabel" : "A",
      "CurrentZone" : "",
      "_Image" : "ge-MMC 838.gif",
      "DeckName" : "",
      "CurrentMap" : "Map",
      "Berserk_Image" : "",
      "Berserk_Level" : "-1",
      "Type" : "Infantry",
      "CurrentBoard" : 1,
      "ObscuredToOthers" : true,
      "Obscured" : true,
      "CurrentX" : 762,
      "InvisibleToOthers" : false,
      "Level" : "Ground",
      "OldMap" : "",
      "Inventory" : true
    },

Aha! So that is what my opponent is hiding from me.

Two Obscured units with the same back MUST return exactly the same representation. If a Unit has the Property Oscured=true, then the only properties you should report are Obscured, LocationName and PieceName and possibly CurrentX, CurrentY, CurrentXZone and CurrentMap.


2. I think this would be better if we combined the Text Saver and the Json Saver into a single 'Export Map Data' component. It should pop up a dialog with options to select
 a) Export data from my point of view or from Opponents point of view?
 b) Output format. Current options would be Plaint Text (Text Saver) or Json.

This gives us the flexibility to easily add other formats later and combines two almost identical components doing the same thing into one.

3. The amount of data output on a moderate scenario in a relatively uncomplicated game was staggering. Most of it was of no interest whatsoever. I think we need some way to select a subset of properties to export. Even a Basic Info and Full Info option would be good (Full = what you have now, Basic = name, position info, not much else?). What would be really nice would be to select and/or list what properties you want exported and have that saved as a preference.

PR #261, opened 2020-08-29
MouseOverStackViewer show n pieces of decks. Previously would only show 1

claudiociardelli:
Previously you could activate the "show top piece in deck" feature to show only the top piece.
Now there are 2 settings for. MouseOverStackViewer.
The checkbox toglles on or off the feature as before. Then the integer field is to specify how many pieces of the deck are displayed. Default to 1

Possible improvement is to use a value as -1 to show all?


vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -620,17 +622,38 @@ public Visitor(Filter filter, Map map, Point pt, boolean showOverlap) {
 
     @Override
     public Object visitDeck(Deck d) {
+      //CC// Display top n pieces of the deck instead of top piece
+
+      //if (foundPieceAt == null) {
```
Cattlesquat, line 627: Go ahead and remove the old crud that you're replacing. It shows up in the diff when we review your PR anyway, and if PR is approved and merged we won't want the old code sitting around.

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -272,7 +272,8 @@ Editor.MouseOverStackViewer.non_stacking=Include non-stacking pieces?
 Editor.MouseOverStackViewer.move_selected=Include move-when-selected pieces?
 Editor.MouseOverStackViewer.non_moveable=Include non-movable pieces?
 Editor.MouseOverStackViewer.unrotated_state=Show pieces in un-rotated state?
-Editor.MouseOverStackViewer.top_deck=Include top piece in Deck?
+Editor.MouseOverStackViewer.show_deck=Show top pieces of a Deck
```
Cattlesquat, line 275: Since all the other related options say "Include..." I suggest changing "Show" to "Include" in this show_deck message too, just for additional clarity that they are all related options. It's fine to leave "show" in the show_deck_depth message though, I think that works well.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -868,7 +892,8 @@ else if (drawSingleDeprecated) {
        Resources.getString("Editor.MouseOverStackViewer.move_selected"), //$NON-NLS-1$
        Resources.getString("Editor.MouseOverStackViewer.non_moveable"), //$NON-NLS-1$
        Resources.getString("Editor.MouseOverStackViewer.unrotated_state"), //$NON-NLS-1$
-       Resources.getString("Editor.MouseOverStackViewer.top_deck"), //$NON-NLS-1$
+       Resources.getString("Editor.MouseOverStackViewer.show_deck"), //$NON-NLS-1$
+      Resources.getString("Editor.MouseOverStackViewer.show_deck_depth"), //$NON-NLS-1$
```
Cattlesquat, line 896: Is there some funny indenty-spacey thing going on with this line?

BrentEaston (approved):

PR #266, opened 2020-08-30
Lobotomize BasicModule, moving all its tasty goodness into the slavering maw of GameModule

Cattlesquat:
It is with the greatest shock that I discover this seemed to go really smoothly. Nobody subclasses BasicModule anyway, right? 

This seems to work, and run, and yay. 

Give it a peek and see what YOU think...

ALSO, I understand GameModule even more better than I ever understood it before.


vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -109,20 +136,24 @@
...
- * <p>It is a <a href="https://en.wikipedia.org/wiki/Singleton_pattern">singleton</a>, and contains access points for many other classes,
- * such as {@link DataArchive}, {@link ServerConnection}, {@link Logger},
- * and {@link Prefs}.</p>
+ * For legacy reasons, {@link BasicModule} extends {@link GameModule}'s abstract class, but
+ * everything of note has been moved inside of GameModule.
+ *
+ * <p>GameModule is a <a href="https://en.wikipedia.org/wiki/Singleton_pattern">singleton</a>, and contains access points for many other classes,
+ * such as {@link DataArchive}, {@link ServerConnection}, {@link Logger}, {@link Chatter}, and {@link Prefs}.</p>
  */
 public abstract class GameModule extends AbstractConfigurable implements CommandEncoder, ToolBarComponent, PropertySource, MutablePropertiesContainer, GpIdSupport {
```
uckelman, line 158: Removing `abstract` doesn't break binary compatibility, so we should do that.

uckelman (changes requested):

uckelman (approved):

uckelman (approved):

PR #268, opened 2020-08-31
13098 - Move pure Compatibility preferences to their own tab

BrentEaston:
Move Compatibility preferences to their own preference tab.
Also added a bit of text to the bug fix compatibility prefs to give an idea of what problems they solve and hopefully make it easier for users to find themselves.


vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
```diff
@@ -524,7 +524,7 @@ GlobalOptions.auto_report=Auto-report moves?
 GlobalOptions.mark_moved=Mark moved pieces?
 GlobalOptions.initial_heap=JVM initial heap (in MB):
 GlobalOptions.maximum_heap=JVM maximum heap (in MB):
-GlobalOptions.bug10295=Drag ghost bug correction?
+GlobalOptions.bug10295=Drag ghost bug correction? (Use if shadow image missing when dragging counters)
```
uckelman, line 527: Does this still happen for anybody?

PR #269, opened 2020-08-31
[bugfix] use month instead of minute for date format

grimreaper:
Problem 
INTERNAL_DATE_FORMATTER is intended to provide a complete datestring but repeats the minute twice

Solution
Use "M" which means month.


uckelman:
This needs testing with a save or log written from before the change, to see what happens when you try to load it.

uckelman:
I got word just now that this was already tested in PR #209, so no need for further testing.

Thanks for the PR, but I'm closing it because PR #209 has the same change.

PR #271, opened 2020-08-31
3.5 - 2569 - Friendlier error message (and fewer ABR reports) for 2569

Cattlesquat:
This should put 2569 to bed. Although the player continuing after this may well lead to another exception of some type, that was already the case if the chose to continue after the default "Eek! A bug!" message, and this way we should get fewer ABR reports on this (because even if they get a later exception they're less likely to want to report)

I suppose we could actually abort after this popup, but that's not our usual practice.

Another idea is "after this" not _offering_ error log report on subsequent exceptions, but getting appropriate guardrails on that might take some work.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
+++ b/vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
@@ -436,7 +436,7 @@ public class ProblemDialog {
   /**
    * Used when SequenceEncoder#next_token runs out of data when it was expecting more.
    * @param usage - information about what was being sequenced
-   * @return Future<?> - pass to get() to wait for dialog to close.
+   * @return Future<?> - Call the get() method of the return value to wait for dialog to close.
    */
   public static Future<?> showOutdatedModule(String usage) {
     return showDisableable(JOptionPane.WARNING_MESSAGE,
--- a/vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
+++ b/vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
@@ -433,4 +433,18 @@ public class ProblemDialog {
   }
 
 
+  /**
+   * Used when SequenceEncoder#next_token runs out of data when it was expecting more.
+   * @param usage - information about what was being sequenced
+   * @return Future<?> - pass to get() to wait for dialog to close.
+   */
+  public static Future<?> showOutdatedModule(String usage) {
+    return showDisableable(JOptionPane.WARNING_MESSAGE,
+      null, null, usage,
+      Resources.getString("Dialogs.module_version_incompatible"),
+      Resources.getString("Dialogs.module_version_incompatible"),
+      Resources.getString("Dialogs.module_version_incompatible_sequence", usage) + "\n\n"
+        + Resources.getString("Dialogs.check_for_updated_vmod")
+    );
+  }
 }
--- a/vassal-app/src/main/java/VASSAL/tools/SequenceEncoder.java
+++ b/vassal-app/src/main/java/VASSAL/tools/SequenceEncoder.java
@@ -21,6 +21,7 @@ import java.awt.Color;
 import java.awt.event.InputEvent;
 import java.util.Arrays;
 import java.util.Iterator;
+import java.util.NoSuchElementException;
 
 import javax.swing.KeyStroke;
 
@@ -235,13 +236,7 @@ public class SequenceEncoder {
     }
 
     public String nextToken() {
-      if (!hasMoreTokens()) {
-        //BR// This tends to happen when a player loads some save/log file written by the wrong version of a module.
-        //BR// Rather than have a trickle of ABR reports about it, print a more helpful error message.
-        ProblemDialog.showOutdatedModule ("");
-        //throw new NoSuchElementException();
-        return "";
-      }
+      if (!hasMoreTokens()) throw new NoSuchElementException();
 
       if (start == stop) {
         // token for "null" is the empty string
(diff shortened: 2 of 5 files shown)


vassal-app/src/main/java/VASSAL/tools/ProblemDialog.java
```diff
@@ -397,4 +397,18 @@ private static Frame getFrame() {
   }
 
 
+  /**
+   * Used when SequenceEncoder#next_token runs out of data when it was expecting more.
+   * @param usage - information about what was being sequenced
+   * @return Future<?> - pass to get() to wait for dialog to close.
```
uckelman, line 403: You don't pass it to get(), you _call_ the Future's get() method.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/tools/SequenceEncoder.java
```diff
@@ -236,7 +235,13 @@ public boolean hasMoreTokens() {
     }
 
     public String nextToken() {
-      if (!hasMoreTokens()) throw new NoSuchElementException();
+      if (!hasMoreTokens()) {
+        //BR// This tends to happen when a player loads some save/log file written by the wrong version of a module.
+        //BR// Rather than have a trickle of ABR reports about it, print a more helpful error message.
+        ProblemDialog.showOutdatedModule ("");
```
uckelman, line 241: Is there another place this could go, where we catch NoSuchElementException, _outside_ of Decoder? By putting the error handling here, we're tying what is a data processing utility class to the GUI.

Cattlesquat: Hmmm. I feel like SequenceEncoder is implemented & called "a myriad different places"? I might ordinarily suggest putting a try/catch in the various ways of Loading games, but then there's the "load in background" issue. 

Thoughts?

uckelman: Yeah, it's used in... basically every game component there is, which makes dealing with this kind of irritating.

Are there really situations where this exception is recoverable? I've always been under the impression that you're totally screwed when it happens.

Cattlesquat: So yes you're probably basically totally screwed. But we should still display THIS error message instead of the other one. *Possibly* we should exit without allowing the option to continue (keep in mind that the Exception box ALSO lets you try to continue, as inadvisable as that often is). Possibly I should make the message I'm showing even more "no really you're screwed, don't try to keep playing" than I already have.

BrentEaston: I'm with Joel, I don't think SequecnceEncoder.Decoder is the right place. It has no context to be able to report a meaningful error message.

Since the 2569 error occurs in Decorator.mergeState() which is only called from ChangePiece.executeCommand(), I think it would be better to do the error reporting in ChangePiece. It can actually report the name of the piece that has the problem, and give some guidance as to what should happen. 

When this happens in ChangePiece, we can be pretty sure that the player is trying to run a log file containing ChangePiece Commands generated on an out of date module and can give some advice on this. And display a Stupid User dialog, not the ABR dialog and don't generate an ABR.

If we stick this error handling in SequenceEncoder.Decoder, it could be caused by anything and the error message will have to be so generic as to be useless. The message you are showing for this error only applies to the ChangePiece case, not to any other SE.D problems.

I would vote to put SE.D back to throwing the NoSuchElementException so that we DO get an ABR if this exception happens anywhere else other than ChangePiece.execute. 

You know what we really need is a Version number in each piece's Type that is incremented each time the Type saved in the PieceSlot changes.

BrentEaston (changes requested):

vassal-app/src/main/java/VASSAL/tools/SequenceEncoder.java
Cattlesquat: Okay I *think* I have set this to do what you describe -- please check it out.

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
```diff
@@ -75,6 +75,9 @@ Dialogs.out_dated_usage=This module contains custom code that is calling into VA
 Dialogs.incompatible.title=Incompatible Custom Code
 Dialogs.incompatible.heading=The Custom Code In This Module Should Be Recompiled
 Dialogs.incompatible.message=This module contains custom Java code (%1$s) which was not compiled to be Java 11 compatible. As a result, this module will not run on all versions of Java which VASSAL itself supports.
+Dialogs.module_version_incompatible=File Written by Different Version of Module
+Dialogs.module_version_incompatible_sequence=The file you are attempting to load may have been written by a different version of this module, and your version is therefore unable to find some of the data it is looking for. Continuing with this version may not faithfully represent the complete game state. 
+Dialogs.check_for_updated_vmod=Please check whether there is an updated version of this module (or if on the other hand your opponent needs an update)
```
BrentEaston, line 80: I would just make that last line:
'Please check check that you and your opponent are using the same version of the module.'

BrentEaston (changes requested):

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
BrentEaston: Otherwise good to go.

uckelman:
Would you rebase this on release-3.4.3? Then I'll take it for the release.

uckelman:
Cherry-picked to release-3.4.3.

PR #273, opened 2020-08-31
3.5 - 13352 - Title Bar of module should always show most recent Save/Load/Log

Cattlesquat:
If you are making a log file you get a nice " - logging to BLAHBLAH.vlog" appended to the titlebar of all your windows.

But if you're doing anything else (like waiting for your opponent to send you the next file, or more importantly trying to figure out WHICH of your Seven paths-of-glory windows you should be loading the brand new log file he just sent you into) then the title bar is infuriatingly silent.

But NO MORE! Now the last file you interacted with (vsav or vlog) and the mode-of-interaction (saved/loaded/logging/logged/replaying/replayed) is displayed.


vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -829,21 +864,97 @@ public JToolBar getToolBar() {
...
+    else if (GameFileMode.LOGGING_GAME.equals(gameFileMode)) {
+      return Resources.getString(key + "_title_logging", name, gameFile); //NON-NLS-1$
+    }
+    else if (GameFileMode.LOGGED_GAME.equals(gameFileMode)) {
+      return Resources.getString(key + "_title_logged", name, gameFile); //NON-NLS-1$
     }
     else {
-      frame.setTitle(frame.getTitle() + s);
+      return "**Unknown Game File Mode** - " + gameFile;
     }
```
uckelman, line 897: You could replace the entire contents of this function with this, by taking advantage of the string names of the enum values:
```
    if (StringUtils.isEmpty(gameFile) || GameFileMode.NEW_GAME.equals(gameFileMode)) {
      return Resources.getString(key + "_title", name);  //NON-NLS-1$
    }
    else {
      return Resources.getString(key + "_title_" + gameFileMode, name, gameFile); //NON-NLS-1$
    }
```

uckelman: We don't want the extremal case to be possible, right? The only way that can happen is if somehow `gameFileMode` has gotten set to null. I think we should just disallow that---throw an NPE in `setGameFile` and `setGameFileMode` if someone tries to set it null.

uckelman (changes requested):

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties.orig
```diff
@@ -0,0 +1,551 @@
+# Module Editor Properties
```
uckelman, line 1: This file shouldn't have been committed?

Cattlesquat: Merge tool detritus

uckelman:
Should the save state be reset when I "Close Game"?

Cattlesquat:
Agreed. Now resets save state on Close Game.

vassal-app/src/main/java/VASSAL/build/module/Map.java.orig
```diff
@@ -0,0 +1,3523 @@
+/*
+ *
+ * Copyright (c) 2000-2012 by Rodney Kinney, Joel Uckelman, Brent Easton
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
```
BrentEaston, line 6: Oops

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties.orig
```diff
@@ -0,0 +1,929 @@
+# VASSAL Properties file
+# Contains all translatable strings for the playable user interface
+# Strings for the Module Editor are in Editor.properties
+#
+# NOTE: Comments to the translator to provide context can be provided
+#       on a # comment line preceding the translation key
```
BrentEaston, line 6: Double Oops, anothrer .orig

BrentEaston (changes requested):
Nice change, looks good. Tested well. Need to resolve the conflicts and get rid of the .orig files and good to go.

vassal-app/src/main/java/VASSAL/build/module/Map.java.orig
Cattlesquat: oops indeed

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties.orig
Cattlesquat: Musta been a mergetool kinda day!

Cattlesquat:
Okay, conflicts and .orig files should be fixed up.

PR #274, opened 2020-08-31
13356 game refresher does not copy marker state

BrentEaston:
Markers are essentially constants and if we change one and run the Refresher, we expected the changed value to be put into the updated piece, not the old value copied from the old piece.


uckelman (approved):

BrentEaston:
That name is technically incorrect, the bug was caused by GameRefresher copying the Marker state, the fix is to not copy the Marker state :)

PR #279, opened 2020-09-01
Javadoc3 - MAP! 

Cattlesquat:
The neverending javadoc project continues, with Map.java coming under the microscope

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -1104,9 +1104,9 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
    * Examples: Drawing a line between two points on a map (see {@link LOS_Thread#draw}. Drawing a piece on the map
    *           (see {@link StackMetrics#draw}.
    *
-   * @param c value to scale
+   * @param c value in Map coordinate space to be scaled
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -1640,30 +2066,52 @@ public void drawPieces(Graphics g, int xOffset, int yOffset) {
...
+  /**
+   * Draws all of our "Drawable" components. Standard examples include {@link CounterDetailViewer}s (aka Mouse-over Stack Viewers),
+   * {@link GlobalMap}s (aka Overview Maps), {@link LOS_Thread}s, {@link MapShader}s, and the {@link KeyBufferer} (to show which
+   * pieces are selected).
+   * @param g target graphics object
+   * @param aboveCounters true means we should draw only the drawables that go above the counters; false means we should draw only the ones that go below
+   */
   public void drawDrawable(Graphics g, boolean aboveCounters) {
     for (Drawable drawable : drawComponents) {
-      if (!(aboveCounters ^ drawable.drawAboveCounters())) {
```
uckelman, line 1645: Wow, I have no idea why we wrote that in such an unobvious way.

Cattlesquat: Probably somebody came in from some mental side-tangent of "did any bits change", and tricked themselves into thinking they needed to xor to find which ones changed

PR #280, opened 2020-09-01
13363 - buildFile renamed to buildFile.xml for 3.5 and beyond

Cattlesquat:
Now featuring a file extension for Extra Sanity!

* Looks for "buildFile" for all versions 3.4.999 and earlier. 
* Looks for "buildFile.xml" for all versions after that
* Writes buildFile.xml
* And then removes old "buildFile" instances lingering in the archive

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -22,6 +22,8 @@ import VASSAL.build.module.GameRefresher;
 import VASSAL.build.module.KeyNamer;
 import VASSAL.build.module.PluginsLoader;
 import VASSAL.build.module.gamepieceimage.GamePieceImageDefinitions;
+import VASSAL.build.module.metadata.AbstractMetaData;
+import VASSAL.build.module.metadata.MetaDataFactory;
 import VASSAL.build.module.properties.GlobalProperties;
 import VASSAL.chat.AddressBookServerConfigurer;
 import VASSAL.chat.ChatServerFactory;
@@ -49,6 +51,7 @@ import java.beans.PropertyChangeSupport;
 import java.io.BufferedInputStream;
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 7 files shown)


vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -337,18 +341,29 @@ protected void build() throws IOException {
       build(null);
     }
     else {
+      final AbstractMetaData data = MetaDataFactory.buildMetaData(f);
+      final String properBuildFileName = (VersionUtils.compareVersions(data.getVassalVersion(), "3.4.999") < 0) ? BUILDFILE_OLD : BUILDFILE;
```
uckelman, line 345: Is there a reason you can't do `VersionUtils.compareVersions(data.getVassalVersion(), "3.5")`? That ought to be negative for anything 3.4 or earlier.

vassal-app/src/main/java/VASSAL/build/module/ModuleExtension.java
```diff
@@ -116,7 +118,14 @@ public boolean getUniversal() {
   }
 
   public void build() {
-    final String fileName = GameModule.BUILDFILE;
+
+    final AbstractMetaData data = MetaDataFactory.buildMetaData(archive.getArchive().getFile());
+    final String fileName = (VersionUtils.compareVersions(data.getVassalVersion(), "3.4.999") < 0) ? GameModule.BUILDFILE_OLD : GameModule.BUILDFILE;
```
uckelman, line 123: Ditto.

vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
```diff
@@ -197,6 +197,20 @@ public void removeImage(String name) {
...
+   * @param name file in the archive to be removed
+   */
+  public void removeFile(String name) {
+    try {
+      archive.remove (name);
+    }
+    catch (IOException e) {
+      WriteErrorDialog.error(e, archive.getName());
+    }
+  }
```
uckelman, line 211: While you're at it, you could replace all but the last line of `removeImage()` with a call to `removeFile(imageDir + name)`.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/GameModule.java
Cattlesquat: That's actually what I had in the original version, I think, and then 3.5.0-SNAPSHOT turned out to be earlier?

Cattlesquat: I need a version that's either "guaranteed lower than ANY OTHER 3.5" or "guaranteed higher than ANY OTHER 3.4".

vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
uckelman: _all but the last line_. You killed the last line.

vassal-app/src/main/java/VASSAL/build/GameModule.java
uckelman: Ah, crud, right. Then you want to use `VersionUtils.trunctateToMinorVersion()` on `data.getVassalVersion()` first, to cut off anything past the minor version.

uckelman (approved):

PR #285, opened 2020-09-02
3.5 - 13368 - preference to control drag-at-edge-of-map width

Cattlesquat:
made it a preference and set default at 60 (was hardcoded at 30 before)

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -255,6 +255,8 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
   protected PieceMover pieceMover;
   protected KeyListener[] saveKeyListeners = null;
 
+  private IntConfigurer preferredScrollConfig;
+
   public Map() {
     getView();
     theMap.addMouseListener(this);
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -786,6 +786,16 @@ public void hierarchyChanged(HierarchyEvent e) {
...
+    GameModule.getGameModule().addSideChangeListenerToPlayerRoster(this);
+    g.getPrefs().addOption(
+      Resources.getString("Prefs.general_tab"), //$NON-NLS-1$
+      new IntConfigurer(
+        PREFERRED_SCROLL_ZONE,
+        Resources.getString("Map.scroll_zone_preference"), //$NON-NLS-1$
+        SCROLL_ZONE
+      )
+    );
+
```
BrentEaston, line 798: Brian, am I missing something? Where does this preference actually get read and used to set the scroll zone? You seem to have just changed the hard-code scroll zone from 30 px to a hard-codes 60px?

BrentEaston (changes requested):

vassal-app/src/main/java/VASSAL/build/module/Map.java
BrentEaston: Yes, but you have only added code to create a preference and to set the initial value of the preference. If a user changes the preference, SCROLL_ZONE will not be updated. In fact it can't be updated since it is final.

You have to either
a) read the preference value back each time you need the value like this:

  public void dragOver(DropTargetDragEvent dtde) {
    int preferredScrollZone = (Integer) GameModule.getGameModule().getPrefs().getValue(PREFERRED_SCROLL_ZONE);
    scrollAtEdge(dtde.getLocation(), preferredScrollZone);
  }

or
b) Read the pref value immediately after you add the preference, then add a PropertyChangeListener to the IntConfigure to check for changes like this:

    // Create the Configurer for the Pref
    preferredScrollConfig = new IntConfigurer(
      PREFERRED_SCROLL_ZONE,
      Resources.getString("Map.scroll_zone_preference"), //$NON-NLS-1$
      SCROLL_ZONE
    );

    // Register the Pref, which copies any existing pref value into the configurer
    g.getPrefs().addOption(
      Resources.getString("Prefs.general_tab"), //$NON-NLS-1$
      preferredScrollConfig
    );

    // Read the current value of the pref
    SCROLL_ZONE = preferredScrollConfig.getIntValue(60);

    // Listen for any changes to the pref
    preferredScrollConfig.addPropertyChangeListener(new PropertyChangeListener() {
      @Override
      public void propertyChange(PropertyChangeEvent evt) {
        SCROLL_ZONE = preferredScrollConfig.getIntValue(60);
      }
    });

a) is simpler, but b) is more efficient if the preference is going to be evaluated often.

Cattlesquat: Oh! Right! I went with option B, "because dragging at edge of screen"

PR #291, opened 2020-09-05
13389: Make Chatter less flickery

Cattlesquat:
Should hopefully reduce "flickering" of the chatter, though I'd like to get Claudio to test since he had a good repro case.


Cattlesquat:
Okay I think this actually works -- I had a very strong repro case going, and this seems to have cleared up all the flickering. The strategy that when we have "a buncha bullshit going on" (huge global key command, or somebody dragged 45 pieces across the map), we sweep all the "update chat log" into a single update that gets picked up by invokeLater at the end of the chain of crappery.

PR #294, opened 2020-09-06
3.4.1 - 13381 - Ignore numeric literals when returning Beanshell Values

BrentEaston:
Prevent values like "1D", "3F", "2L" being converted to numbers. While this is technically correct, Vassal does not implement Double, Float or Long types and no Developer will ever want this to happen.


BrentEaston:
Causing TriggerActionTest to fail. Pulling this and will rework for 3.5.

PR #303, opened 2020-09-10
3.5 - 13415 - MouseoverStackViewer - HTML Compliant! - Holy Actual Crap! 

Cattlesquat:
* "Surgeon General's Notice: some JLabels ___may___ have been harmed in the making of this feature"

Basically you can now use HTML in the happy little text fields. If you're a masochist and manually put in surrounding html tags yourself, it will just "take it". But if you turn on the Special Check Box it will automatically wrap it in html tags and a div and side-padding for you, making HTML tagging Easy For The Whole Family. And of course, Quick Colors(tm) are supported for, you know, quickly coloring. 

There is actually fruitful ground here for doing more stuff too, like making it easier for module designers to do text-wrapping in these, and more control over size/styling of the text boxes. Could be this PR, or some future one. Also really the whole Quick Colors thing could be (a) tied up with a bow in a nice neat package and (b) shipped all over the place to other text features.

Also because I have like 8 different MouseoverStackViewers in one map (because different kinds of thing need different kinds of rollover), I gave this baby a "Description" field so one can aspire to tell them apart in the editor.

image
image

YOU KNOW YOU NEED THIS!!!


Cattlesquat:
THIS PR IS GETTING RIDICULOUSLY GOOD!

BrentEaston:
Just needs a reference manual update as well to explain all the new features :)

Cattlesquat:
> Just needs a reference manual update as well to explain all the new features :)

Oh absolutely! As soon as the 3.5 feature list settles and we start talking about approaching beta, I plan to do doc updates for all my new 3.5 features (Flare, this, Chess Clocks, etc), and will do another "release notes" feature tour as well.

PR #306, opened 2020-09-11
13405: Partially revert "refactored PlayerWindow and its SplitPane handling"

yanlyub:


Cattlesquat:
Do we really want to just undo this? It mostly works there's probably just "one small thing wrong"?

uckelman:
The code's still in the repo if someone wants to have another go. Until then, putting it back seems prudent.

PR #307, opened 2020-09-12
13394 - Fix "1D" interpreted as "1.0" plus Beanshell test suite.

BrentEaston:
Also changes Mockito to 'inline' version to allow static Mocks and explicitly specified byte buddy to get correct version for inline mocks.


vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -269,23 +269,18 @@ public String evaluate(PropertySource ps, boolean localized) throws ExpressionEx
...
           try {
             setVar(var, Integer.parseInt(value));
           }
           catch (NumberFormatException e) {
-            try {
-              setVar(var, Float.parseFloat(value));
-            }
-            catch (NumberFormatException e1) {
-              setVar(var, value);
-            }
```
BrentEaston, line 288: This should be the only actual code change to prevent "1D" and "1F" being converted to "1.0". The Float.parseFloat() check should never have been there.

uckelman (approved):

PR #320, opened 2020-09-14
bug13409 fix board.SnapTo error by calling it again

claudiociardelli:
Recreate PR against 3.4.2

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -911,6 +911,13 @@ public class Map extends AbstractConfigurable implements GameComponent, MouseLis
     snap.translate(-r.x, -r.y);
     snap = b.snapTo(snap);
     snap.translate(r.x, r.y);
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -911,6 +911,13 @@ public Point snapTo(Point p) {
     snap.translate(-r.x, -r.y);
     snap = b.snapTo(snap);
     snap.translate(r.x, r.y);
+
+    //CC bugfix13409
+    // If we snapped to a point outside the board b, call sanpTo again with the board we landed into
+    final Board bSnappedTo = findBoard(snap);
+    if (!b.equals(bSnappedTo)) {
+      snap = this.snapTo(snap);
```
BrentEaston, line 919: This should be snap  = bSnappedTo.snapTo(snap), not snap = this.snapTo(snap). We already know which board we need to do the re-snap, no need to recurse.

PR #325, opened 2020-09-15
13420 - PieceWrapper implementation

BrentEaston:
I think I have cracked the PieceWrapper puzzle. This seems to be a good working implementation and I would be interested for people to try it.

I have temporarily added a KeyCommand to the PieceWrapper to give a visual feedback if a piece has a PieceWrapper not.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/BasicCommandEncoder.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/BasicCommandEncoder.java
@@ -17,6 +17,7 @@
  */
 package VASSAL.build.module;
 
+import VASSAL.counters.PieceWrapper;
 import java.awt.Point;
 import java.util.Map;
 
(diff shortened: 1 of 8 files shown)


Cattlesquat:
So this looks pretty exciting in terms of future potential module performance (especially for modules like mine with 50-100 Decorators on a piece). 

I'd love to try this out for "perf" purposes, but of course the piecewrapper currently seems relatively "inert" in that it isn't intercepting anything and making it faster. And so the question is... what is the "right" way to add some making-it-fast stuff? Like the immediately obvious off-the-top-of-my-head stuff to me would being able to get at the standard "BasicPiece" properties (LocationName, CurrentZone, CurrentMap, OldLocationName, OldZone, OldMap, etc, as well as possibly BasicName). Apart from possibly BasicName I'm not sure any of these are ever meaningfully intercepted by a "hiding/obscuring" decorator.  

Or I could imagine some kind of "cache" system where certain getProperty() calls (to some of those basic things) would be remembered at the top level. 

Would we start coding some stuff right onto PieceWrapper? Or is there some way one would override PieceWrapper with e.g. MyPieceWrapper but make sure that one gets put onto every piece in the normal course of affairs? 

Or there would be some way to configure (in the Editor) the PieceWrapper profile so it would know what to pass through and what not to?

BrentEaston:
Need to include PieceWrappers on temporary pieces created for GamePiece palettes and piece and prototype editor, but not show the PieceWrapper in the Piece Editor.

BrentEaston:
Closing this. Too much work to implement and I was never able to conclusively prove any performance improvements at all.

PR #326, opened 2020-09-16
3.4.2 (!!) - 13425 - MouseoverStackViewer legacy expression matching was broken by a code cleanup

Cattlesquat:
Did some Big Boy Bisecting and we found a Broken Thing. Reverted it, and the problem went away. Yay! 

Marked with comments So We Will Know.


BrentEaston (commented):
This fix looks good to me as it is a straight revert of the code cleanup. The reason the cleanup failed is because though the cleaned code was equivalent in a Java 'equals' sense, it violated the Vassal 'equals' contract for Legacy PME's which is not the same.

PR #327, opened 2020-09-16
3.5 - 13342 - Kill two birds with one save string?

Cattlesquat:
This addresses 13342 by having BasicLogger#setup share a save string/command with GameState#setup.

My only concern is... are we sure that nowhere in the giant chain of gc.setup() for every GameComponent, that there aren't any setup()'s that change the game state (i.e. final save state)? I mean I guess if there ARE then it means BasicLogger's save string isn't any good anyway, but "just asking". As they say on the road signs... Merge With Caution! :D


uckelman:
@BrentEaston Does this fix what you saw?

BrentEaston:
Yes, this does fix what I saw, but Brian's question is a very good one. There is something fishy going on here.

Cattlesquat:
SwampwallabyToday at 6:13 PM
The setup() command of a GameComponent initialises the 'state' of the component that is returned by getRestoreCommand(). So that whole piece of code is complete garbage.

CattlesquatToday at 6:13 PM
So it's actually that earlier string that is bad, and we should really generate the string AFTER all that stuff and assign it to both the logfile and the check-if-modified or whatever that other thing is

SwampwallabyToday at 6:14 PM
The initial 'state' returned to BasicLogger is also completely random, depending on when it is called in the for loop. I think both strings are wrong.
[6:15 PM]
Yes, you are correct I believe.
[6:16 PM]
I don't quite follow why it even seems to work at the moment!
[6:17 PM]
It will need a lot of testing, I think aim it for 3.5.0 rather than 3.4.3. It's been that way for ever.
[6:17 PM]
Someone needs to actually thread through the code and see where and how those strings are being used.

CattlesquatToday at 6:18 PM
Yeah. That PR is against 3.5 anyway I'm pretty sure.
[6:19 PM]
And see the thing is, the string that's going into BasicLogger is getting there during its "setup" method (some indeterminate way through the setup-method-calling process)

SwampwallabyToday at 6:19 PM
Exactly

CattlesquatToday at 6:19 PM
I guess if we could guarantee that BasicLogger.setup() would be the absolutely last setup method called, then that would be "one way".
[6:20 PM]
Another way would be to NOT do that during BasicLogger.setup() and have GameState do a later call after that's done to tell it what save string to use.

SwampwallabyToday at 6:20 PM
BasicLogger itself doesn't have any state to setup, but, for example, the PlayerRoster doesn't to be reported until PlayerRoster.setup() has been called/

CattlesquatToday at 6:21 PM
I sort of think the current PR is basically garbage we need a whole new approach, but I can leave it up for now As A Warning To The Unwary(edited)
NEW

SwampwallabyToday at 6:22 PM
Yeah, just annotate it with this discussion
[6:23 PM]
I think you approach is right though. After setup has completely finished, generate a save String and hand it over to any component that wants to know it. Or even just have the one centrally managed Save String in GameState and BasicLogger uses that one. There should be no reason for two different save strings

Cattlesquat:
Nope, nope, nope. Big Bag of Nope.

PR #330, opened 2020-09-16
3.5 - Global Key Command performance improvements

Cattlesquat:
Check out this idea. I've put some select "pre-filters" on Global Key Commands, where you can "pre-filter" pieces that don't match a specified Map/Zone/Location/XY with a FormattedString (NOT an expression, for reasons which will become obvious). This means a precomputed matching string is used to eliminate potentially hundreds of GKC-target-candidates more quickly than instantiating the whole Beanshell matching expression is.

In my For the People module, this is eliminating 20%-50% of of the total amount of time that certain of my bigger GKC's take (so at the high end nearly a SECOND per GKC on my fairly strapping Windows machine -- significant!). It may not be the biznatch-and-endsnatch, but it's definitely something, and fairly unobtrusive.

If this looks GOOD, I can finish up localization of CounterGlobalKeyCommand as part of the PR :) 

So here in this first example -- if I've got a location already picked out in the property "ReorgLoc", I select "restrict to specific location" and then put $ReorgLoc$ in the Region field. So then it populates that field before it runs the search, and rejects anything for which LocationName does not match the $ReorgLoc$ value. You may notice I still have LocationName == ReorgLoc in my regular beanshell matching expression, but that's just because I was lazy when testing this -- I can totally take that out now, not that it's actually costing me very much now that the location is pre-filtered before beanshell ever runs anyway.
image

Can be configured to match MAP, ZONE, REGION, or X/Y location. Default is "Whole Game" which is like normal GKC.
image


yanlyub:
This can be rebased on latest master where the build breaker is fixed

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -654,6 +654,32 @@ Editor.Stacking.color_nonexpand=Color of pieces when not expanded
 Editor.Stacking.component_type=Stacking options
 
 
+#Global Key Command
+Editor.GlobalKeyCommand.global_key_command=Global Key Command:
```
BrentEaston, line 658: No More colons or white space at the end of prompt strings. We will sort that out in the configurers.

BrentEaston:
Here's a wild thought. There seems to be masses of duplicated code across several classes. How about encapsulating all of the 'Fast match'  stuff into a single class. Give it it's own Configurer (generated via a ConfugurerFactory) and encode/decode and that should greatly simplify the changes you need to make to all classes. You can just treat the 'Fast Match' info as a single entity. It should make development and maintenance much easier. Because you also need to add it to Deck Global Key Commands!

I'm still unclear about having board in the X, Y co-ords option. Is the X, Y relative to the Map or the Board? Is Board needed?

Also, don't forget about my idea for Counter and Deck GKC's to have a 'Location of this counter' option.

Cattlesquat:
> Here's a wild thought. There seems to be masses of duplicated code across several classes. How about encapsulating all of the 'Fast match' stuff into a single class. Give it it's own Configurer (generated via a ConfugurerFactory) and encode/decode and that should greatly simplify the changes you need to make to all classes. You can just treat the 'Fast Match' info as a single entity. It should make development and maintenance much easier. Because you also need to add it to Deck Global Key Commands!
> 
> I'm still unclear about having board in the X, Y co-ords option. Is the X, Y relative to the Map or the Board? Is Board needed?
> 
> Also, don't forget about my idea for Counter and Deck GKC's to have a 'Location of this counter' option.

I don't think my Configurer Kung Fu is up to making a class that will work right both in Traits and in Components. Might be a good refactoring project afterwards for someone who knows what they're doing?

But I will get on all those other things. I'm not 100% clear on the Board thing either -- I was just emulating SendToLocation in reverse, which asked for both a Map and a Board for whatever reason.

PR #336, opened 2020-09-19
3.4.3 - Bug 13433 - ObcuredToOthers now returns null in face-down dec…

BrentEaston:
…ks instead of a value


Cattlesquat:
The code looks okay to me, and I've verified that the problems the "old-old" version used to create in Paths of Glory are still fixed.

uckelman (approved):

PR #340, opened 2020-09-20
Null safety for BoardPicker.currentBoards

yanlyub:


vassal-app/src/main/java/VASSAL/build/module/map/BoardPicker.java
```diff
@@ -86,7 +86,7 @@
 public class BoardPicker extends AbstractBuildable implements ActionListener, GameComponent, GameSetupStep, Configurable, CommandEncoder, ValidityChecker {
   public static final String ID = "BoardPicker"; //$NON-NLS-1$
   protected List<Board> possibleBoards = new ArrayList<>();
-  protected List<Board> currentBoards = null;
```
uckelman, line 89: This will break `ASLBoardPicker` in VASL.

vassal-app/src/main/java/VASSAL/build/module/map/BoardPicker.java
```diff
@@ -175,6 +175,7 @@ public void addTo(Buildable b) {
     for (Board board : possibleBoards) {
       board.setMap(map);
     }
+    // TODO b instanceof Translatable is always true
```
BrentEaston, line 178: Intellij believes that b instanceof Translatable is always true, but I don't see it. Translatable extends Buildable and b is Buildable. All Buildables in Vassal may currently be Translatable, but that isn't guaranteed to be true in the future.

BrentEaston (approved):

PR #344, opened 2020-09-20
3.5 - Default logfile prompt to on

Cattlesquat:
3.5 has a much better & more comprehensible "reminder-to-start-logfile" flow. 

At the time I did it, I didn't realize quite how big of an issue this was in the community (or I'd have done it for 3.4 back when there was time), but ALSO because of that lack of realization, I didn't default the preference to on (and I just used the old preference which everyone already HAD defaulted to off anyway). 

So now this defaults those logfile preferences to "on", and uses a new key so that with 3.5 it will start "defaulted on" for everyone including players upgrading. And it has a "don't ask again" button on the box so the people that don't want it will just click that and everything will be fine.

But for all the PBEM people it will now remind them not only when they finish a log file, but ALSO when they load a saved game or start a new game or whatever.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
@@ -72,8 +72,8 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
   public static final String END = "end_log";  //$NON-NLS-1$
   public static final String LOG = "LOG\t";  //$NON-NLS-1$
   public static final String PROMPT_NEW_LOG = "PromptNewLog";  //$NON-NLS-1$
-  public static final String PROMPT_NEW_LOG_START = "PromptNewLogStart";  //$NON-NLS-1$
-  public static final String PROMPT_NEW_LOG_END = "PromptNewLogEnd";  //$NON-NLS-1$
+  public static final String PROMPT_NEW_LOG_START = "PromptNewLogAtStart"; //$NON-NLS-1$
+  public static final String PROMPT_NEW_LOG_END = "PromptNewLogAtEnd"; //$NON-NLS-1$
   public static final String PROMPT_LOG_COMMENT = "promptLogComment";  //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


Cattlesquat:
I don't think we need a ticket number for this because we already have an item for "improved log file UI flow" of some kind. I can write it up in the wiki entry for 3.5.

BrentEaston (commented):
My only concern is that this effectively wipes out whatever settings that a user currently has for these preferences and resets them to these new defaults for every single module. That will be pretty annoying to anyone with lots of modules who doesn't want these preferences set this way.

I have no problem with setting these as the default setting for a new module a user starts using, but forcing everyone to change will cause a lot of noise, even though it's a one-off change per module.

Cattlesquat:
We were discussing yesterday on the thread that "forgetting to start a logfile" is a major and long-standing complaint for many PBEM users. We have heard this through several different threads.

Unfortunately, because the "old" version of the remind-at-beginning preference existed & didn't work well & was defaulted to off to boot, we can assume this is the case for most users including the ones complaining about this issue.

That means the only way to get any reasonable amount of uptake on the new version of the preference is to treat it as a truly new preference (and default it to "on" this time, in the newfound knowledge that this is a serious issue for many players). It does certainly create an issue of needing to turn it off for non PBEM players, but it is a one-time-per-module issue, and with an easy dont-prompt-again box. I suggest we don't cast it as "we reset your default" but rather as "we added a whole new preference" - which arguably we did because it isn't called what the old one is called and doesn't do what the old one did anyway by a substantial margin. I don't actually expect much noise about that if we handle the rollout carefully.

I do agree it is probably unnecessary (and undesirable). To reset the second preference (remind-AFTER), because that one already worked, is the same, etc. I had changed that one "for symmetry" but that isn't an important reason so I will change it back.

BrentEaston:
Sure, sounds good to me.

Cattlesquat:
Have pushed the change to only reset the "at-start" pref.

BrentEaston (approved):

PR #346, opened 2020-09-21
3.4.3 - 13433 - Card draw facing revisited

BrentEaston:
I think this does the trick.


yanlyub:
This seems to fix the bug that I have encountered, it makes the drawn cards face in the proper direction.

PR #352, opened 2020-09-22
Fixed non-final static fields which should be final

yanlyub:


BrentEaston (commented):
Normally, I would be hesitant to change protected fields to final, but in all these cases, I agree, they really should be final. I doubt very much if any module developer is trying to write to them, and if they are, I would prefer we break their module rather than trying to deal with tracking down the bizarre bugs that might generate.

PR #357, opened 2020-09-24
3.5 - Next Giant Swath Of Localization in Buildables And Their Friends

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/ShapeItem.java
```diff
@@ -100,12 +100,19 @@ public ShapeItem(GamePieceLayout l, String n) {
     );
   }
 
-  public static class ShapeConfig extends StringEnum {
+  public static class ShapeConfig extends TranslatableStringEnum {
```
BrentEaston, line 103: Just above here, the getAttributeDescriptions() need to be done.

BrentEaston (changes requested):
Just a couple of little things

BrentEaston (approved):

PR #363, opened 2020-09-25
3.5 - Allow full-translated strings into TranslatedStringEnum/TranslatingStringEnumConfigurer family as CTOR option

Cattlesquat:
Default is still to pass "keys". Adding extra constructor boolean parameter allows telling it we are sending pre-translated strings so it shouldn't run Resources.get() again on them.

This should give us the flexibility to navigate past situations where translated strings are available but "keys" are not.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/AutoConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/configure/AutoConfigurer.java
@@ -59,12 +59,9 @@ public class AutoConfigurer extends Configurer
 
     target = c;
     setValue(target);
-    target.addPropertyChangeListener(new PropertyChangeListener() {
-      @Override
-      public void propertyChange(final PropertyChangeEvent evt) {
-        if (Configurable.NAME_PROPERTY.equals(evt.getPropertyName())) {
-          setName((String) evt.getNewValue());
-        }
+    target.addPropertyChangeListener(evt -> {
+      if (Configurable.NAME_PROPERTY.equals(evt.getPropertyName())) {
+        setName((String) evt.getNewValue());
       }
     });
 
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/configure/TranslatableStringEnum.java
```diff
@@ -27,4 +27,11 @@ public TranslatableStringEnum() {
     super();
   }
   public abstract String[] getI18nKeys(AutoConfigurable target);
+
+  /**
+   * @return true if names have already been translated; false means keys will be passed to Resources.getString()
+   */
+  public boolean isDisplayNames() {
+    return false;
+  }
```
BrentEaston, line 36: Not needed.

vassal-app/src/main/java/VASSAL/configure/AutoConfigurer.java
```diff
@@ -152,7 +149,7 @@ else if (TranslatableStringEnum.class.isAssignableFrom(type)) {
       if (se != null) {
         final String[] validValues = se.getValidValues(target);
         final String[] i18nKeys = se.getI18nKeys(target);
-        config = new TranslatingStringEnumConfigurer(key, prompt, validValues, i18nKeys);
+        config = new TranslatingStringEnumConfigurer(key, prompt, validValues, i18nKeys, se.isDisplayNames());
```
BrentEaston, line 152: Shouldn't need to be changed, since se.isDisplayName() will always be false anyway. Change the old Constructors on TranslatingStringEnumConfigurer to set the displayNames field to false.

vassal-app/src/main/java/VASSAL/configure/TranslatingStringEnumConfigurer.java
```diff
@@ -51,6 +52,24 @@ public TranslatingStringEnumConfigurer(String key, String name, String[] validVa
...
+  /**
+   * Create a drop-list of localised display values that allows you to select a value from an
+   * underlying list of 'internal' untranslated values.
+   *
+   * @param key Configurer Key
+   * @param name Configurer Name
+   * @param validValues Array of values to maintain
+   * @param i18nKeys Array of Translation Keys used to describe the list of values
+   * @param isDisplayNames True if we have already been given full translated strings; false if they are "keys" for Resources.getString()
+   */
```
BrentEaston, line 64: Change the old Constructors without the isDisplayName parameter to either call the new constructor with isDisplayName=false or if not possible just set this.isDisplayName = false,

BrentEaston (changes requested):

BrentEaston (approved):

PR #365, opened 2020-09-25
3.4.4 Bug 13454 - Fix ClassCastException in SetPersistentPropertyCommand.execute()

BrentEaston:
I'm not sure why it thinks I have replaced the whole SetPersistentPropertyCommand file when I just changed one line, but the result is correct.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/command/SetPersistentPropertyCommand.java
+++ b/vassal-app/src/main/java/VASSAL/command/SetPersistentPropertyCommand.java
@@ -1,7 +1,6 @@
 package VASSAL.command;
 
 import VASSAL.build.GameModule;
-import VASSAL.counters.Decorator;
 import VASSAL.counters.GamePiece;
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/command/SetPersistentPropertyCommand.java
```diff
@@ -1,79 +1,78 @@
...
+    setId (id);
+  }
+
+  protected void executeCommand() {
+    GamePiece target = GameModule.getGameModule().getGameState().getPieceForId(id);
+    if (target != null) {
+      // Not all GamePieces will have persistent Properties
+      if (target instanceof PersistentPropertyContainer) {
+        ((PersistentPropertyContainer) target).setPersistentProperty(getKey(), getNewValue());
+      }
```
BrentEaston, line 33: This is the only change in this file, changing the cast.

PR #366, opened 2020-09-25
3.5 - Localizable stuff, including some using the "new" option

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
```diff
@@ -49,7 +49,7 @@
 import VASSAL.counters.Properties;
 
 public class MenuDisplayer extends MouseAdapter implements Buildable {
-  public static final Font POPUP_MENU_FONT = new Font("Dialog", Font.PLAIN, 10);
+  public static final Font POPUP_MENU_FONT = new Font("Dialog", Font.PLAIN, 10);  //NON-NLS
```
BrentEaston, line 53: I actually change any of these
new Font("Dialog", Font.PLAIN, 10)
I find to
new Font(Font.DIALOG, Font.PLAIN, 10)
Saves on a NON-NLS comment :)
Not vital at all.

vassal-app/src/main/java/VASSAL/build/module/map/TextSaver.java
```diff
@@ -169,11 +161,11 @@ protected void writeMapAsText() {
 
   @Override
   public HelpFile getHelpFile() {
-    return HelpFile.getReferenceManualPage("Map.html", "TextCapture");
+    return HelpFile.getReferenceManualPage("Map.html", "TextCapture"); //NON-NLS
   }
 
   public static String getConfigureTypeName() {
-    return Resources.getString("Editor.TextCapture.component_type"); //$NON-NLS-1$
+    return Resources.getString("Editor.Saver.component_type"); //$NON-NLS-1$
```
BrentEaston, line 168: Type -> Editor.TextSaver.component_type?

vassal-app/src/main/java/VASSAL/build/module/map/Scroller.java
```diff
@@ -38,10 +38,10 @@
...
-  public static final String USE_ARROWS = "useArrows";
-  public static final String ALWAYS = "always";
-  public static final String NEVER = "never";
-  public static final String PROMPT = "prompt";
+  public static final String USE_ARROWS = "useArrows"; //NON-NLS
+  public static final String ALWAYS = "always"; //NON-NLS
+  public static final String NEVER = "never"; //NON-NLS
+  public static final String PROMPT = "prompt"; //NON-NLS
 
   protected String usingArrows = PROMPT;
```
BrentEaston, line 46: I don't think this is your problem Brian, but I just can't work out how this class is supposed to work. usingArrows is maintained as a String attribute with the ALWAYS/NEVER/PROMPT, but there is no way that usingArrows can ever be changed as Scroller is not Configurable. The USE_ARROWS preference is actually Boolean and is checked explicitly in the keyPressed() method.

I think the whole ALWAYS/NEVER/PROMPT stuff can just be thrown away and just set the Boolean configurer. Thoughts?

BrentEaston (changes requested):

vassal-app/src/main/java/VASSAL/build/module/map/TextSaver.java
Cattlesquat: This is how I know I'm _actually_ getting old.

vassal-app/src/main/java/VASSAL/build/module/map/Scroller.java
Cattlesquat: Oh right... WEIRD. I clearly did NOT stare deeply into that class when I came through...

BrentEaston: For now, I'm happy if we just do the i18n and leave any other changes for another day...

Cattlesquat: Well see what you think of this. Also there is that weird "NoEcho" thing which looks like it might have done something once but doesn't seem to do anything now? It's like long ago someone came and cut a chunk out of this class.

I can always just revert this to the "mostly inert" form it was already in.

BrentEaston (approved):

vassal-app/src/main/java/VASSAL/build/module/map/Scroller.java
BrentEaston: Nah, getting rid of clutter is good. Waiting for the build...

BrentEaston (approved):

PR #369, opened 2020-09-27
3.4.4-13458-Allow floats to be compared in Beanshell. (Regression). 

BrentEaston:
Caused by overly aggressive fix for bug 13381.
Replace Float conversion but check for and exclude numeric literals first.
Added Regression test.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
+++ b/vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
@@ -276,8 +276,7 @@ public class ExpressionInterpreter extends AbstractInterpreter implements Loopab
         else if (BeanShell.FALSE.equals(value)) {
           setVar(var, false);
         }
-        // Do not attempt to convert int or float Numeric Literals, return them as Strings.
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -275,12 +276,22 @@ else if (BeanShell.TRUE.equals(value)) {
         else if (BeanShell.FALSE.equals(value)) {
           setVar(var, false);
         }
+        // Do not attempt to convert int or float Numeric Literals, return them as Strings.
+        else if (StringUtils.containsAny(value, "dDfFlL")) { // NON-NLS
```
uckelman, line 280: Shouldn't it check more specifically if any of those are at the end?

BrentEaston: Technically yes, but if those characters appear anywhere except at the end, the parseFloat will fail anyway.
Have a look at this version. If the string contains anything except +-.[0-9] we want to short-circuit the parseInt() and parseFloat() and just return it as a String.

PR #380, opened 2020-09-30
3.4.4-13471-Fix bug in handling null property values in expressions. …

BrentEaston:
…Also disable Calculated Property evaluation for units not on a map.


vassal-app/src/main/java/VASSAL/build/BadDataReport.java
```diff
@@ -37,6 +37,7 @@
   private String message;
   private String data;
   private Throwable cause;
+  private boolean reportable = true;
```
uckelman, line 40: Just a thought: These could all be final, and you wouldn't need `setReportable()`, right? Then `BadDataReport` could be an immutable object.

uckelman (approved):

PR #385, opened 2020-10-04
13469: Temp file cleanup

uckelman:
This turned into quite the excursion, originally trying to fix a problem of temp files failing to be deleted...

### DIFF
--- a/vassal-app/src/main/java/VASSAL/launch/ConfigImpl.java
+++ b/vassal-app/src/main/java/VASSAL/launch/ConfigImpl.java
@@ -1,0 +1,127 @@
+package VASSAL.launch;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import org.apache.commons.lang3.SystemUtils;
+
+import VASSAL.tools.version.GitProperties;
+
+public class ConfigImpl implements Config {
+  private final Path baseDir; 
+  private final Path docDir;
+  private final Path confDir;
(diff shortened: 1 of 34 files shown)


vassal-app/src/main/java/VASSAL/launch/DummyConfigImpl.java
```diff
@@ -0,0 +1,45 @@
+package VASSAL.launch;
+
+import java.nio.file.Path;
+
+public class DummyConfigImpl implements Config {
```
Cattlesquat, line 5: Newly steeped in the ideology that "Impl"-is-always-a-bad-name, I would suggest naming this simply "DummyConfig". Sort of like ArrayList is named ArrayList instead of ArrayListImpl :)

vassal-app/src/main/java/VASSAL/launch/ConfigImpl.java
```diff
@@ -0,0 +1,127 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import org.apache.commons.lang3.SystemUtils;
+
+import VASSAL.tools.version.GitProperties;
+
+public class ConfigImpl implements Config {
```
Cattlesquat, line 11: Configuration? VASSALConfig? SystemConfig? VersionConfig? PathsConfig? InfoConfig? ActualConfig? RealDonaldConfig? Literally-anything-but-Impl-Config?

PR #386, opened 2020-10-05
13706 - Console - Commands from chatter line. (W/ some cheating protection)

Cattlesquat:
Doesn't allow commands if there have ever been two different players in the game, or if there has ever been a server connection.

Meanwhile, some initial commands:
/errorlog show 10                                 # show last 10 lines of errorlog
/errorlog wipe                                      # wipe the errorlog
/errorlog folder                                     # open "containing folder" of errorlog
/errorlog open                                      # open errorlog (e.g. in editor)
/property show GlobalPropName        # show the value of a module level Global Property
/property set GlobalPropName value  # set the value of a module level Global Property

The property stuff is "early stage" obviously since I am still thinking about what the syntax should be. On the one hand I want to be able to do things like:
/property set Map:AlliedReinf available 5
/property set Zone:Victory points 3
But also I would like to be able to "list the values of all properties that match some pattern" and potentially also be able to get/set individual properties on a selected piece, etc.

Lots of stuff we can do if we like this approach. I already love just being able to have the errorlog stuff be easier to deal with.


Cattlesquat:
This PR now closes issue 13444 to let errorlog be sent to chatter

uckelman:
Two trivial conflicts to fix.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -85,6 +85,7 @@
   protected File outputFile;
   protected Action stepAction = new StepAction();
   protected SaveMetaData metadata;
+  private Boolean multiPlayer = false;
```
uckelman, line 88: Use `boolean`---this doesn't need to be an object.

uckelman, line 88: Use `boolean`---this doesn't need to be an object.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -199,6 +200,15 @@ public void add(Buildable b) {
   public void remove(Buildable b) {
   }
 
+  public void setMultiPlayer(Boolean multiPlayer) {
```
uckelman, line 203: Ditto.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -199,6 +200,15 @@ public void add(Buildable b) {
...
+    this.multiPlayer = multiPlayer;
+  }
+
+  public Boolean isMultiPlayer() {
```
uckelman, line 207: Ditto.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -244,9 +254,31 @@ public boolean isReplaying() {
 
   @Override
   public Command getRestoreCommand() {
-    return null;
+    return new MultiplayerStateCommand(isMultiPlayer());
+  }
+
+  private static class MultiplayerStateCommand extends Command {
+    Boolean state;
```
uckelman, line 261: Ditto.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -244,9 +254,31 @@ public boolean isReplaying() {
...
+
+    MultiplayerStateCommand(Boolean state) {
```
uckelman, line 263: Ditto.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -319,7 +319,7 @@ public void queryNewLogFile(boolean atStart) {
 
     final GameModule g = GameModule.getGameModule();
 
-    if ((Boolean) g.getPrefs().getValue(prefName)) {
+    if ((boolean) g.getPrefs().getValue(prefName)) {
```
uckelman, line 322: Wait, not this one! This is one of those cases where you _do_ need `Boolean`, because the `getValue()` returns an ` Object`, so if it returns `null` you'll have `ClassCastException` due to the `boolean` cast.

uckelman: If I marked this one as wrong, that was a mistake.

uckelman: Oh hey, looking at this again---that's still wrong (but was that way before, so not your fault). When you have something that returns an `Object` which could be a `Boolean`, you need to check if it's equal to `Boolean.TRUE`, like so:
```
if (Boolean.TRUE.equals(g.getPrefs().getValue(prefName))) {
```
What can happen with
```
if ((Boolean) g.getPrefs().getValue(prefName)) {
```
is that you get a null and then a `NullPointerException` on the attempt to unbox the `Boolean`.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/Console.java
```diff
@@ -0,0 +1,235 @@
...
+      try {
+        final File dirToOpen = Info.getConfDir();
+        desktop.open(dirToOpen);
+      }
+      catch (IOException e) {
+        System.out.println("File Not Found"); //NON-NLS
+      }
+      catch (IllegalArgumentException iae) {
+        System.out.println("Illegal argument"); //NON-NLS
+      }
```
uckelman, line 97: You've got this twice. Extract to a function?

Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/Console.java
```diff
@@ -59,6 +59,23 @@ private boolean matches(String s1, String s2, int min) {
...
+  private void browseFileOrFolder(File file) {
+    final Desktop desktop = Desktop.getDesktop();
+    try {
+      desktop.open(file);
+    }
+    catch (IOException e) {
+      System.out.println("File Not Found"); //NON-NLS
+    }
+    catch (IllegalArgumentException iae) {
+      System.out.println("Illegal argument"); //NON-NLS
```
uckelman, line 75: One more: Shouldn't these go to the logger instead of stdout?

Cattlesquat: Probably & changed (even though wherever in VASSAL I copied that snippet from was doing it that way - I think we're rife with it)

vassal-app/src/main/java/VASSAL/build/module/Console.java
```diff
@@ -0,0 +1,260 @@
...
+      return false;
+    }
+    if (s2.length() < min) {
+      return s1.toLowerCase().equals(s2.toLowerCase());
+    }
+    return s1.toLowerCase().substring(0, s2.length()).equals(s2.toLowerCase());
+  }
+
+  private String nextString(String def) {
+    return tok.hasNext() ? tok.next() : "";
```
uckelman, line 64: Oops. That should be `def` here, not `""`. The empty string is what you wanted the default to be. My bad.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -485,6 +499,22 @@ public static boolean isSoloSide(String side) {
...
+  /**
+   * @return True if this is currently a multiPlayer game (either connected to a server, or more than one player side allocated)
+   */
+  public boolean isMultiPlayer() {
+    if (GameModule.getGameModule().getServer().isConnected()) return true;
+
+    int allocatedSideCount = 0;
+    for (PlayerInfo p : players) {
+      ++allocatedSideCount; // NB. Intentionally not excluding observers.
+    }
```
uckelman, line 511: I just noticed this. So we're looping to count the items instead of using `players.size()`?

uckelman: Those last four lines could be `return players.size() > 1;`.

uckelman:
- "show" vs "get": Would "get" be better for parallelism since the set command is "set"?
- This makes me want to be able to cycle through command history using the up arrow
- Does `/errorlog wipe` work on Windows? Is wiping the errorLog a good idea even if it does?

Cattlesquat:
Although "get" is of course right in method names, "show" makes more sense in a console/CLI situation.

Wipe does work on windows. It's a good idea if you're e.g. "me" and your errorlog gets to be annoyingly filled with unending crap after a solid two weeks of play.

I too long to cycle back up through the command history (and in fact try to). If this ends up being something we all start using a lot it may be addable.

PR #391, opened 2020-10-09
3.5 - 1804 - Sound "Mute All" preference

Cattlesquat:
As requested, here is a "Mute All Sounds" preference :)

I'm guessing this might already have an issue number, back in the mists of time?


BrentEaston (changes requested):
Issue number is 1804.
This won't mute all sounds though, I can see 2 other places where sounds are generated that you missed :)

Cattlesquat:
I've adjusted to catch the incoming sound commands. I specifically left the SoundConfigurer out because that's where you click "Play" to hear the sound you just added, and it seemed weird to make it so you don't hear a sound when you're manually pressing play on the configurer screen (plus that is the same tab where you have the Mute checkbox, the listener for that doesn't change the preference until after you close the whole dialog, so if you tried to uncheck Mute right away to fix it, it wouldn't work unless you exited and re-entered the pref tab, which would also be weird).

Let me know if I've missed an additional place?

Cattlesquat:
Wakeup sounds split out to a separate preference.

BrentEaston (approved):

PR #393, opened 2020-10-09
Auto update module scenarios

jasonestewart:
changed all VASSAL.counters.*.myGetKeyCommands() for protected to public


pom.xml
```diff
@@ -4,7 +4,7 @@
 
     <groupId>org.vassalengine</groupId>
     <artifactId>vassal-parent</artifactId>
-    <version>3.5.0-SNAPSHOT</version>
+    <version>3.5.0</version>
```
yanlyub, line 7: Better not mess with maven, leave the pom.xml files as they are. Or make temporary changes but revert them before making the PR.

jasonestewart:
mea culpa...

pom.xml
jasonestewart: just ignorance - the pom.xml change crept in from somewhere (I have no idea how). my attempt to restore pom.xml from upstream/master seems to have failed and left it with my local pom.xml...

PR #395, opened 2020-10-10
13467, 13469: HTML img elements refering to images in module hold it open

uckelman:
Our old HTMLEditorKit replacement doesn't handle attributes properly; the JDK's HTMLEditorKit holds open the module file with no way to close it properly. This fixes both.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Chatter.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Chatter.java
@@ -26,7 +26,6 @@ import java.awt.event.ComponentAdapter;
 import java.awt.event.ComponentEvent;
 import java.awt.event.KeyEvent;
 import java.io.IOException;
-import java.net.URL;
 
 import javax.swing.BoxLayout;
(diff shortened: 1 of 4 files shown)


vassal-app/src/main/java/VASSAL/build/module/Chatter.java
```diff
@@ -101,9 +101,12 @@ public Chatter() {
 
     //BR// Conversation is now a JTextPane w/ HTMLEditorKit to process HTML, which gives us HTML support "for free".
     conversationPane = new JTextPane();
-    conversationPane.setContentType("text/html");
+    conversationPane.setContentType("text/html"); //NON-NLS
+//    kit = (HTMLEditorKit) conversationPane.getEditorKit();
```
Cattlesquat, line 105: Can remove commented line.

Cattlesquat (commented):
Code looks good. Obviously it will need to be injected into Labeler and CounterDetailViewer (unless we want to forbid image tags there)

PR #397, opened 2020-10-10
3.4.6 - 12576 and 13456 - setup stack grid locations for 'any' board stacks

Cattlesquat:
I'm not sure what, if anything, the old attempted fix for this accomplished (perhaps something...) but in any event it did not touch the issue for stacks assigned a Grid Location on an "<any>" owning board at all. It was still setting the final position based on the "top board in the list" rather than based on the board picked by the player.

This PR hopefully fixes that -- I mean, it DOES "apparently fix" it for me, but please check the code carefully.

The earlier stuff in 'setAttribute()' **doesn't actually run** during the "start new game" sequence when a player picks a board -- it has already run during the load of the module in the first place. So that whole set of code kind of scares me but in any event I didn't have to interact with it to fix this.


BrentEaston:
I gave this a good test along with my <Any> board change and it looks good to me.

PR #399, opened 2020-10-10
3.4.6 - 13505 - Deck band select fix

Cattlesquat:
There has been a thread on the forums (http://www.vassalengine.org/forum/viewtopic.php?f=3&t=11438) where they are complaining about it doing a band-select when you drag things off a deck. 

For some reason this problem seems "worse" on Macs than on PC, but I am able to see at least some of the problem on PC. 

In any event, I this this PR should fix the problem. Doesn't allow band-selects to start on top of a deck.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/KeyBufferer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/KeyBufferer.java
@@ -104,16 +104,22 @@ public class KeyBufferer extends MouseAdapter implements Buildable, MouseMotionL
 
     final KeyBuffer kbuf = KeyBuffer.getBuffer();
 
-    GamePiece p = map.findPiece(e.getPoint(), PieceFinder.PIECE_IN_STACK);
+    GamePiece p = map.findPiece(e.getPoint(), PieceFinder.DECK_OR_PIECE_IN_STACK);
     // Don't clear the buffer until we find the clicked-on piece
     // Because selecting a piece affects its visibility
     EventFilter filter = null;
     BandSelectType bandSelect = BandSelectType.NONE;
+    Boolean isDeck = (p != null) && (p instanceof Deck);
     if (p != null) {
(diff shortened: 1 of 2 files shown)


BrentEaston:
You seem to have made this fix against Master not Release 3-4-6?

Cattlesquat:
Oh thanks for pointing that out -- I actually made it against Release 3.4.6 but hadn't marked that in the PR.

BrentEaston:
Where you able to test this on a Mac Brian? Otherwise, we will need to get a snapshot build for the Mac guys to try out.

Cattlesquat:
I didn't end up borrowing my wife's Mac, because I was actually seeing the band-select problem happen in Windows using that guy's module (no idea why his does it a lot worse than PoG, but anyway). So I'm certainly not against a snapshot build though I'm not sure it's actually necessary -- since problem was happening for me and now it's not.

vassal-app/src/main/java/VASSAL/build/module/map/KeyBufferer.java
```diff
@@ -104,16 +104,22 @@ public void mousePressed(MouseEvent e) {
 
     final KeyBuffer kbuf = KeyBuffer.getBuffer();
 
-    GamePiece p = map.findPiece(e.getPoint(), PieceFinder.PIECE_IN_STACK);
+    GamePiece p = map.findPiece(e.getPoint(), PieceFinder.DECK_OR_PIECE_IN_STACK);
     // Don't clear the buffer until we find the clicked-on piece
     // Because selecting a piece affects its visibility
     EventFilter filter = null;
     BandSelectType bandSelect = BandSelectType.NONE;
+    Boolean isDeck = (p != null) && (p instanceof Deck);
```
uckelman, line 112: Just make this a `boolean`. No need for the object version of it.

uckelman (changes requested):

PR #408, opened 2020-10-13
13522 - Global Key Command performance improvements

Cattlesquat:


uckelman (approved):

uckelman (approved):

uckelman:
^^^ Conflicty McConflictface

uckelman:
@BrentEaston Is there anything you want to look over on this one?

vassal-app/src/main/java/VASSAL/configure/BooleanConfigurer.java
```diff
@@ -115,7 +115,7 @@ public void setName(String s) {
       box = new JCheckBox();
       box.setSelected(booleanValue());
       box.addItemListener(e -> setValue(box.isSelected()));
-      p.add(box);
+      p.add(box, "growx 0");
     }
```
BrentEaston, line 120: Might be better to revert this change as I have already made this change in master, but done it differently.

vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
```diff
@@ -83,51 +101,303 @@ public void setReportSingle(boolean reportSingle) {
...
+              // This will iterate through actual game pieces
+              for (GamePiece gamePiece : pieceList) {
+                // If a property-based Fast Match is specified, we eliminate non-matchers of that first.
+                if (!passesPropertyFastMatch(gamePiece)) {
+                  continue;
+                }
+
+                // These basic location filters are faster than equivalent filters in the Beanshell expression
+                if ((target.targetType == GlobalCommandTarget.Target.ZONE) || (target.targetType == GlobalCommandTarget.Target.LOCATION) ||
+                  (target.targetType == GlobalCommandTarget.Target.CURZONE) || (target.targetType == GlobalCommandTarget.Target.CURLOC)) {
```
BrentEaston, line 313: I don't think this if statement is needed. All options are covered in the immediately following switch statement. Just add a default: to the switch that does nothing.

vassal-app/src/main/java/VASSAL/build/module/map/DeckGlobalKeyCommand.java
```diff
@@ -137,13 +150,18 @@ public void decode(String s) {
...
   }
 
   @Override
   public String[] getAttributeDescriptions() {
     return new String[]{
       Resources.getString(Resources.NAME_LABEL),
       Resources.getString("Editor.GlobalKeyCommand.command"), //$NON-NLS-1$
+
+      Resources.getString("Editor.GlobalKeyCommand.pre_select"), // Fast Match parameters
+
```
BrentEaston, line 164: I though we where removing Fast Match from DeckGlobalKeyCommand? If we do, the 'Additional Match Properties' label will need to revert to 'Matching properties'.

BrentEaston (changes requested):

BrentEaston:
Otherwise looks great. The Fast Match algorithm looks as good as it can be. Well done.

vassal-app/src/main/java/VASSAL/configure/BooleanConfigurer.java
Cattlesquat: reverted

vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
Cattlesquat: Oh yeah, I think there was some earlier version of this with an intermediate step that made this make sense. Removed.

vassal-app/src/main/java/VASSAL/build/module/map/DeckGlobalKeyCommand.java
Cattlesquat: good point

Cattlesquat:
All cleaned up and perhaps ready-to-go

PR #420, opened 2020-10-14
improve hints for Sum/Count in expression builder

Cattlesquat:
There has been some confusion about what to put in the "Map" field of Sum and Count. I have attempted to clarify the prompts and hints.

image


BrentEaston:
Thanks Brian, Good use of function builder hints, get's it right in the users face. Not that they will read them anyway :) Keep an eye out for other commong Beanshell mis uses.

PR #435, opened 2020-10-19
3.5-Fix POG Slow startup issue

BrentEaston:


BrentEaston:
Broken test. Working on a fix.

BrentEaston:
@Cattlesquat Could you please do a final POG check on this PR Brian? I've just shuffled a few things around, should not have made a difference, just want to be double sure. Thanks.

vassal-app/src/main/java/VASSAL/configure/Configurer.java
```diff
@@ -149,11 +153,37 @@ public void removePropertyChangeListener(PropertyChangeListener l) {
...
+   * The headless check is required in case any Configurers try to initialize during tests.
+   *
+   * @return Screen Size.
+   */
+  protected Dimension getScreenSize() {
+    if (GraphicsEnvironment.isHeadless()) {
+      return new Dimension(1920, 1280);
+    }
+    else {
+      return Toolkit.getDefaultToolkit().getScreenSize();
```
uckelman, line 186: Good solution.

uckelman (approved):

Cattlesquat:
PR runs great with PoG now

PR #436, opened 2020-10-19
13489: Apply device scale factor to system drag ghost

uckelman:


BrentEaston (approved):
Looks good and confirmed that this fixes the issue on Win 10.

uckelman:
I'm waiting to get confirmation that this hasn't broken anything on Macs before merging.

PR #457, opened 2020-10-23
M, N & O traits, all remaining non-trait code in counters, test improvements, small cleanups.

BrentEaston:


vassal-app/src/main/java/VASSAL/counters/MovementMarkable.java
```diff
@@ -303,7 +316,7 @@ public String getState() {
    */
   @Override
   public List<NamedKeyStroke> getNamedKeyStrokeList() {
-    return Arrays.asList(key);
+    return Collections.singletonList(key);
```
uckelman, line 319: `List.of()` also does this, btw.

uckelman (approved):

PR #482, opened 2020-10-25
13560: Don't register mouse listeners from ActionButton.draw()

uckelman:


uckelman:
@BrentEaston Does this look ok to you? I discovered that ActionButton was registering mouse listeners from draw() when hunting for a bug.

BrentEaston (approved):

BrentEaston:
Looks good to me.  And tests as expected

PR #497, opened 2020-10-26
Formality Bites

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/ExtensionsManager.java
```diff
@@ -58,7 +58,7 @@ public ExtensionsManager(File moduleFile) {
   }
 
   public ExtensionsManager(GameModule module) {
-    this.moduleFile = new File(GameModule.getGameModule().getDataArchive().getName());
+    this.moduleFile = new File(module.getDataArchive().getName());
```
uckelman, line 61: Woah, that was goofy. Good to fix that.

uckelman (approved):

PR #505, opened 2020-10-27
3.5 - A SPECIAL PRESENT FOR BRENT! - Fast Match Comparison Modes

Cattlesquat:


vassal-app/src/main/java/VASSAL/counters/GlobalCommandTarget.java
```diff
@@ -67,6 +68,34 @@ public Configurer getConfigurer(AutoConfigurable c, String key, String name) {
...
+  /**
+   * Comparison Modes for property match
+   */
+  public enum CompareMode {
+    EQUALS("=="),
+    NOT_EQUALS("!="),
+    GREATER(">"),
+    GREATER_EQUALS(">="),
+    LESS("<"),
+    LESS_EQUALS("<=");
```
BrentEaston, line 80: What about =~ and !~ ?

BrentEaston (changes requested):
There's also something screwy going on with the encoding/decoding. If I set a FastMatch of 'Strength > 12', save it and immediately re-open the GKC, I get '> == 12'.

vassal-app/src/main/java/VASSAL/configure/GlobalCommandTargetConfigurer.java
```diff
@@ -183,6 +185,12 @@ public Component getControls() {
       controls.add(targetPropertyLabel, "span 2"); // NON-NLS
       controls.add(targetPropertyConfig.getControls(), "wrap"); // NON-NLS
 
+      targetCompareConfig = new CompareConfigurer();
+      targetCompareConfig.addPropertyChangeListener(evt -> getTarget().setTargetProperty(targetCompareConfig.getValueString()));
+      targetCompareLabel = new JLabel(Resources.getString("Editor.GlobalKeyCommand.property_compare"));
+      controls.add(targetCompareLabel, "span 2"); //NON-NLS
+      controls.add(targetCompareConfig.getControls(), "wrap"); //NON-NLS
+
```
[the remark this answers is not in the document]
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/counters/GlobalCommandTarget.java
Cattlesquat: Oh right - I had stared at PropertiesPieceFilter for a while and couldn't decipher the regex mechanism. So I have resorted to "attempting to learn it from google", with results now pushed. See what you think & give it a test -- I never end up using this in my modules because it always takes me too many tries to get the syntax right (among other things my brain always thinks it will be ~= which I literally had to look up *three times* while coding this I guess I could have snuck in a ~= synonym and forgotten about it in peace, haha), but anyway I don't have any convenient test code for it.

vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
```diff
@@ -122,7 +124,53 @@ public GlobalCommandTarget getTarget() {
...
+    return false; // Never gets here, but checkStyle doesn't understand that.
+  }
+
+  /**
+   * Need a super-fast (i.e. non-exception-throwing) plan for detecting valid numbers
+   * @param s string to check
+   * @return true if a value number
+   */
+  private boolean isNumeric(String s) {
+    return s.matches("-?\\d+(\\.\\d+)?");  //match a number with optional '-' and decimal.
```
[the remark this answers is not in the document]
Cattlesquat: Okay I put in the + as an option in the numeric detection (I assume Java's numeric parse functions accept it w/o additional coaxing). 

I used "Double" for the floating points, not sure whether it is preferred or if I should somehow support both, etc. Can change it to Float if that's what we use everywhere else.

Cattlesquat:
The encoders were working but I had forgotten that the configurers need to use the "symbol" rather than the "name" -- should be straightened out now.

vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
```diff
@@ -64,6 +65,9 @@
 
   private String fastProperty = "";     // Used during property Fast Match to hold *evaluated* expressions
   private String fastValue = "";        // Used during property Fast Match to hold *evaluated* expressions
+  private boolean fastIsNumber = false; // Used during property Fast Match to remember if value is numeric
+  private double fastNumber = 0;        // Used during property Fast Match to hold evaluated numerical value
+  private Pattern fastPattern = Pattern.compile(""); // Fast Match regex pattern
```
uckelman, line 70: Why compile a pattern that matches the empty string? (I'm surprised this doesn't throw.)

vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
```diff
@@ -122,7 +126,57 @@ public GlobalCommandTarget getTarget() {
...
+    return false; // Never gets here, but checkStyle doesn't understand that.
+  }
+
+  /**
+   * Need a super-fast (i.e. not dependent on exception-throwing) plan for detecting valid numbers
+   * @param s string to check
+   * @return true if a value number
+   */
+  private boolean isNumeric(String s) {
+    return s.matches("(\\+-)?\\d+(\\.\\d+)?");  //match a number with optional +/- and decimal.
```
uckelman, line 179: It's always faster to use a precompiled pattern.

vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
Cattlesquat: I guess I was thinking there would be some annoying warning without it :) but happily there is not.

vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
Cattlesquat: Good point - AND HERE I JUST LEARNED HOW TO DO THAT!

BrentEaston (commented):
I'm getting an NPE

java.lang.NullPointerException: null
	at java.base/java.util.regex.Matcher.getTextLength(Matcher.java:1770)
	at java.base/java.util.regex.Matcher.reset(Matcher.java:416)
	at java.base/java.util.regex.Matcher.<init>(Matcher.java:253)
	at java.base/java.util.regex.Pattern.matcher(Pattern.java:1135)
	at VASSAL.counters.GlobalCommand.isNumeric(GlobalCommand.java:181)
	at VASSAL.counters.GlobalCommand.passesPropertyFastMatch(GlobalCommand.java:145)
	at VASSAL.counters.GlobalCommand.apply(GlobalCommand.java:360)
	at VASSAL.counters.CounterGlobalKeyCommand.apply(CounterGlobalKeyCommand.java:261)
	at VASSAL.counters.CounterGlobalKeyCommand.myKeyEvent(CounterGlobalKeyCommand.java:162)
	at VASSAL.counters.Decorator.keyEvent(Decorator.java:488)
	at VASSAL.counters.KeyBuffer.keyCommand(KeyBuffer.java:131)
	at VASSAL.counters.KeyCommand.actionPerformed(KeyCommand.java:133)

When tying to do a Fast Match on Strength > 10

image

Cattlesquat:
Oh wow - that didn't happen to me when I was smoke-testing, but apparently because I was testing a property that every eligible piece HAD. As soon as I tried a non-existent property, boom-o. So all the variants that don't use the equals() method needed to have a null check added - pushed that.

PR #512, opened 2020-10-29
3.4.8 - 13567 - if module write fails, make it easier to actually find your tmp file

Cattlesquat:


vassal-app/src/main/java/VASSAL/tools/ErrorDialog.java
```diff
@@ -242,6 +243,17 @@ public static void bug(final Throwable thrown) {
     );
   }
 
+  public static Future<?> showFileOverwriteFailure( // I don't see why it shouldn't have a meaningful name? So much less confusing.
+    FileSystemException thrown,
+    String details,
+    String messageKey,
+    Object... args) {
+
+    return ProblemDialog.showFileOverwriteFailure(
```
uckelman, line 252: Is there a reason for putting the implementation into ProblemDialog? 

ProblemDialog was intended to have general "show stuff" methods in it, and ErrorDialog slightly more specific "error flavored" dialogs. Can the implementation of all of this go into WriteErrorDialog and leave ErrorDialog and ProblemDialog unmodified?

vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -1103,6 +1104,14 @@ protected void save(boolean saveAs) {
 
       lastSavedConfiguration = save;
     }
+    catch (FileSystemException e) {
+      String[] msgs = e.getLocalizedMessage().split("\n");
```
uckelman, line 1108: `final`

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
```diff
@@ -47,6 +48,7 @@
 import java.util.zip.ZipOutputStream;
 
 import org.apache.commons.io.FileUtils;
+import VASSAL.i18n.Resources;
```
uckelman, line 51: Group this with the other `VASSAL.*` imports.

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
```diff
@@ -425,18 +427,50 @@ private void moveFile(Path src, Path dst) throws IOException {
...
-            e.getMessage()
-          ),
-          e
-        );
+      catch (IOException copyException) {
+        Path dst2 = dst.resolveSibling(src.getFileName()); // Our "temp" name but in the destination directory
+        final String fmt = Resources.getString("Editor.ZipArchive.overwrite") + "\n%s"; //NON-NLS
+        try {
+          // Try to at least put it in same directory as the module
+          Files.copy(src, dst2, StandardCopyOption.REPLACE_EXISTING);
```
uckelman, line 435: Should is be `Files.move` instead?

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
```diff
@@ -425,18 +427,50 @@ private void moveFile(Path src, Path dst) throws IOException {
...
-            fmt,
-            dst.toAbsolutePath(),
-            src.toAbsolutePath(),
-            e.getMessage()
-          ),
-          e
-        );
+      catch (IOException copyException) {
+        Path dst2 = dst.resolveSibling(src.getFileName()); // Our "temp" name but in the destination directory
+        final String fmt = Resources.getString("Editor.ZipArchive.overwrite") + "\n%s"; //NON-NLS
```
uckelman, line 432: Can be moved down and inward, nearer to where it's needed?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/tools/ErrorDialog.java
Cattlesquat: I'm pretty sure not, because we both need to pass the "two filenames" thing down, *and* be able to detect that we're in a situation where we need to do the two-filename-thing. I believe I simply followed the chain down to the earliest point that I could actually make the thing happen (or rather, I followed it too far, and then backed up).

Now I would be WHOLLY in favor of having WriteErrorDialog simply _skip_ ErrorDialog (which as far as I can tell contributes nothing to this case except an extra maze-like layer) and directly call ProblemDialog, if that seems to make any sense. 

This has been my first real "adventure" in this section of the code (the error dialog maze), and I do feel like the maze-like-nature of it discourages decent error reporting -- as evidenced for example in various decisions that clearly got made by somebody reporting earlier versions of these same errors I'm dealing with here. I think we would be a lot better off if the methods weren't e.g. all called "showDetails", which neither helps someone coming in and looking to see if any appropriate error reporting methods already exist nor helps someone who knows they need to add a new one. I'm sure it started off well-intentioned, but it doesn't seem to be serving "error reporting customers" very well. It kind of straight-jackets everything into a one-size-fits-all, which I guess works okay for the generic "whoops it crashed, give up now" situation, but anything custom has to be threaded through several different layers and that has clearly discouraged people from doing it.

vassal-app/src/main/java/VASSAL/build/GameModule.java
Cattlesquat: Ah yes - it's funny that it doesn't seem to be failing my build on this -- maybe we forgot to actually turn the rule on.

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
Cattlesquat: Oh right - what you're seeing here is an IDE issue -- when we put in new things the IDE helpfully goes and puts the import right away so the code will compile, but it doesn't do a very good job of picking a spot, unfortunately. So if I forget to go manually check, this happens.

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
Cattlesquat: I lean against it, because in this case we already tried a Files.move and it failed, and then we tried the Files.copy and that failed too. Obviously it is *possible* that a Files.move would now work but is not guaranteed. Given that we're already in the "more rare, error-handling case", adding *another* layer of try/catch in here seemed to needlessly complicate the flow & readability for a relatively minor performance saving in a relatively rare situation.

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
Cattlesquat: I just tried, and then remembered how it ended up here -- it is used for BOTH of the following "catch" statements, so has to be above the first try?

vassal-app/src/main/java/VASSAL/build/GameModule.java
uckelman: It's because PMD is only set up on master... but then this will cause master to fail as soon as it gets merged in to master.

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
uckelman: One of the checkers has an import order rule, I believe, but I haven't found it yet.

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
uckelman: Oh, I see. I misread the nesting.

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
```diff
@@ -425,18 +426,50 @@ private void moveFile(Path src, Path dst) throws IOException {
...
+              dst2.toFile().getName(),
+              copyException.getMessage()
+            )
+          );
+        }
+        catch (IOException didEverythingFail) {
+          if (didEverythingFail instanceof FileSystemException) {
+            // If we picked up a FileSystemException, then it came from the inner loop meaning we at least "got halfway", so
+            // we pass that exception on rather than throwing the "worse" version.
+            throw(didEverythingFail);
```
uckelman, line 460: Shouldn't have parentheses (throw's not a function).

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
```diff
@@ -425,18 +426,50 @@ private void moveFile(Path src, Path dst) throws IOException {
...
+            String.format(
+              fmt,
+              dst.toFile().getName(),
+              dst2.toFile().getName(),
+              copyException.getMessage()
+            )
+          );
+        }
+        catch (IOException didEverythingFail) {
+          if (didEverythingFail instanceof FileSystemException) {
```
uckelman, line 457: The way to do this is with two catch blocks, with the one for FileSystemException first.

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
uckelman: One reason a second `Files.move` may succeed when the first failed is that the destination file shouldn't already exist for the second try, but this is fine.

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/tools/io/ZipArchive.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/tools/ErrorDialog.java
uckelman: Prior to my trying to rationalize the various error dialogs, they weren't uniform and comprised a huge amount of not-quite-duplicate code. That's what the four dialog helper classes were intended to do. I take your point that they're not flexible enough.

ErrorDialog was aimed at providing ways of displaying both a bug dialog and dialogs where you could get the relevant stack trace for serious non-bug errors. It might have been a mistake to route WriteErrorDialog and ReadErrorDialog through ErrorDialog.I tend to see I/O errors as Errors, while the view among users seems to be that they're failures of a lesser sort.

I don't have a sense of which and how frequently the various methods in these classes are even used. Maybe I provided a lot which aren't at present, and it would be clearer if they were deprecated.

uckelman (approved):

PR #514, opened 2020-10-29
PieceDefiner update

BrentEaston:


Cattlesquat (changes requested):
Works nicely for the most part, here are a few problems/suggestions I found:
1 - Biggest problem was the first box I opened was for a card that had full-res GMT graphics. The definer opened to the full pixel height of my *screen*, which meant that all the buttons appeared below my fixed "Task Bar" (the OS thing on the bottom of the screen - many windows configs have that auto-hide, but mine is locked in place and two rows high). So I couldn't actually "get at" the buttons - I had to grab the very bottom pixel of the window title and drag it all the way to the top, and then I could see the top 3-4 pixels of the "OK" button row. There's hopefully some way to get the maximum screen height "minus fixed OS intrusions"? Otherwise probably limit your height to 90% of screen height would be a safe(ish) rule of thumb. 
2 - an additional strategy that might go hand in hand with that is to scale the whole piece image down if its natural size is larger than some fraction of the total pixel space. In my COIN module I have pieces that are overlays for provinces on the mapboard, and they are sometimes a thousand pixels in whatever dimension. So if we're locking ourselves into only 100% size on image display that's probably going to lead to bad things. I should probably try my All Bridges Burning Module... oh yes, haha, total disaster! Have attached a screenshot showing top-to-bottom of my *screen* - the trait portion mostly unrecoverably lost below the bottom! (I put the screenshot link at the bottom of this message so it won't blot out the sun if github renders it full size)
3 - I'll say again for the record I'm not a fan of the arrows-all-by-themselves, especially for the up/down. There's definitely a clarity problem for teaching the add/remove flow to novices, and then for teaching intermediate folks how to reorder (and that reordering "is a thing").

I looked around to see what Apple does in these situations, and it appears to be "drag and drop", with a fixed label nearby that reminds you that is possible. I *LOVE* the fact that you have enabled drag-and-drop, so if there were always-present labels that explained "drag traits here to add them", "drag traits to reorder", that kind of stuff, then I think that would get over the main "teaching hump" for this screen for novice users. 
image

Microsoft of course labels the buttons (move up, move down, add, remove, etc). I suppose one could keep the arrows - maybe  slightly smaller icons, but find a place to incorporate some kind of label about their function, even if in small text, clarity being more important than beauty. 
image
image

The other common paradigm is a small up button and down button on EACH entry. I'm less excited by that one because it's "busy", but at least it is more clear. 
image

At minimum I would suggest having rollover text for all of those buttons (e.g. when I got to the screen I *suspected* that button would move the trait to the bottom of the list, as opposed to scroll me to the bottom of the list, but I had to actually "try it" to make sure, because there was no rollover text). 
image

I think if you added always-present labels reminding about drag-and-drop as the preferred interface, and then either had rollover text or incorporated text in the "legacy buttons", that would be a good solution.

4 - In general it did fairly sane and decent looking things when I stretched and squashed the dialog box to non-standard sizes. It was mildly surprising when it wouldn't keep extending the list-of-traits screen off the bottom when I dragged it below the max height of the "available traits" listbox, but of course that max height is already pretty high and I'm probably one of relatively few module designers who puts 50+ traits in a piece. 

A+++ Would Piece Define Again! 

WARNING GIANT SCREENSHOT AHEAD! 
image

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -382,106 +459,305 @@ public void mouseReleased(MouseEvent e) {
...
+  /**
+   * Return the OS specific name for a control Key.
+   * This will usually be Ctrl+c or Cmd+c on a Mac for character c
+   *
+   * @param c Control Char
+   * @return OS specific name.
+   */
+  private String getCtrlKeyName(char c) {
+    return KeyNamer.getKeyString(SwingUtils.genericToSystem(KeyStroke.getKeyStroke(c, InputEvent.CTRL_DOWN_MASK)));
+  }
```
BrentEaston, line 557: @Cattlesquat Does this make sense?

BrentEaston:
@Cattlesquat Please have a look now and try that big 'piece'. Unfortunately, it isn't possible to 'scale' the image down, as it's not actually an image, it's a dynamically rendered composite GamePiece image, so no easy way to scale it. I just gone for a maximum size of 512x128 with scroll bars if needed.

Since it is so big now, I have also tweaked the PropertiesWindow to always open the PieceDefiner at the highest available screen pixel (taking task and menu bars into account).

All labels, tool tips and button text added as suggest. We have plenty of screen width to play with.

uckelman:
> Unfortunately, it isn't possible to 'scale' the image down, as it's not actually an image, it's a dynamically rendered composite GamePiece image, so no easy way to scale it.

How are you painting this? It should be possible to scale it by wrapping a GamePieceOp in a ScaleOp.

BrentEaston:
@uckelman The drawing is being done directly by the PieceSlot configurer via a custom paint routine. However, there is a whole pile of scaling going on in there to handle HiDPI, so I think the way to go is to build this into the PieceSlot.

BrentEaston:
I love the Ctrl+C / Ctrl+V support. In the give-an-inch-take-a-mile department, now I wish the "Del" key deleted things and the "Ctrl+Z" key undid the add/remove that I just did :slight_smile:
Arrow buttons all look great too
... and the Enter key could open the properties on whatever is selected, if I'm arrow-keying up and down the list.

And then there's the ever-popular "I'm in one of the list-boxes and I start typing e.g. 'Global' and it takes me to those items in the list" (Or without supporting multiple characters in a word you could go to the first word that matched the letter)

No Image: If there has to be something there you could go with no image printed on the "transparent" background (instead of specifically white). That would be maybe consistent with what we do elsewhere?

BrentEaston:
Component splitter to manually re-size image. Remember and re-use maximum image size before using scrollbars as a module preference.

uckelman:
Can you point me to exactly where the piece drawing is happening? If so, I'll take a look at options.

BrentEaston:
Problem solved I think. PieceSlot,paint(), but it already has Scaling capability. I just need to subclass PieceSlot and override getScale().

BrentEaston:
@Cattlesquat give this a try now. New side layout, better handling of large images, more shortcut keys (Enter, Del).

BrentEaston:
I might even think about Ctrl-Z tomorrow, but it would just undo the last change.

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -165,7 +283,8 @@ public void setPiece(GamePiece piece) {
 
       if (!contains) {
         try {
-          availableModel.addElement((GamePiece) pieceClass.getConstructor().newInstance());
+          addAdditionalElement((GamePiece) pieceClass.getConstructor().newInstance());
+          setSortOrder();
```
uckelman, line 287: I know this is awkward to change, but we don't want to catch anything thrown by `setSortOrder()` here, right?

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -195,9 +318,73 @@ private void refresh() {
...
+   * 2. If the image can't fit in the available space, scale it until it can
+   * 3. If the image size is < max split size, then set the divider size = image size to now waste space
+   *
+   */
+  private void resizeSlotPanel() {
+
+    // Reset the piece to 100% scale to get the actual size.
+    final double oldScale = slot.getScale();
+    slot.setScale(1.0);
+    final Dimension newSlotSize = new Dimension(slot.getPiece().boundingBox().width, slot.getPiece().boundingBox().height);
```
uckelman, line 346: Getting the bb could be expensive, maybe get it and store it rather than getting it twice?

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -195,9 +318,73 @@ private void refresh() {
...
+      slotPanel.setMaximumSize(new Dimension(newSlotSize.width, maxPanelHeight));
+      slotPanel.setPreferredSize(newSlotSize);
+      slot.setScale(1.0);
+      scaleLabel.setText("");
+    }
+    else {
+      // The new piece is too large to fit in the available space. Scale it to fit on the most oversized axis
+      final double xRatio = maxSlotWidth / newSlotSize.getWidth();
+      final double yRatio = maxSlotHeight / newSlotSize.getHeight();
+      final double newScale = (Math.min(xRatio, yRatio));
```
uckelman, line 365: Unnecessary parens.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
BrentEaston: Fixed

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
BrentEaston: Fixed

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
BrentEaston: setSortOrder() shouldn't throw anything and if the addAdditionalElement() failed, then there is no need to reset the Sort order.

uckelman: My concern is that if `setSortOrder()` unexpectedly throws something, we'll never see it because it will get swallowed up by the `catch (Throwable t)` just below it. That's the broadest possible catch, so should enclose only the one line which could legitimately throw anything because it could be running the constructor of a custom class.

BrentEaston: Ah true. If setSortorder() ever throws something, we want a Bug report. I'll just move it down below the try.

uckelman:
I made a PR against this branch which cleans up the layout constraints a bit. Give that a try. If it works for you, then merge it and I'll merge this next.

uckelman (approved):

PR #523, opened 2020-10-31
13431: Use ImageOp for caching drawing of grids and grid numbering

uckelman:


uckelman:
Good modules on which to test this are Liberty Roads and VG Vietnam 1965-1975, but anything which displays grid numbering should provide a good comparison.

PR #526, opened 2020-10-31
3.5 - Clean up a few GKC configurers and new Docs for new Global Key Command stuff

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/map/DeckGlobalKeyCommand.java
```diff
@@ -213,6 +213,9 @@ public void decode(String s) {
 
   @Override
   public VisibilityCondition getAttributeVisibility(String key) {
+    if (key.equals(TARGET)) {
+      return () -> false; // No fast match for Deck Global Key Commands
+    }
     return () -> true;
```
uckelman, line 219: Could do this as: `return key.equals(TARGET) ? () -> false : () -> true'`

PR #531, opened 2020-11-02
3.4.8 - 13577 - Stacks properly restore their visual 'layer' level when saved/restored

Cattlesquat:
Stacks are now soulbound to the layer of the first gamepiece added to them, so that they will properly restore themselves to that layer when loading savegame files.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/LayeredPieceCollection.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/LayeredPieceCollection.java
@@ -33,6 +33,8 @@ import VASSAL.counters.Stack;
 import VASSAL.i18n.Resources;
 import VASSAL.tools.TemporaryToolBar;
 
+import static VASSAL.counters.Stack.LAYER_NOT_SET;
+
 /**
  * Defines PieceCollection in which pieces are assigned to an arbitrary number of layers
  * according to a property setting
@@ -228,6 +230,10 @@ public class LayeredPieceCollection extends AbstractConfigurable {
     public Object visitStack(Stack s) {
       GamePiece top = s.topPiece();
       if (top == null) {
+        int layer = s.getLayer();
+        if (layer != LAYER_NOT_SET) {
+          return layer;
+        }
         return layerOrder.length;
       }
       return visitDefault(top);
--- a/vassal-app/src/main/java/VASSAL/counters/Stack.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Stack.java
@@ -523,7 +523,7 @@ public class Stack implements GamePiece, StateMergeable {
       }
       else {
         //BR// This encoding format with the "while" at the end made it challenging to work in a new parameter.
-        if (token.contains(HAS_LAYER_MARKER)) {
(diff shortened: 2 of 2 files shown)


BrentEaston:
This looks good to me Brian.  My only concern is that this will change the way the existing games loading pre-defined scenarios behave. While this is a good solution for counters that move, it could be a problem for pre-defined scenarios that have stacks of non-moving markers in different layers. Under 3.4.7, these counters would all be starting on the 'floating' layer above all other counters. Now, they will be where they are supposed to be, which may not be what the designer wants. 

Do we a) push this fix to 3.5.0 as a major release or b) Anyone who complains, just say, well your module was broken anyway, and now we've fixed it and you would have had to do something in 3.5.0 anyway....

Hard to tell how big (if at all) a problem this might be. I'm tending to b) ?

Cattlesquat:
Right, I'm still suspicious of this on the "possibly disrupting other things we don't expect" front. Have to weigh it against our confidence that this will work, and the existence of the guy who needs the fix, and I'm not quite sure where I come out, though I'm understanding this better and better.

Anecdote from this morning -- I decided to test this with Paths of Glory which does have layer control buttons (just unheard of to save a game with them active). To my surprise/shock/dismay, I found PoG actually works fine with the *unpatched* version! Like it restores things to their right layer! So I'm like WUT?!?! How could that be? At first I thought it was some version of "we must sometimes add the stacks before the units, and sometimes the units before the stacks", and I thought since PoG used At-Start Stacks and he'd used dragging from the GamePiece Palette that therein-lay-the-difference. But no, turns out in ALL code paths I could discover we add the units to the module before the stacks, and I was unable to reproduce the error in PoG even when dragging a theoretically-should-be-affected piece into the game from the Palette. 

And so eventually after a lot of debugging I discover THIS in Stack.java:
  ```java
  /** @return the top visible piece in this stack */
  public GamePiece topPiece() {
    for (int i = pieceCount - 1; i >= 0; --i) {
      if (!Boolean.TRUE.equals(contents[i].getProperty(Properties.INVISIBLE_TO_ME))) {
        return contents[i];
      }
    }
    return null;
  }
```

That's what's being called by the "I Wonder What Layer This Stack Is In" code -- it asks for the top piece of the stack, and then gets its layer. So in other words it becomes immediately obvious that this is a interaction with the Invisible trait -- for his module (unlike Paths of Glory), the piece has an Invisible trait and so it's sending back a null and so using the (spurious) default layer to place the stack -- even though it has pieces in it! 

Now I'm not a super-expert on Invisible and those "traits that hide other traits", so I'm not exactly sure why the Invisible thing is acting up during the load game sequence and thinking the thing is "invisible to me" even though I'm theoretically playing Luftwaffe, but that's what's happening. Maybe my player state is not yet fully loaded/registered into the module at the time of loading the game? Dunno. 

The good news from that (in addition to me understanding this better) is that both this bug and _this change_ are likely to affect far fewer modules than we thought (also perhaps explaining how this bug could have gone unreported so long). Our "layer-aware Stack" fix seems to fix the problem (and _should_ fix the problem), but is far less likely to disrupt other modules ... as long as there isn't something ELSE we haven't thought of :)

PR #534, opened 2020-11-03
Bug 13573 - Allow $$ variables to be used in Sum and Count functions in any expression (not just GKC's)

BrentEaston:


BrentEaston:
Managed to jam 2 into 1

PR #535, opened 2020-11-03
Bug 13573 - Allow $$ variables to be used in Sum and Count functions in any expression (not just GKC's)

BrentEaston:


vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -656,4 +657,44 @@ private Map findVassalMap(String mapName) {
...
+   *
+   * @return Updated expression
+   */
+  private String replaceDollarVariables(String expression, GamePiece source) {
+    if (expression == null || !expression.contains("$")) {
+      return expression;
+    }
+
+    final StringBuilder buffer = new StringBuilder();
+    final SequenceEncoder.Decoder st = new SequenceEncoder.Decoder(expression, '$');
```
uckelman, line 674: Were the expressions you're decoding here assembled by SequenceEncoder?

BrentEaston: No, they weren't, They are text Report Formats entered by the users. I pulled in some old code that used SE.D as a quick way to tokenize the string. But this isn't going to work anyway on a Report like 'The $first$ currency is $ and the $second$ is euro'. I have no way of knowing if there are any embedded non-property '$' signs in the String. It will also strip single '$' signs out. 

Sigh. 

I am going to have to do something a bit more extensive and check every $ delimited token to see if it is a property that has a value and only replace those that are, leaving any excess '$' signs behind.

BrentEaston: This version parses the string 'properly' using a FSA to search for and test potential property references.

PR #536, opened 2020-11-03
Bug 13574 - Fix issue reporting on oldXXXXXX properties in ReportState

BrentEaston:


vassal-app/src/main/java/VASSAL/counters/Decorator.java
```diff
@@ -537,4 +537,25 @@ protected boolean isSelected() {
...
+   *
+   * @param result Map of property values
+   * @return Updated Map of property values
+   */
+  public java.util.Map<String, Object> getProperties(java.util.Map<String, Object> result) {
+
+    for (String propertyName : getPropertyNames()) {
+      if (! result.containsKey(propertyName)) {
+        result.put(propertyName, getLocalizedProperty(propertyName));
+      }
```
uckelman, line 555: You could do this in a single lookup with no branching as:
```
result.computeIfAbsent(propertyName, pn -> getLocalizedProperty(pn));
```

vassal-app/src/main/java/VASSAL/counters/PropertyExporter.java
```diff
@@ -7,22 +7,45 @@
...
+   * @param result Map of properties
+   * @return Update property Map
+   */
+  default Map<String, Object> getProperties(Map<String, Object> result) {
+    final List<String> propertyNames = getPropertyNames();
+
+    for (String propertyName : propertyNames) {
+      if (! result.containsKey(propertyName)) {
+        result.put(propertyName, getLocalizedProperty(propertyName));
+      }
```
uckelman, line 46: Same here.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/counters/PropertyExporter.java
BrentEaston: Fixed

uckelman (approved):

PR #563, opened 2020-11-14
13631 - somebody yanks a piece away

Cattlesquat:
Bug seems likely to have involved pieces being moved/deleted simultaneously in an online multiplayer game.


uckelman:
@BrentEaston Does this look ok to you?

BrentEaston (commented):
I think this is a reasonable change to make, though I am not 100% convinced that it is due to simultaneous online activity, although it is a reasonable hypothesis. 

We should be able to reproduce the issue by firing up 2 clients side-by-side in the debugger and use break-points to exaggerate the time-lag and force different types of simultaneous activity.

I would like to know what actually happens if someone else moves a piece while you have a drag in process? Does the action initiated by them actually happen while you are dragging, or does it wait until you finish the drag, in which case, I don't see why we get this error. I mean, when do command coming in online actually get executed?

PR #566, opened 2020-11-17
13675 - Don't allow lone PieceSlots in a Piece Palette

Cattlesquat:


uckelman:
What will happen if you try to load a module which has a PieceSlot directly in a palette?

Cattlesquat:
I think it will stay & work, you just wouldn't be able to add another (nor re-add it if you delete it). The method I changed just affects what is available to add when you right click the parent component.

PR #569, opened 2020-11-18
13659, 13660: Splitter Positioning & Window Height Preferences

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2622,15 +2622,28 @@ public void windowClosing(WindowEvent e) {
...
+            final int w = window.getWidth();
+            final int h;
             if (Boolean.TRUE.equals(p.getOption(MAIN_WINDOW_REMEMBER).getValue())) {
               p.setDisableAutoWrite(true);
               p.getOption(MAIN_WINDOW_HEIGHT).setValue((int)d.getHeight());
               p.getOption(MAIN_WINDOW_WIDTH).setValue((int)d.getWidth());
               p.saveGlobal();
               p.setDisableAutoWrite(false);
+
+              h = (int)d.getHeight();
```
uckelman, line 2636: d.height will be an int, you can use that and avoid the cast.

vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2622,15 +2622,28 @@ public void windowClosing(WindowEvent e) {
...
+            }
+            else {
+              h = (int)screen.getHeight();
```
uckelman, line 2639: Ditto, screen.height is an int.

vassal-app/src/main/java/VASSAL/build/module/Map.java
Cattlesquat: It was pooching me down with "double" issues for some reason. Or you're saying just directly access the field w/o the getter/setter? So that's an int?

uckelman: Yes, Dimension, Point, Rectangle all have public members you can simply use.

BrentEaston: For some (insert appropriate invective here) reason, in those classes, the width and height are int's, but the getters return doubles.

vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2582,7 +2576,7 @@ public void setup(boolean show) {
 
           splitPane.showBottom();
 
-          //final int divider = (int)g.getToolBar().getPreferredSize().getHeight() + (int)g.getChatter().getPreferredSize().getHeight();
+          //BR// Force the divider to the Chatter's "preferred height"
           final int divider = (int)g.getChatter().getPreferredSize().getHeight();
```
uckelman, line 2580: Here to, just use `.height`.

PR #574, opened 2020-11-23
Use ZipWriter for one-off ZIP writing

uckelman:


jasonestewart:
Looks sexy. Slim and Beautiful.

PR #578, opened 2020-11-25
1826 - Scroll bars for all Config Dialogs.

BrentEaston:


vassal-app/src/main/java/VASSAL/tools/swing/SwingUtils.java
```diff
@@ -62,7 +65,7 @@ public static AffineTransform descaleTransform(AffineTransform t) {
     );
   }
 
-  public static final Map<?, ?> FONT_HINTS = (Map<?, ?>) Toolkit.getDefaultToolkit().getDesktopProperty("awt.font.desktophints");
+  public static final Map<?, ?> FONT_HINTS = GraphicsEnvironment.isHeadless() ? new HashMap<String, String>() : (Map<?, ?>) Toolkit.getDefaultToolkit().getDesktopProperty("awt.font.desktophints");
```
uckelman, line 68: I'd use Collections.emptyMap() instead of creating a mutable HashMap here for the headless case.

Also, is this necessary? Toolkit.getDefaultToolkit() returns a special headless one, so this shouldn't throw when run in headless mode.

BrentEaston: getDefaultTookit() is not the problem, The headleass toolkit returns null when you call  getDesktopProperty("awt.font.desktophints") which blows up the new ImagePicker I wrote on Travis when it initialises in a test.

BrentEaston:
This ready for 3.5.0 now. There are still some old non-standard dialogs that I need to re-write, plus redo the property changer list configurer, but those will take time and I will work in for beta 2.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
```diff
@@ -463,4 +467,60 @@ public static String stripImageSuffix(String name) {
...
+
+    // Determine the size of the translated string
+    final String s = Resources.getString("Editor.ImageUtils.no_image");
+    Image image = new BufferedImage(minWidth, height, BufferedImage.TYPE_4BYTE_ABGR);
+    Graphics2D g2d = (Graphics2D) image.getGraphics();
+    final double os_scale = g2d.getDeviceConfiguration().getDefaultTransform().getScaleX();
+    g2d.addRenderingHints(SwingUtils.FONT_HINTS);
+    g2d.setFont(FONT.deriveFont((float)(FONT.getSize() * os_scale)));
+    final int stringWidth = g2d.getFontMetrics().stringWidth(s);
+    final int imageWidth = Math.max(minWidth, stringWidth + 20);
```
uckelman, line 508: You need to call `g2d.dispose();` here since you're now done with the first Graphics2D.

vassal-app/src/main/java/VASSAL/build/module/properties/PropertyChangerConfigurer.java
```diff
@@ -101,7 +101,7 @@ public Component getControls() {
         updateControls();
       };
       controls = new JPanel();
-      controls.setLayout(new MigLayout("ins 0,hidemode 3", "[]rel[][]rel[fill,grow][fill,grow]")); // NON-NLS
+      controls.setLayout(new MigLayout("debug,ins 0,hidemode 3", "[]rel[][]rel[fill,grow][fill,grow]")); // NON-NLS
```
uckelman, line 104: Did you leave that `debug` there accidentally?

vassal-app/src/main/java/VASSAL/build/module/map/BoardPicker.java
```diff
@@ -176,9 +177,7 @@ public void addTo(Buildable b) {
     for (final Board board : possibleBoards) {
       board.setMap(map);
     }
-    if (b instanceof Translatable) {
-      getI18nData().setOwningComponent((Translatable) b);
-    }
+    getI18nData().setOwningComponent((Translatable) b);
```
uckelman, line 180: Isn't this going to throw a ClassCastException? There's no guarantee that every Buildable is a Translatable.

vassal-app/src/main/java/VASSAL/configure/ImageSelector.java
```diff
@@ -0,0 +1,190 @@
+package VASSAL.configure;
```
uckelman, line 1: Missing license.

vassal-app/src/main/java/VASSAL/configure/ImageSelector.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+    else {
+      icon.setOp(Op.load(s));
+
+      // Is the image too large?
+      if (maxWidth > 0 && (icon.getIconWidth() > maxWidth || icon.getIconHeight() > maxHeight)) {
+        final double xRatio = (double) maxWidth / icon.getIconWidth();
+        final double yRatio = (double) maxHeight / icon.getIconHeight();
+        final double newScale = Math.min(xRatio, yRatio);
+        icon.setOp(new ScaleOpBitmapImpl(Op.load(s), newScale));
```
uckelman, line 99: This will fail if the image is SVG. Use `Op.scale()` instead.

vassal-app/src/main/java/VASSAL/build/module/properties/PropertyChangerConfigurer.java
BrentEaston: Yes indeed! Fixed.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
BrentEaston: Fixed.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
```diff
@@ -463,4 +467,60 @@ public static String stripImageSuffix(String name) {
...
+   * @return Viewable null image
+   */
+  public static Image createViewableNullImage(int minWidth, int height) {
+
+    final int FONT_SIZE = 12;
+    final Font FONT = new Font(Font.DIALOG, Font.ITALIC, FONT_SIZE);
+
+    // Determine the size of the translated string
+    final String s = Resources.getString("Editor.ImageUtils.no_image");
+    Image image = new BufferedImage(minWidth, height, BufferedImage.TYPE_4BYTE_ABGR);
```
uckelman, line 502: We've had graphics glitches when we've tried to paint image types other than the "compatible" one. It would be better to create this image using `createCompatibleTranslucentImage(minWidth, height)` to avoid that.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
```diff
@@ -463,4 +467,60 @@ public static String stripImageSuffix(String name) {
...
+    Graphics2D g2d = (Graphics2D) image.getGraphics();
+    final double os_scale = g2d.getDeviceConfiguration().getDefaultTransform().getScaleX();
+    g2d.addRenderingHints(SwingUtils.FONT_HINTS);
+    g2d.setFont(FONT.deriveFont((float)(FONT.getSize() * os_scale)));
+    final int stringWidth = g2d.getFontMetrics().stringWidth(s);
+    final int imageWidth = Math.max(minWidth, stringWidth + 20);
+
+    // Create a new image large enough to hold the string comfortably
+    image = new BufferedImage(imageWidth, height, BufferedImage.TYPE_4BYTE_ABGR);
```
uckelman, line 511: Ditto here, for creating the image.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
```diff
@@ -463,4 +467,60 @@ public static String stripImageSuffix(String name) {
...
+  /**
+   * Create a viewable representation of a null or empty image to
+   * use as a place holder in Configurers.
+   *
+   * The image will contain the translated text for the key
+   * Editor.ImageUtils.no_image
+   *
+   * @return Viewable null image
+   */
+  public static Image createViewableNullImage() {
```
uckelman, line 480: Would be better to return a BufferedImage, since most everything dealing with images wants one of those.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
```diff
@@ -463,4 +467,60 @@ public static String stripImageSuffix(String name) {
...
+   * use as a place holder in Configurers.
+   * *
+   * The image will contain the translated text for the key
+   * Editor.ImageUtils.no_image
+   *
+   * @param minWidth Minimum width for generated the image
+   * @param height   Height of the generated image
+   * @return Viewable null image
+   */
+  public static Image createViewableNullImage(int minWidth, int height) {
```
uckelman, line 495: Ditto, return type.

vassal-app/src/main/java/VASSAL/build/module/map/BoardPicker.java
BrentEaston: Yes indeed. I fell for Intellij again that insists this is OK, but I agree, While most Buildables end up implementing Translatable, Buildable itself does not and the potential for a Class Cast Exception is there. Reverted.

vassal-app/src/main/java/VASSAL/configure/ImageSelector.java
BrentEaston: Fixed.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
uckelman: Also: As you're using this image just to get a Graphics2D from it, it's size doesn't matter; you can use the pre-existing `compatTransImage` static class member to get your Graphics2D for text sizing and avoid creating this BufferedImage.

vassal-app/src/main/java/VASSAL/configure/ImageSelector.java
BrentEaston: Thanks, fixed.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
```diff
@@ -463,4 +467,60 @@ public static String stripImageSuffix(String name) {
...
+
+    // Create a new image large enough to hold the string comfortably
+    image = new BufferedImage(imageWidth, height, BufferedImage.TYPE_4BYTE_ABGR);
+    g2d = (Graphics2D) image.getGraphics();
+    g2d.addRenderingHints(SwingUtils.FONT_HINTS);
+    g2d.setFont(FONT.deriveFont((float)(FONT.getSize() * os_scale)));
+    g2d.setColor(Color.white);
+    g2d.fillRect(0, 0, imageWidth - 1, height - 1);
+    g2d.setColor(Color.black);
+    g2d.drawRect(0, 0, imageWidth - 1, height - 1);
```
uckelman, line 518: You seem to be completely filling the image---in which case you don't need a translucent one. Use `createCompatibleImage()` and `compatOpaqueImage` instead.

vassal-app/src/main/java/VASSAL/build/widget/PieceSlot.java
```diff
@@ -83,6 +84,8 @@
 public class PieceSlot extends Widget implements MouseListener, KeyListener {
   public static final String GP_ID = "gpid"; //NON-NLS
   private static final int DEFAULT_SIZE = 64;
+  private static Image noImage;
```
uckelman, line 87: Is there a reason for not making this final and assigning it here?

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
BrentEaston: True. Fixed.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
BrentEaston: Fixed

uckelman (changes requested):
See comments. :)

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
BrentEaston: Fixed.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
BrentEaston: fixed

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
BrentEaston: Yes. Fixed.

vassal-app/src/main/java/VASSAL/tools/swing/SwingUtils.java
uckelman: Oh, huh. I suppose it makes sense that this property isn't defined in headless mode. Good catch.

vassal-app/src/main/java/VASSAL/build/widget/PieceSlot.java
BrentEaston: Nope. Fixed.

BrentEaston:
My first foray into image processing for a while, thanks for the tips. Updates pushed to the PR.

vassal-app/src/main/java/VASSAL/tools/image/ImageUtils.java
```diff
@@ -463,4 +467,60 @@ public static String stripImageSuffix(String name) {
...
+  public static Image createViewableNullImage(int minWidth, int height) {
+
+    final int FONT_SIZE = 12;
+    final Font FONT = new Font(Font.DIALOG, Font.ITALIC, FONT_SIZE);
+
+    // Determine the size of the translated string
+    final String s = Resources.getString("Editor.ImageUtils.no_image");
+    Image image = new BufferedImage(minWidth, height, BufferedImage.TYPE_4BYTE_ABGR);
+    Graphics2D g2d = (Graphics2D) image.getGraphics();
+    final double os_scale = g2d.getDeviceConfiguration().getDefaultTransform().getScaleX();
```
uckelman, line 506: Oh, just noticed this---this won't get the screen scale factor, because the Graphics2D isn't coming from the screen. I know what to do here, but it will take a small bit of refactoring to make it work, so you can leave this to me.

BrentEaston: Ah OK, thanks.

uckelman: PR here: https://github.com/BrentEaston/vassal/pull/10

BrentEaston: Done

uckelman (approved):

PR #590, opened 2020-11-30
13792 - Log Surgery - new-style log loading (alongside old-style)

Cattlesquat:
1. "Load Game or Log..." now the preferred option. Always available. Confirms option to save-before-doing-it if any changes are going to be wiped out. And then loads the target file "pure", not on top of something else.
2. "Load Old-Style Continuation..." perpetuates the Bad Old Version in case someone has some scary use case that now depends on it. Everyone Would Be Better Off If No One Used This.

We could consider having little dialogs or messages-of-explanation somewhere, but I couldn't think of non-annoying way off the top of my head. The fact that the preferred option (the new one) is the first one on the list, combined with the other one being marked "Old Style" may help.

Good news: VERY LITTLE CODE ACTUALLY CHANGED.


BrentEaston:
You could add a yes/no dialog to the old-style continuation open option explaining the implications of using this option. Since most people will no longer use this option most of the time, and you don't open log files that often anyway, it shouldn't be much of an issue.

Cattlesquat:
Good point, I can add it with a "don't show again" option. I actually load 15-20 logfiles a day -- not that I would be loading them "old style" any more, but if that were somehow my use case I definitely wouldn't want a constant dreary reminder. But warning at least once with a "stop bothering me" would make a lot of sense.

riverwanderer:
Allow me to chip in with an "old style" use-case, it's the only one I know but it's mundane more than scary so I thought worth mentioning. At least, I think its a use-case, if I understand this change correctly...

I've been playing more multiplayer live games recently and one in particular. For that one (8 players, lots of scenarios), it was useful to have name tags on screen. The module had a suitable Text labels in the palette, so I just made a log file of me putting the labels on the board. In this case, (8 players, lots of scenarios), it was a quick and dirty method of adding this functionality to a module that I don't maintain. Now I have a handy utility that I use before each game and that can be shared with other users.

Mark (marktb1961 on Vassal forums).

Cattlesquat:
I have introduced a descriptive warning that appears when you attempt to load "old style", explaining the risks and preferred flow. It includes a "don't prompt again" box for Mark :)

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -372,10 +385,51 @@ public boolean isGameStarted() {
...
+          JOptionPane.YES_NO_CANCEL_OPTION,
+          JOptionPane.QUESTION_MESSAGE,
+          null,
+          options,
+          options[0]
+        );
+        if (result == JOptionPane.NO_OPTION) {
+          return;
+        }
+        else if (result == 2) { // Turn Preference Off
```
uckelman, line 427: Can we use one of the predefined constants here? It's not obvious what `2` is.

Cattlesquat: I imagine we can. "2" is the condition I found this in but I'll go try to find something appropriate that's equal to 2.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -332,15 +345,15 @@ public void setup(boolean gameStarting) {
...
-      loadGame.putValue(Action.NAME,
-        Resources.getString("GameState.load_continuation"));
+      //loadGame.putValue(Action.NAME, Resources.getString("GameState.load_continuation"));
       g.getWizardSupport().showGameSetupWizard();
     }
     else {
-      loadGame.putValue(Action.NAME,
-        Resources.getString("GameState.load_game"));
+      //loadGame.putValue(Action.NAME, Resources.getString("GameState.load_game"));
     }
```
uckelman, line 353: Remove the empty else. (Probably no reason to keep the commented line in the if either.)

Cattlesquat: Oh right, this mess brought to you by the "let's see if this will work at all" phase

PR #592, opened 2020-12-01
13777 - PolygonEditor improvements

uckelman:
* Vertices may be added by double-clicking existing segments.
* Vertices may be deleted by selecting them and hitting Delete.
* Polygons may be drawn by placing successive vertices and closing by clicking on the initial vertex.


uckelman:
Going with PR 595 instead.

PR #596, opened 2020-12-02
3.5 13783 NonRectangular trait to remember imageName

BrentEaston:


vassal-app/src/main/java/VASSAL/counters/NonRectangular.java
```diff
@@ -58,9 +58,11 @@
 public class NonRectangular extends Decorator implements EditablePiece {
   public static final String ID = "nonRect;"; // NON-NLS
   private static final Map<String, Shape> shapeCache = new HashMap<>();
+  private static final Map<String, String> nameCache = new HashMap<>();
```
uckelman, line 61: Rather than having two parallel maps for this, I would use one map where the value is either a two-member data class with members `name` and `shape`, or an `org.apache.commons.lang3.Pair<String, Shape>`.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/counters/NonRectangular.java
BrentEaston: Updated.

vassal-app/src/main/java/VASSAL/counters/NonRectangular.java
```diff
@@ -142,28 +148,36 @@ private Shape buildPath(String spec) {
...
+            break;
+          // Note the image name is stored as a single token so older clients will ignore it.
+          case 'n':
+            imageName = token.length() > 1 ? token.substring(1) : "";
             break;
           }
         }
         sh = new Area(path);
-        shapeCache.put(spec, sh);
+        cache.put(spec, new ImmutablePair<>(imageName, sh));
```
uckelman, line 164: Tip: Pair.of() will make you an ImmutablePair and is shorter. Not a thing we need often; I only just discovered that.

uckelman (approved):

PR #603, opened 2020-12-07
13571 - Force the image and name arrays in Embellishment to have at least one element

uckelman:


uckelman:
@BrentEaston Do you see any problems with this?

BrentEaston (approved):
Looks good to me Joel.

PR #605, opened 2020-12-07
13805 - counterdetailviewer was showing images when told not to

Cattlesquat:


uckelman:
Cherry-picked to 3.4.12 instead.

PR #614, opened 2020-12-15
List Configurers

BrentEaston:


vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -605,7 +605,7 @@ private void initServer() {
     ChatServerFactory.register("jabber", oncf); //NON-NLS
 
     server = new DynamicClient();
-    final AddressBookServerConfigurer config = new AddressBookServerConfigurer("ServerSelected", "Server", (HybridClient) server); //NON-NLS
+    final AddressBookServerConfigurer config = new AddressBookServerConfigurer("ServerSelected", "", (HybridClient) server); //NON-NLS
```
uckelman, line 608: Why's the "Server" gone away here?

BrentEaston: I am going to recreate this, I seem to have merged in something old at some point.

PR #616, opened 2020-12-15
3.5 13764 list configurers final

BrentEaston:


vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -605,7 +605,7 @@ private void initServer() {
     ChatServerFactory.register("jabber", oncf); //NON-NLS
 
     server = new DynamicClient();
-    final AddressBookServerConfigurer config = new AddressBookServerConfigurer("ServerSelected", "Server", (HybridClient) server); //NON-NLS
+    final AddressBookServerConfigurer config = new AddressBookServerConfigurer("ServerSelected", "", (HybridClient) server); //NON-NLS
```
uckelman, line 608: Did the `"Server"` page name get missed out again?

BrentEaston: That isn't the 'Server' tab preference name, that is the individual Configurer label which we don't want to be displayed anymore. That was the stray 'Server' label on the left of the Server prefs page.

uckelman: Aha. Got it. Thanks.

PR #648, opened 2021-01-17
13910 - Module name column in Module Manager does not follow theme coloring when highlighted

uckelman:


uckelman:
Won't fix. This can't be made to work reliably across all platforms.

PR #676, opened 2021-02-04
3.5.1 - 13977 - Alt+Shift now a configurable option for band-selectability of does-not-stack pieces.

Cattlesquat:
Alt used to be the only allowable modifier key.


Cattlesquat:
That commit title was a typo (what I pasted included a carriage return)

PR #679, opened 2021-02-06
13991- fix GameRefresher stops when NewGame with hasFile flag set to true ex…

claudiociardelli:
…ists in module


vassal-app/src/main/java/VASSAL/configure/RefreshPredefinedSetupsDialog.java
```diff
@@ -133,8 +133,11 @@ private void refreshPredefinedSetups() {
     for (final PredefinedSetup pds : modulePdsAndMenus) {
       if (!pds.isMenu() && pds.isUseFile()) {
         //Exclude scenario folders (isMenu == true)
-        // and exclude any "New game" entries (no predefined setup) (isUseFile == true)
-        modulePds.add(pds);
+        // and exclude any "New game" entries (no predefined setup) (isUseFile == false)
+        // !! Some New Game entries have UseFile = true and filename empty. Check file name too
+        if (pds.getFileName() != null) {
```
BrentEaston, line 138: I would be even more defensive and use

if (pds.getFileName() != null && ! pds.getFileName().isBlank()) {

BrentEaston (changes requested):

PR #684, opened 2021-02-08
13999 more fixy fixy

Cattlesquat:


vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
```diff
@@ -127,7 +127,9 @@ public GlobalCommandTarget getTarget() {
    */
   private boolean passesPropertyFastMatch(GamePiece gamePiece) {
     if (!target.fastMatchProperty || fastProperty.isEmpty()) return true;
-    final String value = gamePiece.getProperty(fastProperty).toString();
+
+    final Object prop = gamePiece.getProperty(fastProperty);
+    final String value = (prop == null) ? "" : prop.toString();
```
uckelman, line 132: Can't you let null just be null instead of the empty string here, since you check for null below?

PR #692, opened 2021-02-09
Remove the "VASSAL Version" column from the Module Manager

uckelman:


uckelman:
Closed in favor of #704.

PR #707, opened 2021-02-13
14026 - Pieces without stacks now refresh too

claudiociardelli:


claudiociardelli:
This one is for 14026

PR #726, opened 2021-02-16
3.5.2 - 14053 - canceling welcome wizard closes app unless show-wizard-at-sta…

Cattlesquat:
…rtup is unchecked

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
@@ -193,13 +193,8 @@ public class WizardSupport {
         if (g.shutDown()) {
           System.exit(0);
         }
-        else {
-          g.getPlayerWindow().setVisible(true);
-        }
       }
-      else {
-        g.getPlayerWindow().setVisible(true);
-      }
+      g.getPlayerWindow().setVisible(true);
     }
   }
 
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
```diff
@@ -188,7 +188,18 @@ protected Void doInBackground() {
...
+        if (g.shutDown()) {
+          System.exit(0);
+        }
+        else {
+          g.getPlayerWindow().setVisible(true);
+        }
+      }
+      else {
+        g.getPlayerWindow().setVisible(true);
+      }
```
uckelman, line 202: Couldn't you eliminate both `else` cases and stick `g.getPlayerWindow().setVisible(true);` below the end of the outer `if`?

uckelman: In fact, just have one `if`: `if (Boolean.TRUE.equals(showWizard) && g.shutDown())`

Cattlesquat: done

Cattlesquat: Alzo Alzo done

uckelman:
@BrentEaston Thoughts on this change?

BrentEaston:
I have no problem with that.

PR #728, opened 2021-02-16
14056 - 3.5.2 - SEARCH - min/max values were getting left out of the search fields list

Cattlesquat:


vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -511,6 +511,12 @@ Editor.DynamicProperty.change_value=Change value
 Editor.DynamicProperty.change_value_of=Change value of %1$s
 # Property <name>: Menu Command 2
 Editor.DynamicProperty.command_description=Property %1$s: Menu command %2$s
+# When displayed in a list of search results
+Editor.DynamicProperty.min_value= (min value)
+# When displayed in a list of search results
+Editor.DynamicProperty.max_value= (max value)
+# When displayed in a list of search results
+Editor.DynamicProperty.initial_value= (initial value)
```
uckelman, line 519: This will lead to someone removing the spaces. I think it would be better to interpolate the values in to prevent that. So, e.g., `%1$d (min value)`.

Cattlesquat: done

PR #732, opened 2021-02-18
14064 - RefreshPredefinedSetups in Extension Editor

claudiociardelli:
Bug 14064
Adds Refresh Predefined Setups to ExtensionEditor 
+ calcel warinings on DoNoT Stack without parent


BrentEaston:
Hi @claudiociardelli looks good except for one thing. When you run Refresh Counters while editing an Extension, it ALSO refreshes the counters in the main module. I don't think it should do this. It is confusing as the main module can't be written back out when you edit an extension. Edit Extension Refresh counters should only refresh any scenarios contained in the extension.

Otherwise, I gave it a good test on my dodgy BfM module and it handled the non-stack pieces well, putting them back into Stacks which was nice instead of deleting them.

claudiociardelli:
Hello Brent. I managed to identify the Extension Setups. I do that by looking up the vmod zip file to check if the setup is in that file or not.

PR #737, opened 2021-02-19
14070 - Array Configurers move cursor to newly created entry

BrentEaston:


uckelman:
Where's a good place to look to test this?

BrentEaston:
Define Sides String Lists
Trigger Action for Hotkey Lists.
Dynamic Property for KeyCommandLists

uckelman:
Maybe I don't understand what this is supposed to do, in that case. I went to the Player Sides in the PoG module and added a new side. The insertion point didn't appear in the new row.

BrentEaston:
Hmm. Yes, that is a good test and it certainly does put the insertion point in a new row on Windows. I will check linux.

BrentEaston:
Works perfectly for me on Ubuntu as well ???

uckelman:
Uh, I have no idea what was wrong before. This behaved as advertised when I checked it just now.

PR #741, opened 2021-02-20
14066 - Remove Unused Image dialog improvements - Layout and file sizes

BrentEaston:


vassal-app/src/main/java/VASSAL/tools/DataArchive.java
```diff
@@ -169,6 +170,26 @@ public InputStream getImageInputStream(String fileName)
       "'" + imageDir + fileName + "' not found in " + getName());
   }
 
+  /**
+   * Return the size in bytes of a file in the Archive
+   * @param fileName File name (include path if required)
+   * @return size in bytes
+   */
+  public long getFileSize(String fileName) {
```
uckelman, line 178: Is there a reason you're doing this instead of using `FileArchive.getSize()`?

BrentEaston: Just unfamiliarity with the code. Will FileArchive,getSize() return the compressed or uncompressed size? Is there an easy way to get the compressed size? That would be better as it relates directly to the issue of reducing the module size.

uckelman: It's uncompressed size---but most images barely compress at all, so uncompressed size is a pretty good proxy for compressed size.

BrentEaston: Turned out to be easy to access the compressed size. Try this new version

uckelman:
This looks ok. Have you tried it when editing an extension?

vassal-app/src/main/java/VASSAL/tools/DataArchive.java
```diff
@@ -202,7 +202,7 @@ public InputStream getInputStream(String fileName)
     // HFS+ filesystem and the filename contains decomposable characters,
     // so it got munged into NFD. Aauugh! Seriously, DIAF Apple!
     final String nfd = Normalizer.normalize(fileName, Normalizer.Form.NFD);
-    if (fileName != nfd) {
+    if (!fileName.equals(nfd)) {
```
uckelman, line 205: Good catch.

BrentEaston:
Ah, good point. Looks like a couple of problems editing extensions. 
1. Images from the module are being included in the list as well as images from the extension
2. All images from the extension are reporting a size of -1 byte (the fall back if ZipArchive.getCompressedSized() can't find the file.

BrentEaston:
That last push should resolve both issues and add support for removing images from Extensions.

PR #743, opened 2021-02-20
14078 - SumStack to report correctly for unstacked counter

BrentEaston:


Cattlesquat (commented):
Are you sure this fixes all the SumStack problems? It would make a single counter that went back to an empty space now report "1" instead of 0. But are there times when multiple undos could send multiple single unstacked counters to the same space and they all end up unstacked? (So it returns "1" when it should return e.g. "2" or more)? Or when the single counter going back gets left unstacked but in a space where there are still other counters? Maybe none of those things are possible?

BrentEaston:
When Undo returns a counter to a space that has an existing counter, they are merged correctly. The Unstacking behaviour only occurs when a single counter is Undone back to an empty space.

PR #744, opened 2021-02-21
14104 - Do not set image name to (No Image) when clearing Image

BrentEaston:
This fixes the bug that is causing the problem, but makes no attempt to hide the error or fix the module. Do we need this?


uckelman:
Closed in favor if PR #745.

PR #755, opened 2021-02-24
14131 - Beanshell recognises all usable Java 11 String functions

BrentEaston:


vassal-app/src/main/java/bsh/BeanShellExpressionValidator.java
```diff
@@ -38,7 +40,54 @@
...
+    supportedStringFunctions.add(".replaceAll");
+    supportedStringFunctions.add(".replaceFirst");
+    supportedStringFunctions.add(".startsWith");
+    supportedStringFunctions.add(".strip");
+    supportedStringFunctions.add(".stripLeading");
+    supportedStringFunctions.add(".stripTrailing");
+    supportedStringFunctions.add(".substring");
+    supportedStringFunctions.add(".toLowerCase");
+    supportedStringFunctions.add(".toUpperCase");
+    supportedStringFunctions.add(".translateEscapes");
```
uckelman, line 87: `translateEscapes` was introduced in Java 15, but we still support Java 11.

vassal-app/src/main/java/bsh/BeanShellExpressionValidator.java
```diff
@@ -38,7 +40,54 @@
...
+  static {
+    supportedStringFunctions = new HashSet<>();
+    supportedStringFunctions.add(".compareTo");
+    supportedStringFunctions.add(".compareToIgnoreCase");
+    supportedStringFunctions.add(".contains");
+    supportedStringFunctions.add(".endsWith");
+    supportedStringFunctions.add(".equals");
+    supportedStringFunctions.add(".equalsIgnoreCase");
+    supportedStringFunctions.add(".format");
+    supportedStringFunctions.add(".formatted");
```
uckelman, line 67: `formatted` was introduced in Java 15, but we still support Java 11.

vassal-app/src/main/java/bsh/BeanShellExpressionValidator.java
```diff
@@ -38,7 +40,54 @@
...
+   *
+   *    stringVariable.function()
+   *
+   * And allow stringVariable to be used directly within beanshell with Java syntax,
+   * rather than having to resort to
+   *
+   *    GetProperty("stringVariable").function()
+   *
+   */
+  private static final Set<String> supportedStringFunctions;
```
uckelman, line 57: You can add all these without a `static` block using `Set.of(...)`:
```
  private static final Set<String> supportedStringFunctions = Set.of( 
    ".compareTo",
    ".compareToIgnoreCase",
    ...
  );
```

uckelman (changes requested):

vassal-app/src/main/java/bsh/BeanShellExpressionValidator.java
BrentEaston: Updated

vassal-app/src/main/java/bsh/BeanShellExpressionValidator.java
BrentEaston: This list is invisible to users, it does not enable or provide access to these functions, or display them on a menu. It just allows a user to use them as 'native' functions on property names.  I am not adding these functions to the Function Builder menu, but if a smarty-pants designer uses one in their design, Beanshell will not fail with the exception reported by Brian. If they use 'formatted' or 'translateEscapes', they will get a different Beanshell error on a pre-Java 15 system, but that's their problem.

BrentEaston:
Should be good to go.

vassal-app/src/main/java/bsh/BeanShellExpressionValidator.java
uckelman: The problem is that they'll appear to work for someone designing on Java 15, but then someone else using Java 11 will get exceptions.

BrentEaston: Ok, I will remove those two from the list.

uckelman (approved):

PR #760, opened 2021-02-25
14139 - Delete trait to set old... properties before deleting unit

BrentEaston:
This one that should have been done ages ago. Delete is really a 'Move' type trait, so the oldxxxx Properties need to be set prior to deleting the counter. There have been many complaints and questions about correctly reporting the location of deleted counters over the years.

Currently, it is impossible to report the location a counter is being deleted from if it has not been moved since being added to the map.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Delete.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Delete.java
@@ -119,7 +119,8 @@ public class Delete extends Decorator implements TranslatablePiece {
           SwingUtilities.invokeLater(runnable);
         }
       }
-      c = new RemovePiece(outer);
+      c = putOldProperties(Decorator.getOutermost(this));
+      c = c.append(new RemovePiece(outer));
       c.execute();
     }
     return c;


riverwanderer:
I guess I have misunderstood something about this topic but I understood that once Delete happens the deleted counter can't execute any further actions, so I was thinking that oldxxxx properties could serve no purpose after a Delete.

Cattlesquat:
Yes I can't quite figure this out myself -- presently once the Delete trait executes it's "flatline" as far as what's going in Vassal is concerned, so if I want to Report on the Delete then I make sure that executes first and THEN fire a key command that actually performs the delete.

Will this make it so that a Report trait will still "fire" after a Delete trait in a piece? Will other traits be able to fire "post mortem" too?

BrentEaston:
Report Actions are executed last after the 'Actionable'  traits have been processed regardless of what has happened the counter. The fact that the deleted counter can perform no more 'actions' is because most actions check if the piece has a map or not, and if not does nothing. Report Action doesn't make this check so that it can report on Deleted counters. This change will allow the Report Action trait to correctly report the delete map and location of a unit, regardless of the state of the unit prior to deleting.

BrentEaston:
Report Actions ALWAYS fire after Delete and all other traits.

PR #762, opened 2021-02-26
14153 - Fix Chatter losing first character of chat when counters selected

BrentEaston:


Cattlesquat (commented):
Looks good to me! Works in PoG.

PR #767, opened 2021-03-02
3.5.4 - 14154 - SEARCH wasn't finding fast-match fields

Cattlesquat:


vassal-app/src/main/java/VASSAL/counters/GlobalCommandTarget.java
```diff
@@ -244,6 +249,90 @@ public boolean equals(Object o) {
...
+    final List<String> expList = new ArrayList<>();
+
+    if (fastMatchLocation) {
+      if ((targetType == Target.MAP) || (targetType == Target.ZONE) || (targetType == Target.LOCATION) || (targetType == Target.XY)) {
+        expList.add(targetMap.getExpression());
+      }
+      if (targetType == Target.XY) {
+        expList.add(targetBoard.getExpression());
+      }
+      if (targetType == Target.ZONE) {
```
uckelman, line 282: Don't you want these to be `else if` from here down, since these are mutually exclusive with the one above it?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/counters/GlobalCommandTarget.java
Cattlesquat: True

uckelman (approved):

PR #769, opened 2021-03-02
14156 - Loading a VLOG file to continue a game shouldn't reset the ma…

Cattlesquat:
…p window

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Decorator.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Decorator.java
@@ -96,7 +96,7 @@ public abstract class Decorator extends AbstractImageFinder implements GamePiece
   }
 
   /**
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/counters/Decorator.java
```diff
@@ -96,7 +96,7 @@ public Map getMap() {
   }
 
   /**
-   * @return the next member "inward" (toward the BasicPiece) of the full GamePiece.
+   * @return the next member "inward" (toward the BasicPiece) of the full GamePiece.i
```
uckelman, line 99: Typo?

Cattlesquat: Apparently.

PR #785, opened 2021-03-11
14199 - Inventory text not translated

javiermunozk:
Fix inventory text translation


vassal-app/src/main/java/VASSAL/build/module/Inventory.java
```diff
@@ -460,7 +460,15 @@ private void buildTreeModel() {
       if (p instanceof Decorator || p instanceof BasicPiece) {
         for (final String s : groupBy) {
           if (s.length() > 0) {
-            final String prop = (String) p.getProperty(s);
```
uckelman, line 463: Is there a reason I'm missing that you can't make the whole PR changing this line to
```
final String prop = (String) p.getLocalizedProperty(s);
```
?

PR #786, opened 2021-03-11
TranslatableMessage - like a Marker, except it's readonly & localizable.

Cattlesquat:


vassal-app/src/main/java/VASSAL/counters/TranslatableMessage.java
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  /**
+   * @return a list of any Property Names referenced in the Decorator, if any (for search)
+   */
+  @Override
+  public List<String> getPropertyList() {
+    return List.of(key);
+  }
+}
```
BrentEaston, line 239: Something like:

  @Override
  public PieceI18nData getI18nData() {
    return getI18nData(message, Resources.getString("Editor.TranslatableMessage.property_value"));
  }

BrentEaston (changes requested):
That should do it for the trait, it is pretty simple.

PR #789, opened 2021-03-12
14207 - Global Translatable Message (these work "kind of like" Global Properties)

Cattlesquat:


vassal-app/src/main/java/VASSAL/i18n/PieceI18nData.java
```diff
@@ -46,6 +46,22 @@ public void add(String value, String description) {
...
   }
+
+  public String translate(String value) {
+    String localisedValue = value;
+    String i18nKey = null;
+
+    for (final Property p : getProperties()) {
+      if (p.getName().equals(value)) {
+        i18nKey = TranslatablePiece.PREFIX + p.getName();
+      }
```
jonathanrwatts, line 57: Would `i18nKey = TranslatablePiece.PREFIX + value` be more efficient?
Also, wouldn't it make sense to exit the loop once a match has been found, or is important to use the last match if there are multiples?

Cattlesquat: @BrentEaston You'd be the one to comment on this one, I think it came from your PR-into-PR (plus I'd recognize that Commonwealth "localised" anywhere, haha!)

BrentEaston: value/p.getName(): In general, that sort of thing is handled by the Java optomizer, you don't really have to worry about efficiency at that level.
Ending loop early: Theoretically, yes, but in practice no Trait has more than one translatable element, so it is moot.

jonathanrwatts: Sounds good, thanks for the explanations.

uckelman: I would put a break in that if block---otherwise that will confuse the next person to look at it.

Cattlesquat: I have done so (and rebased the PR since it has developed conflicts with master)

vassal-app/src/main/java/VASSAL/build/module/properties/GlobalTranslatableMessages.java
```diff
@@ -0,0 +1,177 @@
...
+  public Object getLocalizedProperty(Object key) {
+    return propertySource == null ? null : propertySource.getLocalizedProperty(key);
+  }
+
+  @Override
+  public GlobalTranslatableMessage getTranslatableString(String name) {
+    GlobalTranslatableMessage property = null;
+    for (final GlobalTranslatableMessage prop : getComponentsOf(GlobalTranslatableMessage.class)) {
+      if (prop.getConfigureName().equals(name)) {
+        property = prop;
```
uckelman, line 143: Should we either break out of the loop here or just return?

vassal-app/src/main/java/VASSAL/counters/TraitLayout.java
```diff
@@ -27,7 +27,7 @@
   private static final long serialVersionUID = 1L;
 
   public static final String DEFAULT_TRAIT_LAYOUT_CONSTRAINTS = STANDARD_INSETS + "," +  STANDARD_GAPY + ",hidemode 3,wrap 2"; // NON-NLS
-  public static final String DEFAULT_TRAIT_COLUMN_CONSTRAINTS = "[right]rel[fill,grow]"; // NON-NLS
+  public static final String DEFAULT_TRAIT_COLUMN_CONSTRAINTS = " [right]rel[fill,grow,400::]"; // NON-NLS
```
uckelman, line 30: Nix leading space.

vassal-app/src/main/java/VASSAL/build/module/properties/GlobalTranslatableMessages.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/counters/TraitLayout.java
Cattlesquat: done

PR #795, opened 2021-03-15
3.5.4 - 14217 - $countPieces$ to sum all pieces without having to create a property

Cattlesquat:
Existing functionality only allows $sum(somePropertyName) and then sums the property, which can be useful but is unfortunately infuriating when you only want a sum of how many pieces met the qualifications to be displayed, because then you have to go add a property to every possible piece "StupidProperty" and set its value to "1". So that you can sum it. 

Anyway this fixes that.

### DIFF
--- a/vassal-doc/src/main/readme-referencemanual/ReferenceManual/Map.adoc
+++ b/vassal-doc/src/main/readme-referencemanual/ReferenceManual/Map.adoc
@@ -316,7 +316,7 @@ Otherwise, it will display only if 1 or 2 pieces have been included via the sett
 *Font size:* Fort size for the text drawn by the stack viewer.
 
 *Summary text above pieces:*  A <<MessageFormat.adoc#top,Message Format>> specifying the text to display above the drawn pieces in the viewer.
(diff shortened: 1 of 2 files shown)


vassal-doc/src/main/readme-referencemanual/ReferenceManual/Map.adoc
```diff
@@ -316,7 +316,7 @@ Otherwise, it will display only if 1 or 2 pieces have been included via the sett
 *Font size:* Fort size for the text drawn by the stack viewer.
 
 *Summary text above pieces:*  A <<MessageFormat.adoc#top,Message Format>> specifying the text to display above the drawn pieces in the viewer.
-In addition to standard <<Properties.adoc#top,Properties>>, you can include $sumpieces$ to sum the pieces included in the display. You can also include a property with the name _sum(propertyName)_ where _propertyName_ is a property defined on a Game Piece.
+In addition to standard <<Properties.adoc#top,Properties>>, you can include $countPieces$ to sum the pieces included in the display. You can also include a property with the name _sum(propertyName)_ where _propertyName_ is a property defined on a Game Piece.
```
jonathanrwatts, line 319: Perhaps I'm being overly pedantic, but wouldn't it make sense to change the word "sum" to "count" here, too (as in, "you can include $countPieces$ to **count** the pieces").

PR #796, opened 2021-03-17
14205 - Customized images2

javiermunozk:


vassal-app/src/main/java/VASSAL/i18n/I18nResourcePathFinder.java
```diff
@@ -0,0 +1,81 @@
...
+    //Not found or not localized
+    path = DataArchive.IMAGE_DIR + name;
+    return path;
+  }
+
+  @Override
+  public String findHelpFileName(String name) {
+    String modifiedName = name;
+    if (!language.equals("en")) {
+      if (name.contains("help/")) {
```
uckelman, line 66: Should this be checking if `name` starts with `"help/"` instead? (I'm not sure.)

javiermunozk: I wanted it to be more flexible than images, because "images" is hardcoded in the code, but help is not. If you feel it's better to change it, I can do that without a problem,

uckelman: What I'm concerned about is if `help/` occurs somewhere else in the name other than at the start. Do you intend for `foo/bar/help/stuff.html` to go to this case?

javiermunozk: Yes, that as my intention, but if you prefer I can change to starts-with :)

uckelman: No, that's fine, so long as it's intentional.

javiermunozk: Thanks! If there is anything else to fix, just please tell me :)

PR #798, opened 2021-03-17
3.6 - 14278 - Increased # of default zoom levels (since modern monitors mean people use vassal at a wider variety of resolutions)

Cattlesquat:
http://www.vassalengine.org/forum/viewtopic.php?f=4&t=13014

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
@@ -108,15 +108,21 @@ public class Zoomer extends AbstractConfigurable implements GameComponent {
 
   protected State state;
 
-  // the default zoom levels are powers of 1.6
   protected static final double[] defaultZoomLevels = {
-    1.0 / 1.6 / 1.6,
-    1.0 / 1.6,
+    0.2,
+    0.25,
+    0.333,
+    0.4,
+    0.5,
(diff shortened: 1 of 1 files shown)


BrentEaston:
Waiting for comments from Tim Mcarron

Cattlesquat:
@BrentEaston Hopefully you used some sort of e-mail summoning spell to let Tim know to take a look?

BrentEaston:
@Cattlesquat Have now, just assumed he would read the forum.

I think the < 1.0 levels are good. Do we need a couple of extra > 1.0 levels though? I'm not sure how much they are used?

Cattlesquat:
@BrentEaston I feel like the trend in maps in modules has been mostly to get "even more hi-res than ever before". Which means that on more kinds of monitors you need to zoom out, and mainly that there are more different appropriate levels of zooming out that different users need -- I based my new recommendations on the idea that "many designers won't bother setting these at all" and so therefore trying to provide a good all-around experience when the settings aren't customized. I stuck the "1.25" level in there just because it seemed aesthetically wrong not to - but it's certainly difficult for me to imagine needing a zoom-IN to even 1.6 let alone 2.0+ -- and in the game industry we're also taught to try to avoid zooming IN beyond 100% at all costs because then the artists come to our offices and whine and moan a lot :)

PR #799, opened 2021-03-18
3.5.4 - 14219 - don't traverse INTO prototypes when searching a piece

Cattlesquat:


vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -1916,27 +1915,36 @@ else if (c instanceof SearchTarget) {
...
+      // We're going to search Decorator from inner-to-outer (BasicPiece-on-out), so that user sees the traits hit in
+      // the same order they're listed in the PieceDefiner window. So we first traverse them in the "normal" direction
+      // outer to inner and make a list in the order we want to traverse it (for architectural reasons, just traversing
+      // with getOuter() would take us inside of prototypes inside a piece, which we don't want).
+      final List<GamePiece> pieces = new ArrayList<>();
+      pieces.add(p);
+      while (p instanceof Decorator) {
+        p = ((Decorator) p).getInner();
+        pieces.add(0, p);
+      }
```
uckelman, line 1927: I know this won't generally involve thousands of items, but rather than adding all the pieces to the head of the list and shifting them all on each add, it would be faster to add them to the back and reverse the list with `Collections.reverse()`.

Cattlesquat: I tried changing it to pieces.add(p); and then at the end "pieces.reverse();" but it didn't know what the hell I was talking about. If you can come up with code that compiles I'm happy to use :)

uckelman: `Collections.reverse(pieces);` ought to work.

Cattlesquat: done

PR #803, opened 2021-03-19
DisableTranslatableMessages until 3.6, except if translatableSupport=…

Cattlesquat:
…true found in V_Global


uckelman:
To be clear, those `if (isTranslatableSupport()) {` checks will be removed for 3.6?

Cattlesquat:
> To be clear, those `if (isTranslatableSupport()) {` checks will be removed for 3.6?

Yes. Unless we just made it always return true in some kind of custom-class-supporting religious fervor :)

PR #804, opened 2021-03-20
14164 - Update Fast Match Configurers correctly on Cancel

BrentEaston:


uckelman:
The packaging failed because launch4j has a 50-character limit on version name length, and this branch has a rather long name. Everything else passed, so I'm going to merge this.

PR #805, opened 2021-03-20
14222 - don't lose '<' symbol in displaying search results in Editor

Cattlesquat:


vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -239,6 +239,11 @@ public ConfigureTree(Configurable root, HelpWindow helpWindow, EditorWindow edit
   }
 
 
+  protected static String noHTML(String text) {
+    return text.replaceAll("<", "&lt;")  //NON-NLS // This prevents any unwanted tag from functioning
+               .replaceAll(">", "&gt;"); //NON-NLS // This makes sure > doesn't break any of our legit <div> tags
+  }
+
```
jonathanrwatts, line 246: Why not use the existing EscapeHtml3() method, from org.apache.commons.lang3.StringEscapeUtils?

uckelman: So far as I know, we don't need to escape anything else.

jonathanrwatts: My concern is people might use ampersands without surrounding whitespace in a name, which I would assume would be interpreted as a HTML entity and therefore not display correctly.  Haven't tested that, though (I do have a couple of Maps in Terraforming Mars with ampersands in the names, but they're surrounded by whitespace).

jonathanrwatts: Just did a bit of testing with a 3.5.4 Snapshot (cb4f850e2), and it seems to handle ampersands OK.  I was able to confuse it by actually putting "&amp;" into the name of an object, (the search results just displayed "&", not "&amp;"), but otherwise "<" was the only thing I could find that didn't display properly (I'm assuming that snapshot doesn't include this commit).

PR #809, opened 2021-03-22
14238 - Review, fix and update ValidityCheckers

BrentEaston:


BrentEaston:
Replace with new PR with shorter branch name

PR #812, opened 2021-03-22
14245 - Works around bug in JTabbedPane

Cattlesquat:
When module designers put html (e.g. "<html><b>Bold Boy</b></html>") in the text of the widgets for chart windows, Swing seems to do some grotesque things (losing track of the ordering of the list of tabs). Refreshing them "every time you do anything" seems to be a workaround. Note that when refreshing them you have to kind of "unset" and then "set" their title otherwise Swing has some "oh look it's the same as before so I will do nothing" code that prevents the desired relief.


BrentEaston:
@claudiociardelli We need to test GTS with this change to make sure it doesn't interfere with our tab naming shennanigans,

BrentEaston:
@claudiociardelli Seems to be fine with the latest GTS code for the Scenario card tabs.

PR #816, opened 2021-03-23
"Smooth" version of Load Game-or-Log

Cattlesquat:
(Since we had to revert the old version...)

This suppresses "UI-Fidgeting" by e.g. Map windows during the load-game-over-top-of-game process.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/GameState.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameState.java
@@ -540,11 +540,14 @@ public class GameState implements CommandEncoder {
         //BR// New preferred style load for vlogs is close the old stuff and hard-reset to the new log state.
         if (gameStarted) {
           GameModule.getGameModule().setGameFileMode(GameModule.GameFileMode.NEW_GAME);
+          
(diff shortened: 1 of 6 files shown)


vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -540,11 +535,16 @@ else if (!saveModuleVersion.equals(moduleVersion)) {
...
-        //BR// New preferred style load for vlogs is close the old stuff and hard-reset to the new log state.
-        final boolean foreground = gameStarted;
         g.setGameFile(f.getName(), GameModule.GameFileMode.LOADED_GAME);
-        if (foreground) {
+
+        //BR// New preferred style load for vlogs is close the old stuff and hard-reset to the new log state.
+        if (gameStarted) {
+          GameModule.getGameModule().setGameFileMode(GameModule.GameFileMode.NEW_GAME);
+          GameModule.getGameModule().setLoadOverSemaphore(true); // Stop updating Map UI etc for a bit
+
```
uckelman, line 544: Can there be an exception between here and where you unset the semaphore? If so, it needs to be in a `finally`.

PR #817, opened 2021-03-24
14247 - NPE in FastMatch when CurrentZone returns null

Cattlesquat:
…ome back null for a piece

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
+++ b/vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
@@ -467,13 +467,13 @@ public class GlobalCommand {
                   switch (target.targetType) {
                   case ZONE:
                   case CURZONE:
-                    if (!fastZone.equals(gamePiece.getProperty(BasicPiece.CURRENT_ZONE))) {
+                    if ((fastZone != null) && !fastZone.equals(gamePiece.getProperty(BasicPiece.CURRENT_ZONE))) {
                       continue;
                     }
                     break;
                   case LOCATION:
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
```diff
@@ -467,25 +467,25 @@ else if (!target.targetType.isCurrent() && !fastMap.isEmpty() && !fastMap.equals
                   switch (target.targetType) {
                   case ZONE:
                   case CURZONE:
-                    if (!fastZone.equals(gamePiece.getProperty(BasicPiece.CURRENT_ZONE))) {
+                    if ((fastZone != null) && !fastZone.equals(gamePiece.getProperty(BasicPiece.CURRENT_ZONE))) {
```
uckelman, line 470: Should this to be `fastZone == null ||` instead? If I understand this right, the `continue` case is for a non-match.

Cattlesquat: done

uckelman:
Waiting for user to reply to @Cattlesquat.

BrentEaston:
Brian, Is this ready to go? It seems like a fairly straight-forward defensive programming change.

BrentEaston:
Also, it likes the cause is that fastZone is returning null, not CurrentZone. Is there a problem in our Configurer?

Cattlesquat:
Oh actually this can be close now (and 14247) because of https://github.com/vassalengine/vassal/pull/872

PR #818, opened 2021-03-24
Fix the Refresh Predefined Setups flow that I broke

Cattlesquat:
I had removed a "gameLoadingInForeground" flag that I thought *I* had put in, since it was interfering with using the LoadGameInForeground method. But it turns out @claudiociardelli had put it in for the Refresh Predefined Setups tool -- so in other words I broke that. This restores the semaphore that should get Refresh Predefined Setups back to how it previously behaved while letting LoadGameInForeground still work for me -- by slightly changing where the semaphore is activated.


uckelman:
```
This Is Just To Say

I have clobbered
the semaphore
that was in
GameModule

and which
you were probably
using
for something else

Forgive me
it was handy
so defined
and so boolean
```

PR #828, opened 2021-03-27
14269 Fix poor editor performance using Fast Match

BrentEaston:


BrentEaston:
Mark confirms this resolves the problem. Ready for Merging.

PR #833, opened 2021-03-27
13796 - Fix problem with multiple stacks dragged to a snap-to map

Cattlesquat:
This should be "looked at carefully by experts". 

REALLY I wish the whole "offset" thing didn't exist -- it seems to bounce "p" back and forth between values in a maybe-not-very-understandable way. 

Like couldn't we just use **dragging.getPosition()** for each unit coming through the loop as the "p" for that unit? Maybe I'm missing something important that bouncing around with offset does?

Anyway removing the "offset == null" **seems** to fix 13796. Only adding making the offset when offset == null seemed to be the reason that only the "first stack it started making" got created correctly.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -728,10 +728,8 @@ public class PieceMover extends AbstractBuildable
           p = map.snapTo(p);
         }
 
-        if (offset == null) {
-          offset = new Point(p.x - dragging.getPosition().x,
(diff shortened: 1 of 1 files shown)


uckelman:
I think the reason you can't use `getPosition()` is that it returns the logical position, not the display position.

Cattlesquat:
> I think the reason you can't use `getPosition()` is that it returns the logical position, not the display position.

Ah. Well it would be nice if this loop didn't have the weird delayed-offset-bouncing-arround and just manufactured each "p" correctly/directly at the top of the loop. It's the one remaining "barely comprehensible" part of the logic here.

PR #835, opened 2021-03-28
3.6 - 14276 - expose Mapname_isVisible property

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -2064,6 +2065,20 @@ else if (GameModule.MODULE_CURRENT_LOCALE.equals(key)) {
...
+    //BR// MapName_isVisible property for each map window
+    for (final Map map : Map.getMapList()) {
+      if ((map.getConfigureName() + "_isVisible").equals(key) || (map.getConfigureName().replaceAll(" ", "_") + "_isVisible").equals(key)) { //NON-NLS
+        final Container tla = (map.getComponent() != null) ? ((JPanel)map.getComponent()).getTopLevelAncestor() : null;
+        if ((tla != null) && tla.isVisible()) {
+          return "true"; //NON-NLS
+        }
+        else {
+          return "false"; //NON-NLS
+        }
```
uckelman, line 2078: `return String.valueOf(tla != null && tla.isVisible());` could replace this block.

uckelman: Also, do you want `isShowing()` instead of `isVisible()`? I ran into this when fixing some of those `IllegalStateException`s with components that aren't displayed.

BrentEaston: I've tested this out, works well. I believe for Map views, isVisible() will be equivalent to isShowing(), but isShowing() is definitely safer.

BrentEaston (approved):
Looks good to me, just simplify that block as per Joel's suggestion.

Cattlesquat:
done

PR #844, opened 2021-03-30
13837 - Creating a Global Property with Null name causes NPE on L…

javiermunozk:
…oad Game

Problem fixed and validator added so an error is shown in the editor

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/NotNullConfigureName.java
+++ b/vassal-app/src/main/java/VASSAL/configure/NotNullConfigureName.java
@@ -1,0 +1,39 @@
+/*
+ *
+ * Copyright (c) 2004 by VASSAL team
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License (LGPL) as published by the Free Software Foundation.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU Library General Public
+ * License along with this library; if not, copies are available
+ * at http://www.opensource.org.
+ */
+package VASSAL.configure;
+
+import VASSAL.build.AbstractConfigurable;
+import VASSAL.build.Buildable;
+import VASSAL.i18n.Resources;
+/**
+ * Requires that the object does not have null configurable name
+ */
+public class NotNullConfigureName implements ValidityChecker {
+  private final AbstractConfigurable target;
+
+  public NotNullConfigureName(AbstractConfigurable target) {
+    this.target = target;
+  }
+
+  @Override
+  public void validate(Buildable b, ValidationReport report) {
+    if (b == this.target && target.getConfigureName().equals("")) {
+      report.addWarning(Resources.getString("Editor.ValidationReportDialog.blank_name") + " " + target.getClass().getName());
+    }
+  }
+}
--- a/vassal-app/src/main/java/VASSAL/configure/NotNullConfigureName.java
+++ b/vassal-app/src/main/java/VASSAL/configure/NotNullConfigureName.java
@@ -17,6 +17,7 @@
  */
 package VASSAL.configure;
 
+import java.util.Objects;
 import VASSAL.build.AbstractConfigurable;
 import VASSAL.build.Buildable;
 import VASSAL.i18n.Resources;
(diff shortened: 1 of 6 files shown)


vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -1930,6 +1930,7 @@ Editor.UsernameAndPasswordDialog.choose_your_weapons=Set your username and passw
 Editor.UsernameAndPasswordDialog.turn_your_key_sir=Please confirm your password
 
 # Validation Report Dialog
+Editor.ValidationReportDialog.blank_name=A invalid blank name found at
```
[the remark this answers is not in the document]
javiermunozk: Thanks, will fix.

vassal-app/src/main/java/VASSAL/configure/NotNullConfigureName.java
```diff
@@ -0,0 +1,39 @@
...
+public class NotNullConfigureName implements ValidityChecker {
+  private final AbstractConfigurable target;
+
+  public NotNullConfigureName(AbstractConfigurable target) {
+    this.target = target;
+  }
+
+  @Override
+  public void validate(Buildable b, ValidationReport report) {
+    if (b == this.target && target.getConfigureName().equals("")) {
```
uckelman, line 35: You don't need a `this` here.

vassal-app/src/main/java/VASSAL/configure/NotNullConfigureName.java
```diff
@@ -0,0 +1,39 @@
...
+import VASSAL.build.Buildable;
+import VASSAL.i18n.Resources;
+/**
+ * Requires that the object does not have null configurable name
+ */
+public class NotNullConfigureName implements ValidityChecker {
+  private final AbstractConfigurable target;
+
+  public NotNullConfigureName(AbstractConfigurable target) {
+    this.target = target;
```
uckelman, line 30: `target` should never be null, right? In that case I'd do this:
```
this.target = Objects.requireNonNull(target);
```

vassal-app/src/main/java/VASSAL/configure/NotNullConfigureName.java
uckelman: Would it make sense to test `target.getConfigureName() == null` instead, and then not modify `AbstractConfigurable.getConfigureName()` at all?

javiermunozk: If we do not modify AbstractConfigurable.getConfigureName() we will still have the Null Pointer Exception in already existing modules :(

vassal-app/src/main/java/VASSAL/configure/NotNullConfigureName.java
javiermunozk: ok, thanks!

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -1930,6 +1930,7 @@ Editor.UsernameAndPasswordDialog.choose_your_weapons=Set your username and passw
 Editor.UsernameAndPasswordDialog.turn_your_key_sir=Please confirm your password
 
 # Validation Report Dialog
+Editor.ValidationReportDialog.blank_name=An invalid blank name found at
```
BrentEaston, line 1933: I think this would read better as 

'Component must have a name at '.

To support translations in all languages, you need to include the object of the sentence as well. So this would be better as

'Component must have a name at %1$s'.

Then call is with

Resources.getString("Editor.ValidationReportDialog.blank_name", target.getClass().getName());

vassal-app/src/main/java/VASSAL/configure/NotNullConfigureName.java
```diff
@@ -0,0 +1,40 @@
...
+  private final AbstractConfigurable target;
+
+  public NotNullConfigureName(AbstractConfigurable target) {
+    this.target = Objects.requireNonNull(target);
+  }
+
+  @Override
+  public void validate(Buildable b, ValidationReport report) {
+    if (b == target && ((target.getConfigureName() == null) || (target.getConfigureName().equals("")))) {
+      report.addWarning(Resources.getString("Editor.ValidationReportDialog.blank_name") + " " + target.getClass().getName());
```
BrentEaston, line 37: Resources.getString("Editor.ValidationReportDialog.blank_name", target.getClass().getName());

is better (see below).

BrentEaston (changes requested):
Hi Javier,

Let's get this one and finished before we sort out the rest of the ValidityChecker stuff.

Brent.

Cattlesquat:
Replaced by #10636 which copied the functionality and made the minor text adjustments, and is now merged.

PR #846, opened 2021-03-30
Bug13203 - GameRefresher refreshes Double Deckers as well now. 

claudiociardelli:


vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -333,15 +393,40 @@ private void processGamePiece(GamePiece piece, Command command) {
...
+
+      // Place the new Piece.
+
+      // Here we use a workaround for placing pieces into decks. If 2 decks are defined
+      // with the same x,y position, the piece placement will not be able to determine
+      // the target deck. So 1st move the deck of the piece being replaced to -1, -1
+      // then put it back to its original position
+
+      final Stack stack = piece.getParent();
+      final Deck deck = (Deck) stack;
```
uckelman, line 405: Isn't this going to throw a `ClassCastException` if `stack` is not a `Deck`? You do an `instanceof` check just below this, so either that's possible or the check is redundant.

uckelman (changes requested):

uckelman (approved):

PR #847, opened 2021-03-31
14284 - one last extra bit I foofed 

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -419,7 +419,7 @@ else if (SHOW_DAYS.equals(key)) {
       return showDays;
     }
     else if (ALLOW_RESET.equals(key)) {
-      return ALLOW_RESET;
+      return ((Boolean)allowReset).toString();
```
uckelman, line 422: The right way here would be either `String.valueOf(allowReset)` or `Boolean.toString(allowReset)`.

Cattlesquat: done

PR #848, opened 2021-03-31
13145, 14225, 14288 - Deck Wreck - make Deck "behave properly" in setup(true)/setup(false) flow by letting DrawPile control its listener adding/removing

Cattlesquat:


vassal-app/src/test/java/VASSAL/build/MockModuleTest.java
```diff
@@ -24,8 +24,8 @@ public void init() throws Exception {
     final DataArchive arch = mock(DataArchive.class);
 
     when(module.getDataArchive()).thenReturn(arch);
-    doNothing().when(module).incorporateSourcesAndListeners();
-    doNothing().when(module).markModuleListeners();
+    //doNothing().when(module).incorporateSourcesAndListeners();
+    //doNothing().when(module).markModuleListeners();
```
uckelman, line 28: You can just delete these two lines. (And the import of `doNothing` a bit further up.)

Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -1111,6 +1116,31 @@ public void dumpNewListeners() {
...
+    }
+  }
+
+  public void restoreNewListeners() {
+    if (!isLoadingContinuationSemaphore()) {
+      Collections.copy(keyStrokeSourcesToAdd, keyStrokeSourcesToAddBackup);
+      Collections.copy(keyStrokeListenersToAdd, keyStrokeListenersToAddBackup);
+      Collections.copy(sideChangeListenersToAdd, sideChangeListenersToAddBackup);
+    }
+  }
```
jonathanrwatts: Why do backup and restore use the exact same code?

Cattlesquat: that code is no longer in the PR

Cattlesquat: (but good point otherwise haha)

PR #849, opened 2021-03-31
14290 - GlobalKeyCommands causing unnecessary Deck reshuffles and corruption

BrentEaston:


vassal-app/src/main/java/VASSAL/counters/Deck.java
```diff
@@ -882,17 +882,22 @@ public Command shuffle() {
...
+   * If this is an Always shuffle Deck, then shuffle the list of pieces, otherwise just
+   * reverse the list order so that we deal from the top.
+   *
+   * @return List of pieces in Dealable order
+   */
+  public List<GamePiece> getOrderedPieces() {
+    final List<GamePiece> pieces = asList();
     if (!ALWAYS.equals(shuffleOption)) {
-      return new NullCommand();
+      Collections.shuffle(pieces);
```
uckelman, line 895: Should that be `Collections.shuffle(pieces, gameModule.getRNG());`, to ensure we're using the right RNG?

BrentEaston: Indeed.

jonathanrwatts: Am I misreading this, or is this code only shuffling if the shuffleOption is NOT "Always"?  Isn't that backwards?

BrentEaston: The dangers of cut & paste. You are 100% correct. Good pick-up :)

PR #865, opened 2021-04-06
3.6 - 14346 - Deck reports suppressed during GKC report suppresssion

Cattlesquat:
First reported here: http://www.vassalengine.org/forum/viewtopic.php?f=4&t=12782

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Deck.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Deck.java
@@ -879,9 +879,9 @@ public class Deck extends Stack implements PlayerRoster.SideChangeListener {
     final List<GamePiece> l = Arrays.asList(a);
     DragBuffer.getBuffer().clear();
     Collections.shuffle(l, gameModule.getRNG());
-    final Command c = setContents(l);
+    Command c = setContents(l);
     if (Map.isChangeReportingEnabled()) {
-      c.append(reportCommand(shuffleMsgFormat, Resources.getString("Deck.shuffle"))); //$NON-NLS-1$
+      c = c.append(reportCommand(shuffleMsgFormat, Resources.getString("Deck.shuffle"))); //$NON-NLS-1$
     }
     return c;
   }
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/counters/Deck.java
```diff
@@ -1047,9 +1051,12 @@ public void setState(String state) {
 
   public Command setContentsFaceDown(boolean value) {
     final ChangeTracker t = new ChangeTracker(this);
-    final Command c = new NullCommand();
+    final   Command c = new NullCommand();
     faceDown = value;
-    return t.getChangeCommand().append(c).append(reportCommand(faceDownMsgFormat, value ? Resources.getString("Deck.face_down") : Resources.getString("Deck.face_up"))); //$NON-NLS-1$ //$NON-NLS-2$
+    if (Map.isChangeReportingEnabled()) {
+      c.append(reportCommand(faceDownMsgFormat, value ? Resources.getString("Deck.face_down") : Resources.getString("Deck.face_up")));
```
BrentEaston, line 1057: Try and follow the pattern

c = c.append(newCommand);

as this will cause an appened non-null command to replace an existing NullCommand. Removes spurious NullCommands ending up in the command stream.

It's not needed if you know c will contain a non-null command (e.g. shuffle where it is initialised with setContents(l), but should always be used when you have

if (blah) {
  c = c.append(...);
}

BrentEaston (changes requested):
Looks good to me.

Are we concerned that some modules may actually depend on Deck messages appearing inside GKC's? 

My feeling is the advantage we get out of improving consistency is worth more. If someone complains, we add another GlobalOption!

vassal-app/src/main/java/VASSAL/counters/Deck.java
Cattlesquat: Okay I think I've got it right now, if I'm understanding you correctly.

BrentEaston (approved):

PR #864, opened 2021-04-06
3.6 - 14344 - Deck supports configurable strings for Draw Multiple, Draw Specific, Face Up, Face Down

Cattlesquat:
Requested here - http://www.vassalengine.org/forum/viewtopic.php?f=4&t=12782

PR #868, opened 2021-04-06
14350 - Command allowed to SORT a Deck

Cattlesquat:
From the BIG DECK THREAD: http://www.vassalengine.org/forum/viewtopic.php?f=4&t=12782


Cattlesquat:
Also contains everything that was originally in #864 

Deck supports configurable strings for Draw Multiple, Draw Specific, Face Up, Face Down #864

vassal-app/src/main/java/VASSAL/build/module/map/DrawPile.java
```diff
@@ -82,6 +82,19 @@
...
+  private final VisibilityCondition drawMultipleMessageVisibleCondition = () -> dummy.isAllowMultipleDraw();
+
+  private final VisibilityCondition drawSpecificMessageVisibleCondition = () -> dummy.isAllowSelectDraw();
+
+  private final VisibilityCondition faceUpDownMessageVisibleCondition = () -> Deck.USE_MENU.equals(dummy.getFaceDownOption());
+
+  private final VisibilityCondition sortVisibleCondition = () -> dummy.isSortable();
+
+  private final VisibilityCondition selectOrSortVisibleCondition = () -> dummy.isSortable() || dummy.isAllowSelectDraw();
+
```
BrentEaston, line 94: Is using the same property name for Drawing and Sorting flexible enough if you want to do both? Sorting for drawing will usually by by some sensible 'user meaningful' order, but sorting for game play purposes will quite likely be completely different.

What about if the designer wants two kinds of sorts for different times during the game? Perhaps the Sort Command should be taken out of the DrawPile itself and added as a Child like Deck Global Key Commands are? That will also simplify the rapidly expanding DrawPile config screen. 

I would leave the existing sort property and the new descending? option for card draw display. Add a separate Sort property and descending flag to the new child Sort command.

What about multi-level sorts? Tribe by descending Strength? <evil grin> This can be handled by creating a Calculated Property with the compound sort key so probably not needed.

BrentEaston: In fact, the 'Command to send entire deck to another deck' should probably be a child as well as I can easily see the need for multiple of these. For this, maybe just leave the one that is in the Deck there, but add the facility to add extras as children. This would keep compatibility without a painful version conversion.

BrentEaston (changes requested):

BrentEaston:
@BrentEaston to update DeckSelector to make it a more general purpose component.

Cattlesquat:
@BrentEaston Okay I have made a separate "sort" parameter for the sort-command. Since that one can now be down with the other sort-command stuff, I don't think there's any "reordering the attributes" actually necessary -- the new Sort feature is at the bottom, Selection is back up where it always was.

Meanwhile as far as making multiples of command-to-send-to-another-deck, I think that seems like a separate PR from this sorting-related one, and probably calls into play your superior add-and-subtract-from-list-configurer skills as well...

BrentEaston:
I've got some ideas I'd like to try on this one.

BrentEaston:
Superceded by PR #10271 and PR #10289

PR #872, opened 2021-04-08
3.5.6(!!) - 14286 - prevent fastmatch expression 'bad module data' from throwing exceptions

Cattlesquat:


BrentEaston (commented):
Do we want to go further and ensure tryEvaluate NEVER returns a null, even if the evaluate call somehow returns one?

Cattlesquat:
> Do we want to go further and ensure tryEvaluate NEVER returns a null, even if the evaluate call somehow returns one?

Is... is that POSSIBLE?

PR #876, opened 2021-04-08
3.6 - 14360 - Window title less verbose on subwindows

Cattlesquat:
Per forum discussion here - I basically just implemented Mark's suggestion:
http://www.vassalengine.org/forum/viewtopic.php?p=67213#p67213


uckelman:
Look ok to me.

PR #881, opened 2021-04-10
3.6 - 14372 - OH BLESSED DAY! Duplicate a copy of component 'in place' in editor!

Cattlesquat:
image

So you don't always end up with it all the way at the bottom of your 500 at-start stacks or your 100 global properties or your 90 prototypes!

### DIFF
--- a/vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
+++ b/vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
@@ -386,8 +386,8 @@ Editor.ConfigureTree.add_cards=Add multiple cards
 Editor.ConfigureTree.add_pieces=Add multiple pieces
 # Adding a component (e.g. Add Map, Add Global Key Command, Add Zone, Add Single Piece)
 Editor.ConfigureTree.add_component=Add %1$s
-Editor.ConfigureTree.add_peer=Insert %1$s
-Editor.ConfigureTree.insert=Insert peer item
+Editor.ConfigureTree.add_peer=%1$s
+Editor.ConfigureTree.insert=Insert
 Editor.ConfigureTree.add_duplicate=Duplicate this %1$s
 Editor.ConfigureTree.clone=Clone
 Editor.ConfigureTree.java_name=Enter fully-qualified name of Java class to import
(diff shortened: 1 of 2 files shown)


BrentEaston (changes requested):
The Duplicate option is good.

I think the 'Insert another thing below this thing' option is too limited. I think it should just be be 'Insert' (or Insert Component') which opens the same sub-menu that the parent has so that you can insert any peer component after this one, not just another thing of the same type.

For example, inserting another Global Options component after the existing Global Options component is meaningless, and invalid. What you probably want to insert is a Dice Button or a ToolBar menu etc.

Cattlesquat:
@BrentEaston Good idea - Try new version! I'd thought of listing all the peer items but hadn't thought of the submenu method, so it seemed like it would be really crowded.

Once the drag-and-drop is merged I will also sanity-check the final "index" of insertion using the method I added there (for extensions). I doubt inserting something "up among the grey items" in an extension would actually cause any harm (they would just naturally drop to the bottom the next time a save/load cycle happened), but it's still aesthetically questionable.

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -386,6 +386,9 @@ Editor.ConfigureTree.add_cards=Add multiple cards
 Editor.ConfigureTree.add_pieces=Add multiple pieces
 # Adding a component (e.g. Add Map, Add Global Key Command, Add Zone, Add Single Piece)
 Editor.ConfigureTree.add_component=Add %1$s
+Editor.ConfigureTree.add_peer=Insert %1$s
+Editor.ConfigureTree.insert=Insert peer item
```
BrentEaston, line 390: I don't want no damn components peeing in my module! Can we make this either just 'Insert' or 'Insert component'? Peer is too technical.

And I don't think we need 'Insert ' in front of every sub-menu entry. That's implied by selecting Insert (component) from the menu.

BrentEaston (changes requested):
Works well, just the wording changes.

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
Cattlesquat: Done! (But who doesn't like to frolic with their peers? :D )

BrentEaston (approved):

PR #882, opened 2021-04-11
3.6 - 14373 - (and 1944!) - HELL YES! Drag-and-drop for Editor items.

Cattlesquat:
1. Grab your favorite module components and just DRAG their sorry asses to a new location!
2. OR, Ctrl+Drag ("That's Cmd+Drag on Mac") to make a FREE copy!
3. PROFIT!!!

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
+++ b/vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
@@ -2230,10 +2230,12 @@ public class ConfigureTree extends JTree implements PropertyChangeListener, Mous
   }
 
   /**
-   * Tree Transfer Handler provides drag-and-drop support for editor items. Heavily adapted from Oracle "tutorial" and various StackOverflow and Code Ranch articles, but yeah, google ftw.
-   * (Brian Reynolds Apr 4 2021)
+   * Tree Transfer Handler provides drag-and-drop support for editor items. Heavily adapted from Oracle "tutorial" and
+   * various StackOverflow and Code Ranch articles, but yeah, google ftw. (Brian Reynolds Apr 4 2021)
    */
   class TreeTransferHandler extends TransferHandler {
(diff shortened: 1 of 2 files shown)


BrentEaston (commented):
This is excellent and works perfectly when editing a module.

However, there are some issues when editing an Extension that may be difficult to overcome. 

When you create a new component in the extension, then add sub-components, the DnD works as expected. eg. create a new Map and you can DnD components within the new Map.

However, if you add components to an existing component (e.g. Add new Global Properties to the existing Global Properties container, or add new top-level components like maps, then you can't use drag and drop to move those around, they never have a valid drop target. Neither can you DnD any component into a Component not owned by the Extension, even if you can create it there manually.

For example
image
I have added 3 new Global Variables aaa, bbb & ccc to the Extension. I can't drag and drop aaa to below bbb. Neither can I DnD a Global Variable from an Extension Map into the Module Global Properties.

Cattlesquat:
@BrentEaston Wasn't that hard to fix -- it turned out ExtensionEditor had (unbeknownst to me) overridden the isValidPasteTarget() method and I needed to adjust the overridden version. That then raised the issue that I probably shouldn't let people drag extension items "in among the greyed-out items up at the top of any given component list", so I added a part to make sure nothing ever ends up "earlier in the list than the lowest module-owned item".

BrentEaston (changes requested):
Couple of problems:

1. Minor issue, You stopped dragging extension items above module items, but you can still pick up and drag module items down among the extension items.
2. Major issue, you can pick up and drag greyed out Module components (e.g. dice button) and when you drop it, it just vanishes.

If editing an extension, you should not be able to drag any greyed  out module components at all. That would fix both problems.

Cattlesquat:
@BrentEaston Ah, good point! Have now prohibited any module-owned component from being the "source" of a drag operation during an extension edit. Seems to work right.

PR #885, opened 2021-04-12
3.6 - 14377 - Map Windows - non-toggle hotkeys to show and hide

Cattlesquat:
Requested in forums: http://www.vassalengine.org/forum/viewtopic.php?f=4&t=12809


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -3079,11 +3128,13 @@ public HelpFile getHelpFile() {
       Resources.getString(Resources.TOOLTIP_TEXT),
       Resources.getString(Resources.BUTTON_ICON),
       Resources.getString(Resources.HOTKEY_LABEL),
+      Resources.getString("Editor.Map.show_key"),
+      Resources.getString("Editor.Map.hide_key"),
```
BrentEaston, line 3132: A few interesting things, most which are not new, but might be nice to nail now:

1. For clarity, I think in this case it might be better to change the 'Hotkey' label to 'Hotkey to show/hide':
image

2. I set up this window as above, then deselected the 'Include toolbar to show/hide?' option, saved and reloaded the module. No button and the F5/F6/F7 definitions are no longer visible in the editor, but the F5/F6/F7 to toggle/show/hide still work! [F5 doesn't work in the Editor, but it does in the Player]. I think it is actually better than them not working. I'm thinking that the 3 Hotkey fields should be always visible, since they are working anyway.

3. While you're in there, how about fixing that thing where changing the 'Include toolbar to show/hide?' button doesn't actually change the visibility until the next time you reload the module. That is just painful, let's just make it a standard visibility change.

BrentEaston (changes requested):

vassal-app/src/main/java/VASSAL/build/module/Map.java
Cattlesquat: Okay - I think latest set of commits (aka "BREAK that app! BREAK that app!") should do it. There was a deep rabbit hole of special cases but I think I have it largely sorted (and the "docked" state of a "first map window" in that category of things will still not change until restart -- I have updated messaging appropriately). A good thing is I think in the process I managed to clean up a few minor "bad behaviors". See what you think.

BrentEaston (commented):
I'm getting a weird interaction with the chat box/divider position. Do the following:
1. Open a module for editing:
image
2. Edit the 1st/main/docked Map.
3. Cancel the Edit. You now get this: 
image
And have lost all divider and window position settings, Have to pull window and divider around to get back to the old settings.

Doesn't happen for any other map, or if you click OK to finish the map edit instead of Cancel.

Cattlesquat:
@BrentEaston Newest version fixes that.

PR #887, opened 2021-04-12
3.6 - Default Font Size change

Cattlesquat:
Allows the default font for VASSAL and the Editor to be a different size, via General Preferences tab.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/launch/Launcher.java
+++ b/vassal-app/src/main/java/VASSAL/launch/Launcher.java
@@ -18,11 +18,13 @@
 
 package VASSAL.launch;
 
+import java.awt.Font;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.charset.Charset;
 
 import javax.swing.SwingUtilities;
+import javax.swing.UIManager;
 
 import org.apache.commons.lang3.SystemUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


Cattlesquat:
image

BrentEaston (commented):
That works well to change the font size of the default UI fonts, however, at the risk of turning this from a nice little change into a monster, it looks a bit strange because it doesn't change the font size for many things:
 
- The title bar font size
- Any font sizes set directly by preferences  - Chat font size and Turn Counter Font size
- What do we do about font sizes specified in Components (grid number font, region font, mouse over stack viewer font etc. These should probably be affected
- What do we do about font sizes specified in Label traits? These should probably not be affected.

To do this properly, I think we need to make the Font preference a %  with 100 = default, ranging from , say 50% to 200% (slider perhaps?) We can then apply to all fonts across the board as needed, a bit like the zoom factor is applied in all drawables.

Thoughts?

BrentEaston:
Actually, probably just the Chat Window and the Title Bar would be enough to stop it looking strange. Mos of the rest can already be enlarged at will via map zoom.

Cattlesquat:
Two main things:
(1) Title Bar is controlled by the OS, and we have no way to change our title bar font size. 
(2) Having this change the things that have their own preferences (e.g. Chat Window) I disagree with. It is better to give the players direct control of font size (e.g. 18 point font) than make them mess with a % scale thing. Although going to 24 point "for most things" without changing chat might look unusual when you first boot in, it's easy to then find the preference and once you do it's better that you have direct control (after all the whole reason for having Chat have its own font size in the first place is that presumably some players want that a different size from everything else).

BrentEaston (approved):

PR #890, opened 2021-04-13
3.5.6 - 14386 - Don't show bogus 'module saved' confirmations if save was cancelled by user at a dialog.

Cattlesquat:


uckelman:
For thoroughness, I checked VASL to see if it uses ArchiveWriter directly. It doesn't, so it's very unlikely that anything else does.

uckelman (approved):

vassal-app/src/main/java/VASSAL/tools/ArchiveWriter.java
```diff
@@ -267,26 +267,31 @@ public void addFile(String fileName, byte[] content) {
...
-    if (isTempArchive) saveAs(notifyModuleManager);
-    else write(archive, notifyModuleManager);
+  public boolean save(boolean notifyModuleManager) throws IOException {
+    if (isTempArchive) {
+      return saveAs(notifyModuleManager);
+    }
+    else {
+      write(archive, notifyModuleManager);
+    }
+    return false;
```
uckelman, line 286: Shouldn't we return `true` after `write()`?

uckelman: Is this the cause of the problem noted here? http://www.vassalengine.org/forum/viewtopic.php?p=67818#p67818

Cattlesquat: Yep, had to make a new PR for it, so see #891

PR #893, opened 2021-04-14
Don't start the Module Manager when loading a module from a file association

uckelman:


uckelman:
There's no easy way to get module paths when saves are clicked.

PR #895, opened 2021-04-15
14393 - Don't try to report bugs for test versions

uckelman:
Separate the case of an outdated version from that of a test version for crash reporting.


BrentEaston:
Just tried this using the 3.6.0 snapshot. Forced an NPE in CalculatedProperty. I get the following dialog telling me details are being collected and there is a spinning cursor. Nothing ever happens. Eventually realised I can just click the OK button to continue.

image

Cattlesquat:
I get the infinite snake whenever an NPE happens while the IDE thinks the version is "{git version}" or whatever that thing is that happens sometimes. 

The bad part about it is you can't ever get to see the stack trace, so you can't find out what happens until you've woken it back up. 

That usually stops happening for me with a "regular snapshot" build once the version number gets back to being a "proper snapshot number".

PR #896, opened 2021-04-15
3.5.6 - 14271 - notes window is a keystrokesource so that it won't "eat" keys

Cattlesquat:
…trokes when in focus

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/NotesWindow.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/NotesWindow.java
@@ -275,7 +275,7 @@ public class NotesWindow extends AbstractToolbarItem
   public void addTo(Buildable b) {
     super.addTo(b);
     getLaunchButton().setAlignmentY(0.0F);
-    GameModule gm = GameModule.getGameModule();
+    final GameModule gm = GameModule.getGameModule();
     gm.addCommandEncoder(this);
     gm.getGameState().addGameComponent(this);
     gm.addCommandEncoder(privateNotes);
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/NotesWindow.java
```diff
@@ -273,23 +275,26 @@ public static String getConfigureTypeName() {
...
-    GameModule.getGameModule().addCommandEncoder(secretNotes);
-    GameModule.getGameModule().getGameState().addGameComponent(secretNotes);
+    final GameModule gm = GameModule.getGameModule();
+    gm.addCommandEncoder(this);
+    gm.getGameState().addGameComponent(this);
+    gm.addCommandEncoder(privateNotes);
+    gm.getGameState().addGameComponent(privateNotes);
+    gm.addCommandEncoder(secretNotes);
+    gm.getGameState().addGameComponent(secretNotes);
+    gm.addKeyStrokeSource(new KeyStrokeSource(frame.getRootPane(), JComponent.WHEN_IN_FOCUSED_WINDOW));
```
uckelman, line 285: Is this introducing a leak, as you don't call `removeKeyStrokeSource()` in `removeFrom()` below?

Cattlesquat: There is no `removeKeyStrokeSource()` to call in GameModule. I think you're thinking of the one you made for the listeners so that they would de-register?

uckelman: I thought I had made one of those. So, is this going to be a problem?

PR #897, opened 2021-04-15
14388-Add Scrollbars to Prefs Dialog when needed.

BrentEaston:
Align checkboxes in center of line
Add missing ? in Boolean options

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/BooleanConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/configure/BooleanConfigurer.java
@@ -20,6 +20,8 @@ package VASSAL.configure;
 import javax.swing.JCheckBox;
 import javax.swing.JPanel;
 
+import net.miginfocom.swing.MigLayout;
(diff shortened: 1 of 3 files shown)


Cattlesquat:
Note that most apps (examples at hand: IntelliJ, Firefox, Windows, Adobe Premiere) don't put ? after the names of options that represent binary checkbox choices. Checking the box means the thing in the description is "on" or "happens"; unchecking it means it does not.

BrentEaston:
Created issue 14392 for someone to do a run through the Configurer/Preference texts. @Cattlesquat could you give this a run and check I've not done anything stupid? Bumping the default font size right will cause it to kick in.

Cattlesquat (commented):
Worked right when I ran it -- when I ran at the max 32 it still didn't "need it" on my monitor, but if I dragged my prefs window smaller than the space needed to hold the whole horizontal or vertical dimension the scrollbars appeared and worked.

PR #901, opened 2021-04-16
3.6 - 14372 - This finishes up Duplicate/Insert for Extensions

Cattlesquat:


BrentEaston:
Something weird going on.

When editing a module (e.g. Pog), I can't move any of the top few components in any folder, they have no valid drop targets. For example, can't move [Help Menu] in the module level and can't move [Stacking Options] in a map.

Extension editing seems to be normal.

Cattlesquat:
Oh I see what's happening, it isn't let any of the [things] that aren't on the "getAllowableConfigurableComponents()" list for their parents to be moved.

Cattlesquat:
Okay, the new commit clears this up, allowing "non-addables" to be dragged around within their parents too.

HOWEVER -- when you try this out, be aware that if you drag anything *to* a position that's *above* the first Map in the module (the dockable map), the Player window will disappear. This is actually just a manifestation of a pre-existing bug in the Editor (you can verify the pre-existingness of it by using the "traditional" move function to move any component "above" the first map). The reason it happens is because the `insert()` method does some ZANY things -- everything that is to be displaced lower down the list by the inserted item is first REMOVED from the module and then reinserted after that. And so therefore if the first Map is displaced, the first Map is REMOVED which then makes the Player window decide to disappear (who can blame it, really). The insert method "looks really scary" and so I don't want to fool with that pre-existing bug.

BrentEaston:
Ok, That's wacko. Confirmed that bug exists going back to at least 3.2, so nothing we have done recently.

I agree, let's escape with our lives while we can and come and fight that one another day.

PR #902, opened 2021-04-16
3.6 - FOLDERS!!!! YEEEEEEEE-HAW!!!!

Cattlesquat:
image

For the first version I did folders SetupStack, but it should be super-easy to teach many other things to play well in folders.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/AbstractBuildable.java
+++ b/vassal-app/src/main/java/VASSAL/build/AbstractBuildable.java
@@ -146,23 +146,6 @@ public abstract class AbstractBuildable extends AbstractImageFinder implements B
   }
 
 
-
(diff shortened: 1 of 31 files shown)


Cattlesquat:
image

Cattlesquat:
Now supports folders for At-Start Stacks, Global Properties, and Prototypes. 

In theory I think I could make a folder designed to be the child of a Map and which accepted any of a variety of Map children (however many types of Map children we'd taught how to behave in a folder -- so e.g. At-Start Stack, and then could logically teach Draw Pile, CounterDetailViewer, Global Key Commands, etc). So far I haven't run into anything that seemed very hard to teach how to be in a folder, but I'm still "waiting for the shoe to drop"...

PR #903, opened 2021-04-17
Small fixes to docs

abelsromero:
Just checked the configuration and found some things to improve.
Feel free to ignore this PR this don't make sense to your current approach.

* Fix html file extension override
`filetype` attribute does not apply to file conversion, you need to use `outfilesuffix`.
If you don't want to change the file extension, just remove the attribute.

* Simplify `optimize` attr
This attribute does not need any value. Also if it's causinf trouble due to the issue with asciidoctor-pdf, you can just run the command with `-Dasciidoctor.attributes=!optimize` to unset it from the command line. You can also add this to the properties section in settings.xml to avoid typing it any time.

Also saw, you are composing the user guide. You may be interested to use `:leveloffset: +1` https://docs.asciidoctor.org/asciidoc/latest/directives/include-with-leveloffset/#manipulate-heading-levels-with-leveloffset. That way, the composed parts don't need to add +1 to the section headers. It makes writting simpler and useful if you want to reuse the pages for other convertions, like individual HTML.

### DIFF
--- a/vassal-doc/pom.xml
+++ b/vassal-doc/pom.xml
@@ -40,7 +40,7 @@
                             <backend>html</backend>
                             <preserveDirectories>true</preserveDirectories>
                             <attributes>
-                                <filetype>htm</filetype>
+                                <outfilesuffix>.htm</outfilesuffix>
                             </attributes>
                         </configuration>
                     </execution>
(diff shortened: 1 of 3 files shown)


uckelman:
We actually want `.html` as the extension, and hadn't realized that `filetype` attribute was being set---so the correct thing will be for us simply to delete that line. Thanks for pointing it out, though!

Cattlesquat:
@abelsromero Thx for the note on :leveloffset: -- I have made a PR to incorporate that into the designer guide.

PR #911, opened 2021-04-20
3.6 - Mats & Cargo - "Pieces that loosely prefer to move together"

Cattlesquat:
(1) You give some "big" pieces a "Mat" trait (and a does-not-stack trait)
(2) You give some "little" pieces a "Mat Cargo" trait
(3) In the game you drag some little pieces (cargo) onto one of the big pieces (mats)
(4) Now when you click/drag to select the Mat, any Cargo that's on it will automatically be selected -- so dragging the "Mat" will naturally carry all its cargo with it, unless you take the cargo off.

There is other stuff, but that's the basics. Probably more to come as I think of it.


vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
```diff
@@ -806,6 +809,58 @@ public Command movePieces(Map map, Point p) {
       // Any piece we successfully moved, make sure is now considered a "selected piece" (i.e. a member of KeyBuffer)
       for (final GamePiece piece : draggedPieces) {
         KeyBuffer.getBuffer().add(piece);
+
```
BrentEaston, line 812: We will probably need some selection criteria here. Not all Cargo may be allows on all mats. The Mat should have an optional PropertyExpression to limit the cargo that is allowed on it. Only my pieces on my mat for example.

vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
```diff
@@ -806,6 +809,58 @@ public Command movePieces(Map map, Point p) {
...
+                  comm = comm.append(mat.makeAddCargoCommand(piece));
+                }
+              }
+              else {
+                // We're NOT on a mat, so if we WERE on one, mark ourselves as no longer on it
+                comm = comm.append(cargo.makeClearMatCommand());
+              }
+            }
+          }
+          else {
```
BrentEaston, line 833: I tried various options of unselecting cargo and moving the mat out from under it, then back, etc. all worked as I expected. The basic functionality looks great.

vassal-app/src/main/java/VASSAL/counters/Mat.java
```diff
@@ -0,0 +1,308 @@
...
+    return new Ed(this);
+  }
+
+  @Override
+  public String getDescription() {
+    return buildDescription("Editor.Mat.trait_description", desc);
+  }
+
+  @Override
+  public Object getProperty(Object key) {
```
BrentEaston, line 213: Will need a property to expose number of cargo pieces this mat holds. 

Probably also need the equivalent of Deck  'Count Expressions' and the properties to expose them, keyed off the mat name.

BrentEaston (commented):
Looks good so far. Some food for thought in the comments.

Cattlesquat:
@BrentEaston Okay I think the basic Mat functionality is now ready to go.

BrentEaston (approved):

PR #913, opened 2021-04-22
1949-Add 'Stack Normally' option Does Not Stack trait

BrentEaston:


Cattlesquat (changes requested):
Looks right, but see also line ~359 of GameRefresher.java, which checks the existence of the Immobilized class rather than the property. Fortunately I think that's the only instance of that style.

BrentEaston:
@Cattlesquat have a look now. I have also updated tests and reference manual.

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -356,7 +356,7 @@ private void processGamePiece(GamePiece piece, Command command) {
     final Stack oldStack = piece.getParent();
     if (oldStack == null) {
       final Immobilized i = (Immobilized)Decorator.getDecorator(piece, Immobilized.class);
-      if (i == null) {
+      if (i == null || !i.isCanStack()) {
```
Cattlesquat, line 359: Shouldn't the whole search for Immobilized be removed, and instead just check `Boolean.TRUE.equals(piece.getProperty(NO_STACK))`?

Everywhere else we check for stacking we do so via the property. Which will help when I introduce Mat returning a true to "NO_STACK".

vassal-app/src/main/java/VASSAL/counters/Immobilized.java
```diff
@@ -287,6 +287,10 @@ else if (Properties.BAND_SELECT_EVENT_FILTER.equals(key)) {
     }
   }
 
+  public boolean isCanStack() {
```
Cattlesquat, line 290: Similarly I'm not sure we should introduce this here as a public, because it will lead to checking the stack state of the piece "incompletely". A private method for "is THIS trait preventing stacking" is fine, but for outside-this-trait purposes we should be checking the property, right?

BrentEaston:
@Cattlesquat Good point. Modified as suggested.

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -352,11 +351,10 @@ private void processGamePiece(GamePiece piece, Command command) {
     }
 
     // Piece should have a parent stack (Decks are extensions of Stacks)
-    // Except pieces with a DNS (Immobilized) trait.
+    // Except pieces that return TRUE for the NO_STACK property.
     final Stack oldStack = piece.getParent();
     if (oldStack == null) {
-      final Immobilized i = (Immobilized)Decorator.getDecorator(piece, Immobilized.class);
-      if (i == null || !i.isCanStack()) {
+      if (!Boolean.TRUE.equals(Decorator.getOutermost(piece).getProperty(Properties.NO_STACK))) {
```
Cattlesquat, line 357: It looks like you might have a "double negative" here that you don't want? It seems to check for the *lack* of NO_STACK but then increase the noStackCount? I will admit I have lost the thread about what this little block was trying to do, but probably check back what it originally did and make sure we have the "plus" and "minus" in the right direction :)

BrentEaston: I believe this is correct. This is not supposed to count the non-stacking units, it is counting the number of stackable units that are not actually in a stack. So if a unit has no parent, but NO_STACK is true, then we don't want to count it.

The original code was 
    final Stack oldStack = piece.getParent();
    if (oldStack == null) {
      final Immobilized i = (Immobilized)Decorator.getDecorator(piece, Immobilized.class);
      if (i == null) {
        noStackCount++;
        log(Resources.getString("GameRefresher.refresh_error_nostack", piece.getName(), piece.getId()));
      }
    }

Cattlesquat: Okay yes, makes sense!

PR #915, opened 2021-04-23
MassPieceUploader adoc clarification

claudiociardelli:
14409 MassPieceUploader - mention in help that filenames cannot contain dot + suffix separator cannot be dot

### DIFF
--- a/vassal-doc/src/main/readme-referencemanual/ReferenceManual/MassPieceLoader.adoc
+++ b/vassal-doc/src/main/readme-referencemanual/ReferenceManual/MassPieceLoader.adoc
@@ -72,7 +72,7 @@ The options you have to choose from are:
 * _Same as:_
 * _Use Base image:_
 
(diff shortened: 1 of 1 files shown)


vassal-doc/src/main/readme-referencemanual/ReferenceManual/MassPieceLoader.adoc
```diff
@@ -72,6 +72,10 @@ The options you have to choose from are:
 * _Same as:_
 * _Use Base image:_
 
+==== Version 3.5.5 limitations
```
BrentEaston, line 75: @claudiociardelli This limitation isn't version specific is it? This has always and will always be the case?

claudiociardelli: Removed ref to version.

BrentEaston (approved):

PR #920, opened 2021-04-29
14418 Mass Piece Loader paste correct type of Layer trait

BrentEaston:


vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -1070,7 +1074,7 @@ public Component getListCellRendererComponent(
    * @author rkinney
    *
    */
-  private static class TraitClipboard {
+  protected static class TraitClipboard {
```
uckelman, line 1077: Is this actually necessary?

BrentEaston: Yes, because MPL's subclass of PieceDefiner now accesses this class to check the type of the copied piece. It would not compile unless I changed this.

uckelman: Oh, I see what's going on now. `clipBoard` is a `protected` field in the base class, but its _type_ had been `private` so you couldn't see the class definition in order to call any methods on it.

uckelman (approved):

PR #941, opened 2021-05-12
14477 - MouseoverStackViewer show a single card when used in Player Hand Window

Cattlesquat:
See: http://www.vassalengine.org/forum/viewtopic.php?p=68180#p68180

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
@@ -773,13 +773,10 @@ public class CounterDetailViewer extends AbstractConfigurable implements Drawabl
           s.asList().forEach(this::apply);
         }
         else {
-          for (final GamePiece p : s.asList()) {
-            final int index = s.indexOf(p);
-            if (index >= 0 && index < shapes.length) {
-              if (shapes[index].contains(pt)) {
-                apply(p);
-                break;
-              }
+          for (int i = 0; i < shapes.length; ++i) {
+            if (shapes[i].contains(pt)) {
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -768,7 +769,20 @@ public Object visitStack(Stack s) {
...
         super.visitStack(s) != null : foundPieceAt.equals(s.getPosition());
       if (addContents) {
-        s.asList().forEach(this::apply);
+        if (!(map instanceof PlayerHand)) {
+          s.asList().forEach(this::apply);
+        }
+        else {
+          for (final GamePiece p : s.asList()) {
+            final int index = s.indexOf(p);
+            if (index >= 0 && index < shapes.length) {
```
uckelman, line 778: How can `index` ever be negative or off the end here? `p` has to be in the list, since you got it from iterating over the list.

Cattlesquat: I copied this out of PieceFinder.MOVABLE which had some scary comments about "blah blah blah THREADS, fear fear fear" and then did it that way.

uckelman: Hmm. Alright, I'll go look at that.

uckelman: I think you could do this:
```
for (int i = 0; i < shapes.length; ++i) {
  if (shapes[i].contains(pt)) {
    apply(s.getPieceAt(i));
    break;
  }
}
```

PR #954, opened 2021-05-19
3.5.6 - Don't remove tile cache on module removal if another existing module is using it

Cattlesquat:


uckelman:
Closing in favor of PR 955.

PR #961, opened 2021-05-30
14526 - Don't try rendering empty labels

uckelman:


Cattlesquat (approved):
Looks like a good idea to me.

PR #970, opened 2021-06-08
14570 Let wizard divider float further right

BrentEaston:


vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
```diff
@@ -321,7 +321,7 @@ else if (tutorial != null) {
         box.add(Box.createVerticalGlue());
         final BooleanConfigurer wizardConf = (BooleanConfigurer)
           Prefs.getGlobalPrefs().getOption(WELCOME_WIZARD_KEY);
-        final JCheckBox show = new JCheckBox(wizardConf.getName() + " " + Resources.getString("WizardSupport.Bypass"));
+        final JCheckBox show = new JCheckBox("<html>" + wizardConf.getName() + "<br/>" + Resources.getString("WizardSupport.Bypass"));
```
uckelman, line 324: You don't need the close tag?

BrentEaston: Nope, not in a swing element

PR #980, opened 2021-06-19
14586 - Editor Duplicate Key

Cattlesquat:
Can now "duplicate" an item in the editor by hitting Ctrl+D (similar to keys for cut, paste, etc)


riverwanderer:
i.e. Cmd+D on MacOS, or is that a given?

Cattlesquat:
Yes it will automatically be Cmd+D on MacOS.

BrentEaston (approved):

Issue #4533, opened 2021-06-28
Make Grid Center dot display independent of grid display


uckelman:
| Bugzilla |  |
|----------|----------|
| ID | BZ1940 |
| Reported | 2009-06-03 06:56:00 +0200 |
| Modified | 2020-06-22 00:12:33 +0200 |
| Product | VASSAL |
| Component | Unknown |
| Version | unspecified |
| Platform | All |
| OS | All |
| Status | NEW |
| Resolution | None |
| Priority | medium |
| Severity | enhancement |
| Milestone | --- |
| Reporter | [Joel Uckelman](mailto:uckelman@nomic.net) |
| Assignee | [Bugs](mailto:bugs@vassalengine.org) |

#### [Joel Uckelman](mailto:uckelman@nomic.net) 2010-08-25 14:35:37 +0200
<pre>---- Reported by taruss 2009-06-03 18:56 ----

Change the behavior of the grid visibility so that the center dot of a grid can be displayed independently of the grid lines.

That would mean allowing the "Show center dot" checkbox enable the choice of color and the display of just the center dot.

Use case:  This would support using grids that consist just of points, without the border lines of the cells.  The "Crayon Rail" series of games (Empire Builder, Eurorails, etc.) use this type of grid.

--- Bug imported by uckelman@nomic.net 2010-08-25 14:35 MST  ---

This bug was previously known as _bug_ 2800672 at http://www.vassalengine.org/trackershow_bug.cgi?id=2800672

Unknown Component 
   Using default product and component set in Parameters 
The original reporter of this bug does not have
   an account here. Reassigning to the person who moved
   it here: uckelman@nomic.net.
   Previous reporter was taruss.
The original assignee of this bug does not have
   an account here. Reassigning to the default assignee
   for the component, uckelman@nomic.net.
   Previous assignee is unknown.
Missing Resolution. Setting status to NEW
Unknown resolution "".
   Setting resolution to MOVED</pre>

uckelman:
That's a new features. Should be for 3.6.0?

PR #10210, opened 2021-07-02
#4533 (BZ1940) Grid center dot display independent of grid display

BrentEaston:


uckelman (commented):
I'm not sure if this matters, but after this change, hex grids draw the dots followed by the grid, while rectangular grids draw the grid followed by the dots.

BrentEaston:
It's not really significant I think. Square grid draws all grid lines, then all the dots. The Hexgrid draws each cell individually, the dot then the lines. In either case, dots and lines can never overlap.

BrentEaston:
On 3.6.0, I personally consider it a bug. The configurer has always allowed you to request center dots only to be displayed, but it has never worked.

PR #10245, opened 2021-07-23
Don't force Game Piece Image Definitions

Cattlesquat:
I'm finding the force-re-add of Game Piece Image Definitions to be quite cluttery and annoying. 

We only added this for 3.6 so no ticket number. 

In theory we could allow the manual re-adding of these, although with them being so incredibly obsolete these days it doesn't particularly seem worth it?


BrentEaston (approved):
Fine by me

PR #10257, opened 2021-07-27
Mat Fixes

Cattlesquat:
(1) PlaceMarker traits need to attach cargo to mats where applicable
(2) Apparently checking getProperty(Mat.MAT_NAME) and comparing it to "" wasn't sufficient, because non-Mats were returning null instead of ""

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
@@ -238,11 +238,15 @@ public class PlaceMarker extends Decorator implements TranslatablePiece, Recursi
 
     // Mat support
     if ((c != null) && GameModule.getGameModule().isMatSupport()) {
-      // If a cargo piece has been placed, find it a Mat if eligible
+      // If a cargo piece has been placed, find it a Mat if eligible, and select it if the Mat is selected
       if ("true".equals(marker.getProperty(MatCargo.IS_CARGO))) { //NON-NLS
         final MatCargo cargo = (MatCargo) Decorator.getDecorator(marker, MatCargo.class);
         if (cargo != null) {
           c = c.append(cargo.findNewMat());
+          final GamePiece mat = cargo.getMat();
+          if ((mat != null) && (mat.getProperty(Properties.SELECTED) == Boolean.TRUE)) {
+            KeyBuffer.getBuffer().add(marker);
+          }
         }
       }
     }
--- a/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
@@ -239,7 +239,7 @@ public class PlaceMarker extends Decorator implements TranslatablePiece, Recursi
     // Mat support
     if ((c != null) && GameModule.getGameModule().isMatSupport()) {
       // If a cargo piece has been placed, find it a Mat if eligible, and select it if the Mat is selected
-      if ("true".equals(marker.getProperty(MatCargo.IS_CARGO))) { //NON-NLS
+      if (Boolean.TRUE.equals(marker.getProperty(MatCargo.IS_CARGO))) { //NON-NLS
         final MatCargo cargo = (MatCargo) Decorator.getDecorator(marker, MatCargo.class);
         if (cargo != null) {
           c = c.append(cargo.findNewMat());
--- a/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
@@ -22,7 +22,6 @@ import VASSAL.build.GameModule;
 import VASSAL.build.GpIdSupport;
 import VASSAL.build.module.BasicCommandEncoder;
 import VASSAL.build.module.Chatter;
-import VASSAL.build.module.GlobalOptions;
 import VASSAL.build.module.Map;
 import VASSAL.build.module.documentation.HelpFile;
 import VASSAL.build.widget.CardSlot;
--- a/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
@@ -22,6 +22,7 @@ import VASSAL.build.GameModule;
 import VASSAL.build.GpIdSupport;
 import VASSAL.build.module.BasicCommandEncoder;
 import VASSAL.build.module.Chatter;
+import VASSAL.build.module.GlobalOptions;
 import VASSAL.build.module.Map;
 import VASSAL.build.module.documentation.HelpFile;
 import VASSAL.build.widget.CardSlot;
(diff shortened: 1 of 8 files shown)


vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
```diff
@@ -236,6 +236,21 @@ protected Command placeMarker() {
       c = m.placeAt(marker, p);
     }
 
+    // Mat support
+    if ((c != null) && GameModule.getGameModule().isMatSupport()) {
+      // If a cargo piece has been placed, find it a Mat if eligible, and select it if the Mat is selected
+      if ("true".equals(marker.getProperty(MatCargo.IS_CARGO))) { //NON-NLS
```
uckelman, line 242: Why's this a `String` instead of a `Boolean`?

vassal-app/src/main/java/VASSAL/counters/Translate.java
```diff
@@ -193,17 +193,20 @@ protected Command newTranslate(KeyStroke stroke) {
...
+        if (mat != null) {
+          //BR// Should we ONLY take Cargo that are currently selected (in the KeyBuffer)?
+          contents = new ArrayList<>(mat.getContents());
+          offsets = new ArrayList<>();
+          for (final GamePiece piece : contents) {
+            final Point pt = piece.getPosition();
+            pt.x -= p.x;
+            pt.y -= p.y;
+            offsets.add(pt);
+          }
```
uckelman, line 209: Is there someplace this could go so that it's not repeated?

vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
Cattlesquat: Possibly I just don't know everything about how getProperty things work, I'm used to the inside-VASSAL end of it where they're all strings. So you're saying this would return me a `Boolean`? Or are you saying that it **should have been written to** return me a `Boolean`?

vassal-app/src/main/java/VASSAL/counters/Translate.java
Cattlesquat: If I were to pass those ArrayLists as parameters into a method in e.g. Mat.java, and it filled them up, then upon return would the lists have been updated? So I could call e.g. Mat.makeOffsets(x, y, contents, offsets) and it would fill up "contents" and "offsets" for my later use?

vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
uckelman: I'm not saying any of those things in particular. This is me wondering why that returns a string. :)

vassal-app/src/main/java/VASSAL/counters/Translate.java
uckelman: Yes, because everything that's not a primitive in Java is a pointer.

vassal-app/src/main/java/VASSAL/counters/SendToLocation.java
```diff
@@ -382,18 +382,14 @@ public Command myKeyEvent(KeyStroke stroke) {
...
-          pt.y -= basePt.y;
-          offsets.add(pt);
+    if (GameModule.getGameModule().isMatSupport()) {
+      final String matName = (String)outer.getProperty(Mat.MAT_NAME);
+      if (matName != null && !"".equals(matName)) {
+        mat = (Mat) Decorator.getDecorator(outer, Mat.class);
+        if (mat != null) {
+          final Point basePt = outer.getPosition();
+          contents = new ArrayList<>(mat.getContents());
+          offsets = new ArrayList<>(mat.getOffsets(basePt.x, basePt.y));
```
uckelman, line 392: Do you need to make another copy of the offsets list here? `getOffsets` is already returning you one that it created locally.

Cattlesquat: Oh so I should really just do...

```
contents = mat.getContents();
offsets = mat.getOffsets(basePt.x, basePt.y);
```

Yeah?

uckelman: If `getContents()` is also returning something that's a local, rather than a member of `mat`, then yeah, you don't need to make additional copies.

vassal-app/src/main/java/VASSAL/counters/SendToLocation.java
```diff
@@ -388,8 +388,8 @@ public Command myKeyEvent(KeyStroke stroke) {
         mat = (Mat) Decorator.getDecorator(outer, Mat.class);
         if (mat != null) {
           final Point basePt = outer.getPosition();
-          contents = new ArrayList<>(mat.getContents());
-          offsets = new ArrayList<>(mat.getOffsets(basePt.x, basePt.y));
+          contents = mat.getContents();
```
uckelman, line 391: Oh, oh, oh. That's going to be bad for `contents`. This is what `Mat.getContents()` does:
```
 public List<GamePiece> getContents() {
    return contents;
  }
```
So it's _not_ returning a copy---it's returning the actual contents list held by the `Mat`, which whatever you've given it to can now twiddle at will. That's no bueno.

Cattlesquat: I'm not actually twiddling it, but I guess obviously "someone could someday". 

Rather than keep guessing, maybe just let me know what you'd like this to say?

uckelman: Sorry. I thought you'd confirmed that `getContents()` was already returning a copy. Since it's not, you should do:
```
contents = new ArrayList<>(mat.getContents());
```
Alternatively, you could alter `getContents()` to return a copy. Whether that's a good idea depends on how much `getContents()` is called in contexts where you don't need a copy.

Cattlesquat: Ah - okay since I can't imagine it ever being a good idea for any "recipient" of this to twiddle contents itself, I've changed getContents() to return a copy and changed one case where it would thus have been re-copying the copy to account for that. Yay copies.

uckelman: This is one of those things some languages resolve by having `const` and non-`const` versions of functions, so then you can make the `const` version return the real thing with a guarantee that it won't be modified.

vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
uckelman: Ok, so I've refreshed my memory on properties. You're right that a _lot_ of properties are strings. There are also quite a few which are `Boolean`, though, e.g., `SELECTED`, `OBSCURED_TO_ME`, `OBSCURED_TO_OTHERS`, `NO_STACK`.

So you could have `MatCargo.getProperty()` (and `getLocalizedProperty()`) return `true` instead of `"true"` and change the test on the value here (and in the other three places where `IS_CARGO` is used) to `Boolean.TRUE.equals()`.

Cattlesquat: Gotcha - all fixed

uckelman (approved):

PR #10269, opened 2021-07-31
10272 Don't auto-name new regions a long string

Cattlesquat:
Right new every time you make a new region it gets auto-named "New Region" which means you have to first delete all of that text every single time. So I've changed it to a blank string which seems a lot easier-to-use.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/boardPicker/board/Region.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/boardPicker/board/Region.java
@@ -45,7 +45,7 @@ public class Region extends AbstractConfigurable {
   private boolean selected = false;
 
   public Region() {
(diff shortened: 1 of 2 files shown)


Cattlesquat:
Issue #10272

BrentEaston (changes requested):
I think this would be better if you change the code to explicitly set the name to the null string than setting it to a translation string that has been set to null. This is just confusing and someone in the future is going to put that back thinking it is a missing translation, or a translator may feel the need to translate 'New Region'.

Cattlesquat:
Oh okay I had *thought* about doing that but then worried that would make it harder to change later. But that might be a silly worry since I obviously think the _proper_ result is to have these be blank.

Cattlesquat:
Should be all set now.

BrentEaston (approved):

PR #10270, opened 2021-08-01
10273 Mat Cargo can be set to detect mats way from their exact center

Cattlesquat:
I have cards that I need to "tuck halfway under mats". Obviously making the user get the precise one-pixel center point just onto the mat as opposed to just-not-on-it would be "Bad UX". 

So here's a way to add some extra detection points to the MatCargo when it goes looking for its next mat. 

A bit kludgey but (a) it should solve 99% of conceivable issues in this depatment, (b) I don't have the math to do better, and (c) doing *much* better would presumably require rewriting PieceFinder.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/MatCargo.java
+++ b/vassal-app/src/main/java/VASSAL/counters/MatCargo.java
@@ -424,8 +424,8 @@ public class MatCargo extends Decorator implements TranslatablePiece {
     private final StringConfigurer descInput;
     private final BooleanConfigurer rotInput;
     private final TraitConfigPanel controls;
-    private final IntConfigurer XInput;
-    private final IntConfigurer YInput;
+    private final IntConfigurer xInput;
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/counters/MatCargo.java
```diff
@@ -390,6 +424,8 @@ public HelpFile getHelpFile() {
     private final StringConfigurer descInput;
     private final BooleanConfigurer rotInput;
     private final TraitConfigPanel controls;
+    private final IntConfigurer XInput;
+    private final IntConfigurer YInput;
```
uckelman, line 428: Generally we don't capitalize private members.

Cattlesquat: fixed

BrentEaston (approved):

PR #10271, opened 2021-08-01
9973 - Deck supports configurable strings for Draw Multiple, Draw Specific, Face Up, Face Down

Cattlesquat:
…ecific, Face Up, Face Down

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/DrawPile.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/DrawPile.java
@@ -84,6 +84,12 @@ public class DrawPile extends SetupStack implements PropertySource, PropertyName
 
   private final VisibilityCondition restrictExpressionVisibleCondition = () -> dummy.isRestrictOption();
 
+  private final VisibilityCondition drawMultipleMessageVisibleCondition = () -> dummy.isAllowMultipleDraw();
+
+  private final VisibilityCondition drawSpecificMessageVisibleCondition = () -> dummy.isAllowSelectDraw();
+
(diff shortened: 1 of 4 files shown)


Cattlesquat:
#9973

vassal-app/src/main/java/VASSAL/build/module/map/DrawPile.java
```diff
@@ -279,7 +290,11 @@ public boolean isDisplayNames() {
       EXPRESSIONCOUNTING,
       COUNTEXPRESSIONS,
       RESTRICT_OPTION,
-      RESTRICT_EXPRESSION
+      RESTRICT_EXPRESSION,
+      DRAW_MULTIPLE_MESSAGE,
+      DRAW_SPECIFIC_MESSAGE,
+      FACE_UP_MESSAGE,
+      FACE_DOWN_MESSAGE
```
BrentEaston, line 297: Re-order these (and getDescriptions() and getTypes()) so that the menu command configurers appear up with the config options that select them. So the 'Menu text to draw multiple' should appear directly below the 'Allow multiple cards to be drawn' option.

BrentEaston (changes requested):

vassal-app/src/main/java/VASSAL/build/module/map/DrawPile.java
Cattlesquat: Oh right, I think you'd had me do that "later in the PR history"

Cattlesquat: All fixed

BrentEaston (approved):

PR #10275, opened 2021-08-02
10276 Expose property if we're currently drawing Counter Detail Viewer

Cattlesquat:
So I've got these cards that "tuck underneath the map", and it's working perfectly w/ a CalculatedProperty to tell them whether to use their "top half is transparent" version of the image.

However when one "rolled over" the card for CounterDetailViewer, it of course displays the ugly "half empty field" version of the image in the Mouseover box. 

(Oh and by the way this is NOT the first time this same issue has come up for me making a module - I've run into similar situations with previous modules and just never got around to making this change)

So this PR makes there be a property you can check if you're currently drawing the CounterDetailViewer, and results in the "correct result" below:
image


BrentEaston (changes requested):
Can you add a reference to DrawingMouseOver to any appropriate places in the doco?

Cattlesquat:
Doco added!

BrentEaston (approved):

PR #10277, opened 2021-08-02
10277 - When editing an IrregularGrid, start the map view somewhere near the grid or its parent Zone

Cattlesquat:
"This PR is its own ticket"

When you edit an IrregularGrid, it used to just start its little editor view at the upper left corner of the map. Which is of course annoying when the map is 8000x8000 and your grid/zone is in the lower right or whatever.

So this now attempts to make an at least vaguely sensible centering as the starting point:
(a) If the IrregularGrid has any Regions in it, then the regions are used as a place to scroll the map
(b) If the IrregularGrid has NO Regions, but is part of a Zone, then the Zone shape is used.

Tested this in both cases and it produces pretty reasonably sensible starting points. Definitely open to ways it could be even smarter.

image


BrentEaston (approved):
Solves the immediate problem well enough. Not much more you could do.

PR #10285, opened 2021-08-04
Some additional useful properties for Mats and Mat Cargo, along with documentation

Cattlesquat:
(Following an Almoravid stress dream last night I awoke with a vision of exposing more properties)

This PR adds on to an already-new 3.6 feature, and so is either its own ticket, or doesn't need a ticket since it's just part of "Mats"


Cattlesquat:
It also developed that the Delete and Replace traits needed some Mat support. So I added that.

BrentEaston (approved):

PR #10286, opened 2021-08-04
Add removal/deprecation checker and remove anything deprecated a year ago

uckelman:


vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
```diff
@@ -66,11 +66,21 @@ Dialogs.disable=Don't show this again
 Dialogs.show_details=Show details
 Dialogs.hide_details=Hide details
 Dialogs.dont_show_again=Don't show this again
+
+Dialogs.removed_code_title=Defunct Custom Code
+Dialogs.removed_code_heading=Defunct Custom Code
+Dialogs.removed_code_message=This module uses VASSAL classes, methods, or fields which were deprecated more than a year ago and have since been removed. In order to run this module, uses of the classes, methods, or fields listed below must be replaced and the custom code recompiled.
```
Cattlesquat, line 72: This module cannot be opened because it contains "custom code" that references VASSAL classes, methods, or fields which were announced as deprecated more than a year ago and have since been removed. 

IF YOU ARE A MODULE USER - you will need to contact the module designer/maintainer and get them to correct the problems in their custom code. If the module maintainer is no longer active or cannot be found, please post a message in the Technical Support forums at https://vassalengine.org/forum

IF YOU ARE THE MODULE DESIGNER/MAINTAINER - In order to run this module, uses of the classes, methods, or fields listed below must be replaced and the custom code recompiled (alternatively the custom code can be removed from the module). If you are not the original maintainer and do not have the knowledge of how to do this, please post a message in the Technical Support forums at https://vassalengine.org/forum

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
```diff
@@ -66,11 +66,21 @@ Dialogs.disable=Don't show this again
...
+
+Dialogs.deprecated_code_title=Obsolete Custom Code
+Dialogs.deprecated_code_heading=Obsolete Custom Code
+Dialogs.deprecated_code_message=This module uses VASSAL classes, methods, or fields which are deprecated and will be removed in some future version. The uses of the classes, methods, or fields listed below must be replaced and the custom code recompiled before these classes, methods, or fields are removed.
```
Cattlesquat, line 76: This module contains "custom code" that uses VASSAL classes, methods, or fields which are deprecated and will be removed in some future version. 

IF YOU ARE A MODULE USER - you can run the module for now, but you may want to let the module designer/maintainer know that their custom code needs updating.

IF YOU ARE THE MODULE DESIGNER/MAINTAINER - The uses of the classes, methods, or fields listed below should be replaced and the custom code recompiled before these classes, methods, or fields are removed. If you are not the original maintainer and do not have the knowledge of how to do this, please post a message in the Technical Support forums at https://vassalengine.org/forum

Cattlesquat (changes requested):
New text supplied. Also, see my analysis (in the review comments) of why it is best to "errors and out" at module boot time.

vassal-app/src/main/java/VASSAL/launch/AbstractLaunchAction.java
```diff
@@ -255,6 +260,35 @@ public Void doInBackground() throws InterruptedException,
...
 
+      // Check for usage of removed and deprecated classes, methods, fields
+      final RemovalAndDeprecationChecker rdc = new RemovalAndDeprecationChecker();
+      Pair<Map<String, Set<String>>, Map<String, Set<String>>> rd;
+      try (ZipFile zf = new ZipFile(lr.module.getAbsolutePath())) {
+        rd = rdc.check(zf);
+      }
+
+      final Map<String, Set<String>> deprecated = rd.getRight();
+      if (!deprecated.isEmpty()) {
```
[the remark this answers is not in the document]
uckelman: I've made the removed dialog first.

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
uckelman: Do you think we should mention for the maintainers that they can find notes about replacements in the Javadoc and provide a link to that?

Cattlesquat: Seems like a reasonably good idea

uckelman:
@Cattlesquat What do you think now?

uckelman:
I've thought of one more thing we need to do, which is to check extensions.

PR #10292, opened 2021-08-05
Properties for OldMat, OldMatID, etc.

Cattlesquat:
Mat Cargo now track an OldMat and OldMatID, corresponding with CurrentMat and CurrentMatID.

This also fixes some bugs I found with moving a piece from one mat to another (apparently my testing was mostly moving onto and off of mats, less transferring between mats). Hooray for actually making a module in parallel with the new feature :)

"This PR is its own ticket"

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Decorator.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Decorator.java
@@ -18,6 +18,7 @@
 package VASSAL.counters;
 
 import VASSAL.build.BadDataReport;
+import VASSAL.build.GameModule;
 import VASSAL.build.module.GameState;
 import VASSAL.build.module.Map;
 import VASSAL.build.module.map.boardPicker.Board;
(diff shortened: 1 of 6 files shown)


vassal-app/src/main/java/VASSAL/counters/Decorator.java
```diff
@@ -749,6 +777,12 @@ public static Command putOldProperties(GamePiece p) {
     comm = comm.append(container.setPersistentProperty(BasicPiece.OLD_BOARD, boardName));
     comm = comm.append(container.setPersistentProperty(BasicPiece.OLD_ZONE, zoneName));
     comm = comm.append(container.setPersistentProperty(BasicPiece.OLD_LOCATION_NAME, locationName));
+    if (wasOnMat) {
+      comm = comm.append(container.setPersistentProperty(BasicPiece.OLD_MAT, matName));
+      comm = comm.append(container.setPersistentProperty(BasicPiece.OLD_MAT_ID, matID));
+      comm = comm.append(container.setPersistentProperty(BasicPiece.OLD_MAT_PIECE_NAME, matPieceName));
+      comm = comm.append(container.setPersistentProperty(BasicPiece.OLD_MAT_BASIC_NAME, matBasicName));
+    }
```
BrentEaston, line 785: Do the OLD Mat properties ever get reset to null once a piece is no longer on a mat? Should the next move where !wasOnMat set them to null?

Cattlesquat: Ah yes you're right, I out-clevered myself. Fixed now.

BrentEaston (approved):

PR #10294, opened 2021-08-06
Let FreeRotator start with a specific angle index

Cattlesquat:


vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
```diff
@@ -981,12 +1001,13 @@ public String getType() {
...
       se.append(nameConfig.getValueString());
       se.append(descConfig.getValueString());
+      se.append(startAngleConfig.getValueString());
       return ID + se.getValue();
     }
 
     @Override
     public String getState() {
-      return "0";
+      return startAngleConfig.getValueString();
```
BrentEaston, line 1010: return rotator.getState();

vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
```diff
@@ -859,6 +873,7 @@ public boolean testEquals(Object o) {
     private final StringConfigurer nameConfig;
     private final StringConfigurer descConfig;
     private final JLabel anyLabel;
+    private final IntConfigurer startAngleConfig;
```
BrentEaston, line 876: private final FreeRotator rotator;

vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
```diff
@@ -931,6 +946,11 @@ public Ed(FreeRotator p) {
       rndKeyConfig = new NamedHotKeyConfigurer(p.rotateRNDKey);
       panel.add(rndKeyConfig.getControls(), "wrap"); // NON-NLS
 
+      final JLabel startAngleLabel = new JLabel(Resources.getString("Editor.FreeRotator.start_angle"));
+      panel.add(startAngleLabel);
+      startAngleConfig = new IntConfigurer(p.startAngleIndex);
+      panel.add(startAngleConfig.getControls());
+
```
BrentEaston, line 953: rotator = p;

BrentEaston (commented):
I don;t think this is needed. Can't you just fix the bug in FreeRotator.Ed.getState() (see below) and then rotate the unit in the palette?

vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
BrentEaston: Ah, I know why they returned "0". For my solution, you would need to check if the type has changed from fixed to free or v/v and convert the starting state if it is non-zero.

Cattlesquat:
Okay, so "good news and bad news":

GOOD NEWS - with a few extra details worked out (see current version of PR) this solution now "works"

BAD NEWS - After some hair-pulling thinking there was some weird bug in my code, I have discovered that actually there is a crash bug IN THE ORIGINAL CODE. This PR doesn't exacerbate it, nor does it fix it, but *testing* this PR reveals the original crash bug.

REPRO STEPS (original code, don't need to use this PR):
1. Make a piece with a FreeRotator w/ e.g. 8 valid angle facings (I assume the default 6 works too, I was using 8)
2. Manually rotate the image a step (i.e. modifying the state)
3. Now open the FreeRotator (Can Rotate) trait to edit
4. Check the checkbox to "allow arbitrary facings"
5. Click OK button

EXPECTED: yay you edited your thingy and we cleared the state back to 0
ACTUAL: crash

The crash is somewhere in updating the image/sizing-of-image/bounding-box-of-the-sizing code for redrawing the thing, and has apparently been there for Who Knows How Long. 

Obviously the case where we have a pre-rotated state of a thing and then someone decides to change to an arbitrary facing is so unusual/rare that we've never noticed it until now. 

So I propose mainly that we merge this and move on, since it handles the key issues at hand correctly, enables my use case, and doesn't make the existing bug worse. Fixing the revealed bug is probably your wheelhouse as far as configurers-interacting-with-things; I'm not sure it's even an "important" bug, we might just want to open an issue and leave it for now?

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -675,6 +675,7 @@ Editor.FreeRotator.rotate_ccw_command_description=Rotate CCW command
 Editor.FreeRotator.rotate_random_command_description=Rotate Random command
 Editor.FreeRotator.allow_arbitrary_rotations=Allow arbitrary rotations
 Editor.FreeRotator.number_of_allowed_facings=Number of allowed facings
+Editor.FreeRotator.start_angle=Starting angle index (or angle if arbitrary rotations allowed)
```
BrentEaston, line 678: No longer needed?

vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
```diff
@@ -126,7 +126,7 @@
 
   public FreeRotator() {
     // modified for random rotation (added two ; )
-    this(ID + "6;];[;" + Resources.getString("Editor.FreeRotator.default_rotate_cw_command") + ";" + Resources.getString("Editor.FreeRotator.default_rotate_ccw_command") + ";;;;", null); //$NON-NLS-1$//
+    this(ID + "6;];[;" + Resources.getString("Editor.FreeRotator.default_rotate_cw_command") + ";" + Resources.getString("Editor.FreeRotator.default_rotate_ccw_command") + ";;;;;", null); //$NON-NLS-1$//
```
BrentEaston, line 129: The extra ';' should no longer be needed as we are no longer adding the start  angle option?

BrentEaston (changes requested):
Still some cruft left from the original solution.

Cattlesquat:
Cruft removed!

BrentEaston (approved):

PR #10297, opened 2021-08-07
10297 - EDITOR - After duplicating an item in the Editor and saving it, select the duplicate

Cattlesquat:
Minor touchup of the "Duplicate" action that I believe is already a "new with 3.6 feature"

PREVIOUSLY - if you duplicated an item and accepted/saved from the configurer that appeared, the OLD item (the one you'd duplicated) would stay selected. This got a bit confusing to me, because when you'd just been editing the NEW item it seemed like the new item would be the one selected (and so I'd sometimes end up next editing a child of the OLD item by accident)

NOW - if you duplicate an item and save/accept from the configurer, the NEW item becomes selected (if you cancel, the old item stays selected since the new item ends up not actually being created)

Not sure this even needs a new entry in the change list (since it's just a touchup of a new feature)


uckelman:
Looks good. No need to list this in the change log as it affects something that's never gone out in a release.

PR #10329, opened 2021-08-20
Various Audit Trail fixes

BrentEaston:
Prevents spurious 'Bad Data in Module' errors on startup when 'level follow expression' expressions fail evaluation because not all properties are available to pieces in palettes.

Stopping evaluation of the expressions will change the visual display of some pieces in some modules, so for the sake of compatibility, we will just suppress these messages. 

Prevent recursive cascade of Audit Trails being generated during Audit Trail reporting.

Prevent spurious Bad Data errors when clicking on palette units that use Count and Sum functions.


vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -687,14 +687,13 @@ private String unescape(String expr) {
   }
 
   private List<Map> getMapList(Object mapName, GamePiece sourcePiece) {
-    final List<Map> maps;
+    List<Map> maps = new ArrayList<>();
```
uckelman, line 690: Don't you want to avoid creating the `ArrayList` for the case where you're going to immediately throw it away?

vassal-app/src/main/java/VASSAL/script/expression/AuditTrail.java
```diff
@@ -189,9 +197,30 @@ public void setSource(Auditable source) {
...
+    if (source != null) {
+      // Calling getComponentName() on a GamePiece may cause further expression evaluations, which could fail.
+      // Set the reportingInProgress flag on which will prevent the creation of any more audit trails until
+      // this trail has completed reporting to prevent an infinite recursive cycle of Audit Trail creation.
+      final boolean origReportingState = isReportingInProgress();
+      setReportingInProgress(true);
+      if (source instanceof EditablePiece) {
+        sb.append("\n   ").append(Resources.getString("Audit.source_type", "Piece Trait"));
+        sb.append("\n   ").append(Resources.getString("Audit.source_name", source.getComponentName()));
+        sb.append("\n   ").append(Resources.getString("Audit.source_description", source.getComponentTypeName()));
```
uckelman, line 210: You can chain them all, just drop the `sb` at the start of successive lines and remove the semicolons.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
BrentEaston: True. I have recoded getMapList, got rid of the temporary List alltogether

vassal-app/src/main/java/VASSAL/script/expression/AuditTrail.java
BrentEaston: Good point.

uckelman (approved):

PR #10344, opened 2021-08-23
10226 Force Game Piece Image names to be valid png filenames

BrentEaston:
Newly created GPI's will be forced to be a valid file name containing only 09,a-z,A-Z and _ and ending in .png.

Not much can be done about filenames of existing GPI's as other components will be referencing them. However, if an existing GPI has a valid filename, or is changed to gave a valid filename, then valid filenames will be enforced for that GPI from then on.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/GamePieceImage.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/GamePieceImage.java
@@ -18,6 +18,22 @@
 
 package VASSAL.build.module.gamepieceimage;
 
+import VASSAL.build.AbstractConfigurable;
+import VASSAL.build.AutoConfigurable;
+import VASSAL.build.Buildable;
+import VASSAL.build.GameModule;
+import VASSAL.build.module.documentation.HelpFile;
+import VASSAL.configure.Configurer;
+import VASSAL.configure.ConfigurerFactory;
+import VASSAL.configure.StringConfigurer;
+import VASSAL.configure.VisibilityCondition;
+import VASSAL.i18n.Resources;
+import VASSAL.tools.ArchiveWriter;
+import VASSAL.tools.ErrorDialog;
+import VASSAL.tools.UniqueIdManager;
+import VASSAL.tools.imageop.Op;
+import VASSAL.tools.imageop.SourceOp;
+
 import java.awt.Image;
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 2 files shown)


uckelman:
Why allow `$`?

BrentEaston:
Good Point. $ removed.

vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/GamePieceImage.java
```diff
@@ -463,4 +501,167 @@ protected void rebuildInstances() {
...
+
+    private String clean(String string) {
+      final StringBuilder sb = new StringBuilder();
+      for (int i = 0; i < string.length(); i++) {
+        final char c = string.charAt(i);
+        if (c == '_' || (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')) {
+          sb.append(c);
+        }
+      }
+      return sb.toString();
```
uckelman, line 613: Could you replace this whole function with `string.replaceAll("\W", "");`?

BrentEaston: Better version.

vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/GamePieceImage.java
```diff
@@ -575,45 +575,34 @@ private boolean isPng(FilterBypass fb) {
...
-    private void fixPng(FilterBypass fb, int pos) throws BadLocationException {
+    private void fixPng(FilterBypass fb, int caretPos) throws BadLocationException {
       if (!isPng(fb)) {
         final String text = getText(fb);
         if (text.endsWith(".")) {
           super.replace(fb, text.length() - 1, 1, PNG_SUFFIX, null);
-          config.setCaretPosition(pos);
+          config.setCaretPosition(caretPos);
         }
         else if (text.toUpperCase().endsWith(".p")) {
```
jonathanrwatts, line 585: Am I missing something here, or is this test always going to fail (converting to uppercase, then checking for a lowercase letter)?  (Same for the next "else if", too.)

vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/GamePieceImage.java
```diff
@@ -539,8 +539,8 @@ public void setCaretPosition(int pos) {
   }
 
   /**
-   * ImageNameFilter that controls how the user can change the Image Name
-   * If the GPI is still version 0, then no controls. Not we can't force an image
+   * ImageNameFilter that controls how the user can change the Image Name.
+   * If the GPI is still version 0, then no controls. Note we can't force an image
    * name because other components may reference that image.
    * Once the GPI is version 0, then enfore a .png suffix and
```
jonathanrwatts, line 545: "and" what?

PR #10395, opened 2021-09-06
10387 Do not center on opponents moves if moving units are invisible

BrentEaston:


vassal-app/src/main/java/VASSAL/command/AddPiece.java
```diff
@@ -55,6 +56,19 @@ protected void executeCommand() {
...
+            // Do not center on a Stack of invisible pieces
+            int visibleCount = 0;
+            for (final GamePiece gp : ((Stack) target).asList()) {
+              if (!Boolean.TRUE.equals(gp.getProperty(Properties.INVISIBLE_TO_ME))) {
+                visibleCount++;
+                break;
+              }
+            }
+            if (visibleCount == 0) return;
+          }
```
uckelman, line 71: I think you could replace that whole block with this:
```
if (target instanceof Stack &&
    !((Stack) target).asList().stream().anyMatch(PieceIterator.VISIBLE)) {
  return;
}
```

BrentEaston: Thanks, I new there had to be an easier way.

uckelman (approved):

PR #10398, opened 2021-09-07
10394 Add Command to Rotate trait to rotate to specific facing or degrees

BrentEaston:


vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
```diff
@@ -565,6 +601,48 @@ else if (rotateRNDCommand.matches(stroke)) {
...
+          targetAngle = Double.parseDouble(targetAngleText);
+        }
+        catch (NumberFormatException e) {
+          reportDataError(this, Resources.getString("Error.non_number_error"), "Invalid number: " + targetAngleText, new AuditableException(this, audit));
+          targetAngle = (isFreeRotation() || !directTypeFacing) ? 0 : 1;
+        }
+      }
+
+      // Now have a valid numeric target angle
+      if (validAngles.length == 1 || !directTypeFacing) {
```
uckelman, line 633: Use `isFreeRotation()` here?

vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
```diff
@@ -202,6 +215,10 @@ public double getCumulativeAngleInRadians() {
     return -PI_180 * getCumulativeAngle();
   }
 
+  public boolean isFreeRotation() {
+    return validAngles.length == 1;
+  }
+
   public void setAngle(double angle) {
     if (validAngles.length == 1) {
```
uckelman, line 223: Use `isFreeRotation()` here?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
BrentEaston: Yes indeed, and the other dozen or so places. Fixed.

uckelman (approved):

PR #10406, opened 2021-09-10
10405 Fix issue with Delete trait failing after load log file

BrentEaston:
Clear the pieces in the current game state completely before loading a new log file on top.


uckelman:
Woah, that's wild. How did anything ever work without that?

uckelman (approved):
I hesitate to ask, but is there anything else which needs to be reset at this point as well?

PR #10432, opened 2021-09-19
Store the tile cache in %LOCALAPPDATA% instead of %APPDATA% on Windows.

uckelman:
The latter causes problems for people with a settings directory which
lives on a network share.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/Info.java
+++ b/vassal-app/src/main/java/VASSAL/Info.java
@@ -100,6 +100,10 @@ public final class Info {
     return CONFIG.getConfDir().toFile();
   }
 
+  public static File getCacheDir() {
+    return CONFIG.getCacheDir().toFile();
+  }
+
   public static File getTempDir() {
     return CONFIG.getTempDir().toFile();
   }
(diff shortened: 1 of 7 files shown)


BrentEaston (approved):

BrentEaston:
I've confirmed that your localappdata change creates the tiles in appdata/local instead of appdata/roaming. 

This change will leave windows users with GB of junk orphaned in their appdata/roaming/VASSAL/tiles directory.

I'm wondering if MM -> Clear Tile cache should be changed to clear the tile cache in both roaming and local?

uckelman:
> This change will leave windows users with GB of junk orphaned in their appdata/roaming/VASSAL/tiles directory.

I know. That's exactly the reason I set it to be a draft, as I'm not sure how to deal with that.
 
> I'm wondering if MM -> Clear Tile cache should be changed to clear the tile cache in both roaming and local?

That's a good idea.

uckelman:
That's still not great, though, since people might never clear the tile cache.

BrentEaston:
True. You could have windows machines always clear the roaming cache on startup. Minimal overhead once its done.

jonathanrwatts:
Don't forget that some people insist on keeping old versions around and using them for old modules, so constantly clearing the roaming cache is a bad idea in that case, too.

I think the best solution is just to include a highlighted note in the release notes that describes the issue and how to manually clear the remote cache (or possibly include an option in the MM to do it for the user on demand).

BrentEaston:
Very good point Jonathan.

I agree, I think we just have the Clear cache option clear both and leave it at that. 

And try and highlight how the changed behavour  will affect people. Ideally, users moving to 3.6 should clear their cache when they first start 3.6. The first thing people will notice moving to 3.6 is that all of their modules will re-tile and the questions will flow from there.

PR #10452, opened 2021-09-23
10452 - Fix Mat related crash in Game Refresher

Cattlesquat:
I had a NullPointerException when running GameRefresher on a game with Mats in it. It appears that at some point the "getPieceForId" returned a null instead of a piece, and this got inserted into a Mat's "contents" list, leading to Bad Things later on.

This patches the Mat not to accept a null contents member. Which is correct and reasonable behavior as far as it goes. 

What it does not fix is whatever reason getPieceForId() returns null in the first place, which may well be some "temporary state during GameRefreshing". In which case GameRefreshing a Mat-game might possibly dissociate pieces from their Mats in some cases. But at least that's better than crashing.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Mat.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Mat.java
@@ -119,7 +119,7 @@ public class Mat extends Decorator implements TranslatablePiece {
     final int num = st.nextInt(0);
     final GameState gs = GameModule.getGameModule().getGameState();
     for (int i = 0; i < num; i++) {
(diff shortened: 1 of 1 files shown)


BrentEaston:
Will likely be incorporated into #10468

Cattlesquat:
> Will likely be incorporated into #10468

Yes I agree it is probably good to not leave Mats willing to accept null members (at the very least we should immediately throw exception when the null member is first added)

BrentEaston:
Incorporated into #10468

PR #10474, opened 2021-09-30
10474 - Movement traits need to update cargo's mat after movement before applying apply-on-move keystroke

Cattlesquat:
Send-to-Location, Rotate, and Move-Fixed-Distance all had the "update our mat based on where we moved to" logic _after_ the apply-on-move key was applied to them. That was incorrect and was leading to "Current Mat" property not being updated correctly, wrong messages being printed, etc. 

I've moved the code blocks to the right location and verified that this was removing the "wrong messages" in my module. 

PieceMover was already working correctly (which is why I didn't notice this earlier)

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
+++ b/vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
@@ -504,15 +504,8 @@ public class FreeRotator extends Decorator
     // Move the piece
     c = c.append(map.placeOrMerge(outer, dest));
 
-    if (GameModule.getGameModule().isMatSupport()) {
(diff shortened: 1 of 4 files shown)


uckelman:
It's kind of gross that the same block of code is copied there four times.

Cattlesquat:
> It's kind of gross that the same block of code is copied there four times.

Got a preference where to centralize it?

Cattlesquat:
Cool - all taken care of. Code looks nice now. I'm always afraid to introduce "static" w/o being told to :)

uckelman (approved):

uckelman:
Something got confused here. It's merged.

PR #10484, opened 2021-10-02
10484 - Deselect trait improvements and double-click Mat

Cattlesquat:
I've discovered a few more UX cases for Mats that led me to want to improve the select/deselect options.

I've improved the existing Deselect trait to allow three "modes" of Deselection:
(1) Deselect this piece (default & what the trait formerly only did)
(2) Deselect ALL pieces
(3) Select ONLY this piece

Also double clicking a Mat can be used to ditch selection of cargo items.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/KeyBufferer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/KeyBufferer.java
@@ -53,6 +53,8 @@ import VASSAL.counters.Properties;
 import VASSAL.counters.Stack;
 import VASSAL.tools.swing.SwingUtils;
 
+import static VASSAL.counters.Mat.MAT_NAME;
+
 /**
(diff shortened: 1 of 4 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/KeyBufferer.java
```diff
@@ -258,26 +260,13 @@ public void mousePressed(MouseEvent e) {
...
-            if (p instanceof Decorator) {
-              final Mat mat = (Mat)Decorator.getDecorator(Decorator.getOutermost(p), Mat.class);
-              if (mat != null) {
-                final List<GamePiece> matPieces = new ArrayList<GamePiece>(mat.getContents());
-                for (final GamePiece mp : matPieces) {
-                  kbuf.add(mp);
-                }
-              }
-            }
-            */
```
uckelman, line 280: Yay, cruft removal.

PR #10486, opened 2021-10-03
10486 Better obscurable option for insetting

Cattlesquat:
In a "Mask" trait, the current "Inset" option doesn't work well (at least for me) because it draws the "masked back" over top of 1/4 of the "real" image, meaning that if you need to be able to read the card text of the real image, you can't. 

So I have added a second inset option, which instead uses the "masked back" as the "behind part" and puts the "real image" as the inset, and at a larger amount of space.

image

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Obscurable.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Obscurable.java
@@ -63,6 +63,7 @@ public class Obscurable extends Decorator implements TranslatablePiece {
   protected static final char BACKGROUND = 'B';
   protected static final char PEEK = 'P';
   protected static final char IMAGE = 'G';
+  protected static final char INSET2 = '2';
   protected static final String DEFAULT_PEEK_COMMAND = Resources.getString("Editor.Obscurable.default_peek_command");
 
   protected char obscureKey;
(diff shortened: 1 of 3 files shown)


BrentEaston (changes requested):
looks good, but don't forget the reference manual update :)

Cattlesquat:
Done - including fixing the earlier inset documentation which was wrong

vassal-app/src/main/java/VASSAL/counters/Obscurable.java
```diff
@@ -368,6 +371,14 @@ protected void drawObscuredToOthers(Graphics g, int x, int y, Component obs, dou
           - .5 * zoom * obsBounds.height / 2),
         obs, zoom * 0.5);
       break;
+    case INSET2:
+      obscuredToMeView.draw(g, x, y, obs, zoom);
+      final Rectangle bounds2 = piece.getShape().getBounds();
+      final Rectangle obsBounds2 = obscuredToMeView.getShape().getBounds();
+      piece.draw(g, x - (int) (zoom * .8 * obsBounds2.width / 2  - .8 * zoom * bounds2.width/2),
+                    y - (int) (zoom * .8 * obsBounds2.height / 2 - .8 * zoom * bounds2.height/2),
```
uckelman, line 379: I I hear you can factor expressions like this:
```
0.4 * zoom * (obsBounds2.width - bounds2.width)
```?

Cattlesquat: Haha I imagine. I just copied from some other piece of ass-crud without thinking about the math too much, but happy to clean it up.

BrentEaston (approved):

PR #10506, opened 2021-10-06
10505 Enable Offset for Send To Location Grid option

BrentEaston:


BrentEaston:
Yeah right, like that is ever going to happen :) That comment has already been removed from the documentation, so the documentation now accurately reflects the results of this PR. I did tweak the doco slightly.

PR #10507, opened 2021-10-06
More robust java -version parsing; Fix #10439

### LINKED ISSUE #10439: VASSAL.sh: Incorrectly parsing version
If you have _JAVA_OPTIONS set, the startup script will not correctly identify the version, but instead try to parse the _JAVA_OPTIONS.

```
$ ./VASSAL.sh 
Error: VASSAL requires Java 11 or later. The Java you are using (/usr/bin/java) has an unrecognizable version 'Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=lcd'.
```

```
$ java --version
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=lcd
java 11.0.2 2019-01-15 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.2+9-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.2+9-LTS, mixed mode)
```

```
$ env | grep _JAVA_OPTIONS
_JAVA_OPTIONS=-Dawt.useSystemAAFontSettings=lcd
```

mlassnig:
Replaces a simple `head` with a `grep`.

### DIFF
--- a/dist/linux/VASSAL.sh
+++ b/dist/linux/VASSAL.sh
@@ -13,7 +13,7 @@ if [ ! -x "$JAVA" ]; then
 fi
 
 # Check the Java vesion; we require Java 11+.
-VERSION=$("$JAVA" -version 2>&1 | head -1 | sed -e 's/.* version "//; s/".*//')
+VERSION=$("$JAVA" -version 2>&1 | grep version | sed -e 's/.* version "//; s/".*//')
 
 # Java 8 and earlier formats version x.y as 1.x.y
 # Java 9 and later formats version x.y as x.y


dist/linux/VASSAL.sh
```diff
@@ -13,7 +13,7 @@ if [ ! -x "$JAVA" ]; then
 fi
 
 # Check the Java vesion; we require Java 11+.
-VERSION=$("$JAVA" -version 2>&1 | head -1 | sed -e 's/.* version "//; s/".*//')
+VERSION=$("$JAVA" -version 2>&1 | grep version | sed -e 's/.* version "//; s/".*//')
```
jonathanrwatts, line 16: This isn't going to work--the line with the version number doesn't contain the text "version" anywhere, unfortunately.  Sadly, it isn't even guaranteed to contain the word "java"--on my (Kubuntu 21.04) system, it's `openjdk 11.0.11 2021-04-20`.

mlassnig: Hah, interesting.

I got this on Oracle Java:

```
$ java -version
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=lcd
java version "11.0.2" 2019-01-15 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.2+9-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.2+9-LTS, mixed mode)
```

this on OpenJDK11:
```
$ java -version
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=lcd
openjdk version "11.0.12" 2021-07-20
OpenJDK Runtime Environment (build 11.0.12+7-post-Debian-2)
OpenJDK 64-Bit Server VM (build 11.0.12+7-post-Debian-2, mixed mode, sharing)
```

this on OpenJDK18:

```
$ java -version
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=lcd
openjdk version "18-ea" 2022-03-15
OpenJDK Runtime Environment (build 18-ea+15-Debian-4)
OpenJDK 64-Bit Server VM (build 18-ea+15-Debian-4, mixed mode, sharing)
```



So, the grep for `version` continues to succeed as intended. (never mind the version parsing problem on OpenJDK18).

jonathanrwatts: False alarm--see my edited comment.  (I'm so used to using `java --version`, I forgot that `java -version` also works, but with slightly different output!)

uckelman: Hmm. I wonder if this is the wrong approach. If we tried to load a class file compiled for Java 11 and that failed, we'd know we can't load Vassal. Maybe we should do that instead of trying to parse the version number output.

uckelman: Have a look at #10510.

PR #10551, opened 2021-10-15
10551 - Action Button can open context menu

Cattlesquat:
Action Button trait now features a checkbox that allows (left)-clicking the action button to launch the piece's context menu (aka right click menu)

Doco included.

image


Cattlesquat:
It's for use in specific situations, rather than for e.g. every action button. 

Use case example: I have some "mostly-invisible pieces" that are designed as "slots", into which you muster a lord from a list of available lords. Right-click *works* of course, but launching the menu when you left click "to find out what this box is even for" adds to discoverability. 

Obviously misuse of this feature (like misuse of most UI features) leads to bad UI/UX. 

image

BrentEaston (approved):
I am happy with this change. It makes sense, it's not a single-action button, it's a select-from-menu action button. I think it is in keeping with the spirit of the actionbutton UI. Designers can (and do) abuse anything we do, can't worry about that too much.

Brian

PR #10572, opened 2021-10-19
10572 - Various Drag Threshold Bugs Fixed - dragThreshold preference & etc should work properly now

Cattlesquat:
The Drag Threshold calculation was getting fouled by the final position being offset by the amount the original click was off center on the dragged piece. Fixing it opened up a YAWNING ABYSS of various bugs in dragThreshold things. I think I've fixed most or all of them now.

* The weird offset that was being introduced into the drag-distance calculation is now accounted for
* You can now properly/intentionally drag-and-drop something a short distance once drag threshold is breached
* Drag Threshold "awt" setting is now set properly on entry (not just when you change the preference) 
* Drag Threshold value is now used for the drag-thresholdy part of KeyBufferer (aka lasso selecting units)
* Lasso-selecting units now fully uses drag threshold (like doesn't draw that irritating 3-pixel wide lasso thing for half your clicks)
* Menu Displayer now properly filters out lasso & drag-and-drop

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -246,7 +246,6 @@ public class Map extends AbstractToolbarItem implements GameComponent, MouseList
   protected String markUnmovedTooltip = Resources.getString("Map.mark_unmoved"); //$NON-NLS-1$
   protected MouseListener multicaster = null;
   protected ArrayList<MouseListener> mouseListenerStack = new ArrayList<>(); //NOPMD
-  protected MouseMotionListener mouseMotionListenerLocal = null;
   protected List<Board> boards = new CopyOnWriteArrayList<>();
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -279,6 +280,7 @@
   public Map() {
     getView();
     theMap.addMouseListener(this);
+    theMap.addMouseMotionListener(this);
```
uckelman, line 283: What's the deal with this?

Cattlesquat: I dunno -- it seemed like it *should* be there. This particular bit is one of a couple of things where I *thought* I was going go need some support in Map.java but then it turned out I didn't, but I left some of the cleanupy things in.

BrentEaston (approved):
Gave it a good test, Very nice.

PR #10576, opened 2021-10-20
10576 - Send-to-Location includes option to Cycle through pieces-selected-by-properties, or to find Nearest matching piece (default remains to find first/any match)

Cattlesquat:
Send To Location allows you to send the piece "To a piece selected by a property match expression". 

However it had a very simplistic implementation -- even if multiple pieces match, it _always_ sends to the first one on the list. Which makes the function largely useless in the multiple targets situation. 

New implementation cycles through the matching pieces (so if you keep executing the command it will hop from target to target)

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/SendToLocation.java
+++ b/vassal-app/src/main/java/VASSAL/counters/SendToLocation.java
@@ -260,6 +260,9 @@ public class SendToLocation extends Decorator implements TranslatablePiece {
               }
             }
           }
+          if (destination.equals(DEST_COUNTER.substring(0, 1)) && !targets.isEmpty()) {
+            break; // "Any match" version just takes first hit
+          }
         }
         else {
           if (propertyFilter.accept(outer, piece) && !targets.contains(piece)) {


BrentEaston (commented):
I'm not sure I understand the purpose of this one and also have some concerns.

What is the purpose of cycling through a list of potential targets? 

Also, isn't this vastly less efficient? The filter has to be applied to every single piece in the game every time, even if the 1st checked piece is the one and only piece. If you have a lot of counters doing this at once, there will be a noticeable impact.

Cattlesquat:
Well an example use case for Almoravid is there's a lord who can be sent to "a Cathedral". And there are various Cathedral tokens that may or may not be placed around the map. So rather than _only_ be able to send him to the first Cathedral you ever placed, this allows him to cycle around any Cathedral.

I guess it's less efficient if you only have one. I could see making it a "new option" and leaving the earlier one intact?

BrentEaston:
Any match/next match/closest match

If Any match is the default, that will leave legacy behaviour unaffected.

PR #10578, opened 2021-10-20
10578 - MatCargo now provides OldMatOffsetX and OldMatOffsetY properties

Cattlesquat:
Thought of a good use for "Old" versions of these two properties, so added them!


Cattlesquat:
done

PR #10579, opened 2021-10-20
Fix a little goof in the Trait Summary improvement

Cattlesquat:
Found a little goof in the Trait Summaries improvement (#10569) -- it was including "extra summary information" (in the form of various default field values) in the descriptions of the Available Traits (on the _left_ side of the piece definer). Oops. 

Anyway this fixes it. Would be better if a "pure" request-clean-description could be added to EditablePiece interface, but that of course would break every custom class in the universe. Fortunately we were always consistent with the " - " thing and have not supplied that as an option for translators to change.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
@@ -107,7 +107,8 @@ public class PieceDefiner extends JPanel {
   private static final SortedMap<String, GamePiece> alphaMap = new TreeMap<>();
   private static Boolean sorted = false;
   protected DefaultListModel<GamePiece> inUseModel;
-  protected ListCellRenderer<? super GamePiece> r;
+  protected ListCellRenderer<? super GamePiece> inUseRenderer;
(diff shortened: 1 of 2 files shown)


BrentEaston (commented):
Well, if you are after a spare-time activity, you could implement EditablePiece.getBaseDescription() with a default implementation that calls getDesctription() and does the " - " check. Then update all 40 traits to return the correct base description from the i18nkey. That would get you back to where you are now, but with a pat on the back as well!

Cattlesquat:
But alas EditablePiece is an interface rather than abstract so I can’t make a default, right? It was my original idea of how to fix but I think there’s not a way to avoid breaking all custom traits?

BrentEaston:
Decorator.buildDescription() should really just use getBaseDescription() internally rather than having to pass it the i18nkey a second time. But the String parameter setup makes that difficult. Would need a different method that doesn't need i18nkeys to be passed. But do I care that much to do that to you? Not really :)

Cattlesquat:
Yeah I *almost* started in on redoing buildDescription, but then I saw the i18n key thing and decided... blah. If we start having problems with the trait descriptions we can always do another iteration of reorg.

PR #10633, opened 2021-10-28
10633 - Remove doubled images from Ghost Bug Pref

Cattlesquat:


Cattlesquat:
Okay I eventually got back in and may have repaired this. It can sort of be compared as an alternative to #10642 

Note that for the same reason detailed in #10642 I have not so far changed the behavior related to your item #1 (Game Piece Palette) above -- because basically changing that behavior would change the behavior on all Linux machines from what it has been to date. Maybe we want that, maybe we don't.

BrentEaston (approved):
This is working well for me.

PR #10641, opened 2021-10-28
9226 - When Editor changes a prototype, clear the prototype cache and all pieceslot caches

Cattlesquat:
Something I've wanted to fix for a long time, and certainly gets mentioned a lot by other module designers: until now when you changed a prototype you basically had to completely restart vassal (not even just restart the game -- restart _VASSAL_) because none of your prototype changes would be honored until you did -- they all get cached in various places. 

So this PR clears out the cache from startup stacks and piece palettes and the like whenever a prototype is changed. This not only fixes the problem Claudio originally described (cascading changes among multiple prototypes), it also fixes the palettes and startup stacks.

Doesn't affect performance for normal _Player_ operations because no prototype ever changes during a Player session.

In testing this out I opened a module and changed some prototypes. The changes were _immediately_ felt in pieces then dragged from the Piece Palette. For At-Start Stacks, if I did a "Close Game" and then started a new game, the newly placed At Start stack pieces reflected my changes. Neither of these were true before this PR.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/PrototypeDefinition.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/PrototypeDefinition.java
@@ -183,9 +183,21 @@ public class PrototypeDefinition extends AbstractConfigurable
     return piece;
   }
 
+
+  public void clearCache() {
+    pieces.clear();
+  }
+
+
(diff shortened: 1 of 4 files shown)


BrentEaston (approved):
Nice fix!!!!

PR #10642, opened 2021-10-28
10642 - Compatibility Tab "ghost drag" option repaired a bit

Cattlesquat:
This seems to recover the main repairs I did to recover Ghost Drag's compatibility mode, and update the docs to note the required restart. 

Compatibility Mode doesn't have the same Palette behavior because of the code in the `dragEnter` method in DragHandlerNoImage in PieceMover, which changes the size of dragged items on the fly based on the zoom level of their current window. That code could be removed ... but apparently is the default present behavior on Linux. So changing it would change the current standard behavior for every Linux user ... for better and/or worse. 

See also #10633 for the more-full-featured version which I may have repaired.


BrentEaston (commented):
Ditch this version, I still have double drag ghosts

Cattlesquat:
Ditched!

PR #10647, opened 2021-10-29
6946 - Private Maps visible to non-owning players will correctly display Mouseover text

Cattlesquat:


BrentEaston (approved):
Clever solution just enabling the motion listeners. Works well.

PR #10650, opened 2021-10-29
4381 - Show X,Y in Configurer when e.g. repositioning Stack

Cattlesquat:
#4381 OLDEST BUG IN THE BOOK!!!
image

I might play with adding similar displays to the Define Shape for Zone, etc.


uckelman:
I'm going to merge this and then fix the PMD warnings.

uckelman:
The upshot is: You get strings when the right operand of `+` is a string; you don't need to prime the pump by starting with an empty string on the left.

PR #10660, opened 2021-10-31
9295 Report translated name for Symbolic Die Faces

BrentEaston:
Will have no effect on modules when a translation is not in effect, or if the Die Face names have not been translated.


Cattlesquat (approved):
Well that turned out to be straightforward!

PR #10674, opened 2021-11-01
9890 - Startup Global Key Commands - always fire in order, can't be undone, and can be configured to e.g. "once per game"

### LINKED ISSUE #9890: Change Startup GKC's to run in order and after all other components initialised
| Bugzilla |  |
|----------|----------|
| ID | BZ14244 |
| Reported | 2021-03-22 23:29:00 +0100 |
| Modified | 2021-03-22 23:40:09 +0100 |
| Product | VASSAL |
| Component | Player |
| Version | 3.5.3 |
| Platform | All |
| OS | All |
| Status | ASSIGNED |
| Resolution | None |
| Priority | unspecified |
| Severity | normal |
| Milestone | --- |
| Assignee | [Brent Easton](mailto:b.easton@exemail.com.au) |

#### [Brent Easton](mailto:b.easton@exemail.com.au) 2021-03-22 23:29:39 +0100
<pre>1. Will require a Compatibility option to revert to current broken behavior.
2. Remove invokeLater() from startup GKC setup()
3. After all GameComponent setup()s have been run, then run through each Startup GKC in order.</pre>

#### [Brent Easton](mailto:b.easton@exemail.com.au) 2021-03-22 23:32:02 +0100
<pre>Target for 3.6</pre>

#### [Brent Easton](mailto:b.easton@exemail.com.au) 2021-03-22 23:40:09 +0100
<pre>4. Warn user if they try and Undo back through the Startup GKC's.</pre>

Cattlesquat:
Fixes #9890

* SGKCs are now guaranteed to fire in the order they are listed in the module. There's no need to support returning to the old functionality because the old functionality was (a) officially undetermined, so the fact that we're doing them in the right order fulfills their non-expectations (b) but in fact seemed to fire more or less in order anyway. 
* SGKCs can now be configured as to whether they fire (a) once per player session (the default and the old functionality), (b) every time a game is started or loaded during the session, or (c) once per *game* when the game is being started for the first time, either blank or from a predefined setup (i.e. the flag whether the SGKC has been applied is saved with the game)
* SGKCs cannot be undone past

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
@@ -94,6 +94,7 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
   protected List<Command> logOutput;
   protected int nextInput = 0;
   protected int nextUndo = -1;
+  protected int dontUndoPast = 0;
   protected Command beginningState;
   protected File outputFile;
   protected Action stepAction = new StepAction();
@@ -236,6 +237,7 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
       logOutput.clear();
       nextInput = 0;
       nextUndo = -1;
+      dontUndoPast = 0;
       beginningState = null; // Will create one when we actually start a log
     }
     else {
@@ -429,7 +431,7 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
     File file = fc.getSelectedFile();
 
     // append .vlog if it's not there already
-    if (!file.getName().endsWith(".vlog")) {
+    if (!file.getName().endsWith(".vlog")) { //NON-NLS
       file = new File(file.getParent(), file.getName() + ".vlog"); //NON-NLS
     }
 
@@ -485,10 +487,18 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
   }
 
   /**
+   * Mark the current command as a no-undo-past-this-point.
+   */
+  public void blockUndo() {
+    dontUndoPast = nextUndo + 1; // Need +1 because this is done by a command that is about to be added to the list
+    undoAction.setEnabled(false);
+  }
+
+  /**
    * This handles the UNDO button, executing the actual "Undo".
    */
   protected void undo() {
-    if (nextUndo < 0) {
+    if (nextUndo <= dontUndoPast) {
       return; //BR// Throw away extra keys-held-down when nothing left to do
     }
 
@@ -496,19 +506,19 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
     final Command lastInput = (nextInput > logInput.size() || nextInput < 1) ?
       null : logInput.get(nextInput - 1);
     if (lastInput == lastOutput) {
-      while (nextInput-- > 0) {
+      while (nextInput-- > dontUndoPast) {
         stepAction.setEnabled(true);
         if (logInput.get(nextInput).getUndoCommand() != null) {
           break;
         }
       }
     }
-    while (nextUndo-- > 0) {
+    while (nextUndo-- > dontUndoPast) {
       if (logOutput.get(nextUndo).getUndoCommand() != null) {
         break;
       }
     }
-    undoAction.setEnabled(nextUndo >= 0);
+    undoAction.setEnabled(nextUndo > dontUndoPast);
     final Command undo = lastOutput.getUndoCommand();
     undo.execute();
     GameModule.getGameModule().getServer().sendToOthers(undo);
@@ -527,7 +537,7 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
         nextUndo = logOutput.size() - 1;
       }
     }
-    undoAction.setEnabled(nextUndo >= 0);
+    undoAction.setEnabled(nextUndo > dontUndoPast);
   }
 
   /**
--- a/vassal-app/src/main/java/VASSAL/build/module/GameState.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameState.java
@@ -213,6 +213,7 @@ public class GameState implements CommandEncoder {
         }
 
         setup(true);
+        GameModule.getGameModule().getGameState().freshenStartupGlobalKeyCommands(GameModule.getGameModule());
       }
     };
     // FIXME: setting mnemonic from first letter could cause collisions in
@@ -358,6 +359,22 @@ public class GameState implements CommandEncoder {
   //
   // END FIXME
   //
+
+
+  /**
+   * When we're known to be starting a fresh game from a Predefined Setup, freshen the starting-a-new-game flag for all SGKCs
+   * @param target the Game Module
+   */
+  public void freshenStartupGlobalKeyCommands(AbstractBuildable target) {
+    for (final Buildable b : target.getBuildables()) {
+      if (b instanceof StartupGlobalKeyCommand) {
+        ((StartupGlobalKeyCommand) b).freshGame();
+      }
+      else if (b instanceof AbstractBuildable) {
+        freshenStartupGlobalKeyCommands((AbstractBuildable) b);
+      }
+    }
+  }
 
   /**
    * Searches recursively through the game module for Startup Global Key Commands, applying them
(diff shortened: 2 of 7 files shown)


Cattlesquat:
Okay @BrentEaston my undo scheme is in place - check it out.

riverwanderer:
@Cattlesquat - I am very keen to test this set of changes. I have modules that could be improved by the new SGKC option for "once only" (or conversely, might be vulnerable to some unforeseen consequence or error).

uckelman:
@riverwanderer Try the build for this PR and let us know how it works for you.

riverwanderer:
> @riverwanderer Try the build for this PR and let us know how it works for you.

Yes; I will need help getting the build.

BrentEaston:
Go to https://vassalengine.org/builds/ and download the appropriate architecture build starting with 'VASSAL-3.6.0-SNAPSHOT-f945815-9890-' which are currently on the 2nd page.

riverwanderer:
The SGKC configuration panel appears to have some sort of internal tag in place of the "key command" field name.
image

riverwanderer:
I recall seeing a mention of this "Apply command" layer filter but it's not clear to me from the configuration panel as to what layer is being checked . Also this part isn't mentioned in the documentation yet.
image

riverwanderer:
I set the "When to Apply" field to "At the start of every fresh game only" and saved the module.
However, the field always gets reset back to the default of:
image

Cattlesquat:
Hi @riverwanderer -- thanks, looks like there was some kind of dumpster-fire in the way the underlying GKC structure is organized, and when I tried to add a field at the end of the list ... BOOM! So I have added the new field at the *beginning* of the list instead, which will put it at the top of the configurer but also seems to make it work without everything exploding.

So when a new "9890" build comes up in vassalengine.org/builds (should be something with today's date, and I expect it will be up from about 20 minutes from this message), please give it a try for me. DON'T use any SGKC that you wrote-to-disk with the old version, as it will remain corrupted. Instead create a new SGKC and try that for me. Thanks so much!

riverwanderer:
For an initial test, I set a V3.5 SGKC to "At Start of every fresh game only", and I cloned it to create another SGKC for each of the other two options.

In tests, "On First Game Launch / Load of Session" seemed to work. i.e. fired only on launch of the first game of a module session. "On Every Game Launch / Load of Session" also fired all the time, as expected.

The difficulty I had was with "At Start of Every Fresh Game Only". I was worried that it would not fire for pre-set scenarios but in fact this SGKC fires all the time; including for ordinary saved games and log files.

Cattlesquat:
@riverwanderer Were you by any chance loading *old* saved games? Because this mode uses a piece of game state that didn't previously exist, so if you loaded a game created before this build of VASSAL, it wouldn't have the flag and would initialize it to false.  On the other hand if you created a new saved game with THIS version of VASSAL, and loaded it, and the SGKC fired, then that is definitely a bug. Let me know which scenario you have?

riverwanderer:
> Were you by any chance loading _old_ saved games?
Yes. 

> On the other hand if you created a new saved game with THIS version of VASSAL
Hmmm.... any chance that the act of loading a game as a pre-defined scenario or running Refresh Pre-Defined Scenarios could set this new flag ?

Cattlesquat:
@riverwanderer Loading a game or pre-defined scenario would both run the SGKC and also set the flag (so it shouldn't fire in the future in a further load of that saved game). Refresh predefined scenarios ... I have no idea what that would do I don't know much about the game refresher and don't generally use it myself. But obviously starting from a pre-defined scenario would be a "new game" so it should fire the SGKC when you do it, after which the flag should be set, so that if you save THAT game and then later restore it, the SGKC should know not to fire.

riverwanderer:
@Cattlesquat - I'm running a refresh now, so I will let you know.  I think Refresh bypasses SGKCs somehow, based on what I've seen (which is good).

riverwanderer:
@Cattlesquat - just to clarify, of course, nothing using an old saved game will be able to use the "At Start of Every Fresh Game Only" effectively. Whether Refreshed, re-saved or re-installed as a Pre-Defined Setup.

I can't re-do 150+ scenario files, so cases like mine, this functionality won't be of use without the possiblity to convert old saved files.

Cattlesquat:
@riverwanderer The start-of-new-game version is _designed_ to fire once when a game is started fresh from a pre-defined scenario (since that's the beginning of a new game). So pre-defined scenarios should not need any redoing regardless of old-vs-new status, because it is _correct_ for them to fire once when a play session is launched from them. What would be incorrect is if they fired _again_ when the game was subsequently saved then loaded/restored afterwards during play.

riverwanderer:
> What would be incorrect is if they fired _again_ when the game was subsequently saved then loaded/restored afterwards during play.

@Cattlesquat - thanks for the explanation, which is overall what I hoped. However, I am seeing the start-of-new-game SGKC fire again, from a save game and from a log file.

riverwanderer:
This is the module I added the SGKCs too. There are two "fresh game" SGKCs. One that I just converted from the original v3.5.8 module and another that I added (at the end of the module) that just reports "fresh game": https://www.dropbox.com/s/zfiu5h0adjwx92n/CCNapoEpic3a63UPG.vmod?dl=0

I started from the first preset scenario in the module (Austerlitz), creating the 2 log files and a save game in this archive:
[Archive.zip](https://github.com/vassalengine/vassal/files/7480081/Archive.zip)
You can see that the "fresh game" message appears in both log files and the save game when they load.

Cattlesquat:
@riverwanderer Ah! Well it's super good I had you test -- the fact that your module used multiple SGKCs with different settings revealed the principal flaw: that they were stealing each others' "restore state" commands and needed to be given the "Unique ID Treatment". A few other things as well.

ANYWAY I *think* I have it working now -- I tested doing various things using your sample module and it seemed to all be working at least from my basic checking. So once a new build is up, please try putting it through its paces again! No need to recreate module or anything -- should now work from your current one. 

ALSO I made loading an old game (e.g. a game saved from before-this-feature) be treated as if it is NOT a fresh game. Since of course it isn't really a fresh game -- unless you're loading it as a _Scenario_ in which case it is ALWAYS treated as a fresh game.

riverwanderer:
So far so good, @Cattlesquat. I am just running a refresh counters that takes a while, then I can check that the presets still work as before (there's a flag that will get set if they are started normally; up to now, RC doesn't save them in that state, and I don't expect your change has affected that).

Meanwhile, I notice something that raises a question around log files... if a SGKC generates some output to chat, would you not expect to see that replayed in a log file ? At the moment, it looks to me like a SGKC leaves no trace in chat output of a log file. i.e when I log a demo with a "only once per game" SGKC that generates a report, but I don't see the report output when I re-play the log file. 

At the moment, I think that SGKC report output should always be showing in the log file, even when it means the same message might have already displayed from the current session (in the case of run per session and always). That is up to the designer to sort out.

Cattlesquat:
@riverwanderer Interesting question. The SGKC is part of the startup sequence of the new game, and is executed before the game even has a chance to e.g. ask you if you want to start a new log. Since a new log hasn't been started, it is also not part of the log -- sort of like when loading a Scenario you don't replay the scenario creation commands, nor can you "undo" them. Whether this is gooooooood, I haven't deeply thought about yet.

riverwanderer:
> The SGKC is part of the startup sequence of the new game, and is executed before the game even has a chance to e.g. ask you if you want to start a new log.
Oh yeah, of course. It is probably ok then, and not something to worry about. In the unlikely event that there is important information to share then the player would know they have to share it, because, as you say, they won't have opened the log.

riverwanderer:
Testing a conversion from v3.5.8, I notice that the module is getting assigned this default:
image

That isn't equivalent to the v3.5.8 SGKC is it?
Shouldn't it be "Always" ?

Cattlesquat:
"On First Game Launch / Load of Session" is equivalent to what pre-PR SGKCs do, which is why I made it the default.

riverwanderer:
@Cattlesquat  - Are you sure Brian? It's not been my interpretation of my experience. Plus my startup routine is behaving differently under v3.6 compared to v3.5 - in a way that suggests that it is not running an SGKC that did run before (I can follow this up in detail, but I'm betting on it working if I do a "First Game Launch" and not otherwise.)

Cattlesquat:
Oh are you find the old/original version behaves more like the on-ever-launch-of-session?

riverwanderer:
> Oh are you find the old/original version behaves more like the on-every-launch-of-session?

Yes, exactly.  And I just step-by-step tested under v3.5 / v3.6 and confirmed that.

riverwanderer:
Build VASSAL-3.6.0-5b11299-9890 works in all respects for these tests ("game opened" = .vsav or .vlog):
1. v3.5.8 SGKCs perform as prior to v3.6
2. The "On First Game/Launch of Session" fires the first time a game is opened or created since the module started.
3. The "Every game" SGKC always fires whenever a game is opened or created
3. PreSet Scenarios always fire the "Fresh Game" SGKC when opening.
4. A game loaded using Load Game or Log does NOT fire the "Fresh Game" SGKC.
5. SGKCs fire in the order listed in the module editor.

Looks good to me.

PR #10678, opened 2021-11-01
9759 - UNDO of movement will no longer deny lone stackable pieces their own stack (fixes some obscure stack-related bugs)

### LINKED ISSUE #9759: Undo causes standard counter to become unstacked
| Bugzilla |  |
|----------|----------|
| ID | BZ14078 |
| Reported | 2021-02-19 03:21:00 +0100 |
| Modified | 2021-05-06 08:18:26 +0200 |
| Product | VASSAL |
| Component | Player |
| Version | 3.5.1 |
| Platform | All |
| OS | All |
| Status | ASSIGNED |
| Resolution | None |
| Priority | unspecified |
| Severity | normal |
| Milestone | 3.5.2 |
| Assignee | [Brent Easton](mailto:b.easton@exemail.com.au) |
| CC | uckelman@nomic.net |

#### [Brent Easton](mailto:b.easton@exemail.com.au) 2021-02-19 03:26:02 +0100
<pre>See http://www.vassalengine.org/forum/viewtopic.php?f=3&t=12953 for details of repro module and log.

Symptom - Single Blue Donkeys in bottom right of board Have their Reproduce command grayed out because the Trigger expression
{SumStack("donkey")=~"1|2"}
is returning 0 because the piece is not in a stack.

Move one of these pieces and it gets a stack and command is now available.

Move the piece onto a stack of other blue Donkeys, then hit Undo. The command is now grayed out, Undo has caused the piece to Unstack.</pre>

#### [Brent Easton](mailto:b.easton@exemail.com.au) 2021-02-19 04:25:33 +0100
<pre>Has been an issue for ever. Reproduced on 3.2.

Only coming to light because SumStack(property) returns 0 for pieces not in a stack.</pre>

#### [Brent Easton](mailto:b.easton@exemail.com.au) 2021-05-06 08:18:26 +0200
<pre>This was worked around by modifying SumStack. This is still an issue.</pre>

Cattlesquat:
Fixes #9759 

On this one we should be a bit careful about whether we're solving a problem that needs to be solved. 

The problem is that the MovePiece command used to process undo (unlike PieceMover when processing drag-and-drops) does not place *lone* stackable pieces into a new stack when it doesn't find any existing piece/stack to merge with. 

The PR handles making sure lone stackable pieces are placed into stacks. So it's _probably_ benignly harmless. But you know, stacks and movement and undo -- OH MY! 

I'm not _sure_ what the story is with that getParent()-and-insert-at-0 thing going on in there -- it seems like something that would never actually run or matter in this case.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/command/MovePiece.java
+++ b/vassal-app/src/main/java/VASSAL/command/MovePiece.java
@@ -129,7 +129,16 @@ public class MovePiece extends Command {
         }
         else {
           if (newMap.apply(mergeFinder) == null) {
-            newMap.placeAt(piece, newPosition);
+            if (newMap.getStackMetrics().isStackingEnabled()
+                && !Boolean.TRUE.equals(piece.getProperty(Properties.NO_STACK))) {
+              final Stack s = new Stack();
+              s.add(piece);
+              GameModule.getGameModule().getGameState().addPiece(s);
+              newMap.placeAt(s, newPosition);
+            }
+            else {
+              newMap.placeAt(piece, newPosition);
+            }
           }
           if (piece.getParent() != null) {
             piece.getParent().insert(piece, 0);
--- a/vassal-app/src/main/java/VASSAL/command/MovePiece.java
+++ b/vassal-app/src/main/java/VASSAL/command/MovePiece.java
@@ -105,6 +105,22 @@ public class MovePiece extends Command {
     return playerId;
   }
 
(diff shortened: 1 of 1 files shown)


BrentEaston (approved):
For such a small change, this one is incredible scary. However, your implementation looks good to me. Again, it is an example of the generated Command not exactly emulating what the actual action does. Any Stackable piece that ends in a Stackable location by any means MUST be placed into a Stack. The case of moving back on to a existing Stack is already handled when the PieceMover sets up the Undo Command.

This definitely needs to be in beta for a few weeks.

PR #10679, opened 2021-11-01
9800 - Fixed bug where Place Marker, Send-to-Location, and Translate traits didn't always repaint maps

### LINKED ISSUE #9800: Windows not refreshing
| Bugzilla |  |
|----------|----------|
| ID | BZ14129 |
| Reported | 2021-02-24 00:28:00 +0100 |
| Modified | 2021-02-24 00:30:06 +0100 |
| Product | VASSAL |
| Component | Player |
| Version | 3.5.0 |
| Platform | Macintosh |
| OS | Windows |
| Status | NEW |
| Resolution | None |
| Priority | unspecified |
| Severity | normal |
| Milestone | --- |

#### [Brent Easton](mailto:b.easton@exemail.com.au) 2021-02-24 00:28:53 +0100
<pre>See http://www.vassalengine.org/forum/viewtopic.php?f=3&t=12914</pre>

#### [Brent Easton](mailto:b.easton@exemail.com.au) 2021-02-24 00:30:06 +0100
<pre>See also http://www.vassalengine.org/forum/viewtopic.php?f=6&t=12863#p66282</pre>

Cattlesquat:
Fixes #9800 

SOME situations were repainting the maps (e.g. if a piece was actually added to the map as opposed to a stack), but other situations were not (like merging a piece into an existing stack). 

I'm absolutely flabbergasted that this has existed for apparently forever.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
@@ -239,6 +239,8 @@ public class PlaceMarker extends Decorator implements TranslatablePiece, Recursi
         parent.insert(marker, index);
         c = c.append(ct.getChangeCommand());
       }
+
+      m.repaint();
     }
     else {
       c = m.placeAt(marker, p);
--- a/vassal-app/src/main/java/VASSAL/counters/SendToLocation.java
+++ b/vassal-app/src/main/java/VASSAL/counters/SendToLocation.java
@@ -422,6 +422,7 @@ public class SendToLocation extends Decorator implements TranslatablePiece {
     List<Point> offsets = null;
     Command c = null;
     Point dest = null;
+    final Map oldMap = outer.getMap();
 
     // If we're about to move a Mat, establish the initial relative positions of all its "contents"
     if (GameModule.getGameModule().isMatSupport()) {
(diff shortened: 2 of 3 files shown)


uckelman:
I think this is a good illustration of why it's a better design to have a data model that you update and a renderer that just renders, so this kind of thing can't happen.

Cattlesquat:
@BrentEaston Joel pointed out that another possibility would be having Stacks repaint their maps whenever they change (similar to the way maps repaint whenever a new piece is added to them). But I'm uncertain where to reliably do that and about what any potential downsides would be.

BrentEaston:
Just looking at Send To Location and I'm wondering about the difference between it being executed by the local player, as opposed to a remote player or log replayer who essentially just see the execution of an AddPiece or MovePiece command.

AddPiece and MovePiece both appear to repaint the targetMap, so I am guessing they are not a problem, it is just the trait execution not actually doing what the generate commands do.

Looking at Stack, it is interesting that remove() does a getMap().repaint(), but add and insert do not. Possibly an oversight that should be fixed HOWEVER, this won't help you if you do  SendToLocation on an 'Does Not Stack' piece.

So I think you need to go with what you have now and explicitly replaint in the traits at the appropriate point.

BrentEaston (approved):

PR #10686, opened 2021-11-03
Deck fill out the rest of the commands with hotkeys and report formats

Cattlesquat:
Additional fields for deck:
* hotkeys to flip deck, turn it explicitly face up, turn it explicitly face down
* Different text field for "flip face down" from "flip face up"
* Menu text, hotkey, report format for "Save Deck"
* Menu text, hokey, report format for "Load Deck"


vassal-app/src/main/java/VASSAL/build/module/map/DrawPile.java
```diff
@@ -751,6 +824,51 @@ else if (FACE_UP_MESSAGE.equals(key)) {
     else if (FACE_DOWN_MESSAGE.equals(key)) {
       dummy.setFaceDownMessage((String)value);
     }
+    else if (FACE_UP_HOTKEY.equals(key)) {
+      if (value instanceof String) {
+        value = NamedHotKeyConfigurer.decode((String) value);
+      }
+      dummy.setFaceUpKey((NamedKeyStroke) value);
```
uckelman, line 831: What's with all the assigning to `dummy`?

Cattlesquat: That's just the way the Deck / Draw Pile unholy sandwich works. DrawPile is a game component and holds the fields and saves/loads them. Deck is the runtime element that's "sort of a GamePiece" and "sort of a Stack" and stuff.

vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
```diff
@@ -138,6 +139,13 @@ public static JPopupMenu createPopup(GamePiece target, boolean global) {
       // Maps name to a list of commands with that name
       final HashMap<String, ArrayList<JMenuItem>> commandNames = new HashMap<>();
 
+      if (target instanceof Deck) {
+        if (c.length == 1) {
+          c[0].actionPerformed(new ActionEvent(popup, 0, ""));
+          return null;
+        }
+      }
+
```
BrentEaston, line 148: What's this doing?

It looks like it Is automatically executing the menu option if there is only only one option in the menu of a Deck

BrentEaston (commented):
Looks good overall, Just that query about the one-option menu handling

vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
Cattlesquat: Oh, yeah, that's what it does. Forgot about that. Seemed like "the right thing to do". Should I make it an option instead?

BrentEaston: If you right-click on a Deck, I think the expectation is that you will be displayed a list of options to choose from, eve if it only one. What if you don't want to select that option? You can't get out of it, it just magically gets done.

Cattlesquat: Ah, right good point. I was thinking of the "Draw Multiple" and "Draw Selected" kinds of things where it goes to another choose-or-cancel box anyway (and it's annoying/cumbersome to constantly have to pick from the menu of one thing and THEN choose-or-cancel). I have now restricted it to just those two possibilities.

uckelman:
@BrentEaston Retest? Is this ok now?

BrentEaston (approved):

PR #10691, opened 2021-11-03
10691 - Mouseover Stack Viewer - Unstatic a naughty field in CounterDetailViewer, and add ability to count face down cards in a deck

Cattlesquat:
@uckelman Note this "add fields to deck" -- one rescued from staticness, and the other added.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
@@ -604,7 +604,7 @@ public class CounterDetailViewer extends AbstractConfigurable implements Drawabl
     final GamePiece[] allPieces = map.getPieces(); // All pieces from bottom up
 
     final Visitor visitor = new Visitor(new Filter(), map,
-      map.componentToMap(currentMousePosition.getPoint()), showOverlap, this);
+      map.componentToMap(currentMousePosition.getPoint()), showOverlap, this.showNumberFromDeck, this.showDeckMasked);
     final DeckVisitorDispatcher dispatcher = new DeckVisitorDispatcher(visitor);
 
     /*
@@ -730,9 +730,15 @@ public class CounterDetailViewer extends AbstractConfigurable implements Drawabl
     protected int insertPos = 0;
     protected Point foundPieceAt;
     protected boolean showingOverlap;
-    protected CounterDetailViewer cdv;
+    protected final int showNumberFromDeck;
+    protected final boolean showDeckMasked;
 
-    public Visitor(Filter filter, Map map, Point pt, boolean showOverlap, CounterDetailViewer counterDetailViewer) {
+    @Deprecated
+    public Visitor(Filter filter, Map map, Point pt, boolean showOverlap) {
+      this(filter, map, pt, showOverlap, 1, false);
+    }
+
+    public Visitor(Filter filter, Map map, Point pt, boolean showOverlap, int showNumberFromDeck, boolean showDeckMasked) {
       super(map, pt);
       if (map.getPieceCollection() instanceof CompoundPieceCollection) {
         collection = (CompoundPieceCollection) map.getPieceCollection();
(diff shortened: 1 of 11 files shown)


vassal-app/src/test/resources/clirr-ignored-differences.xml
```diff
@@ -1265,4 +1265,15 @@
...
+        <className>VASSAL/build/module/map/CounterDetailViewer</className>
+        <differenceType>6007</differenceType>
+        <field>showDeckDepth</field>
+    </difference>
+    <difference>
+        <className>VASSAL/build/module/map/CounterDetailViewer$Visitor</className>
+        <differenceType>7004</differenceType>
+        <method>**</method>
+        <justification>Nothing external should use this</justification>
+    </difference>
```
uckelman, line 1278: I should run the dependency checker to see if this will blow up any modules. (I suspect it could blow up VASL.)

Cattlesquat: Yeah worth checking. I could make a parallel one if so.

uckelman: This will break VASL, as it has a class extending CounterDetailViewer.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -722,15 +730,17 @@ else if (displayWhat.equals(EXC_LAYERS)) {
     protected int insertPos = 0;
     protected Point foundPieceAt;
     protected boolean showingOverlap;
+    protected CounterDetailViewer cdv;
 
-    public Visitor(Filter filter, Map map, Point pt, boolean showOverlap) {
+    public Visitor(Filter filter, Map map, Point pt, boolean showOverlap, CounterDetailViewer counterDetailViewer) {
```
uckelman, line 735: The `Visitor` doesn't need the CDG, just the values of two of its fields---and it needs those values only at construction time, not any time later. So, rather than break binary compatibility by changing `Visitor`'s only ctor, we can do the following:

* Add two `protected final boolean` fields to `Visitor`, for `showNumberFromDeck` and `showDeckMasked`.
* Deprecate the old ctor, and in it set those two fields to whatever the defaults in the CDV are for them.
* Make a new ctor that takes values for those two fields as arguments, and use that ctor below in `getDisplayablePieces()`.
* Remove the clirr exception for the `Visitor` ctor, because we don't need it anymore.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
Cattlesquat: Okay pushed that (hopefully)

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -596,7 +604,7 @@ protected double getZoom() {
     final GamePiece[] allPieces = map.getPieces(); // All pieces from bottom up
 
     final Visitor visitor = new Visitor(new Filter(), map,
-      map.componentToMap(currentMousePosition.getPoint()), showOverlap);
+      map.componentToMap(currentMousePosition.getPoint()), showOverlap, this.showNumberFromDeck, this.showDeckMasked);
```
uckelman, line 607: You don't need the `this`es here, but it's fine.

vassal-app/src/main/java/VASSAL/build/module/map/CounterDetailViewer.java
```diff
@@ -722,17 +730,27 @@ else if (displayWhat.equals(EXC_LAYERS)) {
     protected int insertPos = 0;
     protected Point foundPieceAt;
     protected boolean showingOverlap;
+    protected final int showNumberFromDeck;
+    protected final boolean showDeckMasked;
 
+    @Deprecated
     public Visitor(Filter filter, Map map, Point pt, boolean showOverlap) {
+      this(filter, map, pt, showOverlap, 1, false);
```
uckelman, line 738: :+1:

uckelman (approved):

PR #10696, opened 2021-11-05
9981 - Global Key Commands - "Within a Deck, select Fixed number of pieces" field now supports expressions

Cattlesquat:
image


vassal-app/src/main/java/VASSAL/build/module/map/MassKeyCommand.java
```diff
@@ -365,16 +370,18 @@ public DeckPolicyConfig(boolean showPrompt) {
...
+      controls2.add(piecesConfig.getControls());
 
       final PropertyChangeListener l = evt -> {
-        intConfig.getControls().setVisible(FIXED.equals(typeConfig.getValueString()));
-        SwingUtils.repack(intConfig.getControls());
+        piecesConfig.getControls().setVisible(FIXED.equals(typeConfig.getValueString()));
+        SwingUtils.repack(piecesConfig.getControls());
+      };
+      final PropertyChangeListener l2 = evt -> {
+        setValue(getSingleValue());
```
uckelman, line 380: Can be a one-liner without braces.

PR #10699, opened 2021-11-05
6476 Repositioning Stacks/Decks optionally draws ghosted images of other stacks/decks

Cattlesquat:
* You have to turn it on with a checkbox (because it's a bit slow/sluggish) 
* -Possibly- it could be optimized and thus run faster? Also possibly that's not worth the trouble. 
* -Certainly- @BrentEaston could make the configurer layout look better

Even in sluggish mode it is something I will often want to use

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
@@ -1127,7 +1127,7 @@ public class SetupStack extends AbstractConfigurable implements GameComponent, U
     public void paint(Graphics g) {
       // Since this configurer is non-modal, there is always the possibility that user is switching back and forth.
       // In which case they get to eat the not-inconsequential perf hit of re-whatevering all the configurers.
-      if (!myBoard.getName().equals(getCachedBoard())) {
+      if (isShowOthers() && !myBoard.getName().equals(getCachedBoard())) {
         SetupStack.setCachedBoard(myBoard.getName());
         SetupStack.setUsedBoardWildcard(myBoard.getName());
         for (final SetupStack s : otherStacks) {
--- a/vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
@@ -1119,9 +1119,9 @@ public class SetupStack extends AbstractConfigurable implements GameComponent, U
       final int x = (int)(s.pos.x * os_scale);
       final int y = (int)(s.pos.y * os_scale);
 
-      final Rectangle bb = s.stackConfigurer.getCachedBoundingBox();
-      bb.x = x;
-      bb.y = y;
+      final Rectangle bb = new Rectangle(s.stackConfigurer.getCachedBoundingBox());
+      bb.x += x;
+      bb.y += y;
 
       if (vrect.intersects(bb)) {
         s.stackConfigurer.drawImage(g, x, y, null, os_scale);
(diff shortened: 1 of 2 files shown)


BrentEaston (changes requested):
I am working on the Dialog layout, but one thing I noticed is that if you have a Deck or Stack on Board 'Any', then you don't see any Decks or Stacks that are defined on specific boards.

I think an 'Any' Stack should see all other Decks/Stacks regardless of board. And a board specific deck/stack should see 'Any' stack/boards.

Cattlesquat:
@BrentEaston Okay, good thought. I was originally protecting from e.g. seeing two different conflicting versions of a piece at the same time (for instance my PoG armies are "any" since they go at e.g. "Paris" on either board, but the trench markers are hand-adjusted at each individual board to pixel locations, so obviously you don't want to see two different versions of the Paris trench at one time). But I have now made more sophisticated wildcard matching for the "any" -- so you'll only get "one board worth of stuff" but you'll get everything that in any way matches the board that originally got picked (by your master stack that you're currently configuring) to be the the background of the configuration session. 

TL;DR - now it works the way you want ;)

BrentEaston:
@Cattlesquat When I try and reposition a Deck on board [Any], I am still not seeing The Decks that are defined on board Main and Main2.

[Any] board only really makes sense in a situation where the multiple boards are different versions of the same board that will exist in the same space and are basically different artwork of the same board. [Any] doesn't really make sense in a multiple board scenario where it just gets dumped into the top left board.

So I sort of think configuring an [Any] stack should show you everything defined on all other boards, regardless of duplicates.

Or is there another option or two that might make more sense here? 

When moving a Board specific Deck/Stack: 'Show [Any] Decks/Stacks?'
When moving an [Any] stack: Show Deck/Stacks from board None/All/Drop-down list of boards?

OR, is it possible to change the select board to display on-the-fly for [Any] stacks and show the board specific stacks (and othe [Any] stacks) for that board.

Am I overthinking this?

vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
```diff
@@ -138,6 +140,26 @@
   public static final String LOCATION = "location"; //NON-NLS
   public static final String USE_GRID_LOCATION = "useGridLocation"; //NON-NLS
 
+  public static boolean showOthers = false; // Whether we also show other decks/stacks when changing the position of one
```
uckelman, line 143: This is just for the editor, right? Why is this a member of `SetupStack` rather than being a non-static member of some the editor class?

Cattlesquat:
Followup from Brent in dev-chat was -- "Go with the version you have, The main issue is showing all other stack on the same board and that is solved. If a designer comes up with real [Any]  issues, we can tweak it then."

vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
Cattlesquat: It is just for the editor, correct. This situation *only* happens when configuring a SetupStack, and so it didn't seem appropriate to put it somewhere broader?

Cattlesquat: Oh and it's static because I'm in one StackConfigurer (inner class of SetupStack) and I need to tell *all the others that exist* that they need to behave themselves. 

I guess you're saying I could tell e.g. GameState that we're in the middle of a configurer draw and so therefore behave differently. And then read that from a non-static member/method of GameState. It seemed cleaner to me to keep all the code in the same file, but I can do that if you'd prefer.

BrentEaston (approved):

uckelman (approved):

PR #10700, opened 2021-11-05
5489 - Improved bug dialog text

Cattlesquat:
(tried to describe what we wanted a bit better)


BrentEaston (approved):

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
```diff
@@ -119,15 +119,15 @@ BadDataReport.enable_pref=Turn on the Audit Trail preference to generate more de
 #
 # FIXME: title is wrong for bugs not arriving via the ExceptionHandler
 BugDialog.title=Uncaught Exception
-BugDialog.heading=Eeek! A bug!
-BugDialog.message=VASSAL had an internal error
+BugDialog.heading=Help us fix this ghastly bug!
+BugDialog.message=<html>VASSAL had an internal error -- <i>YOU can help!</i></html>
```
uckelman, line 123: We should use proper en- and em-dashes for things like this. `&mdash;`, in this case.

Cattlesquat: fixed

PR #10717, opened 2021-11-07
Deck access - allow access to decks to be restricted by player

Cattlesquat:
Initial structure and configurers all put together - it just doesn't seem to be restricting access correctly


uckelman (approved):

uckelman:
This seems fine, but @BrentEaston should also have a look.

BrentEaston (approved):

PR #10719, opened 2021-11-08
9063 - Deck Refresher

Cattlesquat:
This supports basic deck refreshing:
(1) updates decks in the game to latest settings in module
(2) Optionally deletes decks from the game if the deck has been deleted from the module (any current contents of the deck will be left in a stack at the deck's former position)
(3) Optionally adds decks to the game if they have been added to the module. Does NOT add any contents to the deck, whether or not the module defines starting contents -- an empty deck will be added at the location. Module designer will have to work around adding contents by other means (e.g. dragging them in)

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -1723,11 +1723,20 @@ public class GameModule extends AbstractConfigurable
    * @return true if the game is online or has ever had more than one player.
    */
   public boolean isMultiPlayer() {
-    final ServerConnection sv = getServer();
-    if ((sv != null) && sv.isConnected()) return true;
+    if (isMultiplayerConnected()) {
+      return true;
+    }
 
     final PlayerRoster pr = getPlayerRoster();
     return ((pr != null) && pr.isMultiPlayer());
+  }
+
+  /**
+   * @return true if the game is currently online
+   */
+  public boolean isMultiplayerConnected() {
+    final ServerConnection sv = getServer();
+    return (sv != null) && sv.isConnected();
   }
 
   /**
--- a/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
@@ -487,11 +487,11 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
 
   /**
    * Mark the current command as a no-undo-past-this-point.
+   *
+   * @param block number extra command(s) into the future to block (when this is being executed "from inside a command")
    */
-  public void blockUndo() {
-    // +2 because this is done by a command that is about to be added to the
-    // list, and there's _also_ a sentinal command added after that
-    dontUndoPast = nextUndo + 2;
+  public void blockUndo(int block) {
+    dontUndoPast = nextUndo + block;
     undoAction.setEnabled(false);
   }
 
--- a/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
@@ -280,6 +280,13 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
   }
 
 
+  private boolean isGameActive() {
+    final GameModule gm = GameModule.getGameModule();
+    final VASSAL.command.Logger logger = gm.getLogger();
+    return gm.isMultiplayerConnected() || ((logger instanceof BasicLogger) && ((BasicLogger)logger).isLogging());
+  }
+
+
   /**
    * This method is used by PredefinedSetup.refresh() to update a PredefinedSetup in a GameModule
    * The default execute() method calls: GameModule.getGameModule().getGameState().getAllPieces()
@@ -395,9 +402,7 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
      * 4/ Refresh properties of decks in the game
      */
     if (options.contains("RefreshDecks")) { //NON-NLS
-      final GameModule gm = GameModule.getGameModule();
-      final VASSAL.command.Logger logger = gm.getLogger();
-      if (gm.isMultiplayerConnected() || ((logger instanceof BasicLogger) && ((BasicLogger)logger).isLogging())) {
+      if (isGameActive()) {
         // If somebody feels like packaging all these things into Commands, help yourself...
         log(Resources.getString("GameRefresher.deck_refresh_during_multiplayer"));
       }
@@ -430,6 +435,12 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
                   // If drawPile is owned by a specific board, then we can only match it if that board is active in this game
                   if (drawPile.getOwningBoardName() != null) {
                     if (map.getBoardByName(drawPile.getOwningBoardName()) == null) {
+                      continue;
+                    }
+
+                    // If the drawPile is on a map that doesn't have its current owning board active, then we
+                    // cannot match that drawPile.
+                    if (drawPile.getMap().getBoardByName(drawPile.getOwningBoardName()) == null) {
                       continue;
                     }
                   }
@@ -788,6 +799,16 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
       addNewDecks = new JCheckBox(Resources.getString("GameRefresher.add_new_decks"));
       addNewDecks.setSelected(false);
       panel.add(addNewDecks);
+
+      if (refresher.isGameActive()) {
+        refreshDecks.setSelected(false);
+        refreshDecks.setEnabled(false);
+        deleteOldDecks.setEnabled(false);
+        addNewDecks.setEnabled(false);
+
+        final FlowLabel nope = new FlowLabel(Resources.getString("GameRefresher.but_game_is_active"));
+        panel.add(nope);
+      }
 
       panel.add(buttonPanel, "grow"); // NON-NLS
 
--- a/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
@@ -889,7 +889,7 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
       g.sendAndLog(command);
 
       if (options.contains("RefreshDecks") && !refresher.isGameActive()) {
-        VASSAL.command.Logger log = GameModule.getGameModule().getLogger();
+        final VASSAL.command.Logger log = GameModule.getGameModule().getLogger();
         if (log instanceof BasicLogger) {
           ((BasicLogger)log).blockUndo(1);
         }
--- a/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
@@ -426,6 +426,14 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
               for (final DrawPile drawPile : drawPiles) {
                 final String deckName = deck.getDeckName();
                 if (deckName.equals(drawPile.getAttributeValueString(SetupStack.NAME))) {
+
+                  // If drawPile is owned by a specific board, then we can only match it if that board is active in this game
+                  if (drawPile.getOwningBoardName() != null) {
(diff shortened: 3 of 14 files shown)


Cattlesquat:
@riverwanderer Have at it once an auto-build for this goes up at vassalengine.org/builds. Remember, it's either a Deck Refresher OR a Wreck Defresher, and only testing will determine which...

riverwanderer:
Edging to the "wreck" side at the moment. Tested with C&C Napoleonics v3.42, a straightforward module.  Updating parameters seemed to work (tried a deck command, report and drag and drop filter). However, even though 0 added / deleted decks were done, all the decks on the main map got transposed to the left.  I wonder if this is something to do with the decks being on one board (right half of main map) and the Refresher failing to recognise that? you can see two decks (the counters) and the faint outline of the main decks ... all over the main board where they should not be.
image

riverwanderer:
interesting... the above was from the PDS Refresher (the big one).  I refreshed that corrupted scenario using the one-off tool and the decks all got relocated back to the correct spots.

Cattlesquat:
@riverwanderer I've pushed some new code that might fix some of the "board awareness" issues, but may not be out of the woods yet.

riverwanderer:
Tested snapshot f7ae39f; seeing different behaviour now.
1. The decks remain in the right place.

2. An individual refresh is always failing to match two decks. A test change (on another deck) does get applied. This output is generated:
```The following decks were not refreshed because no matching deck was found in the module (turn on delete-old-decks option to remove them)
- Allies shuffle
- French shuffle
- The following decks exist in the module but not this game/save. (Turn on add-new-decks option to add them)
- Allies shuffle
- French shuffle
- ----------
- - 12 Decks refreshed
- - 2 Decks deletable (exist in current game but not in module)
- - 2 Decks addable (exist in module but not in current game)
```
**The two decks are notable because they have a negative offset from the window/board.**

3. A mass refresh generates this output, failing to match decks in the module for every PreSet file in the module. The test changes are not applied during this refresh run: 
```- REFRESHING DECKS
- ----------
- The following decks were not refreshed because no matching deck was found in the module (turn on delete-old-decks option to remove them)
- Draw Deck
etc....
- ----------
- - 0 Decks refreshed
- - 14 Decks deletable (exist in current game but not in module)
- - 0 Decks addable (exist in module but not in current game)
```

Cattlesquat:
@riverwanderer Okay that was a useful stage to go through. I have made what I think is a more comprehensive fix (and I'm pretty sure I figured out why the full saved game refresher was producing different results). New version *seems* to work in both modes -- give it a try and see what you think?

riverwanderer:
Test with C&C Nap v3.42; modified the main discard deck and performed a single refresh.
Property changes tried that worked; Menu entries, reporting and drag and drop filter.
One change did not work - the Hotkey for Send to Deck. I changed this and it showed in the Menu with the new Command key (see below). The command still works from the menu but not as the listed hotkey. Instead the old hotkey still did the action.
image

riverwanderer:
....ah, but it works with a full Pre-Defined Scenarios refresh.  I will test saving a re-opening a single refreshed scenario.

Cattlesquat:
Yeah I bet it will work when you save and re-open. Sounds like I just need to refresh the key command stuff on the fly when I refresh the deck in an active game.

Cattlesquat:
@riverwanderer I have pushed a new version (23570c3) in which I update the key listeners after the deck refresh, which I believe will fix the problem you found. Give it a try when the build appears (in 15-20 minutes)

riverwanderer:
Re-tested single refressh using 3.6.0-b4c400b , with C&N Nap v3.42 ; new hotkey worked, as predicted by @Cattlesquat.

Tested relocating decks - worked on the same map i.e. just updating the deck location. A test of cut/paste of the deck to a different map partly works; send to deck moves the card to the new location. However, at the location, the card behaves as if just stacked i.e. the standalone card commands appear rather than the deck commands.

Cattlesquat:
@riverwanderer Can you give me an example of how I'd test/recreate that with your module? (e.g. what deck to cut/paste, where to find a return-to-deck trait that returns it almost-but-not-quite)

Cattlesquat:
@riverwanderer Also, when you're doing the cut/paste ... is it treating that as just a "refresh of the deck", or is it treating it as "delete one deck, add a new deck of the same name on the other map"?

riverwanderer:
@Cattlesquat  .... I've now tested moving decks by cut/paste & renaming the decks (and updating all references to the deck names). After a mass refresh, the decks were working.

How to test:
1. Use Commands & Colors: Napoleonics module v3.42 (not the latest version, just to keep things simple)
2. There are two types of card "Command" and "Tactic", so called. The Send to Deck traits for each are in prototypes. For the purpose of testing, use the trait in the "Cards" prototype.
3. All cards are in decks on the "HiddenDeckStorage" map. On this map sits the source test deck "CommandCardsOld"
4. Back on the main window, a blue button invites the user to select a deck. When the "70 Command Card" deck is selected, the "CommandCardsOld" deck is moved by a Send to Deck attribute to the "Draw Deck" deck on this main window.
5. There is a "Discard Pile" next to the Draw Deck where the deck of 70 blue backed cards will have appeared.
6. To test, cut/paste the Draw Deck and Discard Pile from the Main Map to "Special Window". Access the special window using the button "Special" in the toolbar.

Results:-
a) A simple cut/paste followed by a Refresh will result cards locating on the Special Window but the Draw Deck cards are face up, and, when the discard pile forms, cards do not show the Deck commands or respond to the reshuffle hotkey (Ctrl+Alt+Shift+L).

b) If the decks are renamed after cut/paste, and all references updated (Send to Deck and Return to Deck traits), then the decks behave properly after a refresh.

riverwanderer:
> Also, when you're doing the cut/paste ... is it treating that as just a "refresh of the deck", or is it treating it as "delete one deck, add a new deck of the same name on the other map"?

I tried both. I don't think it made any difference.

When I also renamed the decks, then of course, I used delete / add and that worked.

Cattlesquat:
@riverwanderer Okay that makes sense -- it works right when it understands it's deleting it one deck and adding it another deck on a different map, but doesn't understand if a single deck is "just" changing maps. I have put in some code to account for change-of-maps during the refresh of a single deck -- give it a try when the build for cc6927d goes up.

riverwanderer:
Change works for single and mass PDS refreshes. However, even though the refresh seems succesful, I notice the decks being flagged up as deleted:
```
- The following decks were not refreshed because no matching deck was found in the module (turn on delete-old-decks option to remove them)
- Draw Deck
- Discard Pile
- ----------
- - 14 Decks refreshed
- - 2 Decks deletable (exist in current game but not in module)
- - 0 Decks addable (exist in module but not in current game)
```


Also, if I refresh the PDS files, save and close the module, and re-open it... a random scenario looks fine (decks in the new locations, working) but another Refresh produces a log for a deleted deck / new deck situation:
```
- The following decks exist in the module but not this game/save. (Turn on add-new-decks option to add them)
- Draw Deck
- Discard Pile
- ----------
- - 12 Decks refreshed
- - 2 Decks deletable (exist in current game but not in module)
- - 2 Decks addable (exist in module but not in current game)

```

Cattlesquat:
@riverwanderer I assume the "Draw Deck" and "Discard Pile" that you cut-and-pasted are the ones it is showing as deletable/addable?

riverwanderer:
> I assume the "Draw Deck" and "Discard Pile" that you cut-and-pasted are the ones it is showing as deletable/addable?

@Cattlesquat  - Yes.

BrentEaston:
On the whole, looking good, but I am not seeing any Commands being generated to transmit the Deck changes to other clients.

Possibly related, but in my simple module, I refresh a scenario and a deck is updated. If I then hit Undo, the card in the Deck vanishes.

Cattlesquat:
The Deck refresher doesn't run if there is either an active connection or a log being logged. Supporting "live" use of Deck Refresh seems bizzare/unnecessary -- its purpose is to fix predefined scenarios. If there are one-offs needed they can be dealt with offline. 

I will need to look at the undo situation though. Undo must have been of some move you made before running refresher? Again the kind of thing where ... why on earth would we spend time supporting that? (Undo past a counter refresh?!) We do have brand new tool to block undo...

BrentEaston:
The existing Game Refresher does generate and transmit commands correctly and can be undone. Since a player can only refresh counters owned or visible to them, this allows to players to fully refresh a game and transmit the results of each partial refresh to each other.

I have no problem with the Deck refresh being off-line only, just need to make it clear in the doco. Deck refresh is more of a 'single-user' thing.

The current source doesn't have Joel's undo fix in it, so I had to move a counter, then I refreshed, then I hit undo and the piece in my Deck disappeared.

Cattlesquat:
Have pushed the rebase (to pick up the undo fix). 

Meanwhile on @riverwanderer 's deck cut-and-pasting, one thing that's going on is the Deck being cut has an explicit owning board ("Side Board") but is being pasted to a map that has only a "Special Board" and that field has not been updated. The Editor is pernicious on this because if you then open the deck configurer it *looks* like it has been changed to "any" but that is just how the pulldown _displays_ when it is not set to something it recognizes as a valid option, so unless you go do the counterintuitive "change it to something else, then change it back", the deck is set to a board that doesn't own it, and so doesn't respond to commands. 

I'll have to look further to think of what makes sense here.

Cattlesquat:
@riverwanderer @BrentEaston Okay I have pushed the following additional fixes:
(1) Refresher will not "match" a Deck to a Drawpile (module deck) that is owned by a board not currently active on the Drawpile's map. This probably won't *happen* any more (essentially implies bad module data, see below), but if it did happen it would essentially be treating the old deck as a "deletable" (draw pile removed from the map where it was, and no new matchable drawpile). It wouldn't try to "add" it on the other map, because the drawpile wouldn't be active on the other map, because the drawpile says it's owned by a board that the map doesn't have.
(2) If someone has already heretofore cut-and-paste decks (or Setup Stacks) and thus has them pointing at non-existent-for-their-current-map owning boards, then at least when they bring up the configurer it will now show them the real setting, and will make it easier to change it (no longer shows "any" when it's not set to any ... so now you can set it to any and have it work)
(3) (The most important fix) - When a Deck or SetupStack is "pasted" to a new location in the editor, if their marked owning board doesn't exist on their new map, their owning board is automatically changed to "any". You get a little warning in the chat box when that happens. **TL;DR** - if you do the cut-and-paste you were doing before, and then run the refresh, it should be able to work. (Assuming there isn't some _other_ heretofore undiscovered bug) 
(4) I also visibly disabled the Refresh Decks checkboxes if a multiplayer game or log is running, and gave an explanatory message.

image

BrentEaston:
Issue 1: Start a new game, Observer, Main Map. Select tools -> Refresh, DO NOT select Refresh Decks. Refresh. You now have an undo option. Undo. All Deck pieces disappear.

Issue 2: Start a new game, Observer, Main Map. Select tools -> Refresh, DO select Refresh Decks. Refresh. You now have an undo option. Undo. Eek, its a bug.

[BattleforMoscow.zip](https://github.com/vassalengine/vassal/files/7515856/BattleforMoscow.zip)

Cattlesquat:
@BrentEaston 
* Issue 1 is in Beta 5 branch and prior (regular Game Refresh doesn't undo)
* Issue 2 - I wasn't calling BlockUndo at the right place, now I am -- so won't be able to undo back over a Deck Refresh to find out what would happen ;)

riverwanderer:
Retested as follows (using C&C Nap v3.42), I believe successfully:-

1) cut/paste Draw Deck and Discard Pile decks: got the warning message about non-existent board, as expected.

2) Changed some properties

3) BEFORE Refresh, tried a GKC that Sends from another Deck to the now moved Draw Deck. This resulted in an odd result; the cards were moved to the NEW map location and possibly the correct Y coordinate for the deck but a much larger X coordinate. The cards turned faced up as might be expected as they were not in a deck anymore. Perhaps this is just how things are when the module is fundamentally in a state that doesn't match the game file ?

4) UnDo put the cards back to their starting position.

5) REFRESH generated normal messaging (no new / removed decks). GKC to Send cards to the Draw Deck now worked and the new deck properties were in place.

6) UnDo stopped at the refresh deck point, as expected.

Cattlesquat:
@riverwanderer I think you're right about (3) when a desk mismatches its drawpile (or drawpile is missing entirely) then zany effects are possible/likely. The _usual_ Deck/Drawpile things are the milder form of "I edited a few properties but they don't update in my saved game". But actually deleting/removing/renaming the definition entirely ... beware! :)

PR #10721, opened 2021-11-08
dontUndoPast is the last item which can be undone; undoing it is ok

uckelman:


uckelman:
What's here seems not to correctly stop undoing of SGKCs.

Cattlesquat:
This works fine for me.

PR #10727, opened 2021-11-10
Module "Debug" Window

Cattlesquat:
Starting to work on a little doodad, with the initial main goal of being able to find out X/Y positions easily. 

Removed default hotkey but you can add one in GlobalOptions.
Meanwhile it now appears on the Tools menu for the Player.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/DebugControls.java
+++ b/vassal-app/src/main/java/VASSAL/tools/DebugControls.java
@@ -1,0 +1,220 @@
+/*
+ *
+ * Copyright (c) 2000-2013 by Rodney Kinney, Brent Easton
+ *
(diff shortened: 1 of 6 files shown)


vassal-app/src/main/java/VASSAL/tools/DebugControls.java
```diff
@@ -0,0 +1,293 @@
...
+import java.awt.Component;
+import java.awt.Container;
+import java.awt.Point;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.net.URL;
+import java.util.List;
+
+public class DebugControls extends AbstractBuildable implements ActionListener {
+  protected static long MEGABYTE = 1024 * 1024;
```
uckelman, line 57: `final`?

PR #10731, opened 2021-11-11
9635 - Add description fields to most important editor components

Cattlesquat:
#9635

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/Widget.java
+++ b/vassal-app/src/main/java/VASSAL/build/Widget.java
@@ -17,13 +17,13 @@
  */
 package VASSAL.build;
 
-import VASSAL.configure.ComponentDescription;
 import org.w3c.dom.Attr;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NamedNodeMap;
 
 import VASSAL.build.module.documentation.HelpFile;
+import VASSAL.configure.ComponentDescription;
 import VASSAL.i18n.Localization;
 
 /**
--- a/vassal-app/src/main/java/VASSAL/build/Widget.java
+++ b/vassal-app/src/main/java/VASSAL/build/Widget.java
@@ -17,6 +17,7 @@
  */
 package VASSAL.build;
 
+import VASSAL.configure.ComponentDescription;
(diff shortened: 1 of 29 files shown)


vassal-app/src/main/java/VASSAL/build/Widget.java
```diff
@@ -17,6 +17,7 @@
  */
 package VASSAL.build;
 
+import VASSAL.configure.ComponentDescription;
```
uckelman, line 20: Group with the other `VASSAL` imports.

vassal-app/src/main/java/VASSAL/build/module/turn/TurnComponent.java
```diff
@@ -26,13 +26,22 @@
...
-public class TurnComponent extends AbstractConfigurable {
+public class TurnComponent extends AbstractConfigurable implements ComponentDescription {
+  public static final String DESCRIPTION = "description"; //NON-NLS
+
+  protected String description;
+
+  @Override
+  public String getDescription() {
+    return description;
+  }
```
uckelman, line 44: What sets it?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/turn/TurnComponent.java
Cattlesquat: It gets set by the setAttribute()

vassal-app/src/main/java/VASSAL/build/Widget.java
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/build/module/turn/TurnComponent.java
Cattlesquat: (and I should explain that TurnComponent, even though it isn't declared abstract, is ONLY overridden by the various fleshed-out kinds of turn component)

uckelman: The horror, the horror. Ok, np.

uckelman (approved):

PR #10749, opened 2021-11-13
10749 - Move Camera Button can also change Zoom

Cattlesquat:
Ability to zoom the camera when moving it was immediately requested in forums

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/MoveCameraButton.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/MoveCameraButton.java
@@ -44,6 +44,7 @@ import java.util.List;
  */
 public class MoveCameraButton extends AbstractToolbarItem {
   public static String MOVE_CAMERA_MODE = "moveCameraMode"; //NON-NLS
+  public static String ZOOM = "zoom"; // NON-NLS
   public static String BOARD_NAME = "boardName"; //NON-NLS
   public static String X_POS = "xPos"; //NON-NLS;
   public static String Y_POS = "yPos"; //NON-NLS;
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/MoveCameraButton.java
```diff
@@ -124,6 +126,22 @@ private Point getDestination() {
...
+            if (zoomer != null) {
+              zoomer.setZoomFactor(toZoom);
+            }
+          }
+        }
+        catch (NumberFormatException e) {
+          ErrorDialog.dataWarning(new BadDataReport(this, Resources.getString("Error.non_number_error"), Resources.getString("Editor.MoveCameraButton.zoom") + ": " + zoomString, e));
+        }
+      }
+
```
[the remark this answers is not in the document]
Cattlesquat: The zoomer itself allows you to zoom to any level with its custom zoom option.

vassal-app/src/main/java/VASSAL/build/module/map/MoveCameraButton.java
```diff
@@ -124,6 +126,22 @@ private Point getDestination() {
...
+          if (toZoom > 0) {
+            final Zoomer zoomer = map.getZoomer();
+            if (zoomer != null) {
+              zoomer.setZoomFactor(toZoom);
+            }
+          }
+        }
+        catch (NumberFormatException e) {
+          ErrorDialog.dataWarning(new BadDataReport(this, Resources.getString("Error.non_number_error"), Resources.getString("Editor.MoveCameraButton.zoom") + ": " + zoomString, e));
+        }
```
uckelman, line 142: The try/catch should be tighter around the thing which is intended to throw. What you've got here will catch a `NFE` from anything inside the `try`, but if anything other than `parseDouble` throws one, it's a legitimate bug, not a bad value, and we'd never see it this way.

Cattlesquat: fixed

BrentEaston (approved):

PR #10750, opened 2021-11-13
10750 - Place Marker should put stacking markers in a stack (even if parent marker is non-stacking)

Cattlesquat:
Place Marker was placing stacking pieces "Commando" if the placing piece was non-stacking.

Bug report was here:
https://forum.vassalengine.org/t/place-marker-doesnt-stack-counters/72928


BrentEaston (approved):
Excellent pickup.

PR #10755, opened 2021-11-14
9487 - When switching sides, close private hands/windows that we no longer have access to.

### LINKED ISSUE #9487: Hand/Private window security improvement
| Bugzilla |  |
|----------|----------|
| ID | BZ13746 |
| Reported | 2020-11-26 13:04:00 +0100 |
| Modified | 2021-04-17 19:48:45 +0200 |
| Product | VASSAL |
| Component | Unknown |
| Version | 3.4.10 |
| Platform | All |
| OS | All |
| Status | NEW |
| Resolution | None |
| Priority | unspecified |
| Severity | normal |
| Milestone | --- |
| Reporter | [Mark Benson](mailto:markben61@gmail.com) |

#### [Mark Benson](mailto:markben61@gmail.com) 2020-11-26 13:04:10 +0100
<pre>Normally a hand / private window is inaccessible to the non-owning side.

It can happen that a player starts a game by joining one side, opens a window and then swaps to another side. In this state, the player will be able to view the contents of the private window but not access the pieces in it. The window's toolbar buttons will still work for that player.

In a solo game, the solo player may wish to switch between sides as an alternative to a true "Solitaire" role. One way for the module writer to support such a solo mode is to integrate side switching with opening up the active side's hand window. However, without a means of determining the private window's display state (hide/unhide), the method will depend on the user not to close/open hand windows manually.

Both these issues could be resolved by automatically closing private windows when a user switches to a side that does not own that private window.</pre>

#### [Mark Benson](mailto:markben61@gmail.com) 2021-03-28 21:03:56 +0200
<pre>Notification of player side changes (v3.5) reduces the need for this request and the arrival of {MapName}_isVisible in Vassal v3.5.4 helps further.

Automatic closing of windows on joining a playerSide which has no access to the window would still be an improvement.</pre>

Cattlesquat:
Fixes #9487

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/PrivateMap.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/PrivateMap.java
@@ -213,10 +213,8 @@ public class PrivateMap extends Map {
 
     // Close this private window if we've switched to a side not allowed to see it
     final Container tla = theMap.getTopLevelAncestor();
-    if (tla != null) {
-      if (tla.isVisible() && !isVisibleTo(newSide)) {
-        tla.setVisible(false);
-      }
+    if ((tla != null) && tla.isVisible() && !isVisibleTo(newSide)) {
+      tla.setVisible(false);
     }
   }
 
--- a/vassal-app/src/main/java/VASSAL/build/module/PrivateMap.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/PrivateMap.java
@@ -17,6 +17,7 @@
  */
 package VASSAL.build.module;
 
+import java.awt.Container;
 import java.awt.Window;
 import java.awt.dnd.DropTarget;
 import java.awt.event.KeyEvent;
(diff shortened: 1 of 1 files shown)


BrentEaston (approved):

vassal-app/src/main/java/VASSAL/build/module/PrivateMap.java
```diff
@@ -209,6 +210,14 @@ else if (isVisibleTo(newSide)) {
       ((View)getView()).enableMotionListeners();
     }
     getLaunchButton().setEnabled(isVisibleTo(PlayerRoster.getMySide()));
+
+    // Close this private window if we've switched to a side not allowed to see it
+    final Container tla = theMap.getTopLevelAncestor();
+    if (tla != null) {
+      if (tla.isVisible() && !isVisibleTo(newSide)) {
```
uckelman, line 217: Is there a reason not to combine these two `if` conditions?

Cattlesquat: fixed

PR #10756, opened 2021-11-14
Fix bug in Undo of refresh of units in Decks

BrentEaston:


BrentEaston:
Closed. Re-creating from latest master to prevent clirr errors

PR #10757, opened 2021-11-14
10757 - Prevent undo of Refresh Counters from causing contents of Decks to disappear

BrentEaston:


vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -689,25 +618,50 @@ private GamePiece processGamePiece(GamePiece piece, Command command) {
...
+   */
+  private void processDeck(Deck deck, Command command) {
 
+    // Take a copy of the pieces in the Deck
+    final List<GamePiece> pieces = deck.asList();
+
+    // Remove all the pieces from the Deck by moving them to 0,0 then deleting them. This ensures the Undo
+    // ends with a valid MovePiece command to move the piece back into the Deck
+    for (final GamePiece piece : pieces) {
+      final Command move = deck.getMap().placeAt(piece, new Point(0, 0));
```
Cattlesquat, line 634: I guess 0,0 is safe because we're just moving the piece there, not processing it into stacks, decks, etc? (I mean I realize -3000,-3000 wouldn't be safe either, haha, but just checking that it doesn't matter if anything else if "already there" at 0,0)

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -689,25 +618,50 @@ private GamePiece processGamePiece(GamePiece piece, Command command) {
...
+      command.append(move);
+      final Command remove = new RemovePiece(piece);
+      remove.execute();
+      command.append(remove);
+    }
+
+    // Create a set of fresh pieces
+    final List<GamePiece> newPieces = new ArrayList<>();
+    for (final GamePiece piece : pieces) {
+      final GamePiece newPiece = gpIdChecker.createUpdatedPiece(piece);
```
Cattlesquat, line 644: Deleting all the pieces and then adding them again makes me dizzy, like we might lose little chunks of state or something, but it _looks_ like it would work. 

Is the reason you don't just add e.g. "piece" back to the map (but instead create newPiece from it) something to do with the undo?

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -223,22 +171,21 @@ public void log(String message) {
     // ex for dialog msg dialog.addMessage(Resources.getString("GameRefresher.counters_refreshed_test", updatedCount));
     // Log to chatter
     GameModule.getGameModule().warn(message);
-    //msg.append(new Chatter.DisplayText(chatter, message));
     logger.info(message);
   }
 
   public List<GamePiece>  getCurrentGameRefresherPieces() {
+    decks = new ArrayList<>();
```
Cattlesquat, line 178: I remember that Decks carry around a lot of references to a lot of stuff. Probably best to set `decks=null;` once the whole operation is done so that these don't last through a Close Game or whatever.

Cattlesquat (changes requested):
One deck memory thing that may be quasi-important, and a couple misc comments.

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
BrentEaston: map.placeAt() will place the piece there as a non-stacking piece before it is immediately deleted. It ensures the final Command of the undo sequence is a MovePiece moving the piece back into the Deck which is the only way I could find to cleanly undo the refresh. Deleting the piece directly from the Deck leaves the final Undo Command as an AddPiece which does not set the piece correctly into the Deck. Which makes me wonder if there is a bug if you delete cards directly from a Deck using a GKC that they will not correctly Undo?

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
BrentEaston: This is the basic process of the Refresh as it has always worked, Remove the existing piece, Create a new piece, place the new piece back where the original piece was. This just pulls this all together for Decks rather than having Deck pieces refreshed in the standard processGamePiece loop.

Yes, this is possibly to ensure undo-ability as the new type and state of the piece has to be encoded into the Gamestate in  a Command that can be undone.

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
BrentEaston: Good point

Cattlesquat (approved):

BrentEaston:
@Cattlesquat I tracked down the problem to a stoopid bug I introduced into beta2 trying to fix another issue. This also allowed me to get rid of the 'move each counter to 0,0 then delete it' rubbish. Much cleaner handling for the pieces in Decks now.

PR #10761, opened 2021-11-16
10761 - Toolbar button disabler - Buttons on toolbar can be disabled when a specified global property is true

Cattlesquat:
This uses the AbstractToolbarItem class to add the ability to have toolbar buttons depend on a named global property (via listener) and if/while it's "true" then the button is disabled.

This applies to Global Key Command buttons, those Action Button things whatever they're called, and various other ones (disable your dice buttons! whee!). 

Example shot below. Only the checkbox part of the configurer appears, unless you check the box, of course... 

image

NOTE - this adds a binary & an IconConfig field to AbstractToolbarItem, which means that various AbstractConfigurables now have more fields.


BrentEaston (changes requested):
We need to update the doco for the changed components and new component images. Sadly we don't have a central 'Toolbar button basics' entry.

Cattlesquat:
I made an entry in the existing Toolbar.adoc that now covers the "toolbar basics" for configuration and includes a shot of these controls fully-expanded (by default the disable-on-property box is unchecked and so most of these controls you don't see anyway unless you activate them). 

I added documentation of all the fields to each of the articles that have them (e.g. Multi Action Button, Inventory, etc). I did not change the screenshots because it's only the checkbox and I kind of like the quaint collection of old configurer shots when they aren't too confusing (but all of the articles now refer to the Toolbar article which does contain the fully expanded shot).

BrentEaston (approved):

PR #10774, opened 2021-11-18
10774 - Validation messages when Movement Trail traits placed inside of Rotate, Pivot, or Mat Cargo

Cattlesquat:
MOVEMENT TRAILS SHAT THE BED. NEWS AT 11.

image

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Footprint.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Footprint.java
@@ -17,6 +17,7 @@
  */
 package VASSAL.counters;
 
+import VASSAL.build.BadDataReport;
 import VASSAL.build.module.Map;
 import VASSAL.build.module.documentation.HelpFile;
 import VASSAL.command.ChangeTracker;
@@ -29,6 +30,7 @@ import VASSAL.configure.NamedHotKeyConfigurer;
 import VASSAL.configure.StringConfigurer;
 import VASSAL.i18n.PieceI18nData;
 import VASSAL.i18n.Resources;
+import VASSAL.tools.ErrorDialog;
 import VASSAL.tools.NamedKeyStroke;
 import VASSAL.tools.SequenceEncoder;
 import VASSAL.tools.image.ImageUtils;
@@ -144,6 +146,7 @@ public class Footprint extends MovementMarkable {
         pointList.add(new Point(x, y));
       }
     }
+    requireNoOuterRotate();
   }
 
   @Override
@@ -270,6 +273,7 @@ public class Footprint extends MovementMarkable {
    */
   @Override
   public void setMoved(boolean justMoved) {
+    requireNoOuterRotate();
     if (justMoved) {
       recordCurrentPosition();
       final Map map = getMap();
@@ -718,6 +722,23 @@ public class Footprint extends MovementMarkable {
     }
 
     return null;
+  }
+
+  /**
+   * Displays a Bad Module Data warning if a movement trails trait is found inside of a rotation trait
+   */
+  public void requireNoOuterRotate() {
+    Decorator piece = getOuter();
+    while (piece != null) {
+      if ((piece instanceof FreeRotator) || (piece instanceof Pivot) || ((piece instanceof MatCargo) && (((MatCargo)piece).maintainRelativeFacing))) {
+        String name = getName();
+        if ((name == null) || name.isEmpty()) {
+          name = Resources.getString("Decorator.prototype");
+        }
+        ErrorDialog.dataWarning(new BadDataReport(Resources.getString("Decorator.trails_inside_rotate", getBaseDescription(), piece.getBaseDescription()), name));
+      }
+      piece = piece.getOuter();
+    }
   }
 
   @Override
--- a/vassal-app/src/main/java/VASSAL/counters/Footprint.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Footprint.java
@@ -117,6 +117,8 @@ public class Footprint extends MovementMarkable {
   private KeyCommand showTrailCommandOff;
   private KeyCommand showTrailCommandClear;
 
+  private boolean rotateCheckedOnMove;
+
 
   public Footprint() {
     super(Footprint.ID, null);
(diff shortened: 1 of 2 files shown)


BrentEaston (changes requested):
Unfortunately, this is unacceptably slow on my Greatest Day module, which has probably got a few thousand pieces and no movement trails More 6-10 seconds ish.

To do something smarter, we will need to move the Validator from the PieceSlots to the PieceSlots parents (Palette, At-Start Stack, Deck,  prototypeContainer) so we can try and reduce the number of checks that need to be made.

e,g, Once a piece has been checked, we know the answer for any other piece that has the same prototypes in the same order. If we know none of those prototypes has a MT trait, then we know the answer for any other piece that has any subset of those prototypes in any order. We should be able to make the check quite quick for modules that don't use MTs.

Cattlesquat:
Sounds reasonable, but probably beyond my algorithmic skill.

BrentEaston:
It's a nightmare. I'll have to think deeper on this.

A Validation of the GameModule is also run at startup and the results written to the Errorlog.

Bitching to the players at run-time is starting to sound like a better idea. Could be done like a Bad Data in Module error. The designers will normally see it during testing,

BrentEaston:
Definitely thumbs down to using the ValidityChecker, is is just too slow and crazy complicated to try and do any pruning to try an reduce the amount of work done. There should definitely be a check when you save a PieceDefinition in the Editor. That will pick up all new creations of the situation. 

You said:

 'I just realized this won't really work 
Because if I change a single prototype to add a Movement Trail (which is what I would normally do), and the Rotate was in some other prototype (which is where I would normally have it), I would never get the message until I edited an individual piece that included the two (and I hardly ever edit individual pieces)
And we obviously don't want to do "comprehensive prototype cross-checking" every time we edit any prototype.
So I'm coming back around to a runtime check ... I think the lightest footprint might be e.g. in the mySetState for anything in the Rotate family -- something like if (inTheEditor && !alreadyWarned) ...  and then walk inward down the the getPiece() looking for instanceof Footprint and if so do the warning.'

Yes, I think that is going to have to be it.

Cattlesquat:
Okay here's the best I've been able to come up with so far:
* Only Rotate/Pivot/Cargo traits do a check.
* It checks at the point of setInner(), which we mostly do once per trait anyway. 
* At module _load_ time it can only detect problems inside of a single prototype, and it doesn't have a way to name the prototype.
* When you start a game though it can & will name all the pieces with the specific problem.

BrentEaston (changes requested):
Trying to test this using the Napoleon's Triumph module that we fixed during the debug phase, I opened the French  prototype (that now has the movement trail outside the Cargo trait) and dragged the Movement trail above the Mat Cargo. This causes an infinite hanging loop.

Cattlesquat:
Welp, I guess we can't use setInner because of this kind of perfectly reasonable stuff:
```
    // Update the inner pointers
    ((Decorator) piece).setInner(inner);
    if (outer != null) {
      outer.setInner(piece);
    }

```

So I dunno... I guess maybe we just stick to the documentation?

Cattlesquat:
Here is a new version. Having a call from the mySetState picks up prototypes (if the full violation is entirely contained in a prototype). Having the call in setMoved picks up the rest at runtime.

vassal-app/src/main/java/VASSAL/counters/Footprint.java
```diff
@@ -720,6 +724,23 @@ public Command myKeyEvent(KeyStroke stroke) {
...
+  public void requireNoOuterRotate() {
+    Decorator piece = getOuter();
+    while (piece != null) {
+      if ((piece instanceof FreeRotator) || (piece instanceof Pivot) || ((piece instanceof MatCargo) && (((MatCargo)piece).maintainRelativeFacing))) {
+        String name = getName();
+        if ((name == null) || name.isEmpty()) {
+          name = Resources.getString("Decorator.prototype");
+        }
+        ErrorDialog.dataWarning(new BadDataReport(Resources.getString("Decorator.trails_inside_rotate", getBaseDescription(), piece.getBaseDescription()), name));
+      }
```
BrentEaston, line 739: One of the constructors for BadDataReport takes an EditablePiece as an argument and includes the piece name in the report.

vassal-app/src/main/java/VASSAL/counters/Footprint.java
```diff
@@ -270,6 +273,7 @@ public Object getProperty(Object key) {
    */
   @Override
   public void setMoved(boolean justMoved) {
+    requireNoOuterRotate();
```
BrentEaston, line 276: This only needs to be checked once per game.

BrentEaston (changes requested):

vassal-app/src/main/java/VASSAL/counters/Footprint.java
Cattlesquat: Any suggestions how to track that? Just some boolean field that is constructored to false and we set to true when we've done the check?

vassal-app/src/main/java/VASSAL/counters/Footprint.java
Cattlesquat: That wouldn't handle the special case of a Prototype at least showing "Prototype" as the name (it couldn't actually show the real name of the prototype without adding a lot more "really bad code")

vassal-app/src/main/java/VASSAL/counters/Footprint.java
BrentEaston: Just a private boolean rotateCheckedOnMove = false and set it true when you check. I don't see any reason why it should need to be reset. Except possibly if you refresh the counters, but I wouldn't worry about that, it will be picked up the next time.

Cattlesquat: okay, fixed up.

vassal-app/src/main/java/VASSAL/counters/Footprint.java
BrentEaston: Ok, true.

BrentEaston (approved):

PR #10780, opened 2021-11-18
9219 - Expression Builder button added for Message Format fields & Report Trait. Fixes Sum() and Count() expressions in the Report trait's field.

### LINKED ISSUE #9219: Report trait to support Sum() & Count() functions.
| Bugzilla |  |
|----------|----------|
| ID | BZ13357 |
| Reported | 2020-09-01 01:18:00 +0200 |
| Modified | 2020-09-01 01:18:35 +0200 |
| Product | VASSAL |
| Component | Player |
| Version | 3.4.0-beta1 |
| Platform | All |
| OS | All |
| Status | NEW |
| Resolution | None |
| Priority | unspecified |
| Severity | normal |
| Milestone | --- |

#### [Brent Easton](mailto:b.easton@exemail.com.au) 2020-09-01 01:18:35 +0200
<pre>The Report trait currently supports the use of {} expressions to generate text dynamically, even though it was not designed to. However, Sum() and Count() always return 0.

Investigate
1. Can Sum() and Count() be enabled by changing the evaluation call sequence?
2. Should the Expression Builder button be added to this field?
3. What about Message Format fields? Should they be treated the same? What about the Message Format fields that have a drop-down to insert locally handled $...$ variables? [This is probably more work that steps 1 and 2, perhaps leave for another day].

Note that currently, if a non Beanshell expression is entered, it must be unquoted and only use $...$ variables.</pre>

Cattlesquat:
Fixes #9219 SumStack and CountStack were failing because the propertysource was a special old-and-new-locations version that it the ExpressionInterpreter didn't know how to use. I put in a step that allows them to use it properly. 

I have also added the Expression Builder button to the field.

And I did part "c" which is actually the biggest designer-facing change: message format fields throughout Vassal now use include the Expression Builder button. 

image


BrentEaston (changes requested):
We need new images for the changed configurers.

We'll also need an update of the Message Format section in the reference manual to make it clear about the different between Old-Style and New-Style or it is going to be very confusing. People will be hitting the Calculator button and clicking Ok and converting the expression and won't have a clue what is going on.

Cattlesquat:
I have updated the Message Format documentation, including the images on it, and I have included an image of an example using the Beanshell style. 

I don't necessarily recommend updating _every_ configurer-with-a-message-format screenshot just to show the inclusion of the calculator button, partially because I find seeing some of the older screenshots kind of "lovable and quaint" as long as they aren't so old that they're actively confusing. (I will probably on the other hand be updating various configurer screenshots for the other thing--the toolbar disabler--because that's a more significant visual change)

PR #10791, opened 2021-11-20
10791 - Cumulative Rotate for Move-Fixed-Distance is now correct?

Cattlesquat:
The get-cumulative-angle method in FreeRotator was ignoring MatCargo. Fortunately it was only called in Move-Fixed-Distance (Translate.java). So now... maybe it's fixed? LOOK CAREFULLY.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
+++ b/vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
@@ -203,6 +203,8 @@ public class FreeRotator extends Decorator
     return useUnrotatedShape ? 0.0 : validAngles[angleIndex];
   }
 
+  // These are deprecated and also don't work because MatCargo also rotates things
+  @Deprecated
   public double getCumulativeAngle() {
     double angle = getAngle();
     // Add cumulative angle of any other FreeRotator trait in this piece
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
```diff
@@ -203,6 +203,8 @@ public double getAngle() {
     return useUnrotatedShape ? 0.0 : validAngles[angleIndex];
   }
 
+  // These are deprecated and also don't work because MatCargo also rotates things
+  @Deprecated
```
uckelman, line 207: Should tack a `(since = "2021-11-20", forRemoval = true)` onto these so they get picked up by the deprecation checker. Since these are broken, it would be good to encourage people not to use them, even if we don't ever actually remove them.

uckelman (commented):
Is there a good test case for this I can try?

vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
Cattlesquat: Added the for-removals. I don't really have a test case because this was just a drive-by-noticing-the-code rather than someone sent me a broken module.

Cattlesquat:
Closing this as I've merged it all into the new "related" PR about action button rotation

PR #10796, opened 2021-11-21
4496 - Fancier action button definition

### LINKED ISSUE #4496: ActionButton Trait editor interface
| Bugzilla |  |
|----------|----------|
| ID | BZ1903 |
| Reported | 2009-02-25 12:29:00 +0100 |
| Modified | 2020-06-22 00:12:42 +0200 |
| Product | VASSAL |
| Component | Unknown |
| Version | unspecified |
| Platform | All |
| OS | All |
| Status | NEW |
| Resolution | None |
| Priority | medium |
| Severity | enhancement |
| Milestone | --- |
| Reporter | [Rodney Kinney](mailto:rodneykinney@comcast.net) |

<pre>---- Reported by rodneykinney@comcast.net 2009-02-25 00:29 ----

Specifying the area manually is a pain.  Instead, show the user the piece graphically and let them drag a regtangle to define the target area. 

This bug was previously known as _bug_ 2635298 at http://www.vassalengine.org/trackershow_bug.cgi?id=2635298

Unknown Component 
   Previous assignee is unknown.
Unknown resolution "".

Cattlesquat:
Fixes #4496 - always fun to finish one that Rodney put on the list ;) 

Now **MEGA-DELUXE!!!!**
(1) Default for new action buttons will be to simply use the entire shape of the piece as the button. Probably like 90% of use cases.
(2) If you choose to define an explicit shape you get a visual polygon editor (uses same code as Zone editor)
(3) You can now use any polygon (not just a rectangle)
(4) Default explicitly-drawn-shape will attempt to start w/ bounding box of piece if that makes sense. Otherwise it will let you do the drag-to-form-a-rectangle thing like the Zone editor does.

image


BrentEaston (changes requested):
Looks great, but needs Doco mcDocface.

BrentEaston (approved):

riverwanderer:
I'm finding that some Action Buttons no longer work since this change. Some do. I can't yet find the reason, but I'm wondering if it is due to interaction with Layers.

For example, take the C&C Nap module, v4.0.x (from Vassal); join a scenario as Allies. The board includes this display:
COMMAND DECK

There are several action buttons:-
1) Clocks OFF; click to toggle on or back off again.  This button still works.
2) 1 button per deck; click to select a deck (triggers a GKC). These no longer work.
3) The Exit Door; click to retire or switch side (trigger a GHK). This button also no longer works.

The button that works as it appears in the Configurator (looks ok):
image

Yet, other buttons that don't work, also look correct in the configurator:
image

The problem persists when I edit one of the buttons that doesn't work, and Refresh the scenario. Same issue appears another module with a different layout.

uckelman:
@riverwanderer Is this fixed as of the current 3.6.2 master build? https://vassalengine.org/builds/

riverwanderer:
Yes. And it is fixed in v3.6.1 and v3.6.0 too.

PR #10803, opened 2021-11-23
10803 Speeds Game Refresh and general performance in modules with many maps defined

Cattlesquat:
I did a profile on this guys' huge 306mb World in Flames module where he said the Refresh Counters time had gone from "a mere 10 minutes" to more like 50 minutes. Anyway... the guys has some "interesting features" like 57 MAP WINDOWS. But anyway from watching the profiling I was able to isolate this one little place and make his module a fuckton faster. 

So it probably makes everybody's module a non-fuckton faster but still faster.

Anyway we should probably put it in 3.6-beta7 so he can try it out.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -2070,10 +2070,12 @@ public class GameModule extends AbstractConfigurable
     }
 
     //BR// MapName_isVisible property for each map window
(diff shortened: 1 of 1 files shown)


Cattlesquat:
Also after I put this in, a refresh counter *with the profiler running* only took 3-4 minutes which seems like ... progress.

BrentEaston (approved):

PR #10807, opened 2021-11-23
Fix an action button oops

Cattlesquat:
If there was more than one action button on a single piece, it was only checking the first one for clicks.


riverwanderer:
works for me - multi-button pieces are back in action.

PR #10808, opened 2021-11-23
Misc property lookup optimizations

Cattlesquat:
(1) moved INNER and OUTER to the top of the properties check list for Decorator, since they're asked for the most
(2) Adjusted Decorator.getOutermost() to not call getProperty twice for the same thing
(3) Adjusted the ExpressionInterpreter to avoid throwing scads of NPE / NumberFormat exceptions when a piece doesn't have a particular property

None of these make the code _look_ any better but they run faster.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
+++ b/vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
@@ -452,14 +452,7 @@ public class ExpressionInterpreter extends AbstractInterpreter implements Loopab
         final Object prop = ps.getProperty(property);
         if (prop != null) {
           final String s1 = prop.toString();
-          if (NumberUtils.isParsable(s1)) {
-            try {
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -512,12 +528,15 @@ public Object countStack(String property, PropertySource ps) {
...
+              final Object prop = gamePiece.getProperty(property);
+              if (prop != null) {
+                final String val = prop.toString();
+                if (!"".equals(val)) {
+                  result++;
+                }
               }
             }
-            catch (Exception ignored) {
+            catch (NumberFormatException ignored) {
```
uckelman, line 539: This try/catch is unnecessary. There's nothing that can throw inside it.

uckelman (changes requested):

uckelman (approved):

PR #10811, opened 2021-11-23
Optimization: Check for possible keystrokes in Restrict Commands before evaluating match epression

Cattlesquat:
They were evaluating the full beanshell even if the key being passed through wasn't the key they were restricting.


Cattlesquat:
@riverwanderer this might make your module seem a little more spritely...

PR #10812, opened 2021-11-23
Optimize Single Property Beanshell expressions

BrentEaston:


BrentEaston:
Ensures {propName} is converted to a non-beanshell SinglePropertyExpression rather than a generalised BeanshellExpression. Will evaluate orders of magnitude faster.

PR #10835, opened 2021-11-28
10835 - Clear Temp User Id when changing sides

Cattlesquat:
When a player retires from a side, clear any "temp user id" they were using to claim that side (e.g. if they were given the chance to match to a blank password or something). This ensures that they will claim their own new side with their correct password rather than (ack!) the temporary blank password. 

I also needed to introduce a "Remove"-player-from-roster for  this. 

Likewise, I made it so if the first player into the game happens to be a Blank Password person, it doesn't auto-force the next player to start the game into that slot -- instead it "offers it". 

Eventually (for perhaps 3.6.1) we can consider additional flow-improvement measures including:
(1) We can halt players with blank passwords and force them to pick a password before continuing (since they can still auto-match their existing games, and this then prevents them from being a _constant_ problem in this department)
(2) Once players pick real passwords, we could actually _write_ their new passwords into the game state (overwriting their blank password), rather than letting the game limp along with a blank password

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
@@ -65,6 +65,7 @@ public class PlayerRoster extends AbstractToolbarItem implements CommandEncoder,
 
   public static final String SIDES = "sides"; //$NON-NLS-1$
   public static final String COMMAND_PREFIX = "PLAYER\t"; //$NON-NLS-1$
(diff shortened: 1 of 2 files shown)


BrentEaston:
Is PR redundant now that #10835 is in play?

Cattlesquat:
Yes. I posted it as an option to do "a little bit" if we thought the other one was too much.

If #10839 is merged this one can be closed.

Cattlesquat:
It seems like we're coming down on the #10839 side of the fence so I'm going to close this one rather than create potential for further confusion.

PR #10838, opened 2021-11-29
10838 - Password stuff! 

Cattlesquat:
When picking a side in a new game, if the password is currently blank, provide an opportunity to correct that. 

This (alas) doesn't _block_ the ability to leave the field blank, as the "GameSetupSteps" part of the Wizard doesn't seem to utilize the "setProblem" thing that blocks forward progress. 

But if I _do_ set my password in this dialog, it sets it in my preferences and then leaves me alone in the future about my password.

Note -- this PR builds on #10835, and includes everything that's in that.

image

### DIFF
--- a/vassal-app/src/main/java/VASSAL/launch/LoadModuleAction.java
+++ b/vassal-app/src/main/java/VASSAL/launch/LoadModuleAction.java
@@ -95,8 +95,8 @@ public class LoadModuleAction extends GameModuleAction {
       module.getPlayerWindow().setVisible(true);
 
       // prompt for username and password if wizard is off
-      // but no username is set
-      if (!isRealName()) {
+      // but no username is set, or password is blank
+      if (!isRealName() || !isNonBlankPassword()) {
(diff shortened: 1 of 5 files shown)


vassal-app/src/main/java/VASSAL/launch/LoadModuleAction.java
```diff
@@ -114,4 +114,15 @@ private boolean isRealName() {
...
+   * Returns true if user has supplied a real password for current GameModule.
+   *
+   * Test's whether GameModule.SECRET_NAME is non-empty
+   *
+   * @return <code>true</code> if user supplied a real password
+   */
+  private boolean isNonBlankPassword() {
+    final String pwd = (String)GameModule.getGameModule().getPrefs().getValue(GameModule.SECRET_NAME);
+    return (pwd != null) && !pwd.isBlank();
+  }
```
uckelman, line 127: This methods shows up in two places; where could it go so that it's in one?

uckelman (changes requested):

PR #10839, opened 2021-11-29
10839 - Improved flow for users with blank passwords

Cattlesquat:
When picking a side in a new game, if the password is currently blank, provide an opportunity to correct that.

This ALSO blocks the ability to leave the field blank, using the "setProblem()" method of the Wizard controller.

When you set your password in this dialog, it sets it in the preferences and then leaves you alone in the future about your password. (You won't see the password part of the dialog every game, only if you're starting with an empty password)

Note -- this PR builds on #10835, and includes everything that's in that.

image


BrentEaston (approved):
This looks good to me.

I'm just wondering whether the dialog where we force them to enter the password, we add text to re-iterate that this password is what connects them to the side they choose?

Cattlesquat:
@BrentEaston I added some (hopefully improved) text along that line. I think it's slightly an uphill battle against people not really thinking of passwords as linking them to things. But at least we've tried :)

BrentEaston:
Looks good to me.

PR #10852, opened 2021-11-30
10852 - Turn Tracker button sizes configurability

Cattlesquat:
A couple of minor tweaks (and one bugfix) to Turn Tracker, which would let me remove almost all of the remaining custom code from PoG.
(1) Configurable fields for the button sizes in the turn widget (so they can be fit to the icon sizes of the rest of the toolbar)
(2) Fixed a buggish behavior where if advancing the turn changed the values of properties used in the "format" field of the turn display, they weren't shown with their new current values.
(3) Created an opportunity for the module designer to override the default preference settings -- the user preferences are still respected, but the _initial_ settings of them are overridden if module designer includes certain Global Properties.


vassal-app/src/main/java/VASSAL/build/module/turn/TurnTracker.java
```diff
@@ -847,8 +897,9 @@ protected void setControls() {
     private static final long serialVersionUID = 1L;
     private IconButton nextButton;
     private IconButton prevButton;
+    private boolean initialized = false;
 
-    protected final int BUTTON_SIZE = 22; //NOPMD
+    protected final int BUTTON_SIZE = 22; @Deprecated //NOPMD
```
uckelman, line 902: Where's that `@Deprecated` supposed to be?

Cattlesquat: Oh just that BUTTON_SIZE isn't actually used for anything. Should I just get rid of that line? Can I?

uckelman: No, I just mean it's in a strange place. If you want to deprecate `BUTTON_SIZE`, it should be on the line _before_ the declaration.

Cattlesquat: Ah, gotcha -- fixed

PR #10853, opened 2021-11-30
10853 - Prototypes can be selected from popup menu

Cattlesquat:
Your wish is my command

image


vassal-app/src/main/java/VASSAL/counters/UsePrototype.java
```diff
@@ -251,6 +257,53 @@ public Editor(UsePrototype up) {
       nameConfig = new StringConfigurer(up.type.substring(ID.length()));
       controls.add("Editor.UsePrototype.prototype_name", nameConfig);
 
+      final JButton selectButton = new JButton(Resources.SELECT);
```
uckelman, line 260: You need to use the key to get the value here. I get a button that says `General.select`.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/counters/UsePrototype.java
Cattlesquat: "Oops"

uckelman (approved):

PR #10869, opened 2021-12-02
Inventory shouldn't show pictures/text for pieces in private windows not currently accessible

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/Inventory.java
```diff
@@ -997,6 +1021,15 @@ protected Command getCommand(CounterNode node, KeyStroke stroke) {
       final GamePiece p = node.getCounter() == null ? null : node.getCounter().getPiece();
       Command comm;
       if (p != null) {
+
+        // No commands to pieces if we don't have access to their windows
+        final VASSAL.build.module.Map map = p.getMap();
+        if (map instanceof PrivateMap) {
+          if (!((PrivateMap) map).isVisibleTo(PlayerRoster.getMySide())) {
```
uckelman, line 1028: These three lines get repeated. It would be cleaner to pack them up into a function:
```
private boolean isVisibleToMe(GamePiece p) {
   final VASSAL.build.module.Map map = p.getMap();
   return !(map instanceof Private Map) || ((PrivateMap) map).isVisibleTo(PlayerRoster.getMySide());
}
```

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/Inventory.java
Cattlesquat: Cool. Done.

BrentEaston (approved):

uckelman (approved):

PR #10872, opened 2021-12-02
Use piece as property source in auto-move-report where possible

Cattlesquat:
Benkyo was trying to do a property computation in map auto-move-reporting that seemed reasonable, except strangely the movement reports use the Map rather than the GamePiece as the property source, so he couldn't use the piece properties.

This makes use of the piece as the property source. Since it has already been moved to the target (destination) map by the time the movement report is called, those properties remain in scope.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/MovementReporter.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/MovementReporter.java
@@ -205,9 +205,9 @@ public class MovementReporter {
       }
       format.setProperty(Map.MAP_NAME, toMap.getLocalizedConfigureName());
 
-      final List<GamePiece> pieces = ms.getPieces();
+      final GamePiece piece = ms.getFirstPiece();
 
-      final String moveText = format.getLocalizedText(((pieces != null) && !pieces.isEmpty()) ? pieces.get(0) : toMap, toMap, sourceFieldKey);
+      final String moveText = format.getLocalizedText((piece != null) ? piece : toMap, toMap, sourceFieldKey);
 
       if (moveText.length() > 0) {
         c = c.append(new Chatter.DisplayText(GameModule.getGameModule().getChatter(), "* " + moveText));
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/MovementReporter.java
```diff
@@ -313,6 +315,10 @@ public Point getOldPosition() {
       return oldPosition;
     }
 
+    public List<GamePiece> getPieces() {
+      return new ArrayList<>(pieces);
+    }
+
```
uckelman, line 321: That kills me that you have to copy the whole list just to get the first element.

Cattlesquat: Even more hilarious is that, as far as I can tell, we NEVER ACTUALLY PUT MORE THAN ONE PIECE IN THIS!!!

Cattlesquat: Which puts me into this newer/better idea -- I changed it to something that just returns the first piece in the array. That will at least save all the making-a-copy. Sometimes I think too much in terms of "future extensibility"

Cattlesquat: So anyway that's checked in now.

BrentEaston: Why the first piece? Why not the last piece or the second piece? This is the reason that Map level auto-reports never used GamePiece properties, they are meaningless if you are dragging more than one piece.

Cattlesquat: Well that's what I *thought* except then Benkyo pointed out that it never actually does that. It generates one "MoveSummary" for EACH piece moved.

Cattlesquat: Or, well, no.

Cattlesquat: So yes ... if we _want_ to make this work and "non-dodgy", then I'd probably put a checkbox on there telling the Map to report separately for every unit (and not group a single stack together in one autoreport line). And so then if you checked that box we could use the piece as the propertysource for each report. 

We can discuss whether it makes sense to go ahead and do that, or if we should leave folks who want "complex move reporting" to create their own logic in the pieces themselves (which is completely possible - just the "on ramp" from these things is a little hard to find)

Cattlesquat:
User didn't end up pushing hard for us to do this, so I'm going to close this.

PR #10879, opened 2021-12-03
10879 Dice Button raw result and counts reporting

BrentEaston:


BrentEaston:
A small change that was missed from 3.6.0 to add promised $rawRolls$ reporting

PR #10880, opened 2021-12-03
Create HintTextFields lazily

uckelman:


Cattlesquat (approved):

BrentEaston (approved):

vassal-app/src/main/java/VASSAL/configure/NamedHotKeyConfigurer.java
```diff
@@ -64,12 +64,26 @@
...
-  private final HintTextField keyName = new HintTextField(StringConfigurer.DEFAULT_LENGHTH, NAME_HINT);
+  private HintTextField keyStroke;
+  private HintTextField keyName;
   private JPanel controls;
   private String lastValue;
   private JButton undoButton;
 
+  private HintTextField getKeyStroke() {
+    if (keyStroke == null) {
+      keyStroke = new HintTextField(StringConfigurer.DEFAULT_LENGHTH, STROKE_HINT);
```
jonathanrwatts, line 75: `DEFAULT_LENGHTH` is mispelled here (and in `getKeyName()` below)--is it also mispelled in the source location, or is this a bug?

jonathanrwatts: Okay, found StringConfigurer.java, and it is indeed misspelled the same way there (I figured it had to be, or it wouldn't compile!).  If I had a DE set up, I'd fix it myself, but since I don't, I'll leave it up to your judgment if it's worth the effort of fixing.

uckelman: Yeah... so I didn't _change_ that. It's a protected field in StringConfigurer, so I can't just remove it as that might break custom module code. I'll introduce new, correctly spelled fields and deprecate the old ones for the next release.

uckelman: So, funny thing: The PostScript standard has a spelling error in it. The angular French quotation marks, guillemets, were accidentally called "guillemots", which are a kind of marine bird similar to an auk. :)

uckelman: See #10881.

PR #10894, opened 2021-12-05
Warn about movement trails inside of rotates only when editing module

Cattlesquat:


vassal-app/src/main/java/VASSAL/counters/Footprint.java
```diff
@@ -754,11 +755,13 @@ public void requireNoOuterRotate() {
     Decorator piece = getOuter();
```
uckelman, line 755: Would it be better to check `GameModule.getGameModule().isEditorOpen()` right at the top, before doing any other work? This way, we're traversing the trait stack when we could already know we won't report.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/counters/Footprint.java
Cattlesquat: Yep - fixed

uckelman (approved):

PR #10907, opened 2021-12-06
10907 - Disabled Toolbar Buttons on Map Windows didn't initialize starting state properly

Cattlesquat:
I think buttons that initialized based on global properties that were truly global were working okay, but buttons initialized based on global properties that were part of Map windows (and so "didn't exist" until game setup fully activated) were not initializing properly.

https://forum.vassalengine.org/t/global-key-command-disable-bug/73222/5


uckelman:
User confirmed this works.

PR #10933, opened 2021-12-11
10930 - Chess Clock Control command should undo

### LINKED ISSUE #10930: Chess Clock show/hide does not undo
To replicate, open a module with the Chess Clock control button. e.g Paths of Glory.
"Show" the chess clocks from the Chess Clock control button.
The chess clock timers are displayed on the toolbar.
UnDo.
The Chess Clock timers remain displayed.

Impact: This UnDo restriction detracts from the ability of Vassal to hide the Chess Clock control button and use of means (GKC/GHK based) to set the chess clocks up.

Cattlesquat:
Fixes #10930

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
@@ -777,7 +777,7 @@ public class ChessClockControl extends AbstractConfigurable
 
     @Override
     protected Command myUndoCommand() {
-      return null;
+      return new UpdateClockControlCommand(!showing, online);
     }
 
     @Override


uckelman:
Wait, what? Why should _displaying the chess clock_ be something you can undo?

riverwanderer:
I can demonstrate how the **command** (no button) can be used to control whether or not chess clocks appear in the toolbar at all. This means (a) it can be an optional feature and (b) the clock can be reliably started at zero, when the game starts.

Cattlesquat:
It turns out there is NOT a command for displaying the chess clock. There is a command for *something* (I think it's for communication of clock status between two different clients). I thought maybe it was just missing the undo so I put one in, but that's not right. Joel is right that there isn't a "command" (i.e. undoable) for interface features. 

I don't think we're in a position to add new features to Chessclocks right now -- you may need to just use the scheme as it presently exists.

riverwanderer:
Thanks both;  I misunderstood, now I see that Joel meant the underlying Command that gets passed to other players. I think I experienced before that indeed Show/Hide clocks doesn't get passed across to other players but I forgot this when I was re-working this part of my module. I'll take another look now I better understand what is happenning.

PR #10961, opened 2021-12-19
Use heuristic value for initial tiler heap; use 3/4 physical RAM as max heap

uckelman:


Cattlesquat (commented):
Now that this is more of a one-pass, is there a cleaner way to report the "didn't have enough heap/memory" version of failure?

uckelman:
> Now that this is more of a one-pass, is there a cleaner way to report the "didn't have enough heap/memory" version of failure?

Nope. The thing that drove me to this solution was digging into the most recent report of the bug today, which demonstrated that the code which reports that you ran out of memory won't run reliably after an OutOfMemoryException.

The situation following this PR is that if you run out of memory tiling, it's just irrecoverable _on your machine_.  You simply don't have enough RAM and there are no settings you can adjust to fix that.

Cattlesquat:
I'm with you to that point, but for the people who fail out of memory, can it give them a "your system doesn't have enough memory to run this particular module" message (rather than scary bug dialog that people then say "there's a bug in Vassal")? Telling them it's just not possible on their computer is better than them thinking Vassal has a bug or that they should try a bunch more stuff. Because it seems a practical certainty that some module designers will inevitably make modules/maps too big for some users?

uckelman:
We can't reliably distinguish the situation in which they ran out of memory from one where they've had a different problem. That's exactly the thing I was faced with in the bug report I was looking at today.

uckelman:
So, for some additional color: If you have 4GB RAM, this will limit your tiler max heap to 3GB, which means the hard bound on what you could tile is a square image which is a bit over 28k pixels on a side. That's several times larger than the largest we've seen, which is already itself a unique outlier. I have a hard time imagining that we'd see people hit this limit more than a handful of times, ever.

Cattlesquat:
Another possible angle of attack is to work backward from THAT (3GB) and simply not allow maps larger than that.

uckelman:
That's not great for people like me with 32GB RAM. I'm going to to be really sad when I can't have my 80k x 80k map. :)

uckelman:
I recommend trying this out on Windows, e.g., to verify that it doesn't do something horrendous.

PR #10976, opened 2021-12-22
10970 Fix problem where rotated pieces did not move-fixed-distance in correct direction

### LINKED ISSUE #10970: Movement via arrow keys only allows for up/down and moving the arrow on the model makes no difference
Hello, all of these relate to Malifaux 3E.

1. select a model, and use the arrow keys to move in the direction indicated by the green arrow in the centre of the model.
Expected: model moves one inch in the arrow direction.
Actual: model moves up one inch.

1. Activate charge lane.
2. turn green arrow in charge direction.
3. use up/down arrow to move model
Expected: model moves one inch in direction of charge lane.
Actual: model moves up and down only.

Please see attached screen images. NOTE that this is occurring on MAC O/S running 3.6.2 and Malifaux 3e Malifaux Burns V3a and was not an issue on vassal 3.5 prior. This is also occurring on my Linux Mint 20 system.

1. Select model
2. press control + arrow up key
Expected: model moves by half inch
Actual: Screen shrinks and shows the names of each monitor.
Screen Shot 2021-12-21 at 15 30 14
Screen Shot 2021-12-21 at 15 31 08

Cattlesquat:
Fixes #10970

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Translate.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Translate.java
@@ -218,7 +218,7 @@ public class Translate extends Decorator implements TranslatablePiece {
     translate(p);
 
     // Handle rotation of the piece, movement is relative to the current facing of the unit.
-    // Traverse any traits outward from this trait (ones that could rotate this trait),
+    // Traverse any traits inward from this trait (because apparently Translate looks at inward traits only),
     // and find out what the cumulative rotation is
     GamePiece rotateTrait = getInner();
     double cumulative = 0.0;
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/counters/Translate.java
```diff
@@ -220,16 +220,17 @@ protected Command newTranslate(KeyStroke stroke) {
     // Handle rotation of the piece, movement is relative to the current facing of the unit.
     // Traverse any traits outward from this trait (ones that could rotate this trait),
```
uckelman, line 221: Should also fix the comment.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/counters/Translate.java
uckelman: "outward"

Cattlesquat: Fixed

uckelman (approved):

PR #11011, opened 2021-12-30
Fixed typo in DiceButton docs

uckelman:


uckelman:
I also spotted a few bad "text replacements". `<=` needs to be escaped with a backslash to prevent asciidoc from turning it into a double left arrow. See https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#text-replacements for details.

BrentEaston (approved):

PR #11038, opened 2022-01-06
11038 Allow 'Use Same Boards' option on Private Map to be removed

BrentEaston:


vassal-app/src/main/java/VASSAL/build/module/PrivateMap.java
```diff
@@ -116,10 +116,15 @@ else if (SIDE.equals(key)) {
       owners = (String[]) value;
     }
     else if (USE_BOARDS.equals(key)) {
-      for (final Map m : Map.getMapList()) {
-        if (m.getMapName().equals(value)) {
-          surrogate = m;
-          break;
+      if (value == null || ((String) value).isEmpty()) {
```
uckelman, line 119: Do you need to check that it's an instance of String here?

BrentEaston: True, unlikely, but possible to cause a problem. Update pushed

vassal-app/src/main/java/VASSAL/build/module/PrivateMap.java
```diff
@@ -116,7 +116,7 @@ else if (SIDE.equals(key)) {
       owners = (String[]) value;
     }
     else if (USE_BOARDS.equals(key)) {
-      if (value == null || ((String) value).isEmpty()) {
+      if (value == null || (value instanceof String && ((String) value).isEmpty())) {
```
jonathanrwatts, line 119: Maybe I'm over-thinking this, but if `value` *isn't* a String, isn't that going to cause problems with the `else` statement, too?  In which case perhaps this should be `if (value == null || !(value instanceof String) || ((String) value).isEmpty())` instead?

BrentEaston: No, .equals() can compare anything without failing, so the type of value in m.getMapName().equals(value) doesn't matter. It can't be equals() unless it is a String, but it won't cause an error if it isn't a String.

PR #11049, opened 2022-01-09
List key improvements

Cattlesquat:
More improvements for the ListKeyCommands dialog, per Dulgin's suggestions -- doesn't need a new issue number as it's just adding on.

(1) Dialog no longer modal
(2) Filter field no longer case sensitive
(3) "Jump to definition" right click option

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/ListKeyCommandsDialog.java
+++ b/vassal-app/src/main/java/VASSAL/configure/ListKeyCommandsDialog.java
@@ -127,7 +127,7 @@ public class ListKeyCommandsDialog extends JDialog {
         // show row containing the filter as a substring
         for (int i = entry.getValueCount() - 1; i >= 0; i--) {
           final String v = entry.getStringValue(i);
-          if (v != null && v.contains(f)) {
+          if (v != null && v.toLowerCase().contains(f.toLowerCase())) {
             return true;
           }
         }
@@ -182,6 +182,8 @@ public class ListKeyCommandsDialog extends JDialog {
 
     setLayout(new MigLayout("insets dialog, fill")); // NON-NLS
     add(panel, "grow"); // NON-NLS
+
+    setModal(false);
 
     SwingUtils.repack(this);
   }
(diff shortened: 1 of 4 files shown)


vassal-app/src/main/java/VASSAL/configure/ListKeyCommandsDialog.java
```diff
@@ -297,7 +354,7 @@ else if (target instanceof GamePiece) {
               }
             }
 
-            list.add(new String[] { cmd_key, cmd_name, src_type, src_name, src_desc });
+            list.add(new Pair(new String[] { cmd_key, cmd_name, src_type, src_name, src_desc }, configurable));
```
uckelman, line 357: People usually use `Pair.of()` to make pairs.

Cattlesquat: fixed - makes sense. I never know which of these little things incorporates which of the little toys.

vassal-app/src/main/java/VASSAL/configure/ListKeyCommandsDialog.java
```diff
@@ -202,13 +240,38 @@ public void actionPerformed(ActionEvent e) {
...
+    @Override
+    public void actionPerformed(ActionEvent e) {
+      int rowIndex = table.getSelectedRow();
+      if (rowIndex < 0) return;
+
+      rowIndex = table.convertRowIndexToModel(rowIndex);
+
+      final AbstractConfigurable jumpTo = ((MyTableModel)(table.getModel())).getSourceFor(rowIndex);
+
+      if (jumpTo instanceof AbstractConfigurable) {
```
uckelman, line 261: `jumpTo` has to be an `AbstractConfigurable` because you just declared it to be one.

vassal-app/src/main/java/VASSAL/configure/ListKeyCommandsDialog.java
```diff
@@ -222,6 +285,35 @@ public int getColumnCount() {
       return COLUMN_COUNT;
     }
 
+    AbstractConfigurable getSourceFor(int row) {
```
uckelman, line 288: `public`?

vassal-app/src/main/java/VASSAL/configure/ListKeyCommandsDialog.java
Cattlesquat: Oh right there were various earlier versions where it wasn't.

vassal-app/src/main/java/VASSAL/configure/ListKeyCommandsDialog.java
Cattlesquat: yes, fixed.

PR #11053, opened 2022-01-09
11053 Suppress Empty Deck Hotkeys when refreshing game

BrentEaston:


vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -626,6 +626,11 @@ private GamePiece processGamePiece(GamePiece piece, Command command) {
    */
   private void processDeck(Deck deck, Command command) {
 
+    // Prevent any listeners or hot Keys firing while we fiddle the Deck
+    deck.removeListeners();
```
Cattlesquat, line 630: The Deck listeners are all key listeners I think -- is it possible for us to actually receive a key & attempt to process it during the middle of the fiddling-the-deck part? Seems like not?

BrentEaston: No, the Refresh process is atomic,

BrentEaston: Hmmm, now you have me worried.

Cattlesquat: I just meant that since we probably can't actually receive a keystroke during this, that you probably don't need to remove and then re-add the listeners. Just suppressing that one thing should be enough?

BrentEaston:
From Mark: Tested out successful… no Hotkeys fired and a Refresh Counters runs worked with a change to a piece and a change to a deck.

PR #11054, opened 2022-01-09
11054 - Don't let old versions of PoG load and crash

Cattlesquat:


vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
```diff
@@ -265,6 +265,10 @@ Error.VSQL_too_old_title=Incompatible VSQL Version
 Error.VSQL_too_old_heading=Incompatible VSQL Version
 Error.VSQL_too_old_message=<html>The verison of VSQL you are trying to load is incompatible with VASSAL %1$s. Please check <a href="https://vassalengine.org/wiki/Module:Squad_Leader">vassalengine.org</a> for a newer version of VSQL, or load this version of VSQL with VASSAL 3.2.17.</html>
 
+Error.POG_too_old_title=Please update Paths of Glory module
+Error.POG_too_old_heading=Please update Paths of Glory module
+Error.POG_too_old_message=<html>The verison of Paths of Glory you are trying to load is incompatible with VASSAL %1$s. Please check <a href="https://vassalengine.org/wiki/Module:Paths_of_Glory">vassalengine.org</a> for a newer version of Paths of Glory.</html>
```
BrentEaston, line 270: verison

BrentEaston (commented):
I'm wondering whether we should make this a little more generic so when the next module designer asks you to do the same thing...

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
Cattlesquat: Lol on verison -- apparently cut-and-pasting from the VASL stuff was a mistake haha. Fixed VASL and VSQL versions too.

I certainly agree if a more generic way of setting lowest-loadable-version for a module was available that it would be a good idea. My throwaway idea would be something on the wiki page.

BrentEaston:
Looking up something from URL sounds like a bit much work. For now, perhaps just condense the code down into a class and hard-code a list of module names and verison numbers (Just VASL, VSQL and POG for now). That would make it far more maintainable if we need to add more, and would need to be done anyway if we decide to go for something more fancy.

Cattlesquat:
@BrentEaston Fixed it all up pretty nice

vassal-app/src/main/java/VASSAL/launch/Launcher.java
```diff
@@ -162,6 +167,45 @@ else if (t instanceof ExtensionsLoader.LoadExtensionException ||
...
   }
 
+
+  private static class LoadConstraint {
+    public String moduleName;
+    public String minVassalVersion;
+    public String minModuleVersion;
+    public String newVersionLink;
+
+    LoadConstraint(String moduleName, String minVassalVersion, String minModuleVersion, String newVersionLink) {
```
uckelman, line 177: `public` ctor?

vassal-app/src/main/java/VASSAL/launch/Launcher.java
```diff
@@ -162,6 +167,45 @@ else if (t instanceof ExtensionsLoader.LoadExtensionException ||
...
+      this.moduleName = moduleName;
+      this.minVassalVersion =  minVassalVersion;
+      this.minModuleVersion =  minModuleVersion;
+      this.newVersionLink   =  newVersionLink;
+    }
+  }
+
+  private static final List<LoadConstraint> constraints = Arrays.asList(
```
uckelman, line 185: `List.of()`

uckelman: Actually, if you're looking these up by name, they should really be in a `Map<String,LoadConstraint>` instead.

vassal-app/src/main/java/VASSAL/launch/Launcher.java
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/launch/Launcher.java
Cattlesquat: fixed

BrentEaston (approved):
Looks good Brian. Easy to expand or do something fancy with.

PR #11056, opened 2022-01-10
11043 - Don't attempt to send to already-hung-up connection

Cattlesquat:


vassal-app/src/main/java/VASSAL/chat/node/NodeClient.java
```diff
@@ -90,6 +90,7 @@
   protected NodeRoom pendingSynchToRoom;
 
   private SocketHandler sender;
+  private final Object senderLock = new Object(); // We want a dedicated object for synchronizing, rather than synchronizing on sender itself, since sender's object reference can/will change.
```
uckelman, line 93: You don't need this; you can synchronize on `this` instead.

PR #11066, opened 2022-01-13
"Path" column for List Key Commands

Cattlesquat:
(no issue number as it's just part of List Key Commands). 

I also included rollover text on the table, for in case a particular field is too wide for its column.


uckelman:
I might use `>` instead of `->` in the Path, as it's a bit more concise.

Cattlesquat:
done

PR #11073, opened 2022-01-14
11073 - Fix situation where Send-to-Location throws exceptions when it can't find target counters

Cattlesquat:
The "map" field is not e.g. where the piece is, it's just where the SendToLocation trait thinks its target is. And since we have "send to target counter" with filters, it is possible for there to be no valid target, so map can be null. And in this particular case we were inside a "if (map != null)" so it looked like map couldn't be null, except the apply-on-move key can change that situation by changing what matches filters, etc. So I've moved the reference up above the place where it can change, since repaint is just a "schedule this" call anyway.


Cattlesquat:
Mark tested & confirmed it stopped throwing exceptions for him.

PR #11074, opened 2022-01-14
11074 - Support Send-to-Location source/target matching with $...$ expressions

Cattlesquat:
Mark pointed out that the Send-to-Location's "send to nearest counter, filtered by properties" could make proper use of the same kind of $...$ source/target matching we do with global key commands. 

Here I have attempted to do that w/o having to restructure the entire Send-to-Location class -- I borrow the existing property filter, premasticate it for $...$, and then send the result onward to the actual matcher. 

This includes the fix in #11073


Cattlesquat:
Mark tested and confirmed that a $CurrentZone$ == CurrentZone produced his desired result.

BrentEaston (approved):
Looks Good.

PR #11079, opened 2022-01-15
11079 - Main window & map accept Drag-and-Drop for loading saves and logs

Cattlesquat:
Basically you can now drag a VSAV file from your file folder (or whatever) onto an existing Vassal player window, and it will load the saved game (after the usual checks/confirmations, e.g. do you want to save your current game first) 

Will now accept:
(a) Files dragged in OS
(b) Discord link


vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -668,7 +683,70 @@ else if (result == JOptionPane.CANCEL_OPTION) { // "Don't Prompt Again" option.
...
+
+      // If the drop item is a Discord file link (or a text URL), we attempt to open the connection and Load That Shit Right Off The Internet
+      if (flavor.isFlavorTextType()) {
+        try {
+          final String text = transferable.getTransferData(flavor).toString();
+
+          try {
+            final URL url = new URL(text);
+            final URLConnection uc = url.openConnection();
+            final InputStream is = new BufferedInputStream(uc.getInputStream());
```
uckelman, line 710: Use a try-with-resources here to avoid leaks:
```
try (InputStream is = uc.getInputStream(), BufferedInputStream bis = new BufferedInputStream(is)) {
```

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -668,7 +683,70 @@ else if (result == JOptionPane.CANCEL_OPTION) { // "Don't Prompt Again" option.
...
+
+            try {
+              loadGameInForeground(text, is);
+            }
+            catch (IOException e) {
+              ReadErrorDialog.error(e, text);
+            }
+            break; // Once we have a successful load, nothing else.
```
uckelman, line 718: Is this because you can't tell beforehand which flavor will be useful?

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -1048,6 +1126,9 @@ public void loadGameInForeground(final String shortName,
     final GameModule g = GameModule.getGameModule();
     g.warn(Resources.getString("GameState.loading", shortName));  //$NON-NLS-1$
 
+    //TODO It would be nice to unblock EDT long enough to have the above loading message get displayed.
+    //TODO I guess it would mean right here we'd need to launch some "do in 10 milliseconds" thing, but then make sure that the below was then executed on EDT?
```
uckelman, line 1130: The problem here is that you're _already_ on the EDT here, so if you post a Runnable to the EDT queue, it won't run until after this method finishes.

What you'd need to do is post your Runnable to display the message, then put the rest of this method into another Runnable that gets posted----but that's assuming that the caller of this method is not waiting on the completion of everything in it, which is an assumption you'd need to verify first.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
Cattlesquat: done

vassal-app/src/main/java/VASSAL/build/module/GameState.java
Cattlesquat: The "transferrables" that we get seem to come in multiple "flavors" and indeed as far as I can tell the same individual thing will sometimes present itself "in several different flavors". 

And of course all manner of shit _might_ get dragged here (totally inappropriate things that aren't files -- here I dragged a paragraph I copied from my Word document) that come in various comprehensible/incomprehensible flavors. 

So what we're doing here is:
(a) If we got some kind of "text paste", see if it has a URL to a loadable game file. The Discord links that you drag in from the chat rooms come with a lot of extra cruft but they have a URL in them you can dig out. 
(b) If we got a "file" paste then see if it's a loadable game file.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
Cattlesquat: The trick here is the whole idea of "loadGameInForeground" is that we *want* to block all the other parts of Vassal from doing anything while we're loading a game over top of the current game (because there were all manner of horrific bugs we fixed this way). But ideally I'd like to be able to tell Swing to "take 10 milliseconds and just put this one last message up first, so the user knows we're attempting the load". I just haven't figured out how *from inside the EDT* to tell it to take a 10 ms break but _then block the EDT while the file loads_. Can you think of any way forward on that?

PR #11094, opened 2022-01-19
11094 - Exclude GamePieceLayout image names from unused image list

BrentEaston:


uckelman:
Verified that it works with user-provided test module.

PR #11106, opened 2022-01-21
Fix NPE when duplicating GamePieceImage

uckelman:
Put the null check in the right place.


Cattlesquat:
Worked for me the module that was crashing before.

PR #11109, opened 2022-01-22
11099 Correct HexGrid range between points within a hex

BrentEaston:
The range in hexes between 2 points that are not in the centres of hexes can be out by 1 hex. In the worst case, 2 the hex range between two non-central points in the same hex can be 1 instead of 0. This affects the range displayed by the LOS thread and also the GKC range check.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/boardPicker/board/HexGrid.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/boardPicker/board/HexGrid.java
@@ -724,7 +724,7 @@ public class HexGrid extends AbstractConfigurable
 
     final VASSAL.build.module.Map m = getContainer().getBoard().getMap();
 
-    // Snap the pounts to the HexGrid before calculating range
+    // Snap the points to the HexGrid before calculating range. Note: snap via the Map to get correct handling of board offets.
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/boardPicker/board/HexGrid.java
```diff
@@ -722,9 +722,17 @@ public int getRawRow(Point p) {
   @Override
   public int range(Point p1, Point p2) {
 
-    final Hex h1 = OffsetCoord.qoffsetToCube(OffsetCoord.ODD, new OffsetCoord(getRawColumn(p1), getRawRow(p1)));
-    final Hex h2 = OffsetCoord.qoffsetToCube(OffsetCoord.ODD, new OffsetCoord(getRawColumn(p2), getRawRow(p2)));
+    final VASSAL.build.module.Map m = getContainer().getBoard().getMap();
 
+    // Snap the pounts to the HexGrid before calculating range
```
uckelman, line 727: points?

BrentEaston: Fixed

PR #11117, opened 2022-01-23
11117 - Log files can be fast-forwarded & appended to

Cattlesquat:
One of my little side-hobbies is making complete playthrough logs of major PoG games (e.g. tournament finals or whatever). 

A major difficulty with making them is that the whole log must be made during a "single session of Vassal" or it won't be a complete playthrough. But a big game of PoG can have e.g. 700 ACTS entries, each containing many piece moves. So the problems this has presented include:
* Have only finished the first 200 entries, but now my system wants to reboot or I want to install a new version of Vassal or whatever. 
* Or I just want to take a break and know I won't lose everything to a power failure.
* Have finished all 700 entries and suddenly discover I don't have enough heap space to contain the whole game, and so I lose ALL the moves.
* If I make a mistake at e.g. move 400, there's no clean starting point to go back to other than 0. 

SO, with that use case in mind, I have created a log "fast-forward and append" tool.
(1) Loads the log you pick
(2) Lets you pick a name for the new "appended" version of the log
(3) Fast forwards the log to the end state you saved it in -- but retains the *original* starting state, and simply transfers the "logged" commands from the old file to the new log.
(4) So now you're starting from the point you left off, and anything new you do will be appended to the log, but if you replay THIS log later you will have the WHOLE game. 
(5) Also you can omit the new (appended) log name if you just want to fast-forward the log to the end -- some people have occasionally mentioned wanting this.


BrentEaston (approved):
Looks reasonable to me

PR #11120, opened 2022-01-23
Permit hyphens and periods in GamePieceImage names

uckelman:


uckelman:
@BrentEaston Does this look ok to you? I don't see any problem with permitting hyphens and periods in GPI names, as those should still be valid filenames on every system.

BrentEaston (commented):
Looks good to me.

PR #11121, opened 2022-01-24
11121 - Fix problem where undo wasn't allowed when starting a new game with logging turned on

Cattlesquat:


riverwanderer:
I’ve seen this bug but wasn’t able to replicate. I guess because I failed to replicate the “with log file open” but.

Cattlesquat:
> I’ve seen this bug but wasn’t able to replicate. I guess because I failed to replicate the “with log file open” but.

Yes that was precisely my repro problem too -- I'd heard various reports of undo being incorrectly not available, but whenever I'd "check" it seemed like it worked -- until I finally hit the correct repro case.

uckelman:
That's hilarious. That's all it took to fix?

Cattlesquat:
Lol yes, it was just a missed code path

PR #11122, opened 2022-01-24
Console can show/set properties of selected pieces

Cattlesquat:
Request from @riverwanderer

By adding "selected" into the command line the properties of currently selected pieces can be shown/set

`/property selected show LocationName`

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Console.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Console.java
@@ -21,16 +21,20 @@ import VASSAL.Info;
 import VASSAL.build.GameModule;
 import VASSAL.build.module.properties.MutableProperty;
 import VASSAL.command.Logger;
+import VASSAL.counters.GamePiece;
+import VASSAL.counters.KeyBuffer;
 import VASSAL.tools.BugUtils;
+import org.apache.commons.lang3.BooleanUtils;
+import org.apache.commons.lang3.math.NumberUtils;
+import org.slf4j.LoggerFactory;
 
 import java.awt.Desktop;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


BrentEaston (commented):
I see no security issues as console commands are only allowed in the 'only ever single user' case.

One thing, many properties, especially trait properties cannot be set, is there anything in the doco to note this since setting a non-existent or unsettable property just fails quietly?

Cattlesquat:
The Console, so far, doesn't much detail anywhere in the html help -- it's only documented under the "/help" command in the console itself -- so I have added text to that.

vassal-app/src/main/java/VASSAL/build/module/Console.java
```diff
@@ -159,31 +163,74 @@ else if (matches("?", option) || matches("help", option) || ("".equals(option)))
...
-      if (propValue != null) {
-        show("[" + property + "]: " + propValue.getPropertyValue());
+    else if (matches("show", option) || matches("set", option)) { //NON-NLS
+
+      final GameModule gm = GameModule.getGameModule();
+
+      if (useSelected) {
+        final List<GamePiece> selected = KeyBuffer.getBuffer().asList();
+
+        if ((selected != null) && !selected.isEmpty()) {
```
uckelman, line 193: You don't need to check if it's non-empty here---the loop just won't loop over an empty list.

uckelman: Oh, there's an `else` down there. Never mind!

uckelman: Wait, no, that `else` is further out in scope.

Cattlesquat: fixed

PR #11124, opened 2022-01-24
Console supports up-arrow / down-arrow to e.g. repeat commands

Cattlesquat:
I keep wanting this totally standard CLI thing where you can hit up/down arrows to scroll around in previous commands you've typed.


riverwanderer:
me too. thanks!

uckelman:
The list can grow without bound?

Cattlesquat:
Yes. Because it's 2022 :)

PR #11135, opened 2022-01-27
11135 - Prompt to save/discard/cancel before loading the new game. Fixed Discord drag-and-drop bug. 

Cattlesquat:
I set out originally to fix the Discord drag-and-drop situation, and I was thinking that would require changing the place we do the save/not-save/cancel prompts. In the course of working on it I discovered the original (Discord link) problem could actually have been fixed without other changes, however given the various discussions we've had recently about the awkwardness of our existing "wait until we've loaded the whole file before asking if you wanted to save the game or cancel" method, and indeed the fact that the Cancel button didn't produce a coherent result in some of the cases, I decided to proceed with the full order change that I'd embarked upon.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/GameState.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameState.java
@@ -456,7 +456,7 @@ public class GameState implements CommandEncoder {
     }
   }
 
-  public static final int NO_NEED_TO_SAVE = ~(JOptionPane.NO_OPTION & 0x01) | ~(JOptionPane.YES_OPTION & 0x02) | ~(JOptionPane.CANCEL_OPTION & 0x04) | ~(JOptionPane.CLOSED_OPTION & 0x08);
+  public static final int NO_NEED_TO_SAVE = (~JOptionPane.NO_OPTION & 0x01) | (~JOptionPane.YES_OPTION & 0x02) | (~JOptionPane.CANCEL_OPTION & 0x04) | (~JOptionPane.CLOSED_OPTION & 0x08);
 
   /**
    * Offers player the chance to save the game if an unsaved one is active & modified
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -435,6 +435,30 @@ protected Command myUndoCommand() {
     }
   }
 
+  public static final int NO_NEED_TO_SAVE = -2;
+
+  /**
+   * Offers player the chance to save the game if an unsaved one is
```
uckelman, line 441: is what?

Cattlesquat: An extra status for the thing to return (in addition to the ones from the dialog box itself)

uckelman: I mean, it's just not a complete thought. Aren't some words missing?

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -661,6 +681,14 @@ public boolean loadGame(File f, boolean continuation) {
         loadContinuation(f);
       }
       else {
+        int optionToSave = NO_NEED_TO_SAVE;
```
uckelman, line 684: This is dangerous. What happens if the values of the `JOptionPane` constants are changed?

vassal-app/src/main/java/VASSAL/build/module/GameState.java
Cattlesquat: Oh right, fixed

vassal-app/src/main/java/VASSAL/build/module/GameState.java
Cattlesquat: I guess _theoretically_ but the chances of suddenly using -2 seem ... effectively zero? Compared to making a bunch of extra lines of code to translate from one scheme to another?

BrentEaston (approved):

PR #11162, opened 2022-02-01
Turn Tracker does not display a spurious blank line if its report format is left empty

Cattlesquat:


riverwanderer:
Is this something that happens when beanshell computes to null?
If so, I've come across similar spurious blank lines from other reports; things like the report that can accompany a Change-Property action.

PR #11178, opened 2022-02-03
11178 - Omit empty lines in non-blank property setting prompts

Cattlesquat:
Purge blank property prompts.

But could someone actually be *relying* on the ability to set a property to blank from one of these menus?


Cattlesquat:
(hold for potential rework-with-checkbox in the event of 3.7)

fanman80:
To allow for the ability of someone to set a property to blank, could the purge of blank lines leave one blank line at the top or bottom of the list while purging the rest?

PR #11209, opened 2022-02-08
Global Options settings for (a) Inventory Security, and (b) Send-to-Location generating move trails. Inventory now defaults to "old behavior" on seeing private windows, but can be turned to secure.

Cattlesquat:


riverwanderer:
Has the case against the original change been proven? The person who [raised](https://forum.vassalengine.org/t/gpiw-now-reports-unknown-piece-instead-of-actual-pieces/73795) it found another way. I'm not sure if the seconder has yet demonstrated his broader concern.

I'll reiterate that one aspect of the security flaw cries to be fixed because the new GPIW zoom opens up an issue that pre-v3.6 module developers will not have anticipated.  My response on the forum thread for reference: [here](https://forum.vassalengine.org/t/gpiw-now-reports-unknown-piece-instead-of-actual-pieces/73795/18?u=marktb1961).

riverwanderer:
For the Global Option on Inventory security to address the legacy issue, this default setting needs to be "Never".
image

Cattlesquat:
@riverwanderer It is presently slated to default to "always" because that's what older versions of Vassal did. So in other words the new enhanced security can be turned *on* but it defaults to off. Our calculation is that would break the fewest modules?

riverwanderer:
I don't follow. GPIW Zoom is a new feature, and is the one causing the debate and changes. Post v3.6 modules are in the minority and developers will be available to turn the feature on if need be. Are we talking at cross-purposes?

Cattlesquat:
The security is also a new feature, and turning it on appeared to break more modules?

Cattlesquat:
See https://github.com/vassalengine/vassal/pull/11216 for Inventory

PR #11215, opened 2022-02-09
11215 - When cancel from Grid edit, don't change Show Grid options

Cattlesquat:


riverwanderer:
Tested VASSAL-3.6.5-SNAPSHOT-f9753fc-CancelGridEdit.  Cancel is still setting the Show Grid checkbox. Worse, Cancel is now ineffective if Save Grid is checked on.  The Cancel/Show Grid change is maybe the wrong way around somehow ?

Cattlesquat:
That was apparently entirely the wrong idea. Need to find somebody who understands these kinds of grid.

PR #11216, opened 2022-02-09
11216 - Inventory has checkbox to enable zoom (defaults off)

Cattlesquat:


riverwanderer:
Running up a module for the first time with this build (in Player, not Editor and Global Options not yet customised); I observe this two things:
1) the GPIW shows Unknown Piece where appropriate.  Expected: this wasn't going to be the default ?
2) each time the GPIW is opened, the Zoom button shows but when I press +, the zoom controls disappear. Expected: Zoom controls not to show at all.
Let me know if you need me to dig in further.

Cattlesquat:
@riverwanderer 
(1) These changes are in a different branch from the Inventory Security stuff, so *presently* in this branch you're getting code that is out of date. When #11209 and this are both merged, then you'd experience this together with that. Short answer - yes it will behave the way you expect it should, in the end.
(2) Okay that's definitely wrong, let me look into that.

Cattlesquat:
@riverwanderer Okay I put in some "repaint" calls when I show/hide, and that *seems* to do the trick when I play with it in my module, so give it a try and see if it works for you.

riverwanderer:
Build VASSAL-3.6.5-SNAPSHOT-2311261 doesn't improve things for me in C&C Napoleonics v4.0.6, I'm afraid. Same as before. On hitting the + button there's a little delay and then the button disappears.

Cattlesquat:
@riverwanderer Okay when I did same repro steps you just described, that happened to me too. The good news is ... well at least nobody gets to actually zoom the window. But apparently I need to bring in Reluctant Swing Expert @uckelman to see why the buttons don't go away when I `setVisible(false)` them.

riverwanderer:
Good for me:  tested VASSAL-3.6.5-SNAPSHOT-19cb270-InventoryZoomCheckbox with different GPIWs (ones where no zoom is configured and ones with). A legacy module - i.e. default Global Option - shows no zoom setting, achieving the goal securing legacy modules.

PR #11218, opened 2022-02-09
Grid Editor incorrectly changes some grid settings on Cancel

uckelman:


riverwanderer:
That works in my test... i.e. cancelling out of the edit does not alter the setting of "Show Grid" anymore.

PR #11233, opened 2022-02-15
11233 - UI Cleanup: don't scroll-at-edge the map if a file or link is being dragged instead of a piece

Cattlesquat:
This just sorts out confusion between different drag-drop handlers -- prevents dragging a file/link onto the map window from triggering any spurious drag-at-edge-of-screen behaviors.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -158,6 +158,7 @@ import java.awt.Rectangle;
 import java.awt.RenderingHints;
 import java.awt.Window;
 import java.awt.datatransfer.DataFlavor;
+import java.awt.datatransfer.Transferable;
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -1971,7 +1967,25 @@ public void dragEnter(DropTargetDragEvent dtde) {
...
+    if (flavors.contains(DataFlavor.javaFileListFlavor)) return;
+
+    if (flavors.contains(DataFlavor.stringFlavor)) {
+      final String string;
+      try {
+        string = (String) dtde.getTransferable().getTransferData(DataFlavor.stringFlavor);
+      }
+      catch (UnsupportedFlavorException | IOException e) {
+        return;
+      }
```
uckelman, line 1982: Since you use this hunk of code to get the string twice, maybe make it a private member function that returns a String or null?

PR #11243, opened 2022-02-21
11242 - Enable Sum and Count in GKC Report Expressions

BrentEaston:


vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -636,17 +636,33 @@ public Object sum(Object src, Object propertyName, Object propertyMatch) {
...
-    final String matchString = replaceDollarVariables((String) propertyMatch, sourcePiece);
-    final PieceFilter filter = matchString == null ? null : new PropertyExpression(unescape(matchString)).getFilter(sourcePiece);
+    final String matchString = replaceDollarVariables((String) propertyMatch, (PropertySource) src);
+    final PieceFilter filter = matchString == null ? null : new PropertyExpression(unescape(matchString)).getFilter((PropertySource) src);
 
-    for (final Map map : getMapList(mapName, sourcePiece)) {
+    if (src instanceof GamePiece) {
+      mapList = getMapList(mapName, (GamePiece) src);
+    }
+    else if (mapName != null && !((String) mapName).isEmpty()) {
```
Cattlesquat, line 652: Is the piece always the source in spite of a passed "mapName"? Or is a mapName sometimes passed on a piece (for some other reason) and doesn't take over as the map but does do other stuff?

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -679,16 +695,34 @@ public Object count(Object src, Object propertyMatch) {
...
-    final PieceFilter filter = matchString == null ? null : new PropertyExpression(unescape(matchString)).getFilter(sourcePiece);
+    final String matchString = replaceDollarVariables((String) propertyMatch, (PropertySource) src);
+    final PieceFilter filter = matchString == null ? null : new PropertyExpression(unescape(matchString)).getFilter((PropertySource) src);
 
-    int result = 0;
-    for (final Map map : getMapList(mapName, sourcePiece)) {
+    if (src instanceof GamePiece) {
+      mapList = getMapList(mapName, (GamePiece) src);
+    }
+    else if (mapName != null && !((String) mapName).isEmpty()) {
```
Cattlesquat, line 712: (same question as above, here)

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
BrentEaston: This is the original behavior (lines 649-651) that I have left unchanged as the first check. The GamePiece, if supplied, is just used as the property source to build the piece Filter and provide the default map to visit. The mapName, usually null, is used to select an alternate Map to visit.

Cattlesquat:
I guess now just needs the OldAndNewProperties fix.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -658,6 +658,9 @@ else if (src instanceof GameModule) {
     else if (src instanceof Map) {
       mapList.add((Map) src);
     }
+    else if (src instanceof ReportState.OldAndNewPieceProperties) {
```
Cattlesquat, line 661: Could that cause an NPE if the piece isn't on a map?

BrentEaston: I've added some general NPE protection further down

uckelman: `null instanceof Whatever` is always false, so it should not.

PR #11256, opened 2022-02-26
Added a preference for the tiler max heap; tiler will retry up to that limit

uckelman:


Cattlesquat:
Is there a place in e.g. Preferences.adoc that it would make sense to describe how to use that?

uckelman:
@Cattlesquat Preferences.adoc seems to be about the preferences for the Player, not the Module Manager. I'm not sure those are described anywhere.

uckelman:
@Cattlesquat The tiler (and converter) prefs are mentioned in the docs now.

PR #11261, opened 2022-02-28
11071 Improve version number checking

BrentEaston:


vassal-app/src/main/java/VASSAL/chat/node/NodeClient.java
```diff
@@ -703,13 +703,19 @@ public void uninitializeControls(ChatServerControls controls) {
    * @param targetRoom Room to be entered
    * @param compatible Return true if entry allowed
    * @param errors Return a list of error message if entry not allowed.
+   * @deprecated Use boolean checkCompatibility(NodeRoom targetRoom, List<String> errors)
    */
+  @Deprecated(since = "2020-03-01")
```
jonathanrwatts, line 708: Should that be "2022", not "2020"?

BrentEaston:
Summary of changes:
1. On-line rooms now include [Vassal version/module version] after name of room.
2. Attempting to join a room with a module that has a different checksum, a different module version number of a different minor (x.y.z) vassal version from the owner of the room generates an 'Are you sure' dialog.
3. Joining a room with a module that has a different checksum, a different module version number of a different minor (x.y.z) vassal version from the owner of the room generates warning messages in Chat Window to all players.
4. Attempting to load a save game created with a different module version number of a different minor (x.y.z) vassal version displays an 'Are you sure' dialog.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -581,47 +581,65 @@ public boolean isSaveMetaDataValid(File file) {
...
+      saveVassalVersion = VersionUtils.truncateToIncrementalVersion(saveData.getVassalVersion());
+      final StringBuilder message = new StringBuilder();
+      message.append(Resources.getString("GameState.load_mismatch_header",  file.getName())).append("\n\n");
 
       if (!saveModuleName.equals(moduleName)) {
-        message = Resources.getString(
-          "GameState.load_module_mismatch",
-          file.getName(), saveModuleName, moduleName
-        );
+        message.append(Resources.getString("GameState.load_module_mismatch", ++problemCount, saveModuleName, moduleName)).append('\n');
```
uckelman, line 602: I think it looks a bit odd displaying `(1)` when there's only one item. Is there a need to number the items? Could we use a bullet point instead?

BrentEaston: Agreed. Updated.

uckelman (approved):

jonathanrwatts:
Am I interpreting [6126b76](https://github.com/vassalengine/vassal/pull/11261/commits/6126b76a1ba654354864712f89726c0f031477ff) correctly and we're no longer checking the checksums against each other, only the module and VASSAL versions/name?  There have been a couple of occasions where my group discovered a corrupted module download because one player's checksum didn't match the others, so I'm thinking we might want to keep checking those, too.  (If I misinterpreted the changes in that commit then never mind.)

BrentEaston:
> Am I interpreting [6126b76](https://github.com/vassalengine/vassal/pull/11261/commits/6126b76a1ba654354864712f89726c0f031477ff) correctly and we're no longer checking the checksums against each other, only the module and VASSAL versions/name? There have been a couple of occasions where my group discovered a corrupted module download because one player's checksum didn't match the others, so I'm thinking we might want to keep checking those, too. (If I misinterpreted the changes in that commit then never mind.)

Checksums are now compared and reported when joining on-line rooms. 

Checksums have never been compared and reported when loading save games (in fact, by accident, no reporting has been done). I may be able to do something though.

jonathanrwatts:
> > Am I interpreting [6126b76](https://github.com/vassalengine/vassal/pull/11261/commits/6126b76a1ba654354864712f89726c0f031477ff) correctly and we're no longer checking the checksums against each other, only the module and VASSAL versions/name? There have been a couple of occasions where my group discovered a corrupted module download because one player's checksum didn't match the others, so I'm thinking we might want to keep checking those, too. (If I misinterpreted the changes in that commit then never mind.)
> 
> Checksums are now compared and reported when joining on-line rooms.
> 
> Checksums have never been compared and reported when loading save games (in fact, by accident, no reporting has been done). I may be able to do something though.

I missed that they were still being compared when joining an on-line room; I think that should be sufficient.

BrentEaston:
Storing the Checksum in save games (or even the module) is actually problematic without some major work. 

No more work from me on this PR.

PR #11279, opened 2022-03-08
11279 - Module "other" strings displayed in About screen

Cattlesquat:
We added these extra "Other" fields a while back but didn't make a way for a downstream user to report them to us/me.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/documentation/AboutScreen.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/documentation/AboutScreen.java
@@ -17,15 +17,6 @@
  */
 package VASSAL.build.module.documentation;
 
-import java.awt.Image;
-import java.awt.event.ActionEvent;
-import java.io.File;
-import java.util.List;
-import java.util.Collection;
-
-import javax.swing.AbstractAction;
-import javax.swing.Action;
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/documentation/AboutScreen.java
```diff
@@ -79,6 +78,18 @@ public void launch() {
...
+
+    final String o1 = g.getModuleOther1();
+    if ((o1 != null) && !o1.isEmpty()) {
+      sb.append(" - ");
+      sb.append(o1);
+    }
+    final String o2 = g.getModuleOther2();
+    if ((o2 != null) && !o2.isEmpty()) {
+      sb.append(" - ");
+      sb.append(o1);
```
jonathanrwatts: `o1` should be `o2` on line 90.

Cattlesquat: I think I got that in the second commit?

jonathanrwatts: Oops, didn't see the 2nd commit.  All good, then.

PR #11286, opened 2022-03-12
11286 - Improved Setup Stack descriptions in editor

Cattlesquat:
Extends the principle I've been using to improve default descriptions of traits.

Displays...

Given Name - Name of Contained Piece - Location

Anything blank is omitted.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
@@ -401,34 +401,31 @@ public class SetupStack extends AbstractConfigurable implements GameComponent, U
   @Override
   public String getConfigureName() {
     final StringBuilder sb = new StringBuilder("");
-    boolean any = false;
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
```diff
@@ -400,17 +400,44 @@ else if (Y_POSITION.equals(key)) {
    */
   @Override
   public String getConfigureName() {
+    final StringBuilder sb = new StringBuilder("");
+    boolean any = false;
```
uckelman, line 404: Is `any` just for tracking if `sb` is non-empty? If so, you could just call `!sb.isEmpty()`.

Cattlesquat: There's no `isEmpty()` method for StringBuilder, but I can use `sb.length() > 0`

PR #11303, opened 2022-03-20
11303 - Preference to control drag-at-edge scroll rate, plus Amazing New Default Powers

Cattlesquat:
I've continued to gradually extend the ability for module designers to set the "default-for-their-module" for various preferences via Global Properties.

image

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -981,7 +981,7 @@ public class Map extends AbstractToolbarItem implements GameComponent, MouseList
 
     // To establish initial defaults, check to see if module designer has overridden the "default defaults" with Global Properties
     final Object delayProp = g.getProperty(DEFAULT_EDGE_SCROLL_DELAY_PROPERTY);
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2154,9 +2205,9 @@ else if (py < vrect.height && py >= vrect.height - dist) {
        */
       @Override
       public void timingEvent(float fraction) {
-        // Constant velocity along each axis, 0.5px/ms
+        // Constant velocity along each axis, 0.5px/ms default
         final long t1 = System.currentTimeMillis();
-        final int dt = (int)((t1 - t0) / 2);
+        final int dt = (int)((t1 - t0) * NumberUtils.max(EDGE_SCROLL_RATE, 0.01));
```
uckelman, line 2210: Is there a reason why `Math.max()` doesn't work here?

Cattlesquat: It will probably work fine, I guess I just got on a tear of using NumberUtils

Cattlesquat: (I did change it to Math, in case that wasn't clear)

PR #11325, opened 2022-03-28
Implement the missing Play Sound trait 'Send sound to other players' functionality.

BrentEaston:


riverwanderer:
Tested a comparison of v3.6.5 and the build with this fix in using C&C Nap module v4.1.0. Works for me (when a user joins a side they get a local 'click', it sounds remotely under Vassal 3.6.5 but not in the fixed build).

PR #11348, opened 2022-04-06
11348 - Polygon Editor improvements (move whole shape, tab between points)

Cattlesquat:
* If SHIFT held down, moves whole shape instead of just one vertex
* If no vertex selected, then arrow keys likewise move whole shape
* TAB between selected vertices
* ESC deselects

Added info on these features to the window legend


uckelman:
Shouldn't Shfit+Tab cycle the other direction?

Cattlesquat:
Shift+TAB now cycles the other way

PR #11362, opened 2022-04-12
Match expanded Inventory nodes by entry value, not by display text

uckelman:


riverwanderer:
Test successfully with Liberté Personal Display inventory.

PR #11404, opened 2022-04-27
Don't show image scale option for panels inside Chart Windows

uckelman:


uckelman:
Closing. #11406 is better.

Issue #11429, opened 2022-05-09
Deck Empty Hotkey repeats per online player instance


riverwanderer:
Stemming my [forum post](https://forum.vassalengine.org/t/is-this-deck-hotkey-working-as-expected/74547/3), I've developed a canned replication of what I think is a bug with "Hotkey to send when Deck empties" affecting online games (at least). The symptom, as illustrated on the forum is duplicate processing of the hotkey, once per connected player.

To replicate, put the attached module into two or more Vassal instances, start it up on one of them and follow the instructions.
[Empty HK test.zip](https://github.com/vassalengine/vassal/files/8651753/Empty.HK.test.zip)

The bug replicates at least back to Vassal v3.5.8.

riverwanderer:
Updated demo module (v0.2) that illustrates a workaround to the issue and also how the Empty Deck hotkey can be used to trigger independent actions local to other player instances (an Alert in the demo).
[Empty HK test.zip](https://github.com/vassalengine/vassal/files/8672190/Empty.HK.test.zip)

Cattlesquat:
Should be fixed by #11963

PR #11435, opened 2022-05-11
11429 Batch Deck Empty Hotkeys with current Command

BrentEaston:
Ensures that any Commands generated by Deck Emptying are fully executed sent to other clients as a singled Command.


BrentEaston:
See issue #11429

BrentEaston:
Superceded by https://github.com/vassalengine/vassal/pull/11453

PR #11453, opened 2022-05-18
11344 & 11429 Prevent Undo triggering Deck Empty Hotkey & Ensure single Undo for all Commands

BrentEaston:


BrentEaston:
Delved too deep. Reverting to simpler changes

PR #11499, opened 2022-06-09
TurnTracker Forward Only Toggle: This commit contains code to allow module  designers to disable the b…

HaraldFair:
…ackwards "-" button on the TurnTracker/TurnCounter. Preventing backwards changes prevents triggering of automation that is designed only for forwards progress, reducing error states during game-play.


HaraldFair:
Can someone give me a brief understanding of what to expect from the process here? This is my first PR for vassal, so I don't know if I am missing something or if I just need to wait.

uckelman:
We haven't had any external PRs in a while. We'd need to review it and maybe reach a consensus on whether this is something we want to merge. It might be some days before that happens, as we've only done a release a week ago.

HaraldFair:
All good...BrentEaston invited me to submit this PR. I just want to make sure you guys aren't waiting on me to do something that I don't know is expected of me. Just assume I'm ignorant if something needs to be done on my end.

vassal-app/src/main/java/VASSAL/tools/IconButton.java
```diff
@@ -33,6 +34,8 @@ public static Color getDefaultColor(int type) {
       return Color.green;
     case CROSS_ICON:
       return Color.red;
+    case GREY_MINUS_ICON:
+      return Color.lightGray;
```
uckelman, line 38: What's the reason for making a new icon? I thought you were trying to disable the back button---which I don't see happening anywhere.

HaraldFair: I've "greyed-out" and disabled the back button so as not to ugly-up the UI. I started by just removing the button altogether, but thought it doesn't look right and re-worked it as a greyed-out control, which, to my eyes, looks better. The disabling code is definitely in there. I've tested it. You will see conditional execution wrapping the button creation process and the "doPrev" method.

HaraldFair: If you run the code, you will see the final setting in the editor is now a boolean check box that sets it to go forward only, disabling the backward functionality at the designer's discretion.

HaraldFair: I've also put in comments to mark my code...search for // FWD_ONLY

uckelman: I don't find a call to `setEnabled(false)` anywhere in the diff. That's what would be needed to disable the back button.

HaraldFair: That's probably what someone who knows Java Swing would do. The way I did it was on line 1025 of  TurnTracker.java.

uckelman: Ok, so the canonical way to do this would be (1) to leave the action triggered by the button (`doPrev()`) unaltered, and (2) call `setEnabled(!fwdOnly)` on the button.

vassal-app/src/main/java/VASSAL/build/module/turn/TurnTracker.java
```diff
@@ -167,6 +167,7 @@ public class TurnTracker extends TurnComponent implements CommandEncoder, GameCo
 
   protected int plusButtonWidth  = DEFAULT_SIZE;
   protected int turnButtonHeight = DEFAULT_SIZE;
+  protected Boolean fwdOnly = false;
```
uckelman, line 170: This should be a `boolean` instead, as you can take advantage of auto-unboxing.

vassal-app/src/main/java/VASSAL/tools/IconButton.java
HaraldFair: Roger that. I will refactor. Please feel free to add anything else to the list. I'll probably push the new commit tonight.

vassal-app/src/main/java/VASSAL/build/module/turn/TurnTracker.java
HaraldFair: doPrev() still has "if" statement to block hotkey activation.

HaraldFair:
Latest commit 26485eb still has "if" statement in "doPrev()" to block inappropriate HotKey activation.

uckelman:
Checking `fwdOnly` in `doPrev()` isn't necessary---you can't activate a disabled button via a hotkey.

HaraldFair:
I disabled the "if" statement and ran the code to test this. The HotKey definitely does trigger the "doPrev" method, overriding the disabled state of the button.

uckelman:
That should not happen. I'll need to look into why it does.

uckelman:
The reason the hotkey is not blocked by disabling the button is that it's not a _button_ hotkey, but a `NamedKeyStroke`, which is a completely different mechanism.

vassalTTF/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+    <groupId>org.example</groupId>
+    <artifactId>vassalTTF</artifactId>
+    <version>1.0-SNAPSHOT</version>
+
+    <properties>
+        <maven.compiler.source>16</maven.compiler.source>
+        <maven.compiler.target>16</maven.compiler.target>
+    </properties>
+    
+</project>
```
uckelman, line 16: Was this file added intentionally? It looks like it doesn't belong in the PR.

HaraldFair: Thanks for asking...I wasn't sure what I was supposed to do with that and figured you'd let me know...Does Maven generate a new one in each branch? Ignore those in the future?

HaraldFair: Unless you know better, it's possible I accidentally copied it from a short-lived experiment. I wasn't sure if Maven built it or where it came from.

uckelman: I've never seen Maven generate a pom before, so I can't comment about how it got there.

HaraldFair: OK, I'll fix it now.

Cattlesquat (approved):
In principle this looks like it's okay and would even be fine in 3.6 (doesn't require 3.7 as it wouldn't "crash" e.g. 3.6.0). I'd wonder if the turn track with only a "+" looks lopsided, but that's a valid module-designer choice.

uckelman:
> with only a "+" looks lopsided

Note that the PR doesn't remove the button, only disables it.

BrentEaston (approved):
Looks good to me and tests well. Happy for this in 3.6.8

HaraldFair:
> The reason the hotkey is not blocked by disabling the button is that it's not a _button_ hotkey, but a `NamedKeyStroke`, which is a completely different mechanism.

I've discovered an important use-case that makes it a good idea to allow the 'NamedKeyStroke' override to happen. If I need to reject the turn-advancement by the wrong player to prevent cheating, I need to be able to rewind the turn counter one step with a GKC. 
All I need to do is remove the "if" statement that I wrapped the "doPrev" method with to allow the "NamedKeyStroke" to take effect. The net effect is allowing the designer to decrement the TurnCounter but not the player if I use a namedCommand.
Shall I make a new commit to this branch? Or, how would you like me to proceed?

Cattlesquat:
@HaraldFair Since this has been merged to master, the best thing is to pull down latest master, make a *new* branch, and commit your changes there, which would you'd then submit as a new PR.

HaraldFair:
Wilco, thanks.

PR #11503, opened 2022-06-13
Make visual gap in stack when dragging piece into expanded stack.

flivni:
### Overview

This PR is an exploration of a new feature whereby a user can drag a pice (or stack) into the middle of another piece (or stack) and see a visual gap where the piece would drop into the stack if dropped.

**Drag piece into expanded stack**
image

More discussion here:
https://forum.vassalengine.org/t/ux-improvement-stack-management-feature-idea/74661/2

**I'm just at the very early stages of learning the Vassal codebase, so this is merely an exploration at this point. Ideally it will grow into something that can be usability tested in order to find out if such a feature is desirable.**

### Current Progress

* You can drag a single piece into an already-expanded stack.
* You cannot yet drag a stack
* You cannot yet insert into the bottom of a stack
* You cannot yet insert underneath an individual piece
* You must move one piece initially before the code will work (haven't looked into why yet).

### Next up / Q's

* Is it correct to detect piece collision from a throttled `AbstractDragHandler::dragMouseMoved`?

* I’m currently using `map.pickAnyPiece` to decide if a piece overlaps with a stack. This has the benefit of not having to write something new. And it could work for a usability test. But, it doesn’t give exactly what I want (I want piece collision). I’d be happy to implement something else, but just wanted to check that my thinking is correct and that I didn’t miss something elsewhere in the codebase that does this (or something similar). I didn’t see any collision detection code.

* Figure out how to drop a piece underneath a stack. I think this will require that a stack to be able to render the bottom piece offset, which looks like it will require a code change.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -96,6 +96,8 @@ import java.awt.event.MouseListener;
 import java.awt.event.MouseMotionListener;
 import java.awt.geom.AffineTransform;
 import java.awt.image.BufferedImage;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
@@ -860,7 +862,7 @@ public class PieceMover extends AbstractBuildable
    * <br>(1) Moves each piece in the {@link DragBuffer} to its proper destination,
    * based on the mouse having been released at point "p" (if multiple pieces
    * were being dragged after being e.g. band-selected, each individual piece's
-   * destination point will vary with the piece's offset from the anhor point
+   * destination point will vary with the piece's offset from the anchor point
    * that started the drag). This also involves removing each piece from any
    * stacks/decks it was part of, and possibly removing it from its old map if
    * it is changing maps.
@@ -1313,6 +1315,8 @@ public class PieceMover extends AbstractBuildable
    */
   @Override
   public void mouseDragged(MouseEvent e) {
+    // FDL: I would prefer to execute the make-stack-gap logic here, since it would be targeted at the appropriate
+    // window/map, but this method is called unreliably (MacOS).
     if (!breachedThreshold && (dragBegin != null)) {
       final Point pt = map.componentToMap(e.getPoint());
       if (map.mapToComponent(Math.abs(pt.x - dragBegin.x)) > GlobalOptions.getInstance().getDragThreshold() ||
@@ -1321,7 +1325,6 @@ public class PieceMover extends AbstractBuildable
       }
     }
   }
-
 
   /**
    * Moves the group of dragged (in the DragBuffer) pieces to the target point (p).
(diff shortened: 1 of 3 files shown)


BrentEaston:
> Is it correct to detect piece collision from a throttled AbstractDragHandler::dragMouseMoved?

That's fine, it's the only way you can do this on the fly.

> I’m currently using map.pickAnyPiece to decide if a piece overlaps with a stack. This has the benefit of not having to write something new. And it could work for a usability test. But, it doesn’t give exactly what I want (I want piece collision). I’d be happy to implement something else, but just wanted to check that my thinking is correct and that I didn’t miss something elsewhere in the codebase that does this (or something similar). I didn’t see any collision detection code.

Yes, that's what it is designed for. Notice that it takes a PieceFinder as a second argument that defines how to find the pieces and supplies a DropTargetSelector. You can supply your own PieceFinder that finds pieces differently.

Caveat to both these points is that Map.findAnyPiece is expensive and the cost is proportional to the number of the units on the Map, so you could find things slowing down significantly on a Map containing may units. Although it looks fine in your demo, how will it perform on a slower machine with a 1000 pieces on the Map? If we go this route, then I would look at loading all the units on the map into a Qtree at the start of the drag and use this to directly find just the pieces close to the drag location. Ideally Maps would maintain their own Qtree of pieces, but I am not sure we have a clean enough interface to guarantee we could keep it consistent. Hmm, looking at it addPiece(), removePiece() and a couple of places where pieces is cleared should make it viable to shoehorn that in. Could vastly speed up things like ranged GKC's as well.

Alternatively, if we tie it to a meta key like Ctrl, then the auto-stack-expanding and piece jump-ups would only be activated when the Ctrl (say) key is depressed. This also solves the problem of Stacks auto-expanding all over the place as you drag a piece around.

> Figure out how to drop a piece underneath a stack. I think this will require that a stack to be able to render the bottom piece offset, which looks like it will require a code change.

It looks like you may be able to just fiddle with the positions array?

flivni:
Awesome info. Thanks!!

What's a "GKC"?

This might be a dumb question, but: How do you do performance analysis / make performance decisions? E.g. let's say I add a QTree to speed up findAnyPiece ... how do you know it is worth it from a user experience perspective. And do you have a good idea where the biggest performance pain points are?

Note: I think I will need some kind of quick lookup for this feature here based on a gut check.

BrentEaston:
We have access to a limited number of Yourkit java profiler licences which works pretty well for general performance profiling tasks and tracking down specific hotspots in the code, I can supply you with a code if you are interested.

For general performance measurement in cases like this, it is usually suck-it and see. I would stress out the system by loading a big scenario in one of the monster WWII games and see how your change works with 1000 pieces on the board. If performance seems satisfactory, it probably isn't worth digging deeper, but if you are getting noticeable lag in drag and drop and stack jinking, then you will need to look deeper.

I have a small qtree library I used in my GTS modules (https://github.com/BrentEaston/vassal-GTS-Series-Code) that I found. I'ts simple, but effective. There are examples of it's use in BlockingTerrainMap.java and TdcMap.java.

You should be able to use QuadTree.searchWithin to locate all pieces with an area surrounding a supplied point, rather than having to check every point on the board.

>  how do you know it is worth it from a user experience perspective

If the lag goes away and the feature is usable, we're good :) 

> where the biggest performance pain points are?

For what you are working on, it is definitely going to be the map.findAnyPiece  has to check every single piece on the mapevery single time it is called. That's why I think the Qtree has potential. GKC's (Global Key Commands) suffer the same problem. Even if you send a GKC to all pieces with 1 hex of the source piece, every single piece on the map has to be checked to see if it is within 1 hex of the source piece. It's one of the few big optimizations left that is feasible in V3 without major disruption.

BrentEaston:
The build failed due to Checkstyle (code layout) violations, reported in the 'Details' link above. A full maven install of vassal-engine on your machine before committing will pick these up locally.

uckelman:
What I can tell you from having done a lot of performance tuning, both of Vassal and of other things, is that measurement is useful in Vassal mostly for finding what's slow---once you've found the problem, it will typically be so egregious that after correcting the problem you can _see_ how much faster things are without needing to profile again. (For contrast, with a lot of other things I've worked on I'm eking out 5% performance improvements, and for that I _do_ need to measure before and after.)

flivni:
> We have access to a limited number of Yourkit java profiler licences which works pretty well for general performance profiling tasks and tracking down specific hotspots in the code, I can supply you with a code if you are interested.

Cool. I saw there is a free trial of this software. I'll take a look at the free trial!

> For general performance measurement in cases like this, it is usually suck-it and see. I would stress out the system [...]

Thanks. Makes sense.

> https://github.com/BrentEaston/vassal-GTS-Series-Code

This appears to be a private or incorrect link. 

> For what you are working on, it is definitely going to be the map.findAnyPiece has to check every single piece on the mapevery single time it is called. [...]

Yeah agreed!

> Even if you send a GKC to all pieces with 1 hex of the source piece [...]

Do you know of a module that does this? (I've only played two different modules in Vassal so far, and I don't think either make use of ranged GKC).

> The build failed due to Checkstyle (code layout) violations, reported in the 'Details' link above. A full maven install of vassal-engine on your machine before committing will pick these up locally.

Great info. I'll look into maven and get the build to pass!

flivni:
> What I can tell you from having done a lot of performance tuning, both of Vassal and of other things, is that measurement is useful in Vassal mostly for finding what's slow---once you've found the problem, it will typically be so egregious that after correcting the problem you can _see_ how much faster things are without needing to profile again. (For contrast, with a lot of other things I've worked on I'm eking out 5% performance improvements, and for that I _do_ need to measure before and after.)

Thanks. Great info!

BrentEaston:
> This appears to be a private or incorrect link.

Send me your email address and I will send an Invite.

> Do you know of a module that does this? (I've only played two different modules in Vassal so far, and I don't think either make use of ranged GKC).

Good question, not sure off-hand. It is quite common, as until recently, it was the only way to send GKC's to other counters in the same stack.

riverwanderer:
Re Ranged GKC, I have used this a few times - mostly it is probably a redundant use, considering that it does not narrow the search. I can root out an example if it will help, probably in Flying Colors or a unpublished version of C&C Napoleonics.

BrentEaston:
> Re Ranged GKC, I have used this a few times - mostly it is probably a redundant use, considering that it does not narrow the search. I can root out an example if it will help, probably in Flying Colors or a unpublished version of C&C Napoleonics.

It's useful where every unit can potentially be affected by the GKC, but you only want those within a certain range to be affected. E.g. A Leader might send a Highlight command to all units within his Command Range.

riverwanderer:
> It's useful where every unit can potentially be affected by the GKC, but you only want those within a certain range to be affected

Yes, that's the type of use I have in the C&C Napoleonics case I mentioned. The (unreleased) "Combat Assist" feature needs to determine if there are 2 or more units/leaders next to the target unit; it sends a command to any such piece within range =1 to increment a counter, once the count =2, the rest will reject the command.

Issue #11515, opened 2022-06-18
Piece dropped into incorrect position in the stack when dragging within a stack to a lower position


flivni:
Repro:

1. Expand stack.
2. Select a piece at a position >= 3.
3. Drag (where the center of the dragged piece is) on the piece at the bottom of the stack (i.e. on the piece at position 1).
4. Notice that the piece is inserted at the bottom of the stack, rather than inserted at position 2.

I am guessing this is a **very minor** issue. But it relates to the drag-drop feature I'm working on where I am trying to make it easy to manipulate an existing stack.

PR #11516, opened 2022-06-18
Fix incorrect index bug.

flivni:
Fix for #11515

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/StackMetrics.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/StackMetrics.java
@@ -664,7 +664,7 @@ public class StackMetrics extends AbstractConfigurable {
           }
         }
         else {
-          if (moving.getParent() == fixedParent && fixedParent != null) {
+          if (moving.getParent() == fixedParent && fixedParent != null && fixedParent.indexOf(moving) < index) {
             index--;
           }
           fixedParent.insert(moving, index);


BrentEaston:
Hi Felix, What's the effect of this bug?

flivni:
In Vassal, when you drop a dragged-piece onto an expanded stack, the dragged-piece is inserted into the stack on **top** of the piece it is over. 

Except, there is a bug: When dragging a piece in its own stack to a lower position in the stack, it is inserted *underneath* the piece it is dropped on.

How does this impact users?

I think few people know how to insert / adjust the stack, and in practice it can be difficult to do accurately even if you know about it. But, it creates inconsistent behavior and perhaps should be fixed just based on that.

The reason I want it fixed is I am separately working on a feature whereby it becomes easy to accurately manipulate the stack. And when that's the case, the inconsistent behavior becomes a more serious issue because when you drag a piece down into a position clearly marked as open to receive the piece, you'll be sad to see it drop into the position below.

Sorry for the long explanation. I tried to describe it more succinctly in the referenced bug, but I guess it wasn't clear enough.

Here's a short video showing the behavior:
https://www.loom.com/share/170f3f329319487d9026103a2ae8d961

BrentEaston:
I seriously doubt that anyone has ever noticed this, or will be bothered by the change. I am happy for this to proceed.

BrentEaston (approved):

PR #11521, opened 2022-06-20
BoundsTracker - Track bounds and repaint just for the regions of added pieces; rather than the whole map.

flivni:
BoundsTracker was always repainting the entire map. This change has BoundsTracker repaint just the regions of the map that the added Pieces report.

### Discussion

**User Impact**
I don't know the general user impact of this fix. BoundsTracker is not used that many places. 

**Why this PR?**
I am working on a "stack slots" feature and was using BoundsTracker to track changes to the stack. The feature repaints the screen frequently in the case where you're "scrubbing" a piece over an expanded stack and slots are appearing/disappearing rapidly. 

The code was performing in a slow and glitchy way.

When I made the adjustment in this PR, the code performed quickly again.

So, in the case where you're using BoundsTracker and calling repaint often, there would be a big benefit in this adjustment.

**Risks**
This code likely exposes new bugs because it requires that (a) BoundsTracker is used properly, and (b) `Piece.getBoundingBox()` accurately contains all things within the piece/stack/deck.

**In closing**
I can work around not having this PR accepted because I can do my own bounds tracking quite easily without the use of this class.

But, on the other hand, after doing some more general Vassal perf testing, I have a strong hunch that using region updates  in more places (rather than full map updates) could very well result in big performance gains. In order to do this, bounds tracking and getBoundingBox will both need to be done properly, so getting the kinks worked out here might be a step in that direction.

**Discussion in Discord**
There is a discussion of additional related ideas in Discord, Development/General channel, Sunday June 20 4:00 UTC.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
@@ -18,11 +18,11 @@
 package VASSAL.counters;
 
 import java.awt.Rectangle;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.LinkedList;
 
 import VASSAL.build.module.Map;
-import VASSAL.tools.lang.Pair;
+import org.apache.commons.lang3.tuple.Pair;
 
 /**
  * Records the bounding boxes of GamePieces.  Use addPiece() to
@@ -31,10 +31,10 @@ import VASSAL.tools.lang.Pair;
  * added pieces belonged.
  */
 public class BoundsTracker {
-  private final List<Pair<Rectangle, Map>> regions;
+  private final List<Pair<Map, Rectangle>> regions;
 
   public BoundsTracker() {
-    regions = new LinkedList<>();
+    regions = new ArrayList<>();
   }
 
   public void clear() {
@@ -45,13 +45,13 @@ public class BoundsTracker {
     if (p.getMap() != null) {
       final Rectangle region = p.boundingBox();
       region.translate(p.getPosition().x, p.getPosition().y);
-      regions.add(new Pair<>(region, p.getMap()));
+      regions.add(Pair.of(p.getMap(), region));
     }
   }
 
   public void repaint() {
-    for (final Pair<Rectangle, Map> rmPair : regions) {
-      rmPair.second.repaint(rmPair.first);
+    for (final Pair<Map, Rectangle> mrPair : regions) {
+      mrPair.getLeft().repaint(mrPair.getRight());
     }
   }
 }
--- a/vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
@@ -17,10 +17,12 @@
  */
 package VASSAL.counters;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.awt.Rectangle;
+import java.util.List;
+import java.util.LinkedList;
 
 import VASSAL.build.module.Map;
+import VASSAL.tools.lang.Pair;
 
 /**
  * Records the bounding boxes of GamePieces.  Use addPiece() to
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
```diff
@@ -17,10 +17,12 @@
...
 package VASSAL.counters;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.awt.Rectangle;
+import java.util.List;
+import java.util.LinkedList;
 
 import VASSAL.build.module.Map;
+import VASSAL.tools.lang.Pair;
```
uckelman, line 25: Most places we've switched over to `org.apache.commons.lang3.tuple.Pair`.

vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
```diff
@@ -29,25 +31,27 @@
  * added pieces belonged.
  */
 public class BoundsTracker {
-  private final Set<Map> maps;
+  private final List<Pair<Rectangle, Map>> regions;
 
   public BoundsTracker() {
-    maps = new HashSet<>();
+    regions = new LinkedList<>();
```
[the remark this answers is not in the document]
uckelman: Or possibly still a `HashSet`, because that gets you deduplication.

vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
```diff
@@ -29,25 +31,27 @@
...
-    maps.clear();
+    regions.clear();
   }
 
   public void addPiece(GamePiece p) {
     if (p.getMap() != null) {
-      maps.add(p.getMap());
+      final Rectangle region = p.boundingBox();
+      region.translate(p.getPosition().x, p.getPosition().y);
+      regions.add(new Pair<>(region, p.getMap()));
```
uckelman, line 48: `Pair.of()` is nice for this.

vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
```diff
@@ -29,25 +31,27 @@
...
     }
   }
 
   public void repaint() {
-    for (final Map m : maps) {
-      m.repaint();
+    for (final Pair<Rectangle, Map> rmPair : regions) {
+      rmPair.second.repaint(rmPair.first);
```
uckelman, line 54: It reads a bit oddly to me to have the pair go from small to large; mabye put the `Map` first?

vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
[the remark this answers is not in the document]
uckelman: Or, maybe a `HashMap<Map, ArrayList<Rectangle>>`? Depends on just how many elements there are typically. If these things can be huge, this will be more efficient. If small, then the array of pairs is likely better.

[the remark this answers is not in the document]
flivni: I think using a HashMap is not necessary for performance reasons. Might be nice to see your regions separated by map ... but it does make the code more complex.

vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
flivni: Fair enough. In my head I was calling the tuple a "region-map" rather than a "map-region", so wanted the pair to be region then map. But the variable name ended up clearer as `rmPair` rather than `regionMap`, and there is no real difference between `rmPair` and `mrPair`.

flivni: Well, other than future readers might think mrPair is "Mr. Pair" ;)

vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
flivni: @uckelman Did you not switch everywhere because modules can rely on existing code that uses `VASSAL.tools.lang.Pair`?

uckelman: This was a while ago now, so my memory is a bit fuzzy---but I think I switched out all the purely internal usage of `VASSAL.tools.lang.Pair` and left the ones which are in signatures of public methods. If I'd been expecting a _lot_ more longevity with V3 codebase than I am, I would likely have deprecated `VASSAL.tools.lang.Pair` as well and added some methods with signatures using `org.apache.commons.lang3.tuple.Pair`.

vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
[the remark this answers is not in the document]
uckelman: You're probably right that these lists tend to be small.

vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
uckelman: I was working with a Map-Reduce application some years ago where _everything_ was Mr _Something_.

PR #11524, opened 2022-06-21
11524 - Option to 'always use format' on text label (even if label is blank)

Cattlesquat:
Heretofore, a TextLabel where the label is currently empty/blank did not run its Name Format evaluation (just left the name of the piece alone). 

However I have discovered some situations where I want to change the name of a piece based on a property, and using a "dummy" text label that otherwise doesn't display (just runs its name format) is presently the only workaround available (besides typing the name of every piece a whole extra time in a whole different place, effectively "duplicating code"). So this checkbox allows me to use TextLabel as a workaround.


Cattlesquat:
@uckelman This adds an extra value into a trait ... I have vague memories I'm supposed to mention that to you for your converter.

Cattlesquat:
image
Example use case -- An otherwise dummy TextLabel trait is used to commandeer the name of the piece. Which otherwise, irritatingly, can't be set using any kind of "substitution scheme". In some bright future we would offer a separate trait or whatever, but meanwhile... workarounds!

PR #11531, opened 2022-06-24
11523 Area of Effect improvements

BrentEaston:
- Add Active property. 
- Add local/global visibility. 
- Add separate on/off Key Commands

Adds several new fields to the trait.


vassal-app/pom.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
Cattlesquat, line 2: Is this file supposed to be in the PR?

uckelman: Pretty sure not.

uckelman: It's just a whitespace change, so doesn't affect anything.

PR #11536, opened 2022-06-25
11534 - Add Open all Contained Pieces option

BrentEaston:


uckelman:
If I have a thousand pieces in a palette, is this going to open a thousand dialogs when I select it?

BrentEaston:
Absolutely! Part of this change does move the renamed 'Multi-Piece Editor' and the new 'Open All Contained Pieces' options down to the bottom of the item menu to make them look less attractive options but yes, if someone really wants to open masses and masses of pieces, this will do it.

riverwanderer:
> if someone really wants to open masses and masses of pieces, this will do it.

When MacOS is about to do the equivalent (open a huge selection of files), it asks "do you really want to do this?" Could 'Open All Contained Pieces' do similar ?

BrentEaston:
Yep, sounds good.

BrentEaston:
Added a dialog 'This will open up to X separate Piece Editor windows. Are you sure you wish to continue with this action?' if there are 40 or more pieces in the current component.

PR #11539, opened 2022-06-25
11538 - Fix revert of translated text on Tab Widget tabs when changing tabs

BrentEaston:


BrentEaston:
Confirmed by user that this fixes the issue on both Player and induced issue in Editor

BrentEaston:
Merged into other Translation fixes in PR #11542

PR #11541, opened 2022-06-26
Add hotkey to start specific chess clock

andoroth:
Reworked ChessClockControl and ChessClock to have a configurable hotkey for starting a specific ChessClock
Reworked Stop Clock option to stop just the specified clock


vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -996,13 +1037,12 @@ public void actionPerformed(ActionEvent e) {
...
       Command c = new NullCommand();
       final ChessClockControl ccc = ChessClockControl.getInstance();
-      if (command.contains(CHESSMENU_START) || command.contains(CHESSMENU_STOP)) {
-        if ((ccc != null) && ccc.getClocksTicking() > 0) {
-          c = c.append(ChessClockControl.getInstance().startNextClock());
-        }
-        else {
-          c = c.append(updateState(true));
-        }
+      if (ccc == null) return;
```
Cattlesquat, line 1040: The fact that `ccc.getClocksTicking()` is no longer checked seems to change the behavior in all cases rather than in some specific new case?

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -53,16 +53,9 @@
 import VASSAL.command.Command;
 import VASSAL.command.CommandEncoder;
 import VASSAL.command.NullCommand;
-import VASSAL.configure.ColorConfigurer;
-import VASSAL.configure.ComponentDescription;
-import VASSAL.configure.Configurer;
-import VASSAL.configure.ConfigurerFactory;
-import VASSAL.configure.IconConfigurer;
-import VASSAL.configure.TranslatableStringEnum;
+import VASSAL.configure.*;
```
Cattlesquat, line 56: I think our style-thing doesn't let us use .* imports (I guess the auto-build will probably soon tell you that)

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -657,7 +662,8 @@ public void removeFrom(Buildable parent) {
    */
   @Override
   public String[] getAttributeNames() {
-    return new String[] { DESCRIPTION, SIDE, BUTTON_TEXT, ICON, TICKING_BACKGROUND_COLOR, TICKING_FONT_COLOR, TOCKING_FONT_COLOR, TOOLTIP };
+    return new String[] { DESCRIPTION, SIDE, BUTTON_TEXT, ICON, START_HOTKEY, TICKING_BACKGROUND_COLOR,
+            TICKING_FONT_COLOR, TOCKING_FONT_COLOR, TOOLTIP };
```
Cattlesquat, line 666: Did an extra newline work its way in here somehow?

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -682,7 +689,8 @@ public String[] getAttributeDescriptions() {
    */
   @Override
   public Class<?>[] getAttributeTypes() {
-    return new Class[] { String.class, PlayerSidesConfig.class, String.class, IconConfig.class, ColorConfig.class, ColorConfig2.class, ColorConfig3.class, String.class };
+    return new Class[] { String.class, PlayerSidesConfig.class, String.class, IconConfig.class, NamedKeyStroke.class,
+            ColorConfig.class, ColorConfig2.class, ColorConfig3.class, String.class };
```
Cattlesquat, line 693: extra newline?

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
```diff
@@ -1046,16 +1086,14 @@ public void popupMenuWillBecomeVisible(PopupMenuEvent evt) {
       });
 
       JMenuItem item;
-      final String s;
-
-      if (clockTicking) {
-        s = CHESSMENU_STOP;
+      if (isTicking()) {
+        item = new JMenuItem(CHESSMENU_STOP);
```
Cattlesquat, line 1090: If we've added a start clock hotkey, should we also add a "stop clock" hotkey while we're at it? (just a thought)

Cattlesquat: (or maybe that's what pause-all-clocks-hotkey does, in which case maybe we should display that hotkey here?)

Cattlesquat (commented):
Looks plausible to me - I added a couple questions

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
andoroth: May as well.  At the moment, it would be effectively identical functionally to making a call to pause-all-clocks, but there's no reason not to add support to multiple ticking clocks while we're here

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
andoroth: It does - as written they seem like placeholders, as both start and stop commands are undifferentiated overloads of the same functionality available under hotkey control in ChessClockControl.  As written, they provide clearly differentiated functionality from what's available from the Control and more closely match the described functionality.  The intent is that the end state after a call to a ChessClock menu option function will be independent of other clock states, while ChessClockControl commands handle multi-clock and state dependent calls.

Cattlesquat: That seems like it could be confusing to users who are expecting this to work "like a physical chess clock" where starting one clock necessarily stops the other and vice versa. I realize there may be other use cases where different functionality is desired (e.g. can force one clock to start/stop independently of whether another is), but it seems like that might be better addressed by creating additional menu options (e.g. "Start only this clock", or "Force-start only this clock" or something like that) and leaving the other ones working "like a physical chess clock".

andoroth: Interesting - the original impetus of this update was to make the clock work more like a physical chess clock.  Normally, if you keep hitting your clock button, it will never swap the clock back to your side - it always sends it over to your opponent.  Being able to just tap 'your' chess clock button doesn't require you to double check who the clock is now on - it is always going to be the opponent.  I suppose it ultimately doesn't matter what the options are labeled as long as they're all exposed to the user, so whatever is less confusing makes sense.

BrentEaston:
You've got some checkstyle errors:

Error:  src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java:[927,21] (whitespace) WhitespaceAround: WhitespaceAround: '{' is not preceded with whitespace.
Error:  src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java:[940,21] (whitespace) WhitespaceAround: WhitespaceAround: '{' is not preceded with whitespace.
Error:  src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java:[1070,45] (whitespace) ParenPad: ')' is preceded with whitespace.
Error:  src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java:[1071,11] (indentation) Indentation: 'if' child has incorrect indentation level 10, expected level should be 8.
Error:  src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java:[1074,11] (indentation) Indentation: 'if' child has incorrect indentation level 10, expected level should be 8.

vassal-app/src/main/java/VASSAL/build/module/chessclockcontrol/ChessClock.java
Cattlesquat: I was mainly thinking about the ability to have more than one clock running at once (which doesn't happen with a physical clock). But likewise the ability to stop all the clocks -- which you can do -- nonetheless seems like a separate/different function.

andoroth:
I'm not sure if I'm doing something wrong here, or if the build checks are just failing on mainline code.

BrentEaston:
You're getting a PMD violation now. You can run the PMD check directly in Intellij through the Maven menu:

image

The actual rule that has failed is

PMD Failure: VASSAL.build.module.ChessClockControl:593 Rule:LocalVariableCouldBeFinal Priority:3 Local variable 'mySide' could be declared final.

andoroth:
I'm not sure what's configured wrong, but when I try to run any checks or do a maven build the build immediately fails despite being able to build and run just fine through the run configuration.

BrentEaston:
How are you running the Maven builds? I usually run the install target on the root Vassal Engine module. That should do everything. Also make sure you have run a Maven refresh every time the pom.xml is changed, which is the recycle icon just below my cursor in the previous image

andoroth:
Is there anything else I should rework at this point?

vassal-app/src/main/java/VASSAL/build/module/ChessClockControl.java
```diff
@@ -563,6 +585,79 @@ public Command startNextClock() {
...
+    System.out.println(mySide);
+
+    // If we don't have any clocks, can't start one.
+    // If there are not exactly 2 clocks, this function is invalid
+    // If player doesn't belong to a side, they don't have an opponent to send to
+    if (chessclocks.isEmpty() || chessclocks.size() != 2 || mySide == null || mySide.contains("observer")) {
+      return command;
+    }
+
+    Boolean clockTicking = false;
```
uckelman, line 603: `boolean` should always be used in preference to `Boolean` if no object is needed.

PR #11557, opened 2022-06-30
Auto-select contents of Edit boxes when they get focus

Cattlesquat:
In the "making things work like good/normal UI" department along the lines of the Enter/ESC stuff -- this will now automatically select the contents of most edit boxes when the edit boxes gain focus (so when you tab to a field or click in it, you can immediately type a replacement value for the text without having to manually backspace over it all). This is "fairly normal app behavior".

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
@@ -36,26 +36,6 @@ import VASSAL.tools.LaunchButton;
 import VASSAL.tools.NamedKeyStroke;
 import VASSAL.tools.swing.SwingUtils;
 
-import java.awt.BorderLayout;
-import java.awt.Component;
-import java.awt.Dimension;
-import java.awt.Frame;
-import java.awt.GridBagConstraints;
-import java.awt.GridBagLayout;
-import java.awt.Insets;
-import java.awt.Point;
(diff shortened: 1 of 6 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
```diff
@@ -430,6 +429,19 @@ public LevelConfigurer(final Zoomer z, String key, String name) {
       levelField.setMaximumSize(new Dimension(
         Integer.MAX_VALUE, levelField.getPreferredSize().height));
 
+      // Edit box selects all text when first focused
+      levelField.addFocusListener(new java.awt.event.FocusAdapter() {
+        @Override
+        public void focusGained(java.awt.event.FocusEvent evt) {
+          SwingUtilities.invokeLater(new Runnable() {
```
uckelman, line 436: Does this need to be in an `invokeLater`? IIRC, your listeners for Swing GUI elements are executed from the EDT already.

Cattlesquat: The place where I got this code from specifically had some reason for doing it "later but on EDT" that sounded correct but now I've forgotten. There's something we're making sure it gets done *after* basically.

Cattlesquat: image

PR #11568, opened 2022-07-06
tweak to allow designer control of TC backwards function when disabled

HaraldFair:
This adjustment to the disabled TurnCounter "-" behavior allows the designer to maintain control if desired. Previously, the "-" button functionality was observed to be able to be overridden by HotKeys and could lead to accidental player control and was therefore bypassed. However, there are valid use cases where the designer would need to exercise the "-" function while denying it to the player. The bypass is therefore removed in this adjustment, returning control to the designer.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/turn/TurnTracker.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/turn/TurnTracker.java
@@ -1020,9 +1020,9 @@ public class TurnTracker extends TurnComponent implements CommandEncoder, GameCo
     }
 
     protected void doPrev() {
-        captureState();          
-        prev();
-        save();
+      captureState();
+      prev();
+      save();
     }
 
     protected void initComponents() {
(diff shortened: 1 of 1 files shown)


HaraldFair:
Hi, I submitted this PR 20 days ago. No one has commented on it. So, I just wanted to make sure that the devs are aware of it. If additional explanation is needed, I'm happy to provide it.

uckelman:
I'm traveling right now and have unpredictable availability through 31 July. It's not clear to me whether I'll have a block of time to review this before then. We will review this before we release 3.6.8.

HaraldFair:
> I'm traveling right now and have unpredictable availability through 31 July. It's not clear to me whether I'll have a block of time to review this before then. We will review this before we release 3.6.8.

Thank you, and no problem at all. I just wanted to make sure it didn't get missed somehow. There is no urgency on my end. I have the code I need running as a custom class.

PR #11575, opened 2022-07-07
Add AOE bounds to `boundingBox()` method.

flivni:
There is a bug related to boundstracking reported by @Amàndil Eremanth.

The code fix in this PR makes sense to me and should fix the issue. 

**Bug repro steps**
Use 3.6.8 build (from master ok).

1. Load up his module
2. New Game
3. Select "Jugador 1"
4. Select "Adoquinado 1"
5. Click "EJERCITOS" button.
6. Select "Sauron" and drag piece to map
7. Click on piece to select.
8. Move mouse off to the side (away from where the AOE will appear).
9. Click `M` on the keyboard to display the AOE.

You'll see that AEO is only partially displayed (only within the bounding box of the piece).

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/AreaOfEffect.java
+++ b/vassal-app/src/main/java/VASSAL/counters/AreaOfEffect.java
@@ -168,11 +168,10 @@ public class AreaOfEffect extends Decorator implements TranslatablePiece, MapSha
 
   @Override
   public Rectangle boundingBox() {
-    // TODO: Need the context of the parent Component, because the transparency is only drawn
-    // on a Map.View object.  Because this context is not known, the bounding box returned by
-    // this method does not encompass the bounds of the transparency.  The result of this is
-    // that portions of the transparency will not be drawn after scrolling the Map window.
-    return piece.boundingBox();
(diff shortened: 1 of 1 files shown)


BrentEaston:
Fixes my test example

flivni:
Should `getShape(..)` include the AOE shape?

BrentEaston:
Almost certainly not. I believe getShape() is the shape used to determine selection highlighter when a piece is selected and also used when drag and dropping to build the drag cursor and to check if a piece is a merge target when dropping. We don't want any  of these functions to include the AoE shading which is purely visual, it has no effect on the operation of the counter.

flivni:
> Almost certainly not. I believe getShape() is the shape used to determine selection highlighter when a piece is selected and also used when drag and dropping to build the drag cursor and to check if a piece is a merge target when dropping. We don't want any of these functions to include the AoE shading which is purely visual, it has no effect on the operation of the counter.

Of course. I was just making a small joke. (Looks at shoes.)

uckelman:
Right, so the thing I forgot when I tried to fix this is that the area is _not_ translated by the piece's actual position.

PR #11576, opened 2022-07-09
Fix bug in BoundsTracker: Not correctly tracking the redraw bounds of pieces that are within a stack

flivni:
BoundsTracker was recently updated to trigger a redraw just part of the map (rather than the entire map).

This revealed some bugs and oddities. One of these bugs/oddities was the fact that when a piece reports its bounds via `boundingBox()` it does not include its offset position in the stack. In some cases we want the offset position and in some cases not. There were a few options considered:

(1) repaint both the boundingBox of the GamePiece AND the boundingBox+stackOffset of the GamePiece.

(2) have each GamePiece report whether its draw area is offset with the stack or not, perhaps defaulting to true.   Then you could have a boundingBoxWithStackOffset() method or even boundingBox(withStackOffset: true) which would include the stack-offset when needed.

(3) track/repaint the parent stack, not the individual GamePiece.

(4) rethink everything ;)

@uckelman voted for option 3. It seems like a great option to me and it is what is implemented here.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
@@ -44,9 +44,11 @@ public class BoundsTracker {
   public void addPiece(GamePiece p) {
     if (p.getMap() != null) {
       if(p.getParent() != null) {
-        // NB: We track the Stack if there is one. This is because individual pieces within a Stack do not include
-        // their stack-offsets in `boundingBox()` and so the repaint would not include these offsets. This matters
-        // for rotate-piece (for example). Note: other effects (like area-of-effect) are not offset from the Stack.
+        // NB: We track the Stack if there is one. This is because individual pieces
+        // within a Stack do not include their stack-offsets in `boundingBox()` and
+        // so the repaint would not include these offsets. This matters for
+        // rotate-piece (for example). Note: other effects (like area-of-effect) are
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
```diff
@@ -43,6 +43,14 @@ public void clear() {
 
   public void addPiece(GamePiece p) {
     if (p.getMap() != null) {
+      if(p.getParent() != null) {
```
uckelman, line 46: Checkstyle is complaining about the lack of a space after the `if`.

flivni: Yeah, sorry. I should be better about doing a maven build before committing. New habits.

vassal-app/src/main/java/VASSAL/counters/BoundsTracker.java
```diff
@@ -43,6 +43,14 @@ public void clear() {
 
   public void addPiece(GamePiece p) {
     if (p.getMap() != null) {
+      if (p.getParent() != null) {
+        // NB: We track the Stack if there is one. This is because individual pieces
```
flivni, line 47: @uckelman An alternative to writing this comment would be to create a bug in Github and then just reference it there "for more detail". Not sure if you like comments like this in the code or not.

PR #11600, opened 2022-07-19
Player context menus: Show keyboard shortcuts right-aligned and with symbols.

flivni:
Making this available so @uckelman can easily see the results in different modules.

Discussion here:
https://forum.vassalengine.org/t/right-justify-context-menu-commands-v3-6-8/75000/18

**IMPORTANT -- Code does not change the tooltip for menu buttons**

This code does **not** effect the display of keyboard shortcuts for the menu buttons. It does not look possible / easy to update the menu-bottons to show symbols. I think this code is an improvement on what was there previously, but perhaps introducing this mismatch between the menu-buttons and the context menus is worse.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
@@ -31,6 +31,7 @@ import javax.swing.JMenu;
 import javax.swing.JMenuItem;
 import javax.swing.JPopupMenu;
 import javax.swing.KeyStroke;
+import javax.swing.UIManager;
 import javax.swing.event.PopupMenuEvent;
 import javax.swing.event.PopupMenuListener;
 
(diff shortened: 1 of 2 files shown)


Cattlesquat:
nice!

vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
```diff
@@ -107,8 +109,15 @@ protected static String getMenuText(KeyCommand keyCommand) {
   // This both eliminates duplicate code AND makes this critical menu-building functionality able to "play well with others".
   // Menu text & behavior can now be custom-classed without needing to override the monster that is MenuDisplayer#createPopup.
   protected static JMenuItem makeMenuItem(KeyCommand keyCommand) {
-
+    // We remember the oldAcceleratorFont so that we can set it back after creating the JMenuItem
+    final Object oldAcceleratorFont = UIManager.get("MenuItem.acceleratorFont");
+    UIManager.put("MenuItem.acceleratorFont", POPUP_MENU_FONT); // This needs to be set prior to creating the JMenuItem
     final JMenuItem item = new JMenuItem(keyCommand.isMenuSeparator() ? MenuSeparator.SEPARATOR_NAME : getMenuText(keyCommand));
+    if (!NamedKeyManager.isNamed(keyCommand.getKeyStroke())) { // If the KeyStroke is named, then there is no accelerator
```
flivni, line 116: @BrentEaston I think you're the person to ask... is this the best way to detect if keystroke is not set?

The older code was using `NamedKeyManager.isNamed(...)` to detect if the keystroke was named, and if so, was not adding the keyboard shortcut to the menu. This is the approach I used here. But it seems weird to me to do this (and rely on the fact that a key that is not named should have a code) rather than explicitly checking directly if the KeyStroke has not been set.

Is there a way to check if the KeyStroke is valid directly?

I looked, expecting to see some kind of sentinel value, like `NamedKeyStroke.NULL_KEYSTROKE`. But I don't see that. Instead I see a keyCode of `57344`. What does this represent? Can it be used as a sentinel value?

Thanks!

uckelman: The relevant code is this:
```
  public static boolean isNamed(KeyStroke k) {
    if (k == null) {
      return false;
    }
    final int code = k.getKeyCode();
    return code >= NAMED_START && code <= NAMED_END;
  }
```
A `NamedKeyStroke` has a code within the range `[NAMED_START,NAMED_END]`.

The only way to check if `KeyStroke` is a _real_ one is to use `NamedKeyManager.isNamed()`, because that's the only thing that knows what `NAMED_START` and `NAMED_END` are.

PR #11689, opened 2022-09-17
Improve poms

JaviVilarroig:
This is just a little cleanup in the POM files to remove some maven Warning.

### DIFF
--- a/vassal-app/pom.xml
+++ b/vassal-app/pom.xml
@@ -9,7 +9,6 @@
     </parent>
 
     <artifactId>vassal-app</artifactId>
-    <version>3.6.8-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <name>${project.artifactId}</name>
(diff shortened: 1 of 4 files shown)


vassal-app/pom.xml
```diff
@@ -409,7 +408,6 @@
             <plugin>
               <groupId>org.codehaus.mojo</groupId>
               <artifactId>exec-maven-plugin</artifactId>
-              <version>3.1.0</version>
```
uckelman, line 412: Is this really duplicate? Does the other specification actually fix the version _here_?

JaviVilarroig: Yes, absolutely.

I the parent POM file, in the PluginsManagement section, line 142, We are already specifying what version of the plugin must be used in all the projects.

That's that way to allow that every module uses the same plugin versions, unless overriden.

uckelman:
Looks good. Thanks for the contribution.

uckelman:
I've added you to the contributors list, which will show up the next time the site is updated.

JaviVilarroig:
Small contribution :)

Thanks!

PR #11700, opened 2022-09-22
11700 - Irregular Grid editor - add region with Insert or Ctrl+Rightclick

Cattlesquat:
Since the train hasn't left the station, here's a quick editor improvement (fairly thoroughly tested on my end)

When you need to add e.g. 100 locations to an irregular grid, going through a context menu every time gets a little old, so added a couple direct shortcuts.


riverwanderer:
I see that on MacOS Cmd+RightClick works as equivalent to Ctrl+RightClick for this feature, which is great.

I haven't found any equivalent to the INS key on a MacOS keyboard, so I'm not sure that part will apply for Mac users.

uckelman (approved):

PR #11716, opened 2022-10-07
Dereference symbolic link to executable before attemping to find install dir

glciampaglia:
On Linux, I have noticed that VASSAL will not start if the executable script (VASSAL.sh) is run through a symbolic link. 

To reproduce the issue, unpack VASSAL into any location (I used /opt in my case), then create a symbolic link to `/opt/<VASSAL FOLDER/VASSAL.sh`, e.g. with the latest version:
```sh
ln -s /opt/VASSAL-3.6.7/VASSAL.sh vassal
```
Then call the symbolic link:
```sh
./vassal
```
You will get the following error:
```
Error: java is too old to run this version of Vassal. Please use Java 11 or later.
```
I noticed it because I was running java 11.0.8 so the error message could not be possibly correct.

The reason for this behavior is that the executable script (VASSAL.sh) will attempt to identify the installation directory a combination of `which` and `dirname`, but this will work only if calling VASSAL.sh directly, not if (like in my case) one calls a symbolic link to it. 

This commit should fix that by attempting to dereference any symbolic link prior to using dirname.

### DIFF
--- a/dist/VASSAL.sh
+++ b/dist/VASSAL.sh
@@ -15,8 +15,10 @@ if [ ! -x "$JAVA" ]; then
   exit 1
 fi
 
-# Find absolute path where VASSAL is installed
-INSTALL_DIR=$(cd "$(dirname "$0")"; pwd)
+# Dereference any possible symbolic link to executable script, then find
+# absolute path where VASSAL is installed 
+EXEC_PATH=$(readlink "$0")
+INSTALL_DIR=$(dirname ${EXEC_PATH})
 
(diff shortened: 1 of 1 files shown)


uckelman:
Does this work for you using `realpath` instead of `readlink`? IIRC, `readlink` only resolves one level of symlinks, whereas `realpath` resolves them recursively.

glciampaglia:
Good point! Yes, `realpath` would be a better choice over `readlink`. I've updated the PR; please take a look.

PR #11781, opened 2022-11-23
Set PlaceMarker dialog buttohn hotkeys using setMnemonic()

uckelman:


uckelman:
setMnemonic() doesn't provide any visual indication on Windows, which was mostly the point of doing it. Closing.

PR #11790, opened 2022-11-25
Use Java 19 for GitHub builds

uckelman:


uckelman:
In order to do this, we also have to update to Java 19, otherwise jlink fails. We can't update to Java 19 until there's a 64-bit ARM Windows build of it...

PR #11792, opened 2022-11-25
Update launch4j to 3.50

uckelman:


uckelman:
This is broken until [this bug](https://sourceforge.net/p/launch4j/bugs/227/) is fixed in a release.

Issue #11564, opened 2022-07-05
Notes window doesn't initialise nicely


riverwanderer:
The Notes window seems to initialise bottom/right-justified. It would be better if it were top/left-justified. To replicate, you need a line of text at the end of your scenario notes that is longer than the window size (which I think is remembered by Vassal).

Bad:
<img width="531" alt="image" src="https://user-images.githubusercontent.com/58135975/177292896-f609d354-c4a3-4e26-b214-3d13f2519ad8.png">

Good:
<img width="541" alt="image" src="https://user-images.githubusercontent.com/58135975/177293522-ba6c04a7-991d-480a-9c9c-497a3a1c9ccf.png">

Scrolling the scenario notes to the desired position and then Saving the Notes seems to temporarily fix the issue, but it re-emerges once the scenario file is saved, closed and re-opened.

jonathanrwatts:
Surely it's simply opening the window with the last character visible (i.e., if you had a shorter line following that long one, it would be left-aligned)?  Is it also placing the text cursor after the last character?

riverwanderer:
Yes, @jonathanrwatts you are right, it is opening with the last character visible (a simple newline is not enough though).
Also, no it is not placing the text cursor at all - the notes window displays but you have to click in it to enter any text and so your cursor starts in the nearest possible place to where you click.

jonathanrwatts:
It would probably be best if the module stored the last cursor position whenever the window is closed, and reopen with that position in view.  That way, it will remember where you left off when last editing text, but you could move the cursor to the top before closing the window when you're done editing.

riverwanderer:
> It would probably be best if the module stored the last cursor position whenever the window is closed, and reopen with that position in view.

Sounds fair enough. I think, to support legacy better, if the module hasn't stored the last cursor position, it should default to the top left.

BrentEaston:
Can you please try the 11564-NotesTextPosition build when it turns up.

riverwanderer:
> try the 11564-NotesTextPosition build

Seems perfect - Notes position is retained whilst in the module session but resets to the top on a fresh open.
I only tested with one session but this should probably be an individual user setting.

PR #11806, opened 2022-12-02
11806 NotesWindow re-open assigns focus properly to retain caret position. …

Cattlesquat:
…Restored proper caret update policy.

The previous change couldn't work & broke all text configurers. 

I've reverted that part, and instead improved the focus behavior of NotesWindow (when you re-open it after having hidden it) so that it returns focus to the TextArea you had going before, thus retaining caret position (turns out, caret position always worked right if you tabbed into the window; the problem was that unless you used tab key you had to click in the text to get the focus back in there, and that usually moved your cursor somewhere irritating)

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/NotesWindow.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/NotesWindow.java
@@ -62,6 +62,9 @@ public class NotesWindow extends AbstractToolbarItem
   public static final String BUTTON_TEXT = "buttonText"; //NON-NLS // non-standard legacy difference from AbstractToolbarItem
   public static final String DESCRIPTION = "description"; //NON-NLS
 
+  public static final int TAB_INDEX_SCEN = 0;
+  public static final int TAB_INDEX_PUBLIC = 1;
+
   // These three identical to AbstractToolbarItem, and are only here for "clirr purposes"
   @Deprecated(since = "2020-10-21", forRemoval = true) public static final String HOT_KEY = "hotkey"; //$NON-NLS-1$
   @Deprecated(since = "2020-10-21", forRemoval = true) public static final String ICON = "icon"; //$NON-NLS-1$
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/NotesWindow.java
```diff
@@ -62,6 +62,9 @@ public class NotesWindow extends AbstractToolbarItem
   public static final String BUTTON_TEXT = "buttonText"; //NON-NLS // non-standard legacy difference from AbstractToolbarItem
   public static final String DESCRIPTION = "description"; //NON-NLS
 
+  public static final int TAB_INDEX_SCEN = 0;
+  public static final int TAB_INDEX_PUBLIC = 1;
```
uckelman, line 66: Why make these public?

Cattlesquat: Oh definitely no good reason. Was blindly copying what was right above haha

Cattlesquat: (of course I predict several rounds of negotiation with style-enforcement-bot over the "right" way to make something non-public... mark my words)

Cattlesquat: (which is why I tend to copy something already there, for better or worse)

riverwanderer:
Problem #11564, that the failed change was intended to fix, is now back. The reason is because the Saved Game state just leaves the caret at the end of the block of text (in the Scenario Notes panel at least) - even though it does save for the individual session.  

Is it possible to save the caret positions in the saved game state ?

Cattlesquat:
Oh is THAT what people use the "scenario notes" field for?! Never even knew! 

I tend to put any starting messages in Startup GKC's displaying brightly into the chat log, for easy discovery :)

PR #11821, opened 2022-12-11
Include the AreaOfEffect bounds in the piece bounding box only if the AoE is visible

uckelman:


vassal-app/src/main/java/VASSAL/counters/AreaOfEffect.java
```diff
@@ -170,12 +170,15 @@ public void mySetState(String newState) {
...
-    if (a != null) {
-      final Rectangle aoeBounds = getArea().getBounds();
-      final Point mapPosition = getPosition();
-      aoeBounds.translate(-mapPosition.x, -mapPosition.y);
-      r.add(aoeBounds);
+    // include the AoE in the bounding box if it's visible
+    if ((alwaysActive || active) && mapShaderName == null) {
+      final Area a = getArea();
+      if (a != null) {
+        final Rectangle aoeBounds = getArea().getBounds();
```
jonathanrwatts, line 177: Use `a.getBounds()` instead of calling `getArea()` again?

PR #11825, opened 2022-12-13
BoundsTracker once again repaints entier maps, reverting 11521

uckelman:


uckelman:
@flivni The bounds tracker changes caused a slew of problems in the 3.6.8 release, which I don't want to sink more of my own time into troubleshooting. If you want to take another crack at it, we now have ample test cases where you can see how it failed.

Discussion: 

https://forum.vassalengine.org/t/3-6-8-slow-refresh/76044/5
http://www.gamesquad.com/forums/index.php?threads/vassal-3-6-8-display-changes.178302/

One of the problem can be seen by using a key command to rotate a piece in basically any module.

PR #11844, opened 2022-12-28
Switch from deprecated set-output to new $GITHUB_OUTPUT

uckelman:


uckelman:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.

Issue #11846, opened 2022-12-29
NullPointerException during Search


uckelman:
| ABR | |
|-----|-----|
| Version | 3.6.9 |
| Reporter |  |
| Error Log | [48259 bytes](https://vassalengine.org/files/gh/daae160ba7f26ee866b56ec1af45bed7e515ff64) |

<pre>

OCS - Reluctant Enemies valt 1.11 3.6.9

java.lang.NullPointerException: null
 at java.base/java.util.Objects.requireNonNull(Objects.java:233)
 at java.base/java.util.ImmutableCollections$List12.<init>(ImmutableCollections.java:564)
 at java.base/java.util.List.of(List.java:829)
 at VASSAL.build.module.turn.TurnTracker.getMenuTextList(TurnTracker.java:1319)
</pre>

BrentEaston:
The search implementation in the Turntracker is failing in this module

BrentEaston:
The problem is in Turntracker:

`public List<String> getMenuTextList() {
    return List.of(getAttributeValueString(BUTTON_TEXT), getAttributeValueString(TOOLTIP));
  `

List.of returns an ummutable list which is not allowed to contain nulls and the TOOLTIP attribute is returning a null. This can easily be fixed here by changing List.of() to Arrays.asList(), but do we have a bigger problem where List.of is used in other search related code?

Cattlesquat:
@BrentEaston For this one I searched through all our various List.of stuff, and turned any that had any opaque calls (like getAttributeValueString) into Arrays.asList. There weren't ALL that many. Some probably didn't actually need protecting, but I guess big deal. PR is #11866

PR #11866, opened 2023-01-07
11846 - Protect from null property results during search

Cattlesquat:


Cattlesquat:
Fixes #11846

BrentEaston (approved):
Looks good to me.

PR #11867, opened 2023-01-07
11813 - NPE in PieceSlot.getLocalizedConfigureName()

BrentEaston:
getLocalizedConfigureName() is trying to expand the piece, even if one does not exist yet.


Cattlesquat (approved):
Looks good - that was one of those bizarre mysteries!

PR #11868, opened 2023-01-08
11823 - NPE in Gamerefresher refreshing Decks

BrentEaston:
Change equals to Objects.equals to prevent NPE.


BrentEaston:
Too many commits

PR #11869, opened 2023-01-08
11823 - NPE in GameRefresher when refreshing unnamed Decks

BrentEaston:
Change equals() to Objects.equals()

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
@@ -475,7 +475,7 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
           final Board board = drawPile.getConfigureBoard(true);
           if ((boardName == null) || boards.contains(board)) {
             for (final DrawPile drawPile2 : foundDrawPiles) {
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -475,7 +475,7 @@ else if (!decksToDelete.isEmpty()) {
           final Board board = drawPile.getConfigureBoard(true);
           if ((boardName == null) || boards.contains(board)) {
             for (final DrawPile drawPile2 : foundDrawPiles) {
-              if (drawPile.getAttributeValueString(SetupStack.NAME).equals(drawPile2.getAttributeValueString(SetupStack.NAME))) {
+              if (Objects.equals(drawPile.getAttributeValueString(SetupStack.NAME), drawPile2.getAttributeValueString(SetupStack.NAME))) {
```
uckelman, line 478: Will this introduce a new problem if both `drawPile.getAttributeValueString(SetupStack.NAME)` and `drawPile2.getAttributeValueString(SetupStack.NAME)` are `null`? In that case, `Object.equals()` returns `true`, so the two piles will register as matching.

PR #11870, opened 2023-01-08
11759 - Do not allow negative number of Dice or Sides

BrentEaston:


uckelman:
Negative dice are just rotated 90 degrees from positive one, right? :)

PR #11872, opened 2023-01-08
11455 - Prevent NPE when adding empty KeyStrokes to list

BrentEaston:


uckelman (commented):
Could we `return Math.min(selectedEntryIndex, entries.size() - 1);`  and not bother with the conditional?

BrentEaston:
No, I don't think so. We want selectedEntryIndex to be updated with the new reduced index in this case. It is currently pointing past the end of existing entries which may cause issues elsewhere.

uckelman:
I hadn't noticed the assignment there. You're right.

uckelman (approved):

PR #11883, opened 2023-01-09
Fixed order of labels for "Select play mode" and "Enter your name" steps in the Welcome Wizard

uckelman:


uckelman:
Fixes https://forum.vassalengine.org/t/welcome-screen-v3-6-7-name-and-play-mode-in-wrong-order/75074.

PR #11911, opened 2023-01-19
11911 - Debug window shows board-relative coordinates if they are different from the map-relative ones

Cattlesquat:
Noticed the need for this while I was working on Imperial Struggle module. Doesn't display the extra stuff on e.g. maps-that-only-have-one-board, so most people won't see anything extra/spurious/new

image

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/DebugControls.java
+++ b/vassal-app/src/main/java/VASSAL/tools/DebugControls.java
@@ -90,24 +90,19 @@ public class DebugControls extends AbstractBuildable implements ActionListener {
     return cursorLocation;
   }
 
+  private Point getBoardLocation(Point pt, Map map) {
+    if (map == null) return pt;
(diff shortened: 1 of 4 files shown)


vassal-app/src/main/java/VASSAL/tools/DebugControls.java
```diff
@@ -85,12 +90,33 @@ public Point getCursorLocation() {
...
+        bp.translate(-r.x, -r.y);
+        cursorLocationBoard = bp;
+      }
+      else {
+        cursorLocationBoard = pt;
+      }
+    }
+    else {
+      cursorLocationBoard = pt;
+    }
```
uckelman, line 110: Most of this is repeated below... I wonder if there's a way to extract both this and the nearly identical block lower down to a function?

Cattlesquat: Fixed

Issue #11610, opened 2022-07-23
Reference Manual: Startup GKC mis-categorised


riverwanderer:
I've noticed that the [Startup Global Key Command](https://vassalengine.org/doc/latest/ReferenceManual/Map.html#StartupGlobalKeyCommand) entry is categories as a map-level component. It should be categorised as a module-level component.

Cattlesquat:
Fixed by #11918

PR #11918, opened 2023-01-22
11918 - Startup GKC that fires when player joins or changes sides

Cattlesquat:
Is this a good idea or a terrible idea? I *think* it will run once at game start (from perspective of whoever started the game) and then once whenever someone joins/changes-side (from their new side's perspective). 

I was thinking it could be used to solve the classic Vassal problem where when you start a game you have to go immediately "flip all your cards upside down" (or pieces or whatever) because them just starting upside down can't be part of the initial state because of our encoded-by-passwords method of turning things face down.

### DIFF
--- a/vassal-doc/src/main/readme-referencemanual/ReferenceManual/GameModule.adoc
+++ b/vassal-doc/src/main/readme-referencemanual/ReferenceManual/GameModule.adoc
@@ -90,7 +90,6 @@ Any of them can be added by right-clicking on the _[Module]_ component in the Mo
 |===
 
 
-
 '''''
 
 ==== <<HelpMenu.adoc#top,Help Menu>>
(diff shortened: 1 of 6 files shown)


riverwanderer:
I've done some intercepting of side changing and it is a complicated and gap-prone process. e.g. Liberté and (less complex) C&C Ancients. Maybe this change would have helped.

Couple of areas that come to mind and that I'd test this change against are:
1. Whilst the switch sides prompt is waiting a response, a race condition is active, since any other user can join/switch and will be selecting from the same side list, so you can end up with users connecting on the same side (and with this change, I presume, executing the SGKC as that same side). My solution (in Liberté) was to track which side slots are taken and have the switch side function in a loop that only exits when an unused side is chosen.
2. Fairly obvious but the existing "run once" SGKC mode is not able to distinguish an external saved scenario file from an in progress game. In modules that need to support users creating their own scenarios, I create a boolean that has to be reset in order save a game as a new scenario - this means I end up using "run always" SGKC with a check on that boolean.

@Cattlesquat I'm not sure that I've come across your "classic Vassal problem" - can you give an example module?

Cattlesquat:
The Imperial Struggle module is an example of one where people have to manually hide their stuff at the beginning (in the Ministry windows)

I didn't know it was possible for two players to connect in the same "player slot" through simultaneous-whatever -- I would think Vassal would reject the second attempt (though maybe it doesn't do it in a pretty way?). In any event that would be a bug. This particular variant of SGKC would only run if a successful connection attempt made it through to the game state.

Cattlesquat:
Also fixes #11610

Cattlesquat:
Fixes #11610

vassal-doc/src/main/readme-referencemanual/ReferenceManual/GameModule.adoc
```diff
@@ -266,6 +265,30 @@ Applies to <<GamePiece.adoc#top,Game Pieces>> on all <<Map.adoc#top,Map Windows>
...
+==== Startup Global Key Command
+
+Can print a welcome message, or perform some other task that needs to be done whenever the module is started up.
+
+image:images/StartupGlobalKeyCommand.png[]
+
+An extension of <<#GlobalKeyCommand,Global Key Command>> that fires automatically upon completion of module load, once all the key listeners are started up.
+All fields behave identically to the corresponding ones in <<#GlobalKeyCommand,Global Key Command>>, except that those pertaining to the physical representation of a Toolbar button are suppressed as being inapplicable.
+
+As of VASSAL 3.6, multiple start-up global key commands _can_ be depended on to run in the order they are listed in the module.
```
uckelman, line 279: Should that be 3.7, not 3.6?

Cattlesquat: No, 3.6 is correct. Much of this particular document edit is merely moving the whole clump from the wrong place it was to the right place it goes.

PR #11927, opened 2023-01-24
11502 Insert items into formatted strings from drop-down list into the correctly selected position

### LINKED ISSUE #11502: Replace selection not supported by report format "Insert" drop down
The "Insert" drop down resets any selection made by the user in the Report Format field and instead inserts text at some unexpected point that field. The reset of a selection is easily missed and the behaviour runs against "the way things usually work".

Demo/example:
1. User selects some text to replace:
Pasted Graphic 6

2. At the moment of activating the drop down "Insert" menu, the selection is reset:
Pasted Graphic 7

3. An insert option is chosen and the result is not what the user was expecting:
Pasted Graphic 8

AsaiToshiya:
Fixes #11502


BrentEaston (approved):
Works for me, thanks for the contribution Toshi!

PR #11933, opened 2023-01-26
11933 - At-Start stacks were getting confused by blank boards

Cattlesquat:
This might fix the problem Benkyo was experiencing (which manifested while dragging SetupStacks in and out of folders, but really was about blank stupid boards)


vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -2676,6 +2676,7 @@ public boolean importData(TransferHandler.TransferSupport support) {
 
           if (remove(getParent(sourceNode), cutObj)) {
             insert(target, convertedCutObj, childIndex);
+            postPasteFixups(convertedCutObj);
```
BrentEaston, line 2679: HA, This is the actual fix. I don't completely missed that in the other code paths!

vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
```diff
@@ -386,6 +386,9 @@ else if (OWNING_BOARD.equals(key)) {
           owningBoardName = null;
         }
       }
+      else if (value.equals("")) {
+        owningBoardName = null; // Because "" is what null sometimes turns into during Vassal configurers
+      }
```
BrentEaston, line 391: I don't believe this is needed once you add the postPasteFixes() call in. Just adding the postPasteFixes() call fixed the issue for me.

BrentEaston (changes requested):

BrentEaston:
I don't see any problem with this going into 3.6.11rather than 3.7.0? It's basically a just a bug fix.

vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
Cattlesquat: Still needed -- see comment above.

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -301,6 +301,7 @@ Editor.Board.reverse=Reversible
 Editor.Board.width=Board width
 Editor.Board.height=Board height
 Editor.Board.component_type=Board
+Editor.Board.main_board=Main Board
```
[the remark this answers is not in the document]
Cattlesquat: Fixed

vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
Cattlesquat: In the "drag" version, this was (on the one hand) "fixing" the problem, but it was creating a new "bad data" message in the console -- in other words, it WASN'T the thing making the cut/paste thing work. Cut/Paste already didn't generate the problem and so didn't get the extra warning message. So this DOES belong here (for other reasons) but DOESN'T really fix this problem.

Cattlesquat:
Agreed it could go in 3.6.11 if we do one

BrentEaston:
Looks good to me. I've asked Benkyo to do a final check.

vassal-app/src/main/java/VASSAL/build/module/map/SetupStack.java
```diff
@@ -386,6 +386,9 @@ else if (OWNING_BOARD.equals(key)) {
           owningBoardName = null;
         }
       }
+      else if (value.equals("")) {
```
uckelman, line 389: You should  check `"".equals(value)` instead. This way, you're still safe if `value` is null.

Cattlesquat: Fixed

PR #11935, opened 2023-01-26
11935 - Allow option for only first Mouseover-stack Viewer that generates a tooltip will be drawn

Cattlesquat:
Multiple Rollovers creates a visual mess, and somehow we hadn't yet provided a way to let multiple viewer entries fight it out for "first viewer to find a target" and keep the others from also displaying and making a big mess. 

I don't think it's possible to NOT make a big mess with multiple simultaneously-eligible viewers, as things previously stood -- so I don't think there will be any complaint. If there IS, then a field can be introduced into CounterDetailViewer to override this behavior (by making it opt-in) w/o needing to bump the minor version number, but again I don't find it "visually plausible"  that anyone is actually using something like that intentionally.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,9 @@
 3.6.11 - ???
 
+* 11941: Since 3.6.7, French/Belgian AZERTY keyboards on Mac were experiencing
+    problems with Ctrl / Alt keys on the number key row
+* 11938: IndexOutOfBoundsException trying to open a private note
+* 11933: At-Start stacks were getting confused by blank boards
 * 11929: Show translated module name and descriptions in Module Manager
 * 11924: Reinstate translation of Player Sides
 * 11911: Debug window shows board-relative coordinates if they are different
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,8 @@
 * 11900: Clear Tile Cache with more efficient Files.walkFileTree()
 * 11899: Hide FileConfigurer label in preferences dialogs
 * 11890: NPE when current-stack option used on non-stacking piece
+* 11505: Insert items into formatted strings from drop-down list into the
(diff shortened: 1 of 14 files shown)


uckelman:
Two viewers enter, one viewer leaves...

riverwanderer:
Interested to see how this works, as I have got workarounds in place in some modules for this issue. Will test when I get a moment.

riverwanderer:
Testing this with Commands & Colors: Ancients v4.2, which includes a couple of Mouse-Overs that I've configured to work together. The two display together satisfactorily in v3.6.10  - like this:
Pasted Graphic 2

With the mouse-over change, only the underlying piece is displayed but if I swap the mouseovers around in the module the unit piece (only) is (displayed image was the top mouseover but is this the design criteria for the change... because I could not replicate it with another mouseover (cards) ?

I have other mouseovers in C&C Napoleonics module to test that are also designed with the original problem in mind, so I will test that next.

riverwanderer:
Here's an example of two mouseovers working together in C&C Nap module:
image

With the mouseover change, the text element is dropped. The reason for two mouseovers is to do with allowing the zoomed card to be optional. If the zoom part was an evaluated field, it could possibly be a single mouseover.

Cattlesquat:
Okay that means this should go to 3.7 with a checkbox of some kind

Cattlesquat:
@riverwanderer New version now defaults this behavior to "off" -- it must be opted into in each mouseover that wants to be exclusive (causing mouseovers lower in the list to be pre-empted). So it should now have no effect on your module (since all your rollovers will default to the original behavior).

riverwanderer:
Tested with the C&C Ancients main map mouseovers. Works as described by @Cattlesquat.

PR #11941, opened 2023-01-29
Since 3.6.7, French/Belgian AZERTY keyboards on Mac were experiencing problems with Ctrl / Alt keys on the number key row

Cattlesquat:
Fortunately (having just looked through the various Mac keyboard types), there don't seem to maaaany European keyboards that go this weird way on the number key row.

Apparently the "new" treatment began in Vassal 3.6.7

image

image


Cattlesquat:
Bandaid confirmed to work by user

PR #11957, opened 2023-02-06
11957 - Spaces in GKC reports made consistent with all other reports

Cattlesquat:
In every other type of report (e.g. Report traits, DoActionButtons, etc) the report is formatted beginning "* " (a space after the asterisk). But for GKCs it is oddly not, leading to SLOPPY OUTPUT. 

This fixes them to be consistent with the other reports.


uckelman:
Does this need to be 3.7.0, or could it be 3.6 still?

Cattlesquat:
SloppyJoes could totally be 3.6, I just didn’t have a milestone available for that on the list that came up (probably could have created one if I looked) - so anyway yes, merge away

Issue #11714, opened 2022-10-06
Loophole in Deck attribute Send Hotkey when empty prevents processing last card


riverwanderer:
Vassal appears to prevent using Deck "Send when empty" hotkey to send a command to the last card to be removed from the deck.

When the card is removed, the Hotkey triggers and, at that point, the removed card has no location and is not found by a Global Key Command (even using CurrentMap=="" or a GKC without any search criteria).

Demo module:
[Empty HK test2.zip](https://github.com/vassalengine/vassal/files/9723883/Empty.HK.test2.zip)

riverwanderer:
Workarounds:

1) Processing of the last card has to be deferred to a subsequent *independent* action, by using the deck-empty hotkey to set a flag instead of directly acting on the card. This is not a universal solution.

2) The Deck-Empty action could be checked for on the card itself, but this is will require extra logic on the cards.

Cattlesquat:
The empty-deck function is fired off in a gnarly way at a gnarly place -- as Mark describes, it happens *instantly* at the moment when the last card has been unlinked from the deck -- before it complete its move. Really this command should get "queued up" at that point, and then fired off when the card/piece is *finished* moving (e.g. right after it possibly fires its apply-on-move key, etc, and definitely after it has been linked back to a map somewhere). 

Of course that would mean adding something in each of the many places a piece can be moved (PieceMover, SendToLocation, FreeRotator, Translate, Return to Deck, Deck's SendToDeck, and the Deck Send Key command. I can't remember if there are... even more? 

It's definitely pretty scary and messy though to be initiating new commands/hotkeys while a piece is unlinked.

```
protected void removePieceAt(int index, boolean suppressDeckCounts) {
    final int startCount = pieceCount;
    if (!suppressDeckCounts) {
      updateCounts(index);
    }
    super.removePieceAt(index, suppressDeckCounts);
    if (!suppressDeckCounts) {
      fireNumCardsProperty();
      // Do NOT fire a Deck Empty key if it has been caused by an Undo Command
      if (hotkeyOnEmpty && emptyKey != null && startCount > 0 && pieceCount == 0 && ! GameModule.getGameModule().getBasicLogger().isUndoInProgress()) {
        gameModule.fireKeyStroke(emptyKey);
      }
    }
  }
```

BrentEaston:
[...]

PR #11963, opened 2023-02-08
11963 - SendToLocation, Can Rotate, and MoveFixedDistance were all unlinking from deck/stack at wrong time, causing buggy behavior in some instances especially involving decks

Cattlesquat:
CHANGELIST for this PR should be:
11994 - Deck "send-when-empty" hotkeys no longer create extra undo steps
11963 - SendToLocation, Can Rotate, and MoveFixedDistance were all unlinking from deck/stack at wrong time, causing buggy behavior in some instances especially involving decks 
11714 - Deck's Send-when-Empty key now processes AFTER fully moving last card
11429 - Deck Empty Hotkey no longer repeats per online player instance, nor when stepping forward a log

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -772,6 +772,9 @@ public class PieceMover extends AbstractBuildable
    * @return Command encapsulating anything this method did, for replay in log file or on other clients
    */
   protected Command movedPiece(GamePiece p, Point loc) {
+    // Make sure we don't end up sending I'm-empty commands from decks that were emptied by other players (e.g. online or log-step)
+    Deck.clearEmptyDecksList();
+
     Command c = new NullCommand();
     c = c.append(setOldLocations(p));
     if (!loc.equals(p.getPosition())) {
--- a/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -1172,6 +1172,9 @@ public class PieceMover extends AbstractBuildable
       comm = comm.append(applyKeyAfterMove(allDraggedPieces, map.getMoveKey()));
     }
 
+    // If we emptied any decks, let them send their I-am-empty key commands
+    Deck.checkEmptyDecks();
+
     // Repaint any areas of the map window changed by our move
     tracker.repaint();
 
--- a/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -1176,7 +1176,7 @@ public class PieceMover extends AbstractBuildable
     }
 
     // If we emptied any decks, let them send their I-am-empty key commands
-    Deck.checkEmptyDecks();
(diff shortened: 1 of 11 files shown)


riverwanderer:
Brief test with Liberté module's send to location (from a deck) features passed ok.

Cattlesquat:
@riverwanderer The bad behavior that I was noticing before (leading me to find this bug) was... I was getting different results (comparing dragging a card out of a deck to send-to-location) for what the value of ObscuredToOthers was on the card when it arrived at its destination and was doing its "apply after move" key. This now makes the results consistent (and correct). This does NOT however address the issue that ObscuredToOthers while actually *in* the deck is a bit undetermined.

Cattlesquat:
I... I couldn't take it any more--I... I... I... refactored to make all these traits & deck-sends use central "Prepare Move"  and "Finish Move" methods which I implemented in EditablePiece (I know, ?!?!, but was lowest-common-denominator of BasicPiece and all the Decorator traits). 

PieceMover is still "special" because it tries for other efficiencies (and sure, also because it's complex and terrifying). 

But now when we adjust these things we can adjust in TWO places rather than SEVEN.

Cattlesquat:
@riverwanderer I just added some significant stuff to this, most notably including a fix for your #11714, but also a bit of a refactor to make improving/fixing piece movement less of a nightmare in the future ... but like all "modest refactors" it could possibly lead to "modest disasters"... so in addition to checking the fix on #11714 if you had time to put some modules through their paces on Send-to-Location, Return to Deck, and perhaps even Move Fixed Distance and Can Rotate (if you use them), along with the Send-to-other-Deck functions of Deck? Obviously use the new version that just posted (build is available as of when I'm writing this message)

riverwanderer:
@Cattlesquat  - I have modules that use these traits in various places. I'll start doing some testing.

riverwanderer:
The following module tests completed ok:
C&C Ancients (Epic scenario): Moved Fixed Distance: cards moved left at turn end.
C&C Napoleonics: send to loc - 1 block from stack of 2 moved to Victory track when eliminated; scout mode end turn - cards sent from deck to board/hand and also Return to Deck used.
Liberte: Send to Loc from multiple decks; Global HotKeys used for send to deck action & deck empty events.

C&C Ancients - tested the "last card" in deck fix. A GHK triggered by empty deck does now find the last card removed from the deck (if clears a DP `fromSide` held on the card).  **Notably** UnDo is not atomic in this situation but I recall this a different issue/restriction on how the deck empty HK works (what happens in the test is that the UnDo moves the cards back onto the deck but the DP is not restored; a further undo (or 2 or 3) does eventually restore the DP.

Cattlesquat:
Correct, there were explicit markings in the code that empty-deck is not supposed to fire as a result of undo. Thanks for the testing!

Cattlesquat:
@riverwanderer More possible testing for you -- the NEW NEW version I believe fixes your 11429 about repeating Deck Empty hotkeys "per player instance". It looks like it ALSO used to do an extra hotkey when a deck was emptied by stepping through a logfile -- i.e. you'd get the effect from the time the original player did it and then it would happen AGAIN for the player stepping through. Probably too hard for you to test the online version, but you might be able to test the logfile version, which runs the same code architecture.

BrentEaston (approved):
This is looking good to me.

PR #11964, opened 2023-02-08
11964 - Mouse-over Stack Viewers can display an image of underlying terrain

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -3733,6 +3740,45 @@ public String getIdentifier() {
...
+
+  /**
+   * Creates an image (of designated size) of the board(s) at a particular point
+   * @param pt - center point of map location to generate image for
+   * @param width - width of map to include
+   * @param height - height of map to include
+   * @param os_scale - OS scaling factor
+   * @return an image
+   */
+  public BufferedImage getImageMapIcon(Point pt, int width, int height, double os_scale) {
```
uckelman, line 3752: This function isn't used now, right?

Cattlesquat: Removed zombie code!

PR #11969, opened 2023-02-08
11969 - Border Outline trait. What?!?! Oh no it's a new trait! FLEE FOR THE HILLS, HAIR ON FIRE!!!

Cattlesquat:
BorderOutline trait makes a colored border around the piece "possibly contingent on a property" 

image

image


BrentEaston:
So this is sort of a per-piece configurable selection highlighter that doesn't only work when selected?

Cattlesquat:
@BrentEaston yes exactly -- choice between "works always" (no property specified) or "works when property is true"

vassal-app/src/main/java/VASSAL/counters/BorderOutline.java
```diff
@@ -0,0 +1,220 @@
...
+
+  @Override
+  public Shape getShape() {
+    return piece.getShape();
+  }
+
+  @Override
+  public Rectangle boundingBox() {
+    final Rectangle r = piece.boundingBox();
+    r.add(piece.boundingBox());
```
uckelman, line 105: Why are you adding the bounding box when `r` starts out as the bounding box?

Cattlesquat: Because of poor copying and pasting I guess :) All fixed

PR #11970, opened 2023-02-09
11627 - ActionButtons with context menu flag no longer trigger context menu when a different ActionButton on same piece (that doesn't have the flag set) is clicked

### LINKED ISSUE #11627: Action Button context menu option triggers on piece boundary rather than Action Button boundary
Following on from this (Vassal Discord) [thread](https://discord.com/channels/740127632586178572/911006365626036294/1005624649490636861), I produced a demo module for the issue:
[context menu bug.zip](https://github.com/vassalengine/vassal/files/9276960/context.menu.bug.zip)

To explore, run up the module, start a new game and follow the instruction in the chat log.

Cattlesquat:
Fixes #11627

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
@@ -17,30 +17,9 @@
  */
 package VASSAL.build.module.map;
 
-import java.awt.Font;
-import java.awt.Point;
-import java.awt.event.ActionEvent;
-import java.awt.event.MouseAdapter;
-import java.awt.event.MouseEvent;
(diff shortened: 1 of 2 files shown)


riverwanderer:
Tested successfully.

In detail:  C&C Napoleonics module; v4.1.5 has to use a Restrict Command trait on combat markers, to suppress "hits" context menu when second action button is "active" (dice roll) - otherwise the context menu displays when the dice roll icon is pressed. Tested with a modified version that removed the suppression - this works, in that the context menu no longer displays when the dice icon Action Button is pressed but does display if the combat marker Action Button is pressed, even whilst the dice icon is showing.

Cattlesquat:
@riverwanderer Hooray, thanks!

PR #11973, opened 2023-02-09
11880 - Triggers set to count 0 times don't spuriously execute 1 time.

### LINKED ISSUE #11880: Trigger Action's loop a fixed number of times always happen at least once
If a Trigger Action is told to repeat a certain number of times, it will always happen at least once even if told to repeat zero times.

Steps to reproduce:
1) Create a piece.
2) Give it a Report Action
3) Give it a Trigger Action that issues the Report Action's Key Command
4) Click "Repeat this set of Key Commands (Loop)"
5) Set the type of loop to "Repeat fixed number of times"
6) Enter 0 for the number of times
7) Start the game and activate the Trigger Action
Expected result: no reports
Actual result: a report

Unlike the Repeat until condition (which, I assume, is based on C/C++/Java's do-while loops rather than other languages' until loops), this behavior is not documented in the reference manual. I assume zero meaning one is unintended, as it would let me reduce the number of Trigger Actions that exist only to maybe activate other Trigger Actions.

Cattlesquat:
Fixes #11880

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/TriggerAction.java
+++ b/vassal-app/src/main/java/VASSAL/counters/TriggerAction.java
@@ -41,6 +41,8 @@ import VASSAL.tools.RecursionLimiter;
 import VASSAL.tools.RecursionLimiter.Loopable;
 import VASSAL.tools.SequenceEncoder;
 
+import javax.swing.JLabel;
+import javax.swing.KeyStroke;
 import java.awt.Component;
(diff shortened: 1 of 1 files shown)


Cattlesquat:
Bug reporter confirms this fixes it for him

riverwanderer:
@Cattlesquat - should / does this change also apply to Action Button loops? I think that has the same problem.

PR #11978, opened 2023-02-10
11462 - Don't +1 inventory sum if property isn't even defined

### LINKED ISSUE #11462: Inventory sum_propName adds 1 for unit that does not have propName defined instead of 0

Cattlesquat:
Fixes #11462


BrentEaston:
I remember seeing a similar bug report for the Sum() beanshell function. Not sure if that was ever fixed?

Cattlesquat:
@BrentEaston I took a scan through ExpressionInterpreter and I couldn't find anything that looked like it would have that problem, so maybe if it existed we already fixed it?

BrentEaston:
@Cattlesquat , Ah, I was thinking of this same issue!

BrentEaston:
What's the 'ter' file in the commit?

PR #11980, opened 2023-02-10
11437 - Right-clicking on a Deck with no menu items no longer produces an exception

### LINKED ISSUE #11437: [Harpoon Captain's Edition] NullPointerException: Cannot invoke "String.equals(Object)" because "menu_item" is null
| ABR | |
|-----|-----|
| Version | 3.6.6 |
| Reporter | kevinamartell@gmail.com |
| Error Log | [8747 bytes](https://vassalengine.org/files/gh/7f20f079fa3b2231e0a5272ba1e3faaaafb51afd) |

<pre>Right-click on a deck that has no menu commands.

Harpoon Captain's Edition v1.1 3.6.6

java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "menu_item" is null
 at VASSAL.build.module.map.MenuDisplayer.createPopup(MenuDisplayer.java:146)
 at VASSAL.build.module.map.MenuDisplayer.maybePopup(MenuDisplayer.java:277)
 at VASSAL.build.module.map.MenuDisplayer.mouseReleased(MenuDisplayer.java:243)
 at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:297)
</pre>

Cattlesquat:
Fixes #11437 

Apparently menu_item can be null, and the other parts of this method were protecting against that -- so this slightly newer part of the method should protect as well.

Oh right of COURSE it can be null because when there are no key commands rather than send an "empty list" we instead apparently send an array of length one that contains a KeyCommand.NONE

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
@@ -17,30 +17,9 @@
  */
 package VASSAL.build.module.map;
 
-import java.awt.Font;
-import java.awt.Point;
-import java.awt.event.ActionEvent;
-import java.awt.event.MouseAdapter;
-import java.awt.event.MouseEvent;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-
-import javax.swing.Action;
-import javax.swing.JMenu;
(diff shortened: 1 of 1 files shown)


BrentEaston (approved):
Looks good.

PR #11985, opened 2023-02-11
10821 - Symbolic Dice buttons how have full Action Button capabilities (sounds, hotkeys, etc.)

### LINKED ISSUE #10821: Expand Symbolic Die Button
The Symbolic Die button cannot be fully integrated in the way that Action and GKC buttons can because of the dynamic button image feature inherent to its ability. For example, an Action Button that plays a die rolling sound can not be activated simultaneously when a Symbolic Die button is pressed. 

Solution: If the Symbolic Die button included the ability to "Send Hotkeys" (like the Action Button does), this would be a gateway to integration for any features that the module designer might imagine. 

Extra: If it is deemed worth it, then the ability to play a sound file (to all players) might also be added (again, like the Action Button). This ability would still be possible even if only "Send Hotkey" were supported, but playing a sound file is probably the main use case.

Forum thread [here](https://forum.vassalengine.org/t/symbolic-dice-sound-file/12152/2).

Cattlesquat:
Fixes #10821 

I made SpecialDieButton extend DoActionButton and it was surprisingly straightforward to get it working -- so now a Symbolic Dice Button can also emit sounds, send hotkeys, etc.  Oh and I also fixed an irritating behavior that's been around forever where if you edit a Symbolic Dice Button that's a "report in window" one it used to disappear of your window until you restarted the module.

Ought to get some thorough @riverwanderer testing :)


riverwanderer:
Tested the basics - existing module (Flying Colors) with Symbolic Die buttons opens up with the new button features available. I added in a sound to both buttons and removed now redundant Action Buttons - works great.

riverwanderer:
Tested the last commit (only one report), still working but this time it seemed like the Description wasn't displayed in the editor (until component is opened, of course).
Also, for consistency, should Description be at the top of the component window, and Name get some guidance text. e.g. like Layer; 
image

Cattlesquat:
I'm unfortunately constrained by some legacy aspects of the Symbolic Dice Button's implementation as far as the location of the description (notice the existing version has the description in a weird place too) - alas.

vassal-app/src/main/java/VASSAL/build/module/DoActionButton.java
```diff
@@ -116,22 +116,24 @@ public class DoActionButton extends AbstractToolbarItem
...
-    };
-
-    setLaunchButton(makeLaunchButton(
-      getConfigureTypeName(),
-      getConfigureTypeName(),
-      "",
-      rollAction
-    ));
-    launch = getLaunchButton(); // for compatibility
+    if (!(this instanceof SpecialDiceButton)) {
```
uckelman, line 119: Woah, shouldn't this be in the ctor of `SpecialDiceButton`?

Cattlesquat: I agree it looks shitty but I couldn't figure out any way inside of SpecialDiceButton's ctor to prevent it calling the *empty* ctor of DoActionButton, since that seems to be what is done by default? Do you know a way?

uckelman: The ` SpecialDiceButton` ctor calls the nullary `DoActionButton` ctor first because there's no explicty call to `super(something, else)` and some ctor of the parent class must be called. You could get around this by adding another `DoActionButton` ctor for `SpecialDiceButton` to call explicitly.

Cattlesquat: I've added a new version of the constructor with a dummy parameter so I can avoid calling the nullary one. Seems a little more obfuscatory to me, but...

PR #11987, opened 2023-02-11
10820 - Make Action Button available on all windows where other similar buttons (eg GKC, Multi-Action, etc) are available

### LINKED ISSUE #10820: Make Action Button available on all windows where other buttons are allowed.
Up to v3.6beta7, the Action Button is only available from the top level (module). This request is that the Action Button component should also be available where-ever Multi-Action and GKC buttons are allowed.

Forum thread [here](https://forum.vassalengine.org/t/action-button-to-be-available-on-map-and-player-hand-windows/72706).

Cattlesquat:
Fixes #10820

@riverwanderer Another from your wishlist ... seems to work for me but please give it a more thorough spin!

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/AbstractToolbarItem.java
+++ b/vassal-app/src/main/java/VASSAL/build/AbstractToolbarItem.java
@@ -30,6 +30,7 @@ import VASSAL.configure.VisibilityCondition;
 import VASSAL.i18n.Resources;
 import VASSAL.tools.LaunchButton;
 import VASSAL.tools.NamedKeyStroke;
+import VASSAL.tools.ToolBarComponent;
 
 import java.awt.Component;
 import java.awt.event.ActionListener;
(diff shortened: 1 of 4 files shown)


riverwanderer:
So far so pretty good; I replaced a lengthy Multi-Action button with the newly map-friendly Action Button. Might be premature but it seems a little quicker too. 
One issue I've encountered - when I try using map-level Global Properties in the Action Button expression fields, they just return null.

Cattlesquat:
When the build that's building now comes up, try again on the properties -- I think I set it to use the Map as the initial property source for search.

riverwanderer:
@Cattlesquat  - map level GPs are working now.
Next thing I spotted - I'm blocked from pasting an Action Button into a folder within a map. It does work if I try to copy a map-level Action Button onto the module, a folder in a module or another map.

Cattlesquat:
@riverwanderer Good catch, thanks! I forgot to change the map folders to let them in (all folders aren't *actually* the same under the hood...) Try again with the new version building now

riverwanderer:
Working with map folders now! I've converted several "Toolbar Menu" buttons over to Action Buttons - no other problems spotted.

PR #11991, opened 2023-02-12
11982 - Fixed multiplayer problems with detach option on Deselect trait

### LINKED ISSUE #11982: Problems with Detach option on Deselect trait
There are some (possibly related) issues with the Detach option on the Deselect trait.
1) The detach command does not replicate to other online instances nor in a log file.
2) An UnDo of a detach command can result in the detached object being restored to a previous location (though it doesn't seem to be the immediately previous location). This seems to happen if the piece has been moved more than once.
3) When a UnDo of a detach command avoids issue 2, the stack is not restored exactly as before.

Illustrated by the Detach command on Leader pieces in the C&C Ancients module. The command is implemented via this trait:
image

Demo log file for the issues (to replay, use C&C Ancients version 4.2.x): 
[detachUnDoBug.zip](https://github.com/vassalengine/vassal/files/10713820/detachUnDoBug.zip)

To create the demo I open a scenario in the C&C Ancients module, connecting as Solitaire and play a card from the active hand. Then I left-click on a leader piece attached to a unit and move the stack 4 times. I then select the Detach command from the leader piece (see below - ). 

Detach commmand:
image

Demo of issue 1:
a) Detach command result (as seen locally):
image

b) Detach command result (as seen remotely and in the log file):
image

Demo of issue 2:
Play the log file to the end and note the position of the leader piece after the UnDo!

Demo of issue 3:
To observe issue 3, perform the detach command on the Dionysus leader; UnDo - note how the leader piece is restored to the bottom of the stack and the stack is expanded.

Cattlesquat:
Fixes #11982 

Recent report -- this 3.6 feature missed an item that needed to be "inside a command" 

@riverwanderer please give a try :)

### DIFF
--- a/vassal-app/src/main/java/VASSAL/command/StackPiece.java
+++ b/vassal-app/src/main/java/VASSAL/command/StackPiece.java
@@ -1,0 +1,52 @@
+/*
+ *
+ * Copyright (c) 2000-2003 by Rodney Kinney
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License (LGPL) as published by the Free Software Foundation.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU Library General Public
+ * License along with this library; if not, copies are available
+ * at http://www.opensource.org.
+ */
+package VASSAL.command;
+
+import VASSAL.counters.GamePiece;
+import VASSAL.counters.Stack;
+
+/**
+ * This Command adds a {@link GamePiece} to a stack.  Its undo
+ * Command is {@link UnstackPiece}.  */
+public class StackPiece extends Command {
+  private Command undo = null;
+  private final Stack stack;
+  private final GamePiece piece;
+
+  public StackPiece(Stack s, GamePiece p) {
+    stack = s;
+    piece = p;
+  }
+
+  /**
+   * Removes the piece form the stack;
+   */
+  @Override
+  protected void executeCommand() {
+    undo = new UnstackPiece(stack, piece);
+    stack.add(piece);
+  }
+
+  @Override
+  protected Command myUndoCommand() {
+    if (undo == null && stack != null && piece != null) {
+      undo = new UnstackPiece(stack, piece);
+    }
+    return undo;
+  }
+}
--- a/vassal-app/src/main/java/VASSAL/command/UnstackPiece.java
+++ b/vassal-app/src/main/java/VASSAL/command/UnstackPiece.java
@@ -1,0 +1,52 @@
+/*
+ *
+ * Copyright (c) 2000-2003 by Rodney Kinney
+ *
+ * This library is free software; you can redistribute it and/or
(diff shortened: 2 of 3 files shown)


riverwanderer:
Tested VASSAL-3.6.12-SNAPSHOT-76dc918-11982-macos-x86_64. It seems to fix the 2 "UnDo" issues but not the replication issue - i.e. on a success Detach, the logfile shows the object still stacked (I presume that other online users would see the same).

Cattlesquat:
Hunh, I wonder if I just need to queue up a redraw on "receiving end"

riverwanderer:
> Hunh, I wonder if I just need to queue up a redraw on "receiving end"

Maybe - but note that the stack does appear to change 
image


(before a detach, it looks like this)
image

Cattlesquat:
Okay I've tried a different tactic that may help, please let me know with new build?

riverwanderer:
Now if I do a detach and then undo, I get this alert.
image

I'll continue testing.

Cattlesquat:
That usually means your logfile was written by a different version of Vassal (in this case I guess the earlier version of this) -- and I *did* change "the very nature of logs for Detach" in this, so I can see why it would get confused. Try making the log from scratch with the new version?

riverwanderer:
The error occurs in a direct undo (i.e. before any logfile is opened). It also recurs when undoing a detach in a log file written from scratch by the new version.

riverwanderer:
Also, the change didn't affect the logfile showing only a stacked item when the original session does a detach.

Cattlesquat:
There was a horrifying typo. Please try again with latest version (a from-scratch log of course). If still issues, please link me to your module & log since I am probably missing something dumb.

riverwanderer:
Working again but still not reflecting the "detach" in the log.
Here is a [link](https://www.dropbox.com/s/106ukh1z9emk8nl/CCAncientsV4-3-0a04bare.vmod?dl=0) to the module and I attach the log.
[detachUnDoBugTst2.zip](https://github.com/vassalengine/vassal/files/10717294/detachUnDoBugTst2.zip)

Cattlesquat:
@riverwanderer This (new new new) build seems to do it for me now. One question... was this a feature you originally asked me for (in whatever 3.5 or 3.6 it was added)? I'm trying to remember why this needed to exist instead of just the normal thing of "double-click the stack to expand it, then drag the piece you want to move away somewhere else"?

riverwanderer:
Tested with: VASSAL-3.6.12-SNAPSHOT-1dc10d8-11982-macos-x86_64.dmg

Almost perfect. Detach / Undo is exactly reflected in the log file. One outstanding undo quirk, I notice -  if I detach a piece and then move it, the "move" UnDo results in the piece rejoining the stack. I'm not sure that is really a problem - certainly much better than before. The next UnDo restores the stack to how it originally looked, which puts the UnDo sequence back on track. Log file to demo this:

[test2.zip](https://github.com/vassalengine/vassal/files/10725758/test2.zip)

Re your question - it may be that I asked for this originally. It helps with modules that use left-click action buttons but also it allows detach as an automated action. I use it for both.

Cattlesquat:
Yeah I don't think I can do much about the "rejoins stack on undo" thing. 

I would think that when automating things just send-to-location to a different location would detach it from the stack? (I guess I'm not quite sure what the desired difference is)

riverwanderer:
It allows for the piece to stay in the same location but be detached... so can then
 be moved manually to some arbitrary location. 

I think it's fine about rejoining the stack on an undo, no harm done in that as far as I can see.

Issue #12001, opened 2023-02-13
New special property count_XXX


riverwanderer:
Provide a count_XXX property that works similar to sum_XXX (in components such as GPIW and Mouseover), e.g. for [GPIW](file:///Applications/VASSAL-3.6.12-SNAPSHOT-ac00b6e-master.app/Contents/Resources/doc/ReferenceManual/Inventory.html),

Any property of the form count_XXX will be replaced with the count of all pieces within that folder where the Property XXX is non-null.

riverwanderer:
Update: At the moment sum_XXX will evaluate a non-numeric (even null) Property XXX as equal to 1; so, e.g. sum_BasicName is equivalent to count_XXX.  If that is to remain as supported Vassal feature, this request can be closed as met already.

PR #12002, opened 2023-02-13
12001 - count_ supported for counting properties in Inventory windows

Cattlesquat:


Cattlesquat:
Fixes #12001

riverwanderer:
Tested build VASSAL-3.6.12-SNAPSHOT-a971a27-12001 with C&C Ancients Module, "Card Manifest" GPIW.

Observations (possibly intended behaviour?) -

The Card Manifest GPIW has a two level heirachy ("all cards" and "cards by BasicName"):
- at the top level, count_BasicName shows a total of the number of folders in level two (in this case, cards by BasicName, so the number is the number of unique card names; 29).  sum_BasicName on the other hand would should the total number of cards (60)
- at level 2, the cards are listed by BasicName and count_BasicName is the quantity of cards with the BasicName in common. At this level, count_XXX and sum_XXX display the same numbers.

Though not what I expected, on reflection this looks correct and a worthwhile addition to the GPIW reporting ability.

PR #12019, opened 2023-02-17
11967 - When dragging piece out of face-down deck, use unrotated shape

### LINKED ISSUE #11967: Dragged piece (card) image reveals rotate facing
When a masked card is dragged from a deck its outline is revealed to the user. Since this is prior to dropping the piece, the user will just need to release the mouse button over the deck (or out of the window entirely) and the draw will be silently cancelled.

The normal solution is for the module developer to ensure that all cards are returned to deck in the same orientation.

In case this doesn't suit all applications, to keep concealment Vassal ought to display a dragged card outline in upright orientation (as it already does for the masked image).

There is a similar inconsistency when a deck is face up. A card moved in with a non-standard facing will show in the deck in that facing:
image
Moving cards on top of the deck (ie not drawing them) can reset the orientation back to normal.
This second situation is more complex to analyse; perhaps its mostly valid behaviour or maybe more work than is justified would be needed for decks for to fully support rotated pieces. 

Demo module: [deckrotate.zip](https://github.com/vassalengine/vassal/files/10685070/deckrotate.zip)
Load this game: [deckrotate game.zip](https://github.com/vassalengine/vassal/files/10685082/deckrotate.game.zip)
Use the "Union Hand". Open zip files directly in Vassal.

Cattlesquat:
Fixes #11967 

@riverwanderer When you have a chance, give a test and see if this corrects your issue

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -1730,6 +1730,7 @@ public class PieceMover extends AbstractBuildable
           if (piece.getParent() instanceof Deck) {
             owner = (String)piece.getProperty(Properties.OBSCURED_BY);
             piece.setProperty(Properties.OBSCURED_BY, ((Deck) piece.getParent()).isFaceDown() ? Deck.NO_USER : null);
+            piece.setProperty(Properties.USE_UNROTATED_SHAPE, Boolean.TRUE);
           }
 
           final AffineTransform t = AffineTransform.getScaleInstance(zoom, zoom);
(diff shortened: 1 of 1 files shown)


riverwanderer:
Tested VASSAL-3.6.13-SNAPSHOT-12edc0e-11967.  Still the same with the test module and game file i.e. the mask image is correctly orientated but the outline is rotated.

Cattlesquat:
@riverwanderer Oh, the OUTLINE! Got it - try new version when it builds.

riverwanderer:
yes, that works!

BrentEaston (approved):
Looks good to me

PR #12024, opened 2023-02-19
9786 - Text Label traits will respond to GameRefresher (unless trait has a manually-change-label hotkey defined)

### LINKED ISSUE #9786: Refresh Counters fails to update a Text Label trait
| Bugzilla |  |
|----------|----------|
| ID | BZ14114 |
| Reported | 2021-02-22 07:37:00 +0100 |
| Modified | 2021-02-23 23:09:35 +0100 |
| Product | VASSAL |
| Component | Unknown |
| Version | 3.5.2 |
| Platform | All |
| OS | All |
| Status | ASSIGNED |
| Resolution | None |
| Priority | unspecified |
| Severity | major |
| Milestone | --- |
| Reporter | [Mark Benson](mailto:markben61@gmail.com) |
| Assignee | [Claudio Ciardelli](mailto:claudio.ciardelli@gmail.com) |
| CC | claudio.ciardelli@gmail.com |

| Attachments | |
|-------|-------|
| [Example Text Label for Refresh Counters fails to update a Text Label….pdf](http://www.vassalengine.org/files/bz/12775) | [Mark Benson](mailto:markben61@gmail.com) 2021-02-22 07:37:00 +0100 application/pdf 942430 Example of the bug |

#### [Mark Benson](mailto:markben61@gmail.com) 2021-02-22 07:37:08 +0100
<pre>Created attachment 12775
Example of the bug

Certain text label traits do not get picked up by Refresh Counters.

Noticed from Vassal v3.4.13 but likely existed much earlier.

Symptoms:-

A Text Label is updated (the Label Text field in the examples seen).

Refresh Counters on a pre-set scenario does not implement the Text Label change. Closing and restarting the module before doing RC
does not help.

Workaround:
1) Open the pre-set scenario
2) Copy the Text Label trait and Remove it. Refresh Counters. The Text Label is removed from the pre-set scenario/
3) Paste the Text Label trait back into place. Refresh Counters. The changed Text Label will take effect in the pre-set scenario.

Impact: The workaround has a high manual overhead with a large number of pre-set scenarios.

See attachment for an example.</pre>

Cattlesquat:
Fixes #9786

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/GpIdChecker.java
+++ b/vassal-app/src/main/java/VASSAL/build/GpIdChecker.java
@@ -30,6 +30,7 @@ import VASSAL.counters.PieceCloner;
 import VASSAL.counters.PlaceMarker;
 import VASSAL.counters.Properties;
 import VASSAL.i18n.Resources;
+import VASSAL.tools.NamedKeyStroke;
 
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


riverwanderer:
Tested this change in VASSAL-3.6.13-SNAPSHOT-2c5e7a3-master using C&C Ancients v4.2 (dev version)

In C&C Ancients v4.1.6 I have an unusual workaround to this text label refresh issue (I'd removed the change-label command/shortcut from the original text label trait). These labels get blanked by the change. To fix, I just had to move a replacement label's calculation into the text label field and repeat the refresh.  This actually proves the change works and in future I wouldn't need the workaround I'd implemented before.

I think its unlikely the improved refresh would normally rub up against the issue. If so, the solution is to update the module's old text label with the calculated one and repeat the refresh. 

Separately, I could not prevent the text label updating to blank by adding ability to change the label manually. Is the intention that this check only operates if the existing game file has a change-label hotkey or should it operate when the updated trait is adding a hotkey ?   From other testing I think the text label is left alone if a change-label shortcut is already defined but I can test again when I understand the intent better.

PR #12050, opened 2023-02-26
12041 - Rollover border color of 'no color' did not save to file, nor copy correctly

Cattlesquat:
#Fixes 12041

When color configurer was translating "no color" as null. Which worked okay in traits, but in attributes it meant our XML code didn't actually write/copy a record for that attribute. 

Worked around by returning the string "null" instead, which the configurer's StringToColor already turns back into a real `null`


riverwanderer:
worked when I tried it.

PR #12060, opened 2023-02-27
12060 - THE BLACK BOOK OF DECK - Editor allows DECK files to be imported into a deck to become permanent part of module

Cattlesquat:
*** I've based this on 3.6 so that MiRO can test it w/o scrudding his module up with 3.7 beta stuff, but it is obviously a 3.7 type of feature.


uckelman:
@Cattlesquat I resolved two conflicts. You should look at how the menu items are ordered in `ConfigureTree` to see if the result is what you intended.

Cattlesquat:
I believe the menu ordering is all cleared up now

PR #12067, opened 2023-03-01
12067 - The BLACK BOOK OF EXPORT -- Editor allows import/export of groups of items as XML files (eg to transfer between modules)

Cattlesquat:
Appears on Editor menu when right-clicking any editor item

Exports selected object "and all of its children" to an XML file

XML file can then be imported to become the child of whatever is selected at the time of import.

GO FORTH AND SIN NO MORE!!!

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/ModuleExtension.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/ModuleExtension.java
@@ -39,17 +39,10 @@ import VASSAL.tools.ArchiveWriter;
 import VASSAL.tools.DataArchive;
 import VASSAL.tools.swing.SwingUtils;
 import VASSAL.tools.version.VersionUtils;
-
-import java.awt.event.ActionEvent;
-import java.io.BufferedInputStream;
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.net.URL;
-import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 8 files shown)


vassal-app/src/main/java/VASSAL/build/module/ModuleExtension.java
```diff
@@ -416,6 +413,7 @@ public String getName() {
     return name;
   }
 
+  @Override
```
uckelman, line 416: What's this override? I can't find it.

uckelman, line 416: I think you can delete this function, since you're providing an identical copy in the parent class now.

Cattlesquat: All fixed

riverwanderer:
Tested in VASSAL-3.7.0-SNAPSHOT-030fb4f-10822, for the following examples in a Player Hand:
1. Global Property with Change-Buttons  - into a player hand and into main module GP
2. Folder containing Action Button and GKCs - back into a player hand OK.
3. Repeated (2) but importing into main module - no option for Import - should there be one?
4. Repeated (2) but importing into main module folder this is prevented with an error message:
image
This is in line the existing prohibition on pasting folders from a map/hand window to the module level (is that required ?). Even so, 3 & 4 don't seem like a big deal.
5. Export / Import of Prototype - worked fine.

Overall: 🐕 ⚽  ⚽

PR #12071, opened 2023-03-03
10822 - Sounds can be suppressed during GKC processing

### LINKED ISSUE #10822: Sound suppression in Global Key Commands
Global Key Commands allow downstream reporting to be suppressed. A similar requirements exists with Sound. The issue is exacerbated by limitations in sound handling that can result in unreliable sound (multiple .wav) or additive volume (multiple .mp3). The current workaround requires additional traits and logic by which the GKC can prevent sound traits from firing.

This request is for a checkbox like the "Suppress individual reports" checkbox that would instead "Suppress individual sounds".

This is envisaged as a way of providing module developers with a more reliable and familiar method to suppress multiple sounds, giving them more control using another method to generate the sound they might require for the GKC action.

Forum thread [here](https://forum.vassalengine.org/t/sound-suppression-in-gkc/72719).

Cattlesquat:
Fixes #10822 

@riverwanderer Here's one of your oldies -- give it a whirl and let us know if it's ready for prime time :)


vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
```diff
@@ -817,6 +833,7 @@ public int hashCode() {
     result = prime * result
       + ((reportFormat == null) ? 0 : reportFormat.hashCode());
     result = prime * result + (reportSingle ? 1231 : 1237);
+    result = prime * result + (isSuppressSounds() ? 1231 : 1237); // No idea what is appropriate here, just copying the mysterious line above
```
uckelman, line 836: Oh wow, that is the way old school way of doing this. The entire body of this function could be replaced with this:
```
return Objects.hash(keyStroke, reportFormat, reportSingle, selectFromDeckExpression, suppressSounds);
```

Cattlesquat: updated

riverwanderer:
Tested VASSAL-3.7.0-SNAPSHOT-030fb4f-10822 - sound suppression works like a charm. Especially neat with the Action Buttons everywhere change - e.g. multi-card draw; make one sound with the Action Button and suppress the normal GKC sound.

Issue #12077, opened 2023-03-05
Error from GPIW expression when a module starts up


riverwanderer:
Vassal v3.6.14.
GPIW folder expressions appear to be evaluated when a module starts up. On starting a module with a complex folder expression, a boolean error is generated (audit below). The GPIW otherwise works normally.

Observations:
1. The "PropertyValue" is the Description field (ie name) of the GPIW.
2. The Audit identifies the source field as Label for Pieces, but the Source Expression comes from the Label for Folders field.
3. Adding the otherwise unnecessary filter `PropertyValue=="[GPIW Description]" ? "" :` to the expression causes the error to go away. Equivalent filters also work, such as (in this case), Type!="Card" ? "" : to prevent 
4. The problem might be related to the complexity of the expression, as I have been unable to replicate it with a simple boolean expression.
5. In the expression, onPlayerMat and BSetInPlay are Boolean values, defined as false by default (module level). The built-in property ObscuredToOthers is also referenced.

Is it necessary to process a GPIW before a game is opened, when no data will exist for the GPIW to process?
Should / Could GPIW buttons be disabled until the game is opened ?

```
Expression Audit:
   Source field=Label for pieces
[...]

PR #12078, opened 2023-03-05
12077 - ObscuredToOthers can be counted on to be 'false' for pieces that don't even have a Mask property

Cattlesquat:
@riverwanderer I think what was going on in 12077 might have involved ObscuredToOthers, which I think is the only thing you were using "unary !" with. This actually also does MOST of the thing you'd requested in some other item -- gives pieces a "false" to return if they don't have a Mask property.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/BasicPiece.java
+++ b/vassal-app/src/main/java/VASSAL/counters/BasicPiece.java
@@ -310,6 +310,9 @@ public class BasicPiece extends AbstractImageFinder implements TranslatablePiece
     else if (Properties.VISIBLE_STATE.equals(key)) {
       return "";
     }
+    else if (Properties.OBSCURED_TO_OTHERS.equals(key)) {
+      return false;
+    }
 
     // Check for a property in the scratch-pad properties
     Object prop = props == null ? null : props.get(key);


riverwanderer:
Yes, `ObscuredToOthers` is in that expression, and may be the immediate cause, as it is undefined as you say. The change sounds generally helpful. I have one worry - prior to this change it is easy to check if a piece is masked (e.g. to include only cards) to a test like `ObscuredToOthers != ""`. More likely that `Obscured != ""` would be used. I can check the modules I've worked on but just afraid that some modules might get obscurely broken by this change.

Still wonder why a GPIW is getting evaluated at module open, before any data is available.

Cattlesquat:
Well that's a good point that someone could be depending on the existing "bad behavior" of the property

I don't think we can do anything practicable about when GPIW runs its expressions. It's full of listeners and things - once we "build" it, those things fire off when they want to.

riverwanderer:
Tested build VASSAL-3.7.0-SNAPSHOT-22f0e31-12077 but it doesn't have ObscuredToOthers set to anything on the unmasked piece that I tried, and from Debug the property is showing null.

Also, I tried what I think would be the equivalent - setting a module level GP ObscuredToOthers = false. That does show up, but #12077 still occurs.

PR #12080, opened 2023-03-07
12080 - Zooming in and out no longer loses track of player's preferred center point for map

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -3768,25 +3793,41 @@ public String getIdentifier() {
...
+    scroll.unregisterKeyboardAction(
+      KeyStroke.getKeyStroke(KeyEvent.VK_PAGE_UP, 0));
+    scroll.setAlignmentX(0.0f);
+    scroll.setAlignmentY(0.0f);
+
+    layeredPane.setLayout(new InsetLayout(layeredPane, scroll));
+    layeredPane.add(scroll, JLayeredPane.DEFAULT_LAYER);
+
+    // When our Viewport changes for any reason, check if we need to update our "preferred center point"
+    scroll.getViewport().addChangeListener(new ChangeListener() {
```
uckelman, line 3815: Change listeners can be replaced with lambdas:
```
scroll.getViewport().addChangeListener(e -> updateCenter());
```

Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -337,6 +339,32 @@ public void setSuppressSounds(boolean suppressSounds) {
...
 
+
+
+
+  private final Timer zoomTimer = new Timer(1000, new ActionListener() {
+    @Override
+    public void actionPerformed(ActionEvent e) {
+      suppressAutoCenterUpdate = false;
+    }
+  });
```
uckelman, line 350: The listener can be a lambda here:
```
private final Timer zoomTimer = new Timer(1000, e -> suppressAutoCenterUpdate = false);
```

Cattlesquat: Nifty spifty - fixed

PR #12084, opened 2023-03-08
12084 - Help menu shows module-specific help items before generic Vassal ones

Cattlesquat:
Basically if a module designer goes to the trouble of making some specific help for the module, then that is the stuff that ought to be at the top of the help menu.

I based this on 3.6.15 since it could go there. But it could also be punted into 3.7 if we're not looking for more, or not looking for trouble, or whatever.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
+++ b/vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
@@ -125,7 +125,6 @@ public class PlayerWindow extends JFrame {
     helpMenu.add(mm.addMarker("Documentation.Module.end"));  //NON-NLS
 
     helpMenu.add(mm.addMarker("Documentation.VASSAL.start"));  //NON-NLS
-    //helpMenu.add(mm.addKey("General.help"));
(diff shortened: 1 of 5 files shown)


vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
```diff
@@ -166,14 +166,15 @@ public void actionPerformed(ActionEvent e) {
...
-    }
+    // This file doesn't have particularly useful information, and the weird entry ("Help"???) was cluttering the Help menu.
+    //try {
+    //  final URL url = new File(Documentation.getDocumentationBaseDir(),
+    //                           "README.html").toURI().toURL();
+    //  mm.addAction("General.help", new ShowHelpAction(url, null));
+    //}
+    //catch (MalformedURLException e) {
+    //  ErrorDialog.bug(e);
+    //}
```
uckelman, line 177: Is there any reason to leave this code here?

Cattlesquat: Nope. Just removed it.

PR #12100, opened 2023-03-15
12100 - Text Labels support <img> tags

Cattlesquat:
If a Text Label trait is wrapped in <html> tags and ALSO contains an  tag, then we use JTextPane instead of JLabel to draw its HTML, which allows us to use the DataArchiveHTMLEditorKit previously developed for e.g. the Chat Log

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Labeler.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Labeler.java
@@ -44,7 +44,7 @@ import VASSAL.tools.imageop.AbstractTileOpImpl;
 import VASSAL.tools.imageop.ImageOp;
 import VASSAL.tools.imageop.Op;
 import VASSAL.tools.imageop.ScaledImagePainter;
-import VASSAL.tools.swing.DataArchiveHTMLEditorKit;
+import VASSAL.tools.swing.LabelerDataArchiveHTMLEditorKit;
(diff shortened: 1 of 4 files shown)


Cattlesquat:
(alllllmost working, except it isn't presently loading the images UNLESS i run in the debugger & wait at a breakpoint for a while)

vassal-app/src/main/java/VASSAL/counters/Labeler.java
```diff
@@ -607,14 +615,60 @@ public BufferedImage eval() throws Exception {
 
       // paint the foreground
       if (fg != null) {
-        final JLabel l = makeLabel();
-        l.paint(g);
+        if (txt.contains("<img")) { //NON-NLS
```
uckelman, line 618: Should it have a trailing space, i.e., "<img "?

Cattlesquat: No, and in addition to the aesthetic value of supporting a plain  or  tag, I was even realizing that an  tag in a string that only *sometimes* contains an image would have the extra value of making both versions run the JTextPane version rather than the JLabel version, so the text would resolve to the same location. (The JLabel and JTextPane versions don't naturally resolve to appearing in quite exactly the same positions as each other, and I have neither the time nor the beer to make them cooperate perfectly, if you know what I mean)

PR #12105, opened 2023-03-17
Fix property source for Deck's 'Send-to-Deck' component so it can use e.g. PlayerSide

Cattlesquat:


Cattlesquat:
JoelCFC25 tested successfully

Cattlesquat:
Grifo tested second part successfully

PR #12124, opened 2023-03-20
4446 - Dynamic button labels (can use expressions and $...$ in certain toolbar buttons) 

### LINKED ISSUE #4446: Substitute global properties in toolbar button labels
| Bugzilla |  |
|----------|----------|
| ID | BZ1849 |
| Reported | 2006-11-28 06:07:00 +0100 |
| Modified | 2020-06-22 00:11:59 +0200 |
| Product | VASSAL |
| Component | Unknown |
| Version | unspecified |
| Platform | All |
| OS | All |
| Status | NEW |
| Resolution | None |
| Priority | medium |
| Severity | enhancement |
| Milestone | --- |
| Reporter | [Rodney Kinney](mailto:rodneykinney@comcast.net) |

<pre>---- Reported by rodneykinney@comcast.net 2006-11-28 06:07 ----

Toolbar buttons could use the $var$ syntax when specifying their text value, reading properties from the map/module and updating appropriately.

This bug was previously known as _bug_ 1604277 at http://www.vassalengine.org/trackershow_bug.cgi?id=1604277

Unknown Component 
   Previous assignee is unknown.
Unknown resolution "".

Cattlesquat:
image

Fixes #4446

Supports expressions in the button text field of anything that's an "AbstractToolbarItem", so lots of the main candidates e.g. Action Buttons and GKC buttons.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -2312,6 +2312,8 @@ public class GameModule extends AbstractConfigurable
     buttonLabelUpdateScheduled = true;
 
     SwingUtilities.invokeLater(() -> {
+        buttonLabelUpdateScheduled = false;
+
         updateToolbarButtons();
 
         for (final Map map : Map.getMapList()) {
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -142,6 +142,7 @@ import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 import javax.swing.JToolBar;
 import javax.swing.KeyStroke;
+import javax.swing.SwingUtilities;
 import javax.swing.Timer;
 import java.awt.Container;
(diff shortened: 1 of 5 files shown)


vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -316,13 +318,29 @@ public boolean isTrueMovedSupport() {
   }
 
   /**
-   * @param fancyMoveSupport true if a ignore-small-moves trait exists in the module
+   * @param trueMoveSupport true if a ignore-small-moves trait exists in the module
```
uckelman, line 321: Param name doc doesn't match code.

Cattlesquat: I'm absolutely baffled that you're seeing this here, because it's not in my code that I pushed, and I even went back to master & then re-checked it out just in case to make sure ?!?!

Cattlesquat: Oh I see it was a 1-letter-off thing. Fixed.

PR #12128, opened 2023-03-21
12093 - Fix disabling of toolbar items nested in multiple Toolbar Menus

Cattlesquat:


riverwanderer:
Tested VASSAL-3.6.15-SNAPSHOT-6a5d4ac-12093for3.6;

working for menu items within a nested Menu Button (i.e. a Menu Button within another Menu Button) - these items  can now be disabled.
not yet working for the Menu Button itself.

Not yet tested whether the fix is working at deeper levels of nesting.

riverwanderer:
Update: Tried testing a next level down nested Menu Button (i.e. "level 3"). Looks like these aren't supported at all - the Menu Items fail to appear, leaving the level 3 Menu Button empty.

riverwanderer:
@Cattlesquat , I have realised that the fixed component (Module / Toolbar Menu / Toolbar Menu / [buttons]) is not fully working. It is a bit peculiar but perhaps it will ring a bell for you ...

1. WORKS - when the controlling GP is set as part of a feature that is (optionally) triggered by a Startup GKC.
3. DOES NOT WORK - when the same feature is instead triggered manually after scenario startup has completed.
4. Can be made to work in the Editor by certain actions  - e.g. I cleared the "Disable" option on the buttons' parent Toolbar Menu, and then the disabling properties on those buttons took effect. Otherwise in a normal Player session the buttons will not disable. This last one is not new behaviour -we had that situation for other types of toolbar component (e.g. Change-Property) when you first introduced the disabling feature.

In all cases I checked the values of the controlling GPs and all were set true at the appropriate time (i.e. showing that the button in question ought to be disabled).

riverwanderer:
I moved another set of buttons in the C&C Ancients module into a nested Toolbar Menu. These buttons change more dynamically... I can see that they set up correctly but then won't change based on changes to the controlling GPs.

I think that common thread with these examples of "disable/enable" Toolbar Menu contents not working is failure to refresh the Toolbar Menu contents when a change is done to a controlling GP.  Perhaps this also explains the controlling GP on the Toolbar Menu itself not working but I have not tested that one for this, yet.

riverwanderer:
A change that partially fixes the issue for Main Toolbar Menu Buttons went live in Vassal 3.6.15. However, that change introduced issue #12180 , affecting nested buttons in Map Toolbar windows.

PR #12136, opened 2023-03-23
12136 - Change Property Buttons now have a description field, which is displayed in the editor if filled in

Cattlesquat:


vassal-app/src/main/java/VASSAL/build/module/properties/ChangePropertyButton.java
```diff
@@ -111,29 +112,35 @@ public PropertyChanger getPropertyChanger() {
...
-      Resources.getString("Editor.ChangePropertyButton.options")
-    );
+    String[] descs = { Resources.getString("Editor.description_label") };
+    for (final String s : super.getAttributeDescriptions()) {
+      descs = ArrayUtils.add(descs, s);
+    }
+    descs = ArrayUtils.add(descs, Resources.getString("Editor.report_format"));
+    descs = ArrayUtils.add(descs, Resources.getString("Editor.ChangePropertyButton.options"));
+    return descs;
   }
```
uckelman, line 122: A few things here:

* Rather than a loop, you can use `ArrayUtils.addAll()` to do it in one shot: 

    ```descs = super.getAttributeDescriptions(descs, super.getAttributeDescriptions());```

That will be more efficient, because `ArrayUtils.add()` makes a new array on each call.

* But, this is still kind of crappy, making three new arrays when you could do it with one... I'm going to write a little utility function for this...

uckelman: I think this would do the same but with less generation of garbage?
```
final Stream.Builder<String> b = Stream.builder().add(Resources.getString("Editor.description_label"));
Arrays.stream(super.getAttributeDescriptions()).forEach(s -> b.add(s));
b.add(Resources.getString("Editor.report_format"))
 .add(Resources.getString("Editor.report_format"));

return b.build().toArray(String[]::new);
```

vassal-app/src/main/java/VASSAL/build/module/properties/ChangePropertyButton.java
```diff
@@ -111,29 +114,34 @@ public PropertyChanger getPropertyChanger() {
...
-      ReportFormatConfig.class,
-      PropChangerOptions.class
-    );
+    Class<?>[] types = { String.class };
+    for (final Class<?> c : super.getAttributeTypes()) {
+      types = ArrayUtils.add(types, c);
+    }
+    types = ArrayUtils.add(types, ReportFormatConfig.class);
+    types = ArrayUtils.add(types, PropChangerOptions.class);
+    return types;
```
uckelman, line 133: Here, too.

Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/build/module/properties/ChangePropertyButton.java
```diff
@@ -111,29 +114,34 @@ public PropertyChanger getPropertyChanger() {
...
-      REPORT_FORMAT,
-      PROPERTY_CHANGER
-    );
+    String[] names = { DESCRIPTION };
+    for (final String s : super.getAttributeNames()) {
+      names = ArrayUtils.add(names, s);
+    }
+    names = ArrayUtils.add(names, REPORT_FORMAT);
+    names = ArrayUtils.add(names, PROPERTY_CHANGER);
+    return names;
```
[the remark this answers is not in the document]
Cattlesquat: fixed

PR #12143, opened 2023-03-24
12140 - Action Buttons on pieces that have just been drag-moved are suppressed for 1/2 second for user-error-avoidance

Cattlesquat:


Cattlesquat:
I've cherry-picked it over to 3.6.15

PR #12149, opened 2023-03-27
BorderOutline trait accepts two fields

Cattlesquat:
This is an improvement to a feature that didn't exist until 3.7.0-beta1.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/BorderOutline.java
+++ b/vassal-app/src/main/java/VASSAL/counters/BorderOutline.java
@@ -178,36 +178,36 @@ public class BorderOutline extends Decorator implements TranslatablePiece {
     return true;
   }
 
+  protected boolean checkProperties() {
+    final boolean p1 = checkProperty(propertyName);
+    if (propertyName.isEmpty()) return true;
+
+    if (propertyName2.isEmpty()) {
+      return p1;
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/counters/BorderOutline.java
```diff
@@ -55,6 +59,47 @@ public BorderOutline(String type, GamePiece p) {
...
+    XOR("XOR"), //NON-NLS
+    NOR("NOR"); //NON-NLS
+
+    private static final String[] KEYS = { "Editor.AND", "Editor.OR", "Editor.XOR", "Editor.NOR" };
+
+    String symbol;
+
+    LogicalCompareMode(String symbol) {
+      this.symbol = symbol;
+    }
```
uckelman, line 77: Shouldn't these two have visibility specified?

Cattlesquat: It let me make "symbol" private, but compiler doesn't seem to want/allow visibility specification for the constructor

uckelman: There is some weird language thing about enums that I always forget... I think you can't set visibility on them.

PR #12176, opened 2023-04-03
12176 - Preference to not *display* Mark Moved status of pieces

Cattlesquat:


uckelman:
https://forum.vassalengine.org/t/mark-moved-bug-report/76866

PR #12184, opened 2023-04-05
12184 - Pivot can have two key commands/angles

Cattlesquat:
Requested by Benkyo

image


Cattlesquat:
Ben Kyo tested & was happy with it

uckelman:
What's the reason for adding a second angle instead of just adding a second Can Rotate trait?

Cattlesquat:
(FWIW this is Pivot, not Rotate) Being able to pivot forward-then-reverse (e.g. +90, -90) seemed like a logical use case. 

I will admit I don't fully understand the inner relationships of Pivot and Can Rotate and so wasn't personally sure two Pivot traits would even work? 

Ultimately I was just adding it since I was asked :)

PR #12191, opened 2023-04-07
12191 - Dunh Dunh DUNH! -- New "Attachment" trait

Cattlesquat:
image

It's like the byblow of a Global Key Command and a Mat!!!
* First you use a GKC-like filter to "attach" to one or more pieces.
* The pieces you attach to go in a list on the piece (sort of like mat cargo are in a list, except you don't have to co-locate the attached pieces, and they don't move together or anything)
* Now you can use "Current Attachments" fast match to very quickly send GKCs to your attachments!
* ALSO, the piece with the Attachment trait can read all the properties of the first attached piece in the list (for any given Attachment trait -- you can always have multiple Attachment traits with different commands/names, and so you could read properties from multiple pieces at once that way)
* <s>If the attached piece has Dynamic Properties, then you can also _write_ their values using the same naming convention, using a "Set Global Property" trait.</s>
* A SetAttachmentProperty trait allows you to set any Dynamic Property of any attachment
* You can clear the attachment if you want w/ a different key command
* CountAttachment & SumAttachment -- also MaxAttachmeent & MinAttachment
* CountAttachmentExpression("MyTurrets", "{ Ammo > 2 }") 
* GetAttachmentProperty(attachmentName, propertyName, index)   // Index starts at 1
* Have added "clear *matching* attachments" which allows clearing individual attachments without clearing all
* Have added "on attach" policy when another piece attaches to this w/ same Attachment Name -- options are either "do nothing", "attach back", "attach back to everything that piece is attached to (w/ same Attachment Name)"
* Have added "on detach" policy when using the Clear All or Clear Matching commands -- options are either "do nothing" or "sever the mutual attachment back to this piece, if any"
* Sibling_AttachCount property tells the number of attachments (for an Attachment named "Sibling")

Use cases will be especially in situations where you have a piece that wants to communicate with a particular other piece (or pieces) frequently -- now the perf-intensive GKC search only has to be done once. OR, attach to a single piece whose properties you want to be able to read (e.g. use inside of expressions)


BrentEaston (commented):
I feel the use of index numbers in the getAttachProperty to be counter-intuitive. Why not (or as well) by Basic Name of the piece? If you want the value of the property of the piece named 'X', how do you know what index it is? You will have no idea what order any attachments will have been made in.

BrentEaston:
How about generalizing the 'Set Attachment Property' trait to be a 'Set Property' trait, which is a kind of GKC, of which you can use a fast-match Current attachments option to do what the 'Set Attachment Property' does now? 

That would be really cool, if not terribly efficient, but will allow pieces to set properties on any other pieces without that whole 'put the value in a GP and send a GKC' rigmarole.

My only fear is that because the whole setting up the infrastructure of attachments and issuing the necessary commands to initialise the attachments is a bit clunky, people will tend to avoid them. 

What would be REALLY nice is if the initial attachments could be automated somehow. Not sure how you would do this. An Attachment Manager that handles this as pieces are added to the gamestate and the attachment traits reach out to each other? (Yes, does need a trait in each piece).

Have a think about this, something's been bugging me about them and I have finally put my finger on it. That's the biggest issue I see with attachments, having to organise the GKC's/Triggers to initialize the attachments in cases where the attachments are essentially static. Not everyone will want the full monty dymanic attach/detach that Grifo wants. A dumbed down super easy option would be handy. E.g. you define an attachment trait 'X' and add it to two counters. Vassal will then assure that those two counters will be attached whenever both are in the gamestate. 

I still like the idea a lot and it is great for advanced designers, but can we do a no-brains version as well?

Cattlesquat:
@BrentEaston Good stuff - I've added:
(1) GetAttachmentProperty can use either index or BasicName
(2) SetAttachmentProperty can likewise use either index or BasicName (or an expression involving either or both)

As far as automating the initial attachments... in the most basic case I would suggest a Startup Global Key Command of e.g. "SETUP" sent with a filter or fastmatch of { AttachName != "" }  -- that could activate every attachment trait in the game to "lock on" to its target. That's actually a lot less complex than a lot of things people have to do regularly already, and at the end of the day Attachment will always be "slightly more advanced than Global Key Commands" by their very nature. I mean we *could* provide pieces with an "At Start" _trait_ to schedule themselves to fire something off inside themselves e.g. at game start (that would of course require some extra handling and searching through the pieces at game start, but would have the advantage of being general purpose for all this-piece-needs-something-at-startup things) -- I'll ponder on it for either 3.7 or 3.8.

In terms of your general purpose SetProperty ... we'd probably want some other name because that makes it sound like the converse of GetProperty which it isn't ... so maybe SetPropertyOfOther or something. I think I'd rather keep that separate from the Set Attachment Property trait which has some specialized code/properties for handling attachments -- and also I think adding all the extra "global key command" boxes on top of everything else seems like it would get pretty confusing. But I *do* like the idea of a general purpose set-property-of-other so I'll put it on my list of good 3.7-or-3.8 features. Doesn't seem too hard to do.

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -299,6 +330,12 @@ Editor.BeanShell.sum6=Counts the total number of pieces in the same stack.
 Editor.BeanShell.sum7=Count the number of pieces with any non-blank value of named property, in the same stack.
 Editor.BeanShell.sum8=Sum the values of the named property among a Mat and its Mat Cargo pieces. 
 Editor.BeanShell.sum9=Count the number of pieces with any non-blank value of the named property, among a Mat and its Mat Cargo pieces.
+Editor.BeanShell.sum10=Sum the values of the named property across all pieces attached to this piece through the named Attachment trait. 
+Editor.BeanShell.sum11=Count the number of pieces with any non-blank value named property across all pieces attached to this piece through the named Attachment trait.
+Editor.BeanShell.sum12=Counts the number of pieces attached through the named Attachment trait that match a an expression.
```
uckelman, line 335: Typo? `a an`

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/module/AttachmentManager.java
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Maintains a list of pieces with attachment traits
+ */
+public class AttachmentManager {
+
+  /**
+   * Map of the attachment traits with the same attachment names
+   */
+  final Map<String, List<Attachment>> attachments = new HashMap<>();
+  final List<GamePiece> newPiecesWithAttachments = new ArrayList<>();
```
uckelman, line 39: Should these be `private`?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -536,6 +579,190 @@ public Object countStack(String property, PropertySource ps) {
...
+    ps = translatePiece(ps);
+
+    if (ps instanceof GamePiece) {
+      GamePiece p = Decorator.getOutermost((Decorator)ps);
+      while (p instanceof Decorator) {
+        if (p instanceof Attachment) {
+          final Attachment a = (Attachment)p;
+          if (a.getAttachName().equals(attachment)) {
+            final String matchString = replaceDollarVariables(expression, ps);
+            final PieceFilter filter = matchString == null ? null : new PropertyExpression(unescape(matchString)).getFilter(ps);
```
uckelman, line 751: Should the above two lines be outside the loop?

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Cattlesquat: no -- p changes during the loop

vassal-app/src/main/java/VASSAL/counters/Attachment.java
```diff
@@ -0,0 +1,1059 @@
...
+    }
+  }
+  */
+
+
+  public Command clearMatching() {
+    final GamePiece outer = Decorator.getOutermost(this);
+
+    clearTarget.fastMatchLocation = true;
+    clearTarget.fastMatchProperty = false;
```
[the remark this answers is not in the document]
Cattlesquat: yes?

vassal-app/src/main/java/VASSAL/build/module/AttachmentManager.java
Cattlesquat: fixed

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -105,6 +108,10 @@ public List<DrawPile> getModuleDrawPiles() {
     return theModule.getAllDescendantComponentsOf(DrawPile.class);
   }
 
+  final java.util.Map<GamePiece, GamePiece> updatedPieces = new HashMap<>();      //BR// maps old pieces to new pieces
+  final java.util.Map<GamePiece, GamePiece> formerPieces = new HashMap<>();       //BR// maps new pieces to old pieces
+  final java.util.Map<String, List<GamePiece>> attachmentIndex = new HashMap<>(); //BR// maps old attachment traits to their original targets
+
```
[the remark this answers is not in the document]
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/counters/Attachment.java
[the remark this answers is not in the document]
Cattlesquat: Those set up a "hardcoded" fast-match to always be "Current Attachments".

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Cattlesquat: Oh I misunderstood which lines you meant -- I chose not to instantiate the beanshell until we actually find a matching attachment -- we should only ever find precisely 0 or 1

Cattlesquat: In theory I guess I could put a "break' after finding a match, but that saves virtually nothing and conceivably some dodobrain could put 2 attachment traits with the same name

uckelman: I had not appreciated that detail about there never being more than one match. That might be worth a comment.

uckelman (approved):

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Cattlesquat: added

uckelman (approved):

PR #12197, opened 2023-04-08
12197 - OPmotize - faster calls to getMap(), getID(), getParent(), getPosition()

Cattlesquat:
We never subclass these four in any Decorator and it's difficult to imagine a viable subclassing of any of them existing in the wild. 

This little experiment gives every Decorator direct access to them instead of chaining a a bunch of calls/returns all the way in and out every time. 

Probably relatively mild perf games, but still (also easier to debug things in the debugger when a few more things don't go digging all the way to the core of the earth on every step-in) 

I ran through a whole game of PoG with it and it _seemed_ to work, but  ¯\_(ツ)_/¯

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Decorator.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Decorator.java
@@ -73,6 +73,7 @@ public abstract class Decorator extends AbstractImageFinder implements EditableP
   PropertyExporter, SearchTarget, ImageSearchTarget {
 
   protected GamePiece piece;
+  private GamePiece innermost;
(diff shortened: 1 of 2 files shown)


BrentEaston:
I did try an even more extensive (and invasive) version of this before, but I was never able to measure any appreciative performance difference. My version was based on property names,  the ones I knew where satisfied by the BasicPiece or the module, I jumped direct to BasicPiece. Joel suggested an even more audacious scheme where you identified and cached a direct link to the decorator that held the properties.  Not saying no...

BrentEaston:
Incorporated into #12267

PR #12220, opened 2023-04-14
Add BorderOutline's property names to the searchable target list

Cattlesquat:
(Minor fix/improvement to new 3.7 feature -- it wasn't adding the property names to the searchable list)

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/BorderOutline.java
+++ b/vassal-app/src/main/java/VASSAL/counters/BorderOutline.java
@@ -336,6 +336,6 @@ public class BorderOutline extends Decorator implements TranslatablePiece {
    */
   @Override
   public List<String> getPropertyList() {
-    return Arrays.asList(propertyName, propertyName2);
+    return List.of(propertyName, propertyName2);
   }
 }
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/counters/BorderOutline.java
```diff
@@ -327,4 +329,13 @@ private static class LogicalCompareConfigurer extends TranslatingStringEnumConfi
...
   }
+
+  /**
+   * {@link SearchTarget}
+   * @return a list of any Property Names referenced in the Decorator, if any (for search)
+   */
+  @Override
+  public List<String> getPropertyList() {
+    return Arrays.asList(propertyName, propertyName2);
+  }
```
uckelman, line 340: Tip: Since you already have `List` imported, you can use `List.of()` and not need to import `Arrays` additionally.

Cattlesquat: All fixed. I'll admit I can't keep all the different ways to assemble List/Array/Collection things in my head.

PR #12237, opened 2023-04-18
12237 - Fix edge case Mat Refresh bugs

Cattlesquat:
Brent noticed this edge case last night ... right after we released 3.6.16 hohoho

Basically things like ... when you refresh a piece it might turn out the new version _doesn't have a Mat trait any more_ and stuff like that.


BrentEaston (approved):

BrentEaston:
Looks good to me.

BrentEaston:
Was able to reproduce the problem and confirmed that this fix resolves the error

PR #12240, opened 2023-04-18
Update GitHub action environment to Ubuntu 22.10

uckelman:


uckelman:
No support for 22.10 right now, so this can't work.

Issue #11998, opened 2023-02-13
 Startup Action Button & added features


riverwanderer:
From a forum [post](https://forum.vassalengine.org/t/startup-action-button-added-features/74632)

Similar to getting Action Buttons at Map level, it would be super helpful to have a Startup Action Button.
Working like the StartUp GKC but with the scope of the Action Button.

One big advantage of this is that Action Buttons can generate Hotkeys directly, unlike a Startup GKC which requires logic on a piece to do anything useful.

Related to this, it could be useful to add these abilities to Startup GKC and Startup Action Buttons:-
- Ability to call the components from a Hotkey directly*
- Ability to disable via a property (= true)
*e.g. for integration with “in game” side switching, as this does not go through startup routines but might need to use the same logic.

PR #12243, opened 2023-04-19
12243 - Startup Global Key Commands can also send a Global Hotkey

Cattlesquat:


riverwanderer:
#11998 relates to this, suggesting instead adapting the Action Button to achieve Startup GHKs, and proposes other features similar to normal buttons.

Cattlesquat:
@BrentEaston Asked your review here mainly because I want to make sure the "sending the global hotkey" part is done "the right way"

BrentEaston (commented):
Happy with the way you are handling the execution of the Global Hotkey.

However, I see one issue.

In the configure, the Global HotKey appears before the Global Key Command. However, in apply(), you call super.apply() to run the GKC BEFORE the GHK runs. Either they need to be reversed, or possibly a better solution is to have a drop-down to select whether you want a GHK or a GKC? Having both is a bit confusing since all the match stuff does not apply to the GHK and I would think you would usually want one or the other. If you need both, you can always add a second one of the other type?

Cattlesquat:
@BrentEaston GREAT idea, I was worried about the confusion myself -- have changed it into a binary (you either send a hotkey or you send a global key command)

vassal-app/src/main/java/VASSAL/configure/VisibilityCondition.java
```diff
@@ -20,4 +20,19 @@
...
+    public VisibilityAND(VisibilityCondition a1, VisibilityCondition a2) {
+      and1 = a1;
+      and2 = a2;
+    }
+
+    @Override
+    public boolean shouldBeVisible() {
+      return ((and1 == null) || and1.shouldBeVisible()) && ((and2 == null) || and2.shouldBeVisible());
+    }
+  }
```
uckelman, line 37: `VisibilityAND` should probably be in its own file, and have `final` members.

vassal-app/src/main/java/VASSAL/build/module/StartupGlobalKeyCommand.java
```diff
@@ -139,16 +169,37 @@ public VisibilityCondition getAttributeVisibility(String key) {
...
+      //};
+      // THINGS THAT DON'T WORK
+      //return () -> (SEND_KEY_COMMAND.equals(hotkeyOrKeyCommand) && (super.getAttributeVisibility(key)).shouldBeVisible());
+      return new VisibilityCondition.VisibilityAND(() -> SEND_KEY_COMMAND.equals(hotkeyOrKeyCommand), super.getAttributeVisibility(key));
+
+      // THINGS THAT DO "WORK"
+      //return super.getAttributeVisibility(key);
+      //return () -> SEND_KEY_COMMAND.equals(hotkeyOrKeyCommand);
+      //return new VisibilityCondition.VisibilityAND(() -> SEND_KEY_COMMAND.equals(hotkeyOrKeyCommand), () -> SEND_KEY_COMMAND.equals(hotkeyOrKeyCommand));
     }
```
uckelman, line 192: We probably don't need all the comments here.

vassal-app/src/main/java/VASSAL/configure/VisibilityCondition.java
Cattlesquat: fixed

vassal-app/src/main/java/VASSAL/build/module/StartupGlobalKeyCommand.java
Cattlesquat: lol clean up the body bags

PR #12245, opened 2023-04-20
4382 - Allow limits of the total *number* of pieces a GKC can affect, and a property to find out how many it *did* affect

### LINKED ISSUE #4382: Report number of Units that matched a Global Key Command
| Bugzilla |  |
|----------|----------|
| ID | BZ1782 |
| Reported | 2008-12-09 10:39:00 +0100 |
| Modified | 2020-06-22 00:13:04 +0200 |
| Product | VASSAL |
| Component | Unknown |
| Version | unspecified |
| Platform | All |
| OS | All |
| Status | NEW |
| Resolution | None |
| Priority | medium |
| Severity | enhancement |
| Milestone | --- |

<pre>---- Reported by b.easton@exemail.com.au 2008-12-09 22:39 ----

Add an addition variable $Count$ to Global Key Command Report Formats that returns the number of GamePiees that matched the Property Expression.

This bug was previously known as _bug_ 2412649 at http://www.vassalengine.org/trackershow_bug.cgi?id=2412649

Unknown Component 
   Previous assignee is unknown.
Unknown resolution "".

Cattlesquat:
image

Always good to hit "the oldest bug in the buglist" :)


BrentEaston:
Options that would be useful:
 1. Stop GKC after {expression} number of matches found
 2. Stop GKC after {expression} number of pieces have been checked
 3. Perform GKC until {expression} is false
 4. Perform GKC while {expression} is true

If you are also adding properties that track the number of checks and number of matches, then those expressions could be combined in the while/until expression. e,g, perform GKC while expression is true or until you have applied it to 10 pieces.

In fact 1 & 2 are probably not needed, but it makes sense to include them as 'simple' options. 3 & 4 for more advanced users.

PR #12255, opened 2023-04-24
Work around Java's use of incorrect gamma for grayscale + alpha PNGs

uckelman:


uckelman:
Messed up the base. Will reopen.

PR #12327, opened 2023-05-25
12327 ! operator will always return false, not error if arg is invalid

BrentEaston:
Modified Beanshell parser to treat anything that is not true/false or "true"/"false" as false, instead of throwing a Data Module Error.

### DIFF
--- a/vassal-app/src/main/java/bsh/BSHUnaryExpression.java
+++ b/vassal-app/src/main/java/bsh/BSHUnaryExpression.java
@@ -82,14 +82,19 @@ class BSHUnaryExpression extends SimpleNode implements ParserConstants
 
     private Object unaryOperation( Object op, int kind ) throws UtilEvalError
     {
-        if (op instanceof Boolean || op instanceof Character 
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/bsh/BSHUnaryExpression.java
```diff
@@ -82,14 +82,19 @@ private Object lhsUnaryOperation( LHS lhs, boolean strictJava )
 
     private Object unaryOperation( Object op, int kind ) throws UtilEvalError
     {
-        if (op instanceof Boolean || op instanceof Character 
+        // VASSAL - Special handling for ! operator. Don't thow errors
+        // Anything that is not a Boolean should return false, unless it is the string "true"
+        if (kind == ParserConstants.BANG ) {
+          return primitiveWrapperUnaryOperation((op instanceof Boolean ? op : new Boolean("true".equals(op))), kind );
```
uckelman, line 88: The `Boolean` ctors are deprecated. Instead of `new Boolean("true".equals(op))`, what you probably want is `Boolean.valueOf(op)` instead.

PR #12340, opened 2023-06-02
Fix Inaccurate Hex Grid location reports

### LINKED ISSUE #12070: Inaccurate Hex Grid location reports
A hex grid set up with $gridLocation$ reporting will report an inaccurate LocationName. Initial investigation suggests that Vassal is determining the $gridLocation$ on the basis of a rectangular grid.

Perhaps this problem only affects mouse-over most of the time, as typical grids will have snap-to-location turned on. Impact on Mouse-over is typically that the mouse-over will seem difficult to trigger but also the mouse-over will be liable to trigger on the adjacent location instead. As shown by this simple example; a hexgrid with locations drawn in and a mouse-over reporting the LocationName incorrectly:
image

Confirmed in Vassal v3.6.14 and in v3.2.17.

Demo module, with maps for both orientations of hex grid: 
[test hexgrid locations.zip](https://github.com/vassalengine/vassal/files/10881994/test.hexgrid.locations.zip)

yetanotherdevaccount:
I am not familiar with the Vassal codebase, so this might not be a good approach to fixing the issue, or it may have bugs in it. I have only tested the specific case with `HexGrid`, not the other types of grids, or other situations than the one described in #12070 .

In particular:
- Are the changes in `RegionGrid` correct?
- Are the changes in `SquareGrid` correct?

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -1276,7 +1276,7 @@ public class Map extends AbstractToolbarItem implements GameComponent, MouseList
    * @see Board#snapTo
    * @see VASSAL.build.module.map.boardPicker.board.MapGrid#snapTo
    */
-  public Point snapTo(Point p, boolean force) {
+  public Point snapTo(Point p, boolean force, boolean onlyCenter) {
     Point snap = new Point(p);
 
     final Board b = findBoard(p);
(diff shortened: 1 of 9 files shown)


yetanotherdevaccount:
The issue in #12070 is not fixed completely, just the issue that BrentEaston is describing in https://github.com/vassalengine/vassal/issues/12070#issuecomment-1465316534 .

yetanotherdevaccount:
Thanks and credit to @riverwanderer , SwampWallaby and @uckelman : I have added two commits, which fixes the last missing inaccuracy of #12070 .

uckelman (approved):

PR #12362, opened 2023-06-08
12362 - Rollover component help button goes to MouseOver.html instead of Map.html

Cattlesquat:


riverwanderer:
Looking at the original code here: 
  'return HelpFile.getReferenceManualPage("Map.html", "StackViewer"); //NON-NLS` 

Clearly the Help button was intended to target the StackViewer anchor, on the Map page. The anchor is there and can be accessed from the browser by adding suffix `#StackViewer`. However, the Help button wasn't producing that result; instead it was displaying the top of the map reference manual page.

So why wasn't it ? I took a look at `getReferenceManualPage` and nothing obvious jumps out. Is there something wrong with the way `anchor` is handled in the return from that function ?

Whatever it is, it looks like a pervasive issue. I can't find any working use of an anchor from a Help button and some components don't have their own page so they need to go their anchor point.

Then there is the question of whether going to the detailed page is better than going to the anchor and letting the user click there for details. Some components already do this but is that just because they were changed to workaround the anchor issue?

PR #12402, opened 2023-06-26
12392 - Add option to Replace With Other trait to copy Marker and Dynamic Property values.

BrentEaston:


BrentEaston:
Re-coding from scratch

PR #12405, opened 2023-06-26
12401 - Allow Ctrl-V to be entered as a Key Command

BrentEaston:
Issue introduced into 3.7 beta. Waiting for Mark to check it works OK on MacOS.


BrentEaston:
This works on Windows and Mark has confirmed no problems on Mac.

PR #12407, opened 2023-06-26
12347 - Refresh Dynamic Properties correctly when they have no value.

BrentEaston:


BrentEaston:
Confirmed by Mark that this resolves the issue.

PR #12413, opened 2023-06-27
12403 - Extension Editing fixes: Cut & Paste now deletes cut element, Drag & Drop between folders in Extension are saved, stopped UnsupportedOperationException during search

BrentEaston:
Should also be back-ported to 3.6.20

1. Cut and Paste now deletes the cut element when Extension is saved.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/DoActionButton.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/DoActionButton.java
@@ -759,7 +759,9 @@ public class DoActionButton extends AbstractToolbarItem
    */
   @Override
   public List<String> getFormattedStringList() {
-    return Arrays.asList(reportFormat.getFormat());
+    final List<String> l = new ArrayList<>();
+    l.add(reportFormat.getFormat());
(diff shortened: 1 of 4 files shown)


vassal-app/src/main/java/VASSAL/build/module/DoActionButton.java
```diff
@@ -759,7 +759,9 @@ else if (LoopControl.LOOP_COUNTED.equals(loopType)) {
    */
   @Override
   public List<String> getFormattedStringList() {
-    return Arrays.asList(reportFormat.getFormat());
+    final List<String> l = new ArrayList<>();
+    l.add(reportFormat.getFormat());
+    return l;
```
uckelman, line 764: Why's this changing?

BrentEaston: Because it is now being over-ridden in SpecialDiceButton which attempts to add an additional entry to the returned list. Arrays.asList() returns an immutable list, so you get an UnsupportedOperation exception.

PR #12419, opened 2023-06-28
12392 - Fix Replace With Other state matching

BrentEaston:
- State matching is no longer dependent on order of traits in replacement
 - Markers will never be state matched
 - Dynamic Properties will be match on property name only, not full type.


BrentEaston:
Successfully tested by Mark.

vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
```diff
@@ -651,13 +657,25 @@ protected Ed(PlaceMarker piece) {
         p.add(aboveLabel, aboveConfig);
         aboveConfig.getControls().setVisible(piece.matchRotation);
         aboveLabel.setVisible(piece.matchRotation);
+
+        copyConfig = createCopyConfig();
+        copyLabel = new JLabel(copyConfig.getName());
```
uckelman, line 662: Isn't `copyConfig.getName()` going to NPE because `createCopyConfig()` returns `null`?

BrentEaston: Technically yes, but we are in an `if (aboveConfig != null)` block and if createAboveConfig() has returned not null, then createCopyConfig() will also return not null.

uckelman: I don't follow. The only thing in the body of `createCopyConfig()` below is `return null;`.

vassal-app/src/main/java/VASSAL/counters/PlaceMarker.java
```diff
@@ -713,6 +731,10 @@ protected BooleanConfigurer createAboveConfig() {
       return null;
     }
 
+    protected BooleanConfigurer createCopyConfig() {
+      return null;
+    }
```
uckelman, line 736: ^^^^ HERE ^^^^

BrentEaston: This is where an IDE helps you out. For me, I see:

image

Which tells me that this method is overidden in some subclass. Clicking on the little symbol (or clicking on the method name and Ctrl-V) takes me to:

image

Personally, I would necessarily not have coded it this way, as the superclass needs too much knowledge of what subclasses might do, but Decorator Editors are a bitch to subclass.

uckelman: I see. This is an awkward design, but seems to have been the way it is since 2008; not much we can do about it now. Thanks for the explanation.

PR #12439, opened 2023-07-04
Uniformly handle addition of standard and imported Configurables

uckelman:
Uniformly handle addition of tree elements.


BrentEaston (approved):
Good solution.

PR #12447, opened 2023-07-06
12434 - FastMatch by Location now respects Deck limits

BrentEaston:

PR #12448, opened 2023-07-06
12433 Attachment Fastmatch: match each piece at most once, limit checking by name

BrentEaston:
- Fastmatch by Attachment will match each piece at most once
- Add Fastmatch Attachment name (or blank for all) to limit Attachments checked

NOTE: This PR is dependent on PR #12447 (12434 - FastMatch by Location now respects Deck limits) which must be merged first


BrentEaston:
Tested successfully by Mark.

vassal-app/src/main/java/VASSAL/counters/GlobalCommand.java
```diff
@@ -959,6 +993,103 @@ else if (!reportFormat.equals(other.reportFormat))
...
+        // Piece is not in a Deck, just handle as normal
+        dispatcher.accept(piece);
+      }
+
+    }
+
+    /**
+     * Class to track the acceptance count of pieces from multiple Decks
+     */
+    class DeckInfo {
```
uckelman, line 1058: Probably should be `protected`.

BrentEaston: Done

PR #12456, opened 2023-07-09
12454 BeanShell Editor improvements: larger, multi-line, more properties

BrentEaston:
- Larger
 - Multi-line
 - More Vassal properties


riverwanderer:
Performed a simple test on build b2582f9-12454-BeanshellEditor a multi-line field in a DP trait. Entered some newlines to break the (commented) expression up.
Passed the following tests (a) saved field with no data loss/changes, (b) saved module with no data loss/changes and (c) still runs the expression without errors.

PR #12497, opened 2023-07-26
12495 - Mouse-over Stack Viewer 'Show move limited pieces' option no longer dependent on 'Show non-stacking' option.

BrentEaston:
Prior to allowing Does Not Stack pieces to Stack, movement limitations in the Stack Viewer where dependent on the piece already being Non-stacking. When Stacking was added to Does Not Stack, the Stack View configurer was not updated to reflect the fact that the non-movement options should now be selectable independently from the non-stacking option.


Cattlesquat (approved):
Looks good

Issue #12482, opened 2023-07-21
Map Shading toolbar entries are placed into wrong toolbar


JaviVilarroig:
I have experienced that in already two modules. One of them just a synthetic module for testing.

Quite easy to replicate.

* Add a Map Shading with Button Text defined.
* You will have a new entry in the toolbar
* Create a Toolbar Menu and add an entry with the same text as the Map Shading Button Text property

The expected result is that the menu option to toggle the Map Shading appears in the Toolbar Menu, but is continues to appear in the toolbar straight.

It can be very annoying if you have multiple Map Shadings

BrentEaston:
Hi,

I believe I have found the problem.

Vassal maintains multiple Toolbars, one for the module and one for each map. Module level components add their button to the module level toolbar and Map level components (are supposed to) add their button to the toolbar for their Map.

Unfortunately Map Shaders are not following this pattern and are adding their button to the Module level toolbar instead of the Map level one. This means that only Toolbar menus defined at the Module level will be able to grab Map Shader buttons, not toolbar buttons defined on it's own Map.

Very confusing and stoopid, but unfortunately, it has always been this way and I'm not sure we can change this without breaking every module that has ever added Map shaders to module level toolbar menus.

The work-around is to create a Module level toolbar for your Map Shaders. Not ideal, as you can't mix them with other Map level buttons, but hopefully this works for you?

Regards.

BrentEaston:
Ok. This seems to be the one and only map-level component that does this. Components that can be either Map or Module level already handle this correctly.

So one option is to add a checkbox 'Add button to Map-level toolbar' to this one component, defaulting to Yes, but remaining No for any existing modules. 

A second option would be to add a Compatibility option 'Map Shaders add button to Map-level toolbar' but that just makes it really hard to find. 

So option 1 would be my vote.

JaviVilarroig:
Hi @BrentEaston 

Thanks for your analysis and your answer :)

I will try the workaround as soon as I can.

For the fix, I think is better option 1.

BrentEaston:
Fix will be in 3.7 beta 5

JaviVilarroig:
Thanks!!!

PR #12500, opened 2023-07-27
12482 - Map Shader now able to add button to Map Toolbar

BrentEaston:
This bug has been around forEVER.

Map Shader buttons have always been added to the Module toolbar instead of the Map toolbar, so cannot be included with other buttons from the same Map on a map-level toolbar menu. Force changing the toolbar will break existing modules that have included Map Shader buttons on a Global Toolbar

This adds an option to enable selecting which toolbar to create the buttons on. Default for existing Map Shaders is the existing defined toolbar (which will be Module for all pre-3,.6 modules). Default for newly created Map Shaders is the Map toolbar.


Cattlesquat (approved):
Brent this looks good to me. Does this fix that ancillary "Toolbar Menu" issue that originally drew our attention to it?

BrentEaston:
Brian, do you mean issue #12482? This doesn't change any behaviour in existing modules as this would break any module where someone has already add Map Shaders to a Global Toolbar. This just gives you the option to select which Toolbar to add the button to, and defaults to the correct one (Map) when creating a new Map Shader.

Of course, the next thing that is going to happen is someone will want all the other Map buttons to optionally be on the Module toolbar, but I am pretending that's not going to happen.

Cattlesquat:
Ah, okay, makes sense

PR #12508, opened 2023-07-30
11104 Refresh Can Rotate trait by Rotator Name match option for Game Refreshers

BrentEaston:


BrentEaston:
Tested successfully by Mark.

PR #12513, opened 2023-07-31
12270 Scenario Options tweaks

BrentEaston:


riverwanderer:
Tested this build. Couple of observations...

1) After locking, the unlock button is greyed out against use.  However, closing and re-entering, it becomes available and henceforth seems to behave itself.
image

2) This one may have existing before as its a while since I tested Lock/Unlock; the associated chat log message doesn't cater for the Tab name having a html tag... e.g:
image

ideally, either the lock/unlock report should detect and reposition the `<html>` tag or the entire Scenario Options component should respect the Global Option setting for html in chat, or just allow html as standard.

I tested the right/left justify, that worked fine for me.

BrentEaston:
Have pushed an update which should resolve both the HTML tab name issue and the Unlock button not enabling first time round.

riverwanderer:
Test of VASSAL-3.7.0-SNAPSHOT-23b08e2-12270-ScenarioOptionsChanges succeeds for html tag and Unlock button issue.

One further issue - when used once, the Lock/Unlock button was initially accessible to `<observer>` but once used again, became disabled. To demo;  connect as non-observer; lock or unlock scenario options; switch to observer and repeat scenario options, noting lock/unlock is enabled. The game can be saved in this state and the button will still be available to observer in the re-opened game.  Press the button once again (as observer) and it becomes disabled until another player role enables it.


dI was able to switch into Observer mode and still

riverwanderer:
> i.e. if you Lock tab, then retire to Observer, you will be able to unlock that tab if you don't change the password.

That applied to my test. All good then.

PR #12517, opened 2023-08-01
12393 - Fix Auto-attach breaking on Undo

BrentEaston:
Major re-work of Attachment handling
 - All Auto-attaching is now handled locally on individual clients, no Commands are generated.
 - Fixes issues with Undo of Auto-attach causing strange problems
 - Deleting/Replacing units now generates correct un-attach Commands to undo any attachments in play
 - Should prevent potential weird problems when UNDOing the delete of pieces before un-attaching them.


riverwanderer:
Tested build VASSAL-3.7.0-SNAPSHOT-dc3f738-12393-AutoAttachUndo … it breaks  attachment-based functions in the current C&C Nap dev module. 

I suggest not to include this in Beta5 until I have time to provide more targeted testing / diagnostics. 

Note that, so far for my uses at least, I have generally been able to workaround undo limitations. This may not be possible for all use cases; mine are generally 1-1 or 1-many attachments. It is when both sides of the relationship use “auto” that I was seeing undo issues. I will try to extract an example from my module.

BrentEaston:
Based on Discord chat with Mark, the issue was due to an Attachment bug creating reverse attachments when they should not have been. His module took advantage of this. The bug corrected this issue, which broke his module. The module has been corrected and looks to be working

BrentEaston:
Testing successfully completed by Mark

PR #12531, opened 2023-08-06
12444 - Implement real Unique Piece Id property UniqueID that does not change during refresh. 

BrentEaston:
Add ParentId property to placed markers.
Update all related documentation

My apologies for this extra one, but it is big issue for more advanced module designers and makes using place markers far, far easier without needing reams of traits to process communication back to the parent piece.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/BasicPiece.java
+++ b/vassal-app/src/main/java/VASSAL/counters/BasicPiece.java
@@ -946,7 +946,14 @@ public class BasicPiece extends AbstractImageFinder implements TranslatablePiece
     // Persistent Property values will always be String (for now).
     // Create the HashMap as lazily as possible, no point in creating it for pieces that never move
     if (persistentProps != null) {
+      // Maintain the value of UNIQUE_ID, as it will have been defaulted for legacy units when they are first
+      // created, prior to getting their state set (which does not specify a value for UNIQUE_ID).
+      // Post UniqueID units will just overwrite this default from the new state being set
+      final Object uniqueId = persistentProps.get(UNIQUE_ID);
       persistentProps.clear();
+      if (uniqueId != null) {
+        persistentProps.put(UNIQUE_ID, uniqueId);
+      }
     }
     final int propCount = st.nextInt(0);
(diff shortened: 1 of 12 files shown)


riverwanderer:
Tested **VASSAL-3.7.0-SNAPSHOT-33165e9-12444-Add_ParentPieceUID**; both `ParentId` and `UniqueId` are reported as null (from the piece that was placed).

BrentEaston:
Ah, I had the property names all over the place. 

New version has the property names in code and doco as `UniqueID` and `ParentID` (This matches the existing ones best. In the version you have they are `parentId` and `uniqueId`.

riverwanderer:
UniqueID works now  but ParentID (and parentId and parentID) reports as null in the placed piece. Maybe UniqueID is mis-capitalised in the code that sets ParentID ?

BrentEaston:
New build fixes both problems you found:
 - Pieces in old save games now get the UniqueID set
 - Clonde pieces get an updated UniqueID.

riverwanderer:
Tested VASSAL-3.7.0-SNAPSHOT-20a6e6d-12444-Add_ParentPieceUID - UniqueID now set in old saved game pieces. A GKC back to the Place Marker piece on `UniqueID == $ParentID$` now works fine.

Cloned pieces get a new UniqueID, ParentID null, CloneID set to original piece's UniqueID. I'm wondering if ParentID should be defined for Cloned piece anyway.

BrentEaston:
> I'm wondering if ParentID should be defined for Cloned piece anyway.

It's not clear what should happen when you start Cloning Placed Markers.

riverwanderer:
Just realised that you probably set it so that ParentID gets cloned through if the original piece was created via Place Marker.  Fair enough.

BrentEaston:
Yes, exactly. The Clones should all be the same. Are you happy with this one now?

riverwanderer:
> Yes, exactly. The Clones should all be the same. Are you happy with this one now?

Yes. I don't have an immediate use for the Clone aspect of this change but I understand the logic of CloneID.

PR #12601, opened 2023-08-22
12578 - Process Set Piece Property expressions correctly

BrentEaston:


BrentEaston:
Tested successfully by Mark

vassal-app/src/main/java/VASSAL/configure/BeanShellExpressionConfigurer.java
```diff
@@ -210,7 +210,7 @@ public Component getControls() {
       validator = new Validator();
       nameField = new JTextArea(1, 100);
 
-      nameField.setFont(new Font("Monospaced", Font.PLAIN, 14));
+      nameField.setFont(new Font("courier new", Font.PLAIN, 14));
```
uckelman, line 213: "courier new" isn't guaranteed to exist. What happens when it doesn't?

uckelman: Also, why have this differ from the monospaced fonts everywhere else?

BrentEaston: oops, that was a test to confirm that on my system, Courier New was in fact the selected monospaced on windows. Fixed.

PR #12612, opened 2023-08-24
Trutype Font test for Beanshell editor

BrentEaston:


BrentEaston:
Replaced with Updated version

PR #12630, opened 2023-09-01
Truetype Font Support

BrentEaston:
- Includes a high quality monospaced font JetBrains Mono for use in the expanded Calculated Property window
 - Allows Designers to add TrueType/OpenType Fonts to a module (New top level Fonts component)
 - All new fonts (Vassal supplied and designer supplied) can be used in the following contexts:
   o Text Label selected font
   o Game Piece Image font
   o Chat Window font
   o Any Swing renderered HTML
 - Fonts added to a module will be used on all target machines the module is run on UNLESS the user already has that font installed on their system, in which case the installed font will be used.


BrentEaston:
Trying different approach

PR #12633, opened 2023-09-02
12235 - Allow expression for HTML Help Starting Page

BrentEaston:


riverwanderer:
Tested build VASSAL-3.7.0-SNAPSHOT-992392a-12235-HTML-Help-Expression with dev module for C&C Nap. A variable HTML start page worked with simple expression and beanshell. A problem in earlier builds with re-evaluating the expression on each invocation has been resolved.

vassal-app/src/main/java/VASSAL/configure/BeanShellExpressionConfigurer.java
```diff
@@ -323,6 +323,18 @@ protected void updateParentBuilder() {
...
+      final String fileName = "/fonts/JetBrainsMono.ttf";
+      final URL url = getClass().getResource(fileName);
+      final Font ms = Font.createFont(Font.TRUETYPE_FONT, url.openStream());
+      return ms.deriveFont(style, size);
+    }
+    catch (Exception ignored) {
+    }
+    return new Font("monospaced", style, size);
+  }
+
```
uckelman, line 337: It looks like you got some font changes in here accidentally.

BrentEaston: Not so accidentally :) This is the minimal font additional code to provide a decent font in the Expression editor. Completely compatible with what I want to do later. Plus it will annoy the shit out of P until 3.8.

vassal-app/pom.xml
```diff
@@ -401,6 +402,7 @@
                 <artifactId>clirr-maven-plugin</artifactId>
                 <configuration>
                     <ignoredDifferencesFile>src/test/resources/clirr-ignored-differences.xml</ignoredDifferencesFile>
+                    <comparisonVersion>${clirr.nofork.comparison.version}</comparisonVersion>
```
BrentEaston, line 405: I've made this change because it is pointless doing clirr nofork comparisons between the betas, which is happening now. The default seems to be the last release which includes betas.

Just means that <clirr.nofork.comparison.version>will need updating for each real release.

I also stripped the majority of the difference rules out of the ignore file as they are no longer relevant.

vassal-app/src/main/java/VASSAL/configure/BeanShellExpressionConfigurer.java
uckelman: This will prevent the system monospace font from being used if the hard-coded font is present in one of the JARs, right? Is that really what we want?

BrentEaston: For the one component where getEditorFont() is called, yes. that's the point. I'll make it a configurable selection in 3.8. It won't stop the system monospace font being used anywhere else.

uckelman: I think I don't understand what this is for, in that case. Why do we want to use something other than the system default _by default_?

BrentEaston:
Resubmitting

Issue #12632, opened 2023-09-02
Expand scope of "solo/referee" sides 


riverwanderer:
I propose a change to `isSoloSide()` to allow for multiple solo / referee sides within a module, and to allow for auxiliary sides with arbitrary names. An "auxiliary side" is one, for example, that may move pieces on the board within a turn but does not control the turn itself. An example of this is found in Commands & Colors modules (Epic & LGB format scenarios).

Currently, isSoloSide() returns true if the side equals "Solitaire", "Solo:", "Moderator" or "Referee".

I propose that isSoloSide() will return true if the side is "Solo" or **_begins_** "Solitaire", "Solo:", "Moderator", "Referee" or ".".

The "." prefix specifies an "auxiliary side".

isSoloSide() is used by PlayerRoster, ChessClockControl and CheckClock.
Within PlayerRoster, the change has the desired effect of skipping the now expanded "solo" sides.

The association with Chess Clock logic, implies that any sides starting "." will be impacted in the Chess clock functionality in the same way as the Solo sides (I recall that Solitaire "sides" don't get offered by the clock component). A designer will be unable to use the auxiliary side exclusion prefix if this is not acceptable for their module; this is likely very rare and no worse than the current situation.

In the unlikely event that an existing module contains a side name starting with a period (.), such a side will be excluded from the side switching default and chess clock. These impacts are minimal and easily addressed in an update to the module, if the new behaviour is undesirable.

riverwanderer:
Example use-cases for the expanded number of solo and other roles to be skipped when hot-seating.

1. A module may have special solo roles that can be started automatically if a user starts a game in that role. Example: C&C Ancients has 3 solitaire modes. These roles can be skipped by the "starts with" changes (e.g. starts with "Solo:".

2. Some roles don't apply for particular scenarios. Example: C&C Ancients has 6 auxiliary roles for Epic scenarios. These roles are not required if hot-seating. These roles can be excluded by defining the property _VassalSkipSidesList_ (which replaces the idea of using a special character to prefix side names.

PR #12636, opened 2023-09-02
12632 - Broaden search for "solo" PlayerSide.

riverwanderer:
Attempt improved side switch filtering, for issue #12632.


riverwanderer:
Tested successfully using C&C Ancients module for skipping names starting "Solo:" and "." as well as equals "Solitaire".

riverwanderer:
Doco update pending.

riverwanderer:
Component reference manaul [updated](file:///Applications/VASSAL-3.7.0-SNAPSHOT-8e34ce6-12632_moreSoloSides.app/Contents/Resources/doc/ReferenceManual/GameModule.html#Definition_of_Player_Sides).

Changed special prefix from dot to space.

Re-tested OK.

riverwanderer:
Further change avoids potential conflict with ChessClock use by creating a dedicated IsRefereeSide() function, separate from IsSoloSide().

vassal-doc/src/main/readme-referencemanual/ReferenceManual/GameModule.adoc
```diff
@@ -116,6 +116,9 @@ Simply type a name for each side and refer to that name in the restricted compon
 Only one player may be assigned to a side.
 When joining a game, players will be prompted to take one of the remaining available sides.
 Any number of observers (players who belong to no side) are allowed.
+
+When prompting for a side, a next default entry is offered. Certain names are skipped in this process; these are "Solo" and names beginning "Solitaire", "Solo:", "Moderator", "Referee" or name found in the <<GlobalTranslatableMessages.adoc#top, Global Translatable Messages>> property _VassalSkipSideList_. If used, the later should be defined as a comma-separated list of Sides.
+
```
BrentEaston, line 121: Why a Global Translatable message?

riverwanderer: It was my solution to the question of what happens if Translation(s) are added to the Definition of Sides component.
As the list of sides is translated, VassalSkipSideList will need to be translated also... or is there a better way ?

riverwanderer: > Why a Global Translatable message?

Now that Ive learnt a bit more about side translation in PlayerRoster, I want to take another go at this.

riverwanderer: To allow the PR to proceed, I will withdraw the extendable (VassalSkipSideList) element from this change and leave that to the wider design of side control features.

I'll also tackle the issue of side names that remain untranslated in a translated module.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -622,13 +622,36 @@ protected boolean allSidesAllocated() {
...
-           Resources.getString("PlayerRoster.referee").equals(side);
+    // FIXME: A future version of Vassal may provide a Side Type attribute in Definition of Sides that would allow this function to be rationalised.
+    // See also isRefereeSide()
+    return side.startsWith(Resources.getString("PlayerRoster.solitaire")) ||
+            side.equals(Resources.getString("PlayerRoster.solo")) ||
+            side.startsWith(Resources.getString("PlayerRoster.solo") + ":") ||
+            side.startsWith(Resources.getString("PlayerRoster.moderator")) ||
+            side.startsWith(Resources.getString("PlayerRoster.referee")) ||
+            side.startsWith(Resources.getString("PlayerRoster.auxiliarySidePrefix")) || // auxiliary side prefix
+            side.startsWith(SOLITAIRE) || // these checks needed in case a translated module has not localised these side names
```
Cattlesquat, line 638: Are these really needed? If a translated module hasn't localized the side names don't they come through in English anyway?

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -622,13 +622,36 @@ protected boolean allSidesAllocated() {
...
-    return Resources.getString("PlayerRoster.solitaire").equals(side) ||
-           Resources.getString("PlayerRoster.solo").equals(side) ||
-           Resources.getString("PlayerRoster.moderator").equals(side) ||
-           Resources.getString("PlayerRoster.referee").equals(side);
+    // FIXME: A future version of Vassal may provide a Side Type attribute in Definition of Sides that would allow this function to be rationalised.
+    // See also isRefereeSide()
+    return side.startsWith(Resources.getString("PlayerRoster.solitaire")) ||
+            side.equals(Resources.getString("PlayerRoster.solo")) ||
+            side.startsWith(Resources.getString("PlayerRoster.solo") + ":") ||
+            side.startsWith(Resources.getString("PlayerRoster.moderator")) ||
```
Cattlesquat, line 635: Why "startsWith" here where it used to only check equals? Are we expecting "Moderator Left" and "Moderator Right"? 

If so, why are we requiring a different format here from Solo (where it requires the colon e.g. "Solo:")? And for that matter why do we only eg allow "Solo:" and not "Solitaire:"

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -622,13 +622,36 @@ protected boolean allSidesAllocated() {
...
+            side.startsWith(REFEREE);
+  }
+
+  /**
+   * @param side Name of a side to see if it's a "referee side"
+   * @return True if the side is begins "Moderator" or "Referee"
+   * To test for non-player sides in general, use isSoloSide()
+   */
+  public static boolean isRefereeSide(String side) {
+    return side.startsWith(Resources.getString("PlayerRoster.moderator")) ||
```
Cattlesquat, line 651: Again I'm fuzzy on the change to startsWith and also the need to compare to the raw strings

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
```diff
@@ -1031,6 +1031,7 @@ PlayerRoster.solitaire=Solitaire
 PlayerRoster.solo=Solo
 PlayerRoster.moderator=Moderator
 PlayerRoster.referee=Referee
+PlayerRoster.auxiliarySidePrefix=~
```
Cattlesquat, line 1034: This appears not to be used?

Cattlesquat (changes requested):
See code notes -- I'm fuzzy on a few things

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
riverwanderer: The raw strings comparison arises from a concern that I have about this along the following lines...

1. Say a module has partial translation but not in the Definition of Sides component. This is not so outlandish, as translated PlayerSides can open up other complications if the module depends on PlayerSide, so I've found.
2. The translation resources for solitaire etc exist regardless.... therefore when Vassal isSoloSide() scans the sides array it will attempt to find the translated side name. This will fail if the solo side names don't match in translation, but the raw string will then catch it.

Have I got that right?

riverwanderer:
This PR doesn't create any v3.7 conflicts, can it be re-classified for the 3.7.1 milestone?

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
BrentEaston: I think the actual problem is in the ChessClock which calling isSoloSide with the translated Sides, it should be checking untranslated sides, but displaying translated sides,

The concern is that ChessClocks now have translated sides recorded in them?, they should be untranslated sides.

Cattlesquat: @BrentEaston So you think this usage is okay? Also are you okay with the various changes to "startsWith"?

Cattlesquat: @BrentEaston And we're comparing both Resources.getString("Some String") and the literal "Some String" in here. Surely only one is right? Or is one way being left in because ChessClock is broken? (are other things broken too?)

riverwanderer:
I am putting this PR on hold as #12732 & #12743 offer more generic and flexible solutions to the issue that this PR seeks to address.

Issue #12545, opened 2023-08-09
Allow a module to exclude one or more Sides


riverwanderer:
Multi-side modules may need to exclude 1 or more sides for some scenarios. Currently this can only be done by using up the unwanted sides each with a unique password. This method is tedious, error-prone, difficult to maintain and under-mines other (v3.7) Vassal features that rely on side-state (e.g. Scenario Options allows observer changes if no sides have been taken).

This request envisages a mechanism based on one or more special module-level Global Properties that the module may set or clear to control which unallocated sides are available to be selected in the side-switcher (PlayerRoster.java). The feature is intended to be used when a scenario is being set up. Possibly it might be integrated into the Scenario Options feature.

The number of Sides excluded in this way will also need to be accounted for as "unavailable" in any checks against the number of available sides, even though these sides _may_ not have a player allocated - e.g. the side-switcher seems to do this.

Suggested control:
 A module-level Global Property of the form `Vassal_Exclude_<Side>` (where `<Side>` is one of the Sides configured in the module)  is checked for each unused side. If true, the side is excluded from the side switch available sides.

PR #12652, opened 2023-09-05
12545 Exclude sides under module control

riverwanderer:
Attempt to address issue #12545 by allowing sides to be excluded under module control using properties of the form` VassalExcludeSide_<side>`

Function installed in the Wizard as well as the core side switcher, since the feature applies to game setup as well as in-game switching.

Scenarios may be be configured using Scenario Options or set in-game via a standard Global Property.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
@@ -787,12 +787,12 @@ public class PlayerRoster extends AbstractToolbarItem implements CommandEncoder,
       For scenario-specific exclusion, define VassalExcludeSide_ properties in Scenario Options
       For dynamic exclusion, define as module Global Properties
     */
-     for (int i = 0; i <= sides.size(); i++) {
-       final String s = sides.get(i);
-       if (!alreadyTaken.contains(s) && (boolean) GameModule.getGameModule().getProperty("VassalSideExcluded_" + s)) {
+    for (int i = 0; i <= sides.size(); i++) {
+      final String s = sides.get(i);
+      if (!alreadyTaken.contains(s) && (boolean) GameModule.getGameModule().getProperty("VassalSideExcluded_" + s)) {
         alreadyTaken.add(s); // exclude this side
-       }
-     }
+      }
+    }
 
     availableSides.removeAll(alreadyTaken);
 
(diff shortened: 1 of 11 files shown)


riverwanderer:
Need to add language support.

riverwanderer:
Translation now working; need to handle standard (ie no translation), as then translated side is null.

riverwanderer:
Translation now working.

Cattlesquat:
This looks okay to me.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -493,6 +494,21 @@ public Component getControls() {
...
+     Scan for excluded sides based on translation of property VassalExcludeSide_<Side>
+     add such to the alreadyTaken list
+     Note that <observer> can't be excluded as it is added later
+     Usage:
+      For scenario-specific exclusion, define VassalExcludeSide_ properties in Scenario Options
+      For dynamic exclusion, define as module Global Properties
+    */
+    for (int i = 0; i < sides.size(); i++) {
+      final String s = sides.get(i);
+      if (!alreadyTaken.contains(s) && StringUtils.equalsIgnoreCase((String) GameModule.getGameModule().getProperty("VassalExcludeSide_" + (untranslatedSides == null ? s : untranslatedSides[i])), "true")) {
```
uckelman, line 507: Don't call `GameModule.getGameModule()` each time through the loop. Get the `GameModule` once, before entering the loop.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -764,6 +780,21 @@ protected String promptForSide() {
...
+        alreadyTaken.add(s); // exclude side
+      }
```
uckelman, line 795: Rather than repeat this code, extract it to a function.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
riverwanderer: resolving in next push.

riverwanderer:
Cancelling this PR. I will re-do it from the ground up once v3.7.1 is released.

Issue #12645, opened 2023-09-04
Translated PlayerSide not matched during process to filter dropdown list


riverwanderer:
As standard, a Player's existing side is omitted from the dropdown list presented for "Retire / Switch Side".
However, if side names are translated, the existing side will not be omitted unless that translated side name is identical to the original (e.g. english and french "Orange").

I haven't tested the more general case of "in use" side names, but it seems likely that these won't be filtered either.

riverwanderer:
Tested the isSoloSide() function - this also fails to work with translated sides.

Failure demonstrated in the Liberte module by translating colours to French. A French run of the module will work correctly for "Orange" (same as english) but fails to find next side when the current PlayerSide is a different coloiur.

PR #12655, opened 2023-09-05
12645 Respect localisation when filtering sides

riverwanderer:
Possible fix for #12645 - suspect that Side is not localized before testing for removal from Available sides.


riverwanderer:
Still works in standard mode but doesn't seem to fix localization issue.

riverwanderer:
Fix is predicated on understanding that PlayerRoster.java _availableSides_  is a list of translated sides and that therefore adds to _alreadyTaken_ should also be translated. Therefore 3 instances of `alreadyTaken.add(p.Side)` are replaced with `alreadyTaken.add alreadyTaken.add(p.getLocalizedSide())`.

uckelman:
Is this ready to review?

riverwanderer:
> Is this ready to review?

yes

Cattlesquat:
Looks okay to me

Issue #12660, opened 2023-09-06
Side names not translated in change message 


riverwanderer:
Example of issue from language demo in Liberté (side names are colours translated to French):

<img width="307" alt="image" src="https://github.com/vassalengine/vassal/assets/58135975/e5bfb5f5-1ef1-4f82-a26e-6e32d75dd2a4">

riverwanderer:
PR #12661 fixes.

PR #12661, opened 2023-09-06
Translate side names in change sides message

riverwanderer:
Fix for #12660.


riverwanderer:
Tested OK in translated and non-translated situations.

BrentEaston (approved):
Looks good to me.

PR #12666, opened 2023-09-07
12678 - Truetype font support

BrentEaston:


BrentEaston:
This branch be broken, generating invalid builds.

PR #12667, opened 2023-09-07
12542 - Mitigate race condition when selecting player sides

riverwanderer:


riverwanderer:
Latest build achieves objective as a demo for the Retire/Switch Side button. includes translations for all but Polish (unresolved issue with Commit), Japanese and Chinese.

Further work needed to 
1. Validate text for side choice unavailable (too long?).
2. Test translations
3. Migrate this enhancement to wizard / initial connect side choices.
4. Investigate possible issue where an available side doesn't show after trapping a side-switch by another player (the side dropped by the other player?).

vassal-app/src/main/resources/VASSAL/i18n/VASSAL_da.properties
```diff
@@ -1023,6 +1023,7 @@ PlayerRoster.join_another_side=Skift til en anden side
 PlayerRoster.give_up_position=Vil du opgive din position som %1$s?
 PlayerRoster.join_game_as=Join spillet p� hvilken side?
 PlayerRoster.switch_sides=Din nuv�rende side er %1$s. Skift til hvilken side?
+PlayerRoster.switch_sides2=%1$s er nu ikke tilg�ngelig. Din nuv�rende side er %2$s. Skift til hvilken side?
```
BrentEaston, line 1026: Normally, unless you are very confident (native language speaker level) of the translation, don't worry about adding non-English translations. This will be (is supposed to be) done 'later' (!) by a native speaker.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -759,50 +759,59 @@ public List<String> getAvailableSides() {
...
-    final String mySide = getMySide(); // Get our own side, so we can find the "next" one
-    final int myidx = (mySide != null) ? sides.indexOf(mySide) : -1; // See if we have a current non-observe side.
-    int i = (myidx >= 0) ? ((myidx + 1) % sides.size()) : 0;   // If we do, start looking in the "next" slot, otherwise start at beginning.
-    for (int tries = 0; i != myidx && tries < sides.size(); i = (i + 1) % sides.size(), tries++) { // Wrap-around search of sides
-      final String s = sides.get(i);
-      if (!alreadyTaken.contains(s) && !isSoloSide(s)) {
-        found = true; // Found an available slot that's not our current one and not a "solo" slot.
+      if (!newSide.isEmpty() && !alreadyTaken.contains(newSide)) {
+        // found side is still available
+        // sides must always be stored internally in English.
```
BrentEaston, line 769: Could you just describe (maybe include in comments in the text) what you are doing here? I think what is happening is that if you choose a side, you then recheck to see if that side has been taken by someone else while you where sitting looking at the change side dialog. 

Have you confirmed that messages from other players are actually being processed (and not just queued up) while the change sides dialog is visible?

vassal-app/src/main/resources/VASSAL/i18n/VASSAL_ja.properties
```diff
@@ -522,6 +522,8 @@ PlayerRoster.become_observer=見学者になる
 PlayerRoster.join_another_side=別な陣営で参加
 PlayerRoster.give_up_position=%1$s の立場を放棄しますか？
 PlayerRoster.join_game_as=どの陣営で参加しますか？
+PlayerRoster.switch_sides=Your current side is %1$s. Switch to which side?
+PlayerRoster.switch_sides2=%1$s is now unavailable. Your current side is %2$s. Switch to which side?
```
BrentEaston, line 526: You definitely don't need to do this. The base language version will always be displayed as a fall-back if no language-specific version exists for that key.

BrentEaston (changes requested):

vassal-app/src/main/resources/VASSAL/i18n/VASSAL_da.properties
riverwanderer: Point taken. I felt I needed at least one translation for testing. If this thing goes live I plan to highlight the changes so hopefully at least the recent translators (Danish for example) will check the Google's work.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
riverwanderer: > Have you confirmed that messages from other players are actually being processed (and not just queued up) while the change sides dialog is visible?

If I understand the question right, then messages (changes?) from other players are processed. So for example;

"Fred" and "Jane" and two players in "<observer>" roles.
Fred attempts to switch side and is presented with the dialogue, offering "Player 1" and "Player 2".
Before Fred makes a choice, Jane also attempts to switch and is offered the same two available sides.
Fred selects "Player 2". Jane at this point still sees both sides available. (Nothing I know of can change that).

The check is done after the player selects a side. So if Jane also selects "Player 2", the re-iterated check detects this (test confirms), and the prompt is re-entered, this time without offering "Player 2".

If I misunderstood your question, please explain what you would like me to check.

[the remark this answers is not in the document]
riverwanderer: This comment and the code beneath it were simply relocated. I guess it depends what "stored internally" means.

The `alreadyTaken` and `availableSides` lists are in native language but what is returned and subsequently shared with other players is in english.

I'll clarify that comment accordingly.

riverwanderer:
The last PR attempts to resolve outstanding review points and includes "missing side" fix (initialising alreadyTaken list before repopulating it).

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
BrentEaston: While Fred is sitting in the Select side dialog and Jane selects Player 1 out from under him and sends a message to Fred saying 'I have selected player 1'. Are you sure that Fred's Vassal client actually processes that message while he is sitting there looking at the select side dialog? When he selects Player 1 also, are you sure that his client will already have been updated with Jane's message? Or will Jane's message only be processed AFTER Fred sends his 'I have select player 1 message'. It's a question of multi-threading and whether anything actually happens in Fred's client while the dialog is open.  

You may have tested this already and know it works?

BrentEaston: Don't worry, I tried it out. Works well. You see the 'Changed sides' message in the background and the get the 'Unavailable' option when I select the missing side.

riverwanderer: Jane's message appears on Fred's chat log whilst Fred is still in the dialogue.

In this example, MTB is the other player who switched to Player 2 whilst the current player was in the dialogue. Then current player tries switching to Player 2 and is denied by the new check.

image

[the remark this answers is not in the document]
BrentEaston: Not following. If you have a module developed in German and translated into French, how do the sides end up being shared to other players in English?

BrentEaston: Yes, I tried it myself. Vastly improved. Should narrow the race window to the current latency between the two clients in question. The best we can do with out adding a more complex 2 level procedure.

riverwanderer:
There is still a problem in the following situation (re-initialising `alreadyTaken` has not solved it):
1. Fred connects to Player 1 whilst Jane is also in the switch dialog.
2. Jane tries to connect to Player 1 and is rejected.
3. Fred disconnects from Player 1 and connects to Player 2 instead.
4. Jane tried to connect to Player 2 and is rejected. However... Player 1 does not appear on her refreshed dailog.

This is an issue within the While... loop because if Jane cancels and retries, Player 1 will be offered as expected.

riverwanderer:
Latest build fixes the re-initi

> There is still a problem in the following situation (re-initialising `alreadyTaken` has not solved it):
> 
> 1. Fred connects to Player 1 whilst Jane is also in the switch dialog.
> 2. Jane tries to connect to Player 1 and is rejected.
> 3. Fred disconnects from Player 1 and connects to Player 2 instead.
> 4. Jane tried to connect to Player 2 and is rejected. However... Player 1 does not appear on her refreshed dailog.
> 
> This is an issue within the While... loop because if Jane cancels and retries, Player 1 will be offered as expected.

Latest build fixes this issue.

This PR should now be effective for switch/retire button. Further work required to integrate with the Wizard.

riverwanderer:
@BrentEaston - Back to the drawing board. I may need some help / guidance as to how I can implement the side check within the Wizard.

BrentEaston:
My suggestion would be that when we get into PlayerRoster.finish() and we find the selected side already taken, that we just launch a whole new promptForSide() with the `side no longer available' message and do this outside of the Wizard context

riverwanderer:
Wizard is now supported. As suggested by @BrentEaston, it diverts to the standard switcher if it finds the selected side already taken. Once a different side is chosen, the standard side switcher closes and then the Wizards also closes. Passes my basic testing. Let me know if you think of any improvements I can make.

vassal-app/src/main/resources/VASSAL/i18n/VASSAL_es.properties
```diff
@@ -704,6 +704,7 @@ PlayerRoster.join_another_side=Unirse a otro bando
 PlayerRoster.give_up_position=¿Dejar su lugar como %1$s?
 PlayerRoster.join_game_as=Unirse a la partida como... 
 PlayerRoster.switch_sides=Tu bando actual es %1$s. ¿Cambiar de bando?
+PlayerRoster.switch_sides2=%1$s ahora no est� disponible. Tu bando actual es %2$s. ¿Cambiar de bando?
```
uckelman, line 707: This looks like mojibake. The encoding of (probably) á seems bad.

vassal-app/src/main/resources/VASSAL/i18n/VASSAL_fr.properties
```diff
@@ -782,6 +782,7 @@ PlayerRoster.join_another_side=Changer de camp
 PlayerRoster.give_up_position=Abandonner sa position comme %1$s
 PlayerRoster.join_game_as=Rejoindre la partie en prenant quel camp ?
 PlayerRoster.switch_sides=Votre camp actuel est %1$s. Basculer vers quel camp?
+PlayerRoster.switch_sides2=%1$s n?est pas disponible. Votre camp actuel est %2$s. Basculer vers quel camp?
```
uckelman, line 785: Bad encoding here, too.

vassal-app/src/main/resources/VASSAL/i18n/VASSAL_it.properties
```diff
@@ -601,6 +601,8 @@ PlayerRoster.become_observer=Diventa osservatore
 PlayerRoster.join_another_side=Cambia fazione
 PlayerRoster.give_up_position=Abbandoni la fazione di %1$s?
 PlayerRoster.join_game_as=Scelta fazione  
+PlayerRoster.switch_sides=Il tuo lato attuale � %1$s. Passare da che parte?
+PlayerRoster.switch_sides2=%1$s ora non � disponibile. Il tuo lato attuale � %2$s. Passare da che parte?
```
uckelman, line 605: Bad encoding.

vassal-app/src/test/resources/test-images/cc.png
uckelman, line 1: Why's this image being removed? It's used in the tests for `GeneralFilter`.

riverwanderer: My error. Somehow it appeared in a commit and my attempt to roll it back must have gone badly.

vassal-app/src/main/resources/VASSAL/i18n/VASSAL_it.properties
```diff
@@ -601,8 +601,8 @@ PlayerRoster.become_observer=Diventa osservatore
 PlayerRoster.join_another_side=Cambia fazione
 PlayerRoster.give_up_position=Abbandoni la fazione di %1$s?
 PlayerRoster.join_game_as=Scelta fazione  
-PlayerRoster.switch_sides=Il tuo lato attuale � %1$s. Passare da che parte?
-PlayerRoster.switch_sides2=%1$s ora non � disponibile. Il tuo lato attuale � %2$s. Passare da che parte?
+PlayerRoster.switch_sides=Il tuo lato attuale e' %1$s. Passare da che parte?
+PlayerRoster.switch_sides2=%1$s ora non e' disponibile. Il tuo lato attuale e' %2$s. Passare da che parte?
```
uckelman, line 605: That doesn't look right either? Is it supposed to be `é`?

IIRC, your options here are UTF-8 or `\u` escapes.

riverwanderer: Thanks. I will review again using UTF-8 and escapes. I noticed that the polish translation was using escapes. It seemed to me that most of the others were not (using UTF-8 perhaps?).

Also, some of the translation files now contain extra indecipherable (to me) characters. I'd noticed these before in the jp / zh translations but not in es or fr. For example, the middle line here:

image

vassal-app/src/test/resources/test-images/cc.png
riverwanderer: Fixed?

BrentEaston (approved):
Looks good to me. I tried out the new Wizard fixes and the worked exactly as expected.

BrentEaston:
Just need to get those conflicts sorted and this is good to go.

riverwanderer:
Retested OK after resolving conflicts (tests in English and French).

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -469,19 +467,23 @@ public void finish() {
...
-        final Command c = new Chatter.DisplayText(GameModule.getGameModule().getChatter(), Resources.getString(GlobalOptions.getInstance().chatterHTMLSupport() ? "PlayerRoster.joined_side_2" : "PlayerRoster.joined_side", GameModule.getGameModule().getPrefs().getValue(GameModule.REAL_NAME), newSide));
+        final Command c = new Chatter.DisplayText(GameModule.getGameModule().getChatter(), Resources.getString(GlobalOptions.getInstance().chatterHTMLSupport() ? "PlayerRoster.joined_side_2" : "PlayerRoster.joined_side", GameModule.getGameModule().getPrefs().getValue(GameModule.REAL_NAME), translateSide(newSide)));
         c.execute();
       }
 
       final Add a = new Add(this, GameModule.getActiveUserId(), GlobalOptions.getInstance().getPlayerId(), newSide);
       a.execute();
       GameModule.getGameModule().getServer().sendToOthers(a);
+
+      getLaunchButton().setVisible(getMySide() != null);
```
uckelman, line 484: Is this the Retire button? Should this be calling `setEnable` now?

riverwanderer: Yes, I need to adjust this PR.  Also, I am considering whether setVisibility should still have a role for when no sides are defined.

riverwanderer:
Tests out ok.

Additional fixes:
1. Resolved an issue where connection to `<observer>` was initially rejects (side taken), arising from using the common function PromptForSide() to validate initial connection.
2. Rather than allow a connection into a game as a null side in the finish() routine, the side is changed to `<observer>`. Possibly this should instead just not connect to the game, as it results from the user cancelling the input panel.

I have done nothing to prevent the Retire button displaying if no sides are defined. This can be achieved by not populating the side button text/image.

PR #12671, opened 2023-09-08
12670 - Prevent essential components being accidentally deleted

BrentEaston:


uckelman:
I don't see anything here that would be backwards incompatible. Could this be in 3.7?

BrentEaston:
I'm not actually very happy with this, I am going to redo it, back to the Buildable level and implement a set of new default methods isMandatory() = false, isMovable() = true and isUnique() = false so all Buildables can register themselves as  to how they want to be treated by the ConfigureTree.

And yes. it will be 3.7 compatible.

uckelman:
Is this the way you want it to be now?

BrentEaston:
Yes, this is good to go. More widespread changes, but it is actually much simpler and supports custom components.

vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -1551,7 +1563,33 @@ public DefaultMutableTreeNode getTreeNode(Configurable target) {
...
+  public boolean isMoveAllowed(Configurable target) {
+    return target.isMovable();
+  }
+
+  /**
+   * Is the Component allowed to be deleted?
+   * A mandatory component can be deleted if there is more than one of them with the same parent
+   * */
+  protected boolean isDeleteAllowed(Configurable target) {
+    if (target instanceof AbstractBuildable) {
```
uckelman, line 1576: Could you simplify this by checking `target.isMandatory()` right at the start and returning `true` if not?

uckelman: I think this?
```
if (!target.isMandatory()) {
  return true;
}

if (target instanceof AbstractBuildable) {
  final Buildable parent = ((AbstractBuildable) target).getAncestor();
  if (parent instanceof AbstractBuildable) {
    final int count = ((AbstractBuildable) parent).getComponentsOf(target.getClass()).size();
    return count > 1;
  }
}

return false; 
```
Maybe that's not really simpler?

BrentEaston: Slightly more efficient though :) Change made.

PR #12689, opened 2023-09-13
12688-Replace with other should not update text of unchangeable labels

BrentEaston:
If `match current state` is specified, then non-changeable Text Labels should be excluded from the state match process.


BrentEaston:
Original reporter confirms change resolves problem

PR #12694, opened 2023-09-14
12260-Pre & Post Refresh Hotkeys

riverwanderer:
Test hotkey VassalRefreshGHK.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
@@ -341,6 +341,10 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
       }
     }
 
+    // About to commence refreshing the game, allow a custom start...
+    GameModule.getGameModule().fireKeyStroke(NamedKeyStroke.of("VassalPreRefreshGHK"));
+
+
     /*
      * 2. Build a list in visual order of all stacks, decks, mats and other pieces that need refreshing
      */
@@ -541,7 +545,7 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
       }
     }
     // After all refreshing, allow a custom finish...
-    GameModule.getGameModule().fireKeyStroke(NamedKeyStroke.of("VassalRefreshGHK"));
+    GameModule.getGameModule().fireKeyStroke(NamedKeyStroke.of("VassalPostRefreshGHK"));
   }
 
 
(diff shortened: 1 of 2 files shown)


riverwanderer:
I have tested this PR with the dev version of C&C Nap as follows:

1. Add an Action Button to the module that responds to hotkey VassalRefreshGHK, simply putting up a message.
2. Add a GKC button also responding to hotkey VassalRefreshGHK that sends a conversion command to a text label.

Simple Refresh Counters and Refresh Predefined Setups trigger both 1 & 2. 

This PR adds another tool for the module maintainer. As per the linked request, it provides a hotkey hook that can be used to automatically convert multiple scenario files in ways that a refresh alone cannot do. The concept is that the mod dev will use this feature to perform such work "once off" and then disable or remove the maintenance functions.

One caveat, if I understand Refresh correctly, any refresh done using "Refresh Predefined Setups" by design is done without triggering SGKCs. As long as the mod dev bears that in mind, VassalRefreshGHK could be useful to them too.

I believe that the WiF developer may have been asking for this type of functionality recently, so I intend to offer him the build for testing & feedback.

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -536,8 +537,11 @@ else if (!decksToDelete.isEmpty()) {
         log(Resources.getString("GameRefresher.refreshable_decks", refreshable));
         log(Resources.getString(options.contains("DeleteOldDecks") ? "GameRefresher.deletable_decks" : "GameRefresher.deletable_decks_2", deletable)); //NON-NLS
         log(Resources.getString(options.contains("AddNewDecks") ? "GameRefresher.addable_decks" : "GameRefresher.addable_decks_2", addable)); //NON-NLS
+
       }
     }
+    // After all refreshing, allow a custom finish...
+    GameModule.getGameModule().fireKeyStroke(NamedKeyStroke.of("VassalRefreshGHK"));
   }
```
BrentEaston, line 545: A simple, useful change. I have no problem with it being a pre-defined keystroke, nothing would be added by making it configurable.

BrentEaston (commented):
Just need some doco

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
riverwanderer: Refresher Doc updated.

PR #12701, opened 2023-09-15
New font branch

### LINKED ISSUE #12678: Add Truetype/Opentype font support to Vassal
1. Add new Vassal Logical fonts VassalSans, VassalSerif, VassalMono, VassalCondensed, VassalEditor. These fonts will be re-badged versions of suitable open-licenced freeware fonts. (currently looking at Roboto variations for Sans, Mono and Condensed, Possibly Merriweather for Serif, JetNrains Mono for Editor).

2.  Allow Designers to add TrueType/OpenType Fonts to a module (New top level Fonts component)

 All new fonts (Vassal supplied and designer supplied) can be used in the following contexts:
    o Text Label selected font
    o Game Piece Image font
    o Chat Window font
    o Any Swing rendered HTML
    o Change GUI default font.

The existing Java logical fonts Dialog/Serif/SansSerif/Monospaced will not be touched. This will be complete opt-in.

Fonts added to a module will be used on all target machines the module is run on UNLESS the user already has that font installed on their system, in which case the installed font will be used.

### LINKED ISSUE #12702: Chat preferences font size field does not show current value (too narrow?)
image

On MacOS, using v3.8 "fonts" test build.

Also, the example text window (which accepts input) does not reset when preferences is closed, but it does clear between restarts of the module. Maybe as designed ?

BrentEaston:


BrentEaston:
Comments from Discord:
1. When I load some fonts, there's a clue to the vassal_plurals issue... intended text is displayed when there's only 1 font... [also notice all the fonts I loaded have "Registration failed", but as below, they seem to work in the text label!] 
image

2. The module fonts appear in the text label dropdown. Maybe logically, they should appear above Vassal fonts... module...vassal...java.

3, Change 'Registration failed' to 'Already installed'

4. Report Duplicate ModuleFont entries via validator.

5. Make all base Vassal fonts 'Regular'

delatbabel:
I would like to have this merged. The rationale is that for modules that have dynamic values shown on counters (e.g. the Europa modules, specifically the v0.9.0 and later New Maps module), the layout of each counter depends on every player having exactly the same mapping between Java native fonts (e.g. "Dialog" or "Sans") to the same font as the module designer has that font mapped to. If one player has "Sans" mapped to "Helvetica" and another player has "Sans" mapped to "Arial" then the counters will look different for each player, and some of the counter values may overlap, affecting readability.

Nearly all systems support TrueType fonts these days so having a TrueType font mapping available to the module designer will greatly assist the functionality of those modules.

PR #12707, opened 2023-09-15
12706-keep the "retire" button disabled unless a game is running.

riverwanderer:


riverwanderer:
Tested OK as followsL:

Open module [retire button disabled]
open scenario - after first connect, button is enabled.
close scenario - button is disabled.
open scenario - button becomes available again.

PR #12723, opened 2023-09-19
12705 Prevent creation of duplicate Translations by Add Translation button

BrentEaston:


riverwanderer:
Works.  Spelling error in error message: "allready" instead of "already".

riverwanderer:
While you are there... red text is illegible on blue background. What does it signify? Can it be white on blue or something else readable? 

Affects the top item and the currently selected attribute.

BrentEaston:
The latest commit for this makes an interim change to the selection background color for Macs in the Translation Window screen. For 3.8, I will extend this to cover all of the UI as part of my changes to allow the default UI font and size to be modified dynamically.

PR #12727, opened 2023-09-19
Ensure that tiles are cobbled for fixed-size board scaling

uckelman:


BrentEaston:
That resolves the issue for me. Memory load and initial tile load performance back to 3.6.19 levels.

PR #12735, opened 2023-09-21
12734 investigating failure of getReferenceManualPage() to position at anchor points

riverwanderer:
Testing only.


BrentEaston:
I tried looking into this once before and am pretty sure I found it just wan't supported.

riverwanderer:
> I tried looking into this once before and am pretty sure I found it just wan't supported.

That's what I found too. Certainly beyond me.  I will close this off.

PR #12736, opened 2023-09-21
12545 side change features (full suite)

riverwanderer:
Includes the features of PR #12732 and in addition:

Provides * Choose a random side * feature in the sides list dropdown. 

When retire/switch sides button or hotkey is activated:
- Global Option Property _VassalForceSide_  is checked and PromptForSide() will switch to that side directly, if valid.
- Global Option Property  _VassalRandomSide_  is checked; when true, random side option is automatically processed.

To prevent persistence, both Global Option properties are cleared both on first connection and after each use.
These features fail-over to the standard sides list dialog.

Use cases: Connection to a randomly selected side; Ability to choose a side directly within a module.


riverwanderer:
This PR is starting to shape up. So far it provides mechanisms for comprehensive control over side switches, detailed in the Definition of Sides documentation. In particular, the UI now offers random side selection as an option.

To do: Integrate random side choice into the Player Roster initial connection routine.

I am exercising the new automated features in the[ v4.3.12 release](https://www.dropbox.com/scl/fi/wpz2mtthxc3piouiiw212/CCAncientsV4-3-12.vmod?rlkey=mz4ixwflijrcw8nobcej34rxa&dl=0) of the C&C Ancients module, as follows:
- Epic sides ("FG" roles) are hidden until an Epic scenario is opened.
- The side choice button (P1/P2 icon) will set VassalRandomSide to automatically select a random side (TO DO)
-  Non-Solitaire roles are hidden once Solitaire is selected
- Solitaire role is hidden once a game is joined by more than 1 player (WIP)

PR #12743, opened 2023-09-24
12545+side controls ultimate edition

### LINKED ISSUE #12741: Random side choice from choose a side dialog
Provide an option in the Choose a side dialog that, when selected, with choose a random side from amongst the available player sides (ie excluding observer and solo or moderator sides).

Include a mechanism to directly access this feature from a module, bypassing the initial dialog.

Failover to the standard dialog.

Do not display the random sides option unless at least one player side is available for selection.

### LINKED ISSUE #12742: Direct connection to a side from a module
Provide a feature that will allow a module to directly switch the player to an available side.

Use cases might include, directly switching to Solitaire to start a solo mode game from a choice of solo modes or setting up a game with players taking roles by clicking on action buttons.

Consider failover to standard dialog vs return with an error message and no side change.

riverwanderer:
Resetting PR #12736. Includes the features of PR #12732 and in addition:

Provides * Choose a random side * feature in the sides list dropdown - if at least two true player sides are available.

When retire/switch sides button or hotkey is activated:
Module Global Property VassalForceSide is checked and promptForSide() will switch to that side directly, if valid.
Module Global Property VassalRandomSide is checked; when true, random side option is automatically processed.
In each case, the special property is cleared immediately after use.
If the request cannot be satisfied, control returns to the user and a warning is logged to chat. No side change occurs.

Use cases: Connection to a randomly selected side; Ability to choose a side directly within a module.

Additional changes: 
Observer role is not displayed if player is already in the role (consistent with behaviour of other sides)
Error message if no sides are available when player is switching from observer role.


riverwanderer:
This PR is starting to shape up. It now provides mechanisms for comprehensive control over side switches, detailed in the Definition of Sides documentation. In particular, the UI now offers random side selection as an option - both for initial connection and side switch from the retire/switch button.

To do - consider alternative failure approach for auto-switching features; returning side unchanged rather than entering the side switch dialog.

I am exercising the new automated features in the[ v4.3.12 release](https://www.dropbox.com/scl/fi/wpz2mtthxc3piouiiw212/CCAncientsV4-3-12.vmod?rlkey=mz4ixwflijrcw8nobcej34rxa&dl=0) of the C&C Ancients module, as follows:

Epic sides ("FG" roles) are hidden until an Epic scenario is opened.
The side choice button (P1/P2 icon) will set VassalRandomSide to automatically select a random side (TO DO)
Non-Solitaire roles are hidden once Solitaire is selected
Solitaire role is hidden once a game is joined by more than 1 player (WIP)

riverwanderer:
This PR is ready for review. One unresolved issue is that I changed getChangeCommand from Protected to Public, as I couldn't get access to this routine working another way (lack of knowledge). I could achieve these parts of the PR using Global Option preferences instead.

uckelman:
If this is still wanted for 3.8, then the merge conflicts need to be resolved.

riverwanderer:
I would like to include it, subject to any adaptations from feedback. 

I will need to schedule time to work through the conflicts.

PR #12753, opened 2023-09-27
Last file name property

riverwanderer:
Creates a property _VassalGameFileName_ (akin to existing read-only properties such as ModuleVersion).
It is set to the file name of the most recent VSAV/VLOG.


riverwanderer:
As well as introducing VassalGameFileName, both bugs reported in #12755 are fixed by this PR.

PR #12756, opened 2023-09-28
Minor code doc correction

riverwanderer:
File extension  = vsav.


vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -496,6 +496,7 @@ public void finish() {
       if (GameModule.getGameModule().isMultiplayerConnected()) {
         final Command c = new Chatter.DisplayText(GameModule.getGameModule().getChatter(), Resources.getString(GlobalOptions.getInstance().chatterHTMLSupport() ? "PlayerRoster.joined_side_2" : "PlayerRoster.joined_side", GameModule.getGameModule().getPrefs().getValue(GameModule.REAL_NAME), translateSide(newSide)));
         c.execute();
+        GameModule.getGameModule().getServer().sendToOthers(c);
```
uckelman, line 499: Is this change intended to be included?

riverwanderer: Not really. That change is in PR #12750.

uckelman: It looks like the only commit which should be in this PR is 635bc268, in that case.

Would you be able to drop the other commits? (The simplest way is probably an interactive rebase.)

uckelman: That's not what I meant---the unnecessary commits are still there, plus one more which undoes them. What I was asking for was a rebase that _takes those commits out of the PR_.

PR #12757, opened 2023-09-28
12260 refresh pre & post

riverwanderer:
Re-submission of PR #12694. 

Special hotkeys VassalGHKpreRefresh & VassalGHKpostRefresh provide maintenance hooks for module maintainers..

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
@@ -341,7 +341,7 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
       }
     }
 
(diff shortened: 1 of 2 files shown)


BrentEaston:
I think this one needs to be 3.8 as well. Again, it is new functionality that can be added to a module that will not work in all versions of 3.7,

riverwanderer:
> I think this one needs to be 3.8 as well. Again, it is new functionality that can be added to a module that will not work in all versions of 3.7,

This PR is not really applicable to game players, if that makes any difference. It's only useful as an adjunct to predefined refreshes and those are run by module developers, as part of the process of releasing an update to a module.

riverwanderer:
Superceded by #12890 which includes a control via the refresh control panel.

PR #12760, opened 2023-09-28
Pds refresh progress bar

riverwanderer:
Exploration... attempting to mash PDS refresher into progress bar from imagesaver


riverwanderer:
Superceded by a simpler method. See #12896.

PR #12763, opened 2023-09-29
Added VassalGameFileName property

riverwanderer:
Creates a property VassalGameFileName (akin to existing read-only properties such as ModuleVersion).
It is set to the file name of the most recent VSAV/VLOG.

Also fixes two issues with managing window title names from the Wizard which would have impeded the PR.


BrentEaston:
This really needs to be a 3.8.0 enhancement. We can't have functionality that a module may use that is present in some versions of 3.7, but not others.

riverwanderer:
> This really needs to be a 3.8.0 enhancement. We can't have functionality that a module may use that is present in some versions of 3.7, but not others.

Thanks for the feedback. Shall I split out the bug fix element from the new functionality or just leave it all until v3.8.0 ?

BrentEaston:
Split out the additional functionality. Good to get any bug fixes in asap, even small.

PR #12768, opened 2023-09-30
"Player joins as side" message now broadcast to all players

riverwanderer:


riverwanderer:
Tested  - as expected message is output to other players but not for the first player who joins a room.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -496,6 +496,8 @@ public void finish() {
       if (GameModule.getGameModule().isMultiplayerConnected()) {
         final Command c = new Chatter.DisplayText(GameModule.getGameModule().getChatter(), Resources.getString(GlobalOptions.getInstance().chatterHTMLSupport() ? "PlayerRoster.joined_side_2" : "PlayerRoster.joined_side", GameModule.getGameModule().getPrefs().getValue(GameModule.REAL_NAME), translateSide(newSide)));
         c.execute();
+        GameModule.getGameModule().getServer().sendToOthers(c);
+
```
BrentEaston, line 500: Mark, the preferred way to do this is:

GameModule.getGameModule().sendAndLLog(c);

Which handles writing the Command into a log if it is currently open, and handles pausing of logging. You should generally never call getServer().sendToOthers() directly.

BrentEaston (changes requested):

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
BrentEaston: Hi @riverwanderer ,

I would say that these sort of messages should go into the log if they are being sent to other players, so sendAndLog() is the preferred way to go for the existing messages as well.

riverwanderer: Is there a distinction that needs to be made between the messaging and the actual side change commands? Eg at the moment change of side is appended with messaging. I suspect that sendToOthers has been used to prevent undo.

riverwanderer: Changes made and given a basic test (2 different users exchanging a log file). Messages appear when expecrted.

vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
```diff
@@ -493,14 +495,18 @@ public void finish() {
 
     // null is a cancel op - player will not connect to the game
     if (newSide != null) {
-      if (GameModule.getGameModule().isMultiplayerConnected()) {
-        final Command c = new Chatter.DisplayText(GameModule.getGameModule().getChatter(), Resources.getString(GlobalOptions.getInstance().chatterHTMLSupport() ? "PlayerRoster.joined_side_2" : "PlayerRoster.joined_side", GameModule.getGameModule().getPrefs().getValue(GameModule.REAL_NAME), translateSide(newSide)));
+      if (gm.isMultiplayerConnected()) {
+        final Command c = new Chatter.DisplayText(gm.getChatter(), Resources.getString(GlobalOptions.getInstance().chatterHTMLSupport() ? "PlayerRoster.joined_side_2" : "PlayerRoster.joined_side", gm.getPrefs().getValue(GameModule.REAL_NAME), translateSide(newSide)));
         c.execute();
+       // trying with logging in place of gm.getServer().sendToOthers(c);
```
BrentEaston, line 501: Looking good, but can you clean up these interim comments? 3 different places

riverwanderer: Done.

BrentEaston (approved):

PR #12775, opened 2023-10-02
More efficient Regex search with improved user experience.

riverwanderer:
Try a re-vamp of Regex search.


riverwanderer:
I've noticed the my attempt to detect a lack of Regex and apply a default doesn't seem that reliable. I am looking into it.

Cattlesquat:
Sounds like still a "Work in Progress"? pending regex detection?

riverwanderer:
> Sounds like still a "Work in Progress"? pending regex detection?

Yes, I will be submitting a branch from this PR that offers normal, word start and full regex as radio button options along with some other enhancements.

PR #12791, opened 2023-10-06
12784 - Fix rendering of external Text Label images

### LINKED ISSUE #12784: Labels are not rendering HTML that reference external images correctly.
Using VASL 6.6.6.

The following renders correctly in VASSAL 3.6.14:
```
﻿<html> 
<head> <style> body { color: red ; } </style> </head>
xxx 
<img src="https://pacman-ghost.com/images/pacman-ghost.png"> 
yyy
```
vassal-3 6 14

But incorrectly in VASSAL 3.7.0 and 3.7.1:
vassal-3 7 0+1

Remove the `` tag and it renders correctly:
no-img

The `body { color: red ; }` text is CSS, so I suspect the `<head>` and/or `<style>` tags are being removed - broken HTML santitization, triggered by the `` tag...?

Cattlesquat:
Fixes #12784

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Labeler.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Labeler.java
@@ -618,7 +618,8 @@ public class Labeler extends Decorator implements TranslatablePiece, Loopable {
 
       // paint the foreground
       if (fg != null) {
-        if (txt.contains("<img")) { //NON-NLS
+        // Support images pre-existing in the module
+        if (txt.contains("<img") && !txt.contains("http://") && !txt.contains("https://") && !txt.contains("file://")) { //NON-NLS
           final DataArchiveTextPane p = new DataArchiveTextPane(txt, "label", fg, font); //NON-NLS
           p.paint(g);
         }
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/counters/Labeler.java
```diff
@@ -618,7 +618,8 @@ public BufferedImage eval() throws Exception {
 
       // paint the foreground
       if (fg != null) {
-        if (txt.contains("<img")) { //NON-NLS
+        // Support images pre-existing in the module
+        if (txt.contains("<img") && !txt.contains("http://") && !txt.contains("https://") && !txt.contains("file://")) { //NON-NLS
```
uckelman, line 622: I think you might want to match `"src=\"http://"`, etc. just to be safe.

uckelman: On reflection, this would be best as a regex, to save checking the string multiple times.

uckelman: I think `]+ )? src="(https?|file)://` is the regex we want.

Cattlesquat: done

uckelman:
That last change doesn't work and I'm not sure why not.

PR #12795, opened 2023-10-09
Editor search regex revamp radio buttons

riverwanderer:

PR #12798, opened 2023-10-11
Editor search revamp: more options and enhanced navigation

riverwanderer:
Supersedes PR #12795 and related PRs.

Summary of new features:
- Word and full Regex search options
- Ability to back-track
- Output includes index number of listed items, position (number) of traits within their container and highlighting of matched text.
- PgUp / PgDn hotkeys to support back-tracking, in addition to the existing Enter / Esc ones.
- Image name and GPID are appended to report for Basic Piece traits hits 
- Fixes a couple of issues.


riverwanderer:
PR is ready for independent testing & review.

Known error - Prev shortcut (PgUp) is ignored after some period of use. Possibly associated with using the "Next" function a few times.

Outstanding - Ideally, the Next button needs PgDn as a shortcut. I am not sure if "PrevAction" was the correct approach to get PgUp working but if so, to apply the same to the Next (i.e. "find") actionlistener requires resolving a scope issue.

There is also issue #12800 & #12801 which I will investigate further. So far as I can tell both arise under earlier versions of Vassal.

riverwanderer:
PgUp / PgDn keys should now be working. Just the couple of possibly old issues mentioned above to investigate.

riverwanderer:
Just working to fix/ 
finalise the search initialisation, so that the navigation index is set correctly when search is started mid-module. Then I will try and get some external testing via the forum.

riverwanderer:
I will do a preview release of this shortly. Any feedback prior to that appreciated.

vassal-doc/src/main/readme-referencemanual/ReferenceManual/Search.adoc
```diff
@@ -18,32 +24,33 @@ The default search is to search for everything, but the Advanced Searh options c
...
 ===== Exact Case
 Force an Exact case search (i.e. abc is different to ABC).
 
-===== Regular Expression
-The specified search string is a https://en.wikipedia.org/wiki/Regular_expression[Regular Expression]. In the simplest example, a search for ABC will find matches that equate wholly to "ABC", and lower or mixed case variants (unless _Exact Case_ is checked as well).
-
-===== Advanced Search
-Show the advanced search options. All search options are on in a simple search. The advanced options allow you to turn off options to narrow your search.
+===== Advanced Filters
+Show the advanced filter options. All search filters are on in a simple search. The advanced options allow you to de-select various categories to narrow your search.
```
Cattlesquat, line 35: I'm not sure "All search filters are on in a simple search" is actually true? I think simple search maybe only searches for names/types of components and doesn't e.g. search into traits and stuff? I can't remember exactly but that's my impression.

riverwanderer: > I'm not sure "All search filters are on in a simple search" is actually true?
This is what the doc always said but the help button didn't work until the last vassal release, so perhaps that is moot. 

> I think simple search maybe only searches for names/types of components...
Yes, that is right. However, isn't it natural for a user to assume that search finds all it can by default? This latest PR does that in its current state.  If you think that names & types are the natural filters, then how about I make them the default for "advanced filters" at a restart of the edit session ?

Cattlesquat: We definitely don't want the advanced filters defaulting back to fewer options at an edit session restart -- advanced users are likely to usually want them to stay checked, doubly so since that's been their default. (There's an argument for making the advanced filters remember their settings from session to session within the same module -- but if we don't already have that it would obviously require some work)

Changing the way a simple search operates would have the following challenges:
* Novice users potentially confused by searches into "inner bits" rather than the basic names (I don't have any data on this because we don't collect any; it's just my intuition to give beginners a simpler experience and let them search for things they actually see directly in the edit "tree")
* The code is presently written to ignore the checkboxes if we're not in advanced mode (because obviously we don't want to use checkbox settings that the user can't see and may not even know about), so that code would have to be changed to change the behavior of simple search. 

My preference in this case is definitely to make the documentation match the present behavior, rather than vice versa.

riverwanderer: > There's an argument for making the advanced filters remember their settings from session to session within the same module -- but if we don't already have that it would obviously require some work

That's what's there; yes, makes sense. The one improvement might be if it could be a global Vassal option.

> * Novice users potentially confused by searches into "inner bits" rather than the basic names (I don't have any data on this because we don't collect any; it's just my intuition to give beginners a simpler experience and let them search for things they actually see directly in the edit "tree")

Novice users will still be interested in "inner bits" like properties and message formats though, surely? There's no telling whether a novice user will want to search on a Name/Class Name or on something they vaguely remember putting in an expression somewhere. As one such user, I can say I've ended up using Advanced as my default option without knowing why until just recently.  The default search position of most apps I have come across is to search for everything and offer the user the ability to filter.

> * The code is presently written to ignore the checkboxes if we're not in advanced mode (because obviously we don't want to use checkbox settings that the user can't see and may not even know about), so that code would have to be changed to change the behavior of simple search.

I already prep'd that change. It meant finding all tests on the "advanced" settings and allowing them if the "Advanced" setting is off.

I would like to get user feedback on the potential change and see if there is a strong preference on any aspect like this one.

riverwanderer: On second thoughts, let's go with your original design, to which I will add an option for full search.

Cattlesquat: Okay let us know when ready for next review

riverwanderer: Ready for review. See also PR#12835.

riverwanderer:
I have been using this PR routinely (as part of a test merge PR #12889), with no issues. It makes use of the trait line-numbering released in v3.7.5 and offers several other benefits deliverable in v3.7.6, outlined in the opening comment.

vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -2638,62 +3092,161 @@ else if (c instanceof PrototypeDefinition) {
...
+          logger.error(getString("Editor.search_badWord", noHTML(e.getMessage()))); //NON-NLS
+          return null;
         }
       }
+
+      // search string is full regex
+      try {
+        return Pattern.compile(searchString, flags);
+      }
+      catch (java.util.regex.PatternSyntaxException e) {
```
uckelman, line 3246: Import `java.util.regex.PatternSyntaxException`.

vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -2513,119 +2880,206 @@ void startNewTrait() {
...
+            final String startPage = c.getAttributeValueString("startingPage");
+            final String pdfFile = c.getAttributeValueString("pdfFile");
+            final String textFile = c.getAttributeValueString("textFile");
+            stringListHits(true, Collections.singletonList(startPage), regexPattern, matchString, item, "", "Starting page", progress); //NON-NLS
+            stringListHits(true, Collections.singletonList(pdfFile), regexPattern, matchString, item, "", "PDF file", progress); //NON-NLS
+            stringListHits(true, Collections.singletonList(textFile), regexPattern, matchString, item, "", "Text File", progress); //NON-NLS
+          }
+        }
       }
+      catch (Exception NullPointerException) {
```
uckelman, line 2987: What are the exceptions which can actually be thrown here? NPE should never be caught---the code which can throw it should be fixed.

vassal-app/src/main/java/VASSAL/configure/ConfigureTree.java
```diff
@@ -2401,40 +2721,93 @@ private boolean checkSearchTarget(SearchTarget st, String searchString) {
...
+
+            final String pdfFile = c.getAttributeValueString("pdfFile");
+            if (pdfFile != null) return checkString(pdfFile, regexPattern);
+
+            final String textFile = c.getAttributeValueString("textFile");
+            if (textFile != null) return checkString(textFile, regexPattern);
+          }
+        }
+      }
+      catch (Exception NullPointerException) {
```
uckelman, line 2783: * Fix the NPE, don't catch it.
* Narrow Exception to whatever is actually thrown.

uckelman (changes requested):

uckelman (approved):

PR #12808, opened 2023-10-12
12808 - Horizontal scrolling now supports 2-finger-trackpad and mice with additional horizontal wheels

lenzls:
# 12808 - Horizontal scrolling now supports 2-finger-trackpad and fancy mice w/ additional horizontal wheels
Vassal already supports vertical scrolling of the map via the mouse wheel. This MR adds horizontal scrolling via the mouse wheel as well.

Amongst real mouse wheels, this also enables omnidirectional scrolling via
- two-finger trackpads usage on laptops, and
- middle mouse + [trackpoint](https://de.wikipedia.org/wiki/Trackpoint) movements

Here is a video of the change in action. I think it feels really nice.
[Screencast 2023-10-13 00 15 31.webm](https://github.com/vassalengine/vassal/assets/263263/1f40580a-c17e-44a3-b14e-db1ab5f4c80f)

## Notes
- in general, zooming and scrolling the map feels quite juddery. I wonder if there is anything we can do about that
- Don't ask me why swing exposes horizontal mouse wheel movements as mouse release events :shrug: 
- The change is inspired from a very similar change some time ago in the TripleA game engine:
  - issue: https://github.com/triplea-game/triplea/issues/1443
  - implementation: https://github.com/triplea-game/triplea/pull/1451

Happy for any feedback!

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/ScrollPane.java
+++ b/vassal-app/src/main/java/VASSAL/tools/ScrollPane.java
@@ -23,6 +23,8 @@ import VASSAL.tools.swing.SwingUtils;
 import javax.swing.JScrollBar;
 import javax.swing.JScrollPane;
 import java.awt.Component;
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
 import java.awt.event.MouseWheelEvent;
 import java.awt.event.MouseWheelListener;
 
@@ -121,6 +123,31 @@ public class ScrollPane extends JScrollPane {
           bar.getValue() +
           e.getUnitsToScroll() *
           bar.getUnitIncrement()
+        );
+      }
+    });
+
+    // adds support for horizontal mouse wheel scrolling.
+    view.addMouseListener(new MouseAdapter() {
+      @Override
+      public void mouseReleased(MouseEvent e) {
+        int scrollDirection = 0;
+        if (e.getButton() == 4) {
+          scrollDirection = -1;
+        }
+        else if (e.getButton() == 5) {
+          scrollDirection = +1;
+        }
+
+      
+        if (!horizontalScrollBar.isVisible()) return;
+
+        GameModule.getGameModule().setSuppressAutoCenterUpdate(false);
+
+        horizontalScrollBar.setValue(
+          horizontalScrollBar.getValue() +
+          scrollDirection *
+          horizontalScrollBar.getUnitIncrement()
         );
       }
     });
--- a/vassal-app/src/main/java/VASSAL/tools/ScrollPane.java
+++ b/vassal-app/src/main/java/VASSAL/tools/ScrollPane.java
@@ -127,30 +127,32 @@ public class ScrollPane extends JScrollPane {
       }
     });
 
-    // adds support for horizontal mouse wheel scrolling.
-    view.addMouseListener(new MouseAdapter() {
-      @Override
-      public void mouseReleased(MouseEvent e) {
-        int scrollDirection = 0;
-        if (e.getButton() == 4) {
-          scrollDirection = -1;
+    if (view != null) {
+      // adds support for horizontal mouse wheel scrolling.
+      view.addMouseListener(new MouseAdapter() {
+        @Override
+        public void mouseReleased(MouseEvent e) {
+          int scrollDirection = 0;
+          if (e.getButton() == 4) {
+            scrollDirection = -1;
+          }
+          else if (e.getButton() == 5) {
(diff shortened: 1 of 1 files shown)


uckelman:
Vassal already has horizontal scrolling on the mouse wheel, using Shift.

lenzls:
> Vassal already has horizontal scrolling on the mouse wheel, using Shift.

I'm aware. And as this is a common pattern, I think that should stay of course.

But scrolling horizontally via trackpad etc. is a ubiquitous user interface concept as well. So I think it's worth supporting it. 

Also, I find it a much smoother experience. Mainly because one does not have to scroll both axis sequentially, but rather in one continuous motion.

Cattlesquat:
Can I ask how one is intended to activate the "new" way of horizontal scrolling? Some modifier key? Some button on the mouse? It wasn't immediately obvious from the code-diff what the new functionality is.

lenzls:
> Can I ask how one is intended to activate the "new" way of horizontal scrolling? Some modifier key? Some button on the mouse? It wasn't immediately obvious from the code-diff what the new functionality is.

Sure. The goal of the MR is to support horizontal scrolling functionality of mouse like devices. How that works specifically is dependent on the device.

I was mainly concerned about laptop usage. 

On a trackpad, you can use the standard 2-figer-scroll gesture to scroll horizontally. So putting two fingers on the trackpad and sliding them sideways. 

Also, on my thinkpad, I can press the middle mouse button and move the trackpoint in the direction I want to scroll ([this video demonstrates vertical scrolling this way — horizontally works exactly the same though](https://youtu.be/H9ya4dUZfTQ?t=82)).

Both gestures already work in plenty of other application. Here is a small test page which contains a section with horizontal and vertical scroll bars. I'm on a linux laptop at the moment, and in all my browsers the gestures mentioned above work here: https://86387w.csb.app/

Note that Vassal already supports vertical scrolling via those methods. This MR simply adds support for horizontal and to some degree diagonal scrolling.


I haven't tested it, but the MR should also support regular mouses with horizontal scrolling support like
nb-75-1b
asd



Let me know if there are further question.

Cattlesquat:
What I'm trying to understand is what the functionality change is for "standard mouse with mouse wheel".

For example in your test page, if I "roll the mouse wheel" it scrolls vertically. I can't figure out how to scroll horizontally, however (using the mouse wheel), except if I hold down the shift key, which is the functionality we already have. 

Are you saying there's *no* functionality change for "standard mouse with mouse wheel" and this only changes the laptop trackpad experience (and maybe exotic mice like the ones shown above that have a separate way of applying horizontal input)? If so that seems fine to me, your original description just didn't make that clear; whereas if there *is* functionality change for standard mouse we'd want to know and document. Mostly it's a matter of describing/documenting the new behavior accurately for users (ie if they don't know it's there, they can't take advantage of it etc)

lenzls:
I realize that I wasn't very clear in my original description. 

> Are you saying there's no functionality change for "standard mouse with mouse wheel" and this only changes the laptop trackpad experience (and maybe exotic mice like the ones shown above that have a separate way of applying horizontal input)?

Yes, that pretty much sums it up.

The normal desktop PC configuration with a standard mouse should not notice any change in behaviour. For laptop users, however, it's a game changer (in my humble opinion anyway).

----

I also wanted to stress, that I didn't come up with this interaction method. It's a widely used pattern. Few documentation examples from some vendors:

- https://www.dell.com/support/kbdoc/en-us/000133650/touch-gestures-for-the-alps-touchpad
- https://www.asus.com/me-en/support/FAQ/1041464/#A3
- https://help.gnome.org/users/gnome-help/stable/mouse-touchpad-click.html.en#twofingerscroll

Cattlesquat:
Okay yeah that clears it up for me. I've marked it ready-to-merge.

Cattlesquat (approved):

uckelman:
Is there somewhere this should be mentioned in the docs?

Cattlesquat:
I actually looked for a spot documenting any of that and didn't find it -- I guess it's just supporting a standard pattern, kind of like we don't document "roll your mouse wheel to zoom vertically". The change-note will be the notice that we've essentially "fixed the bug"?

uckelman:
> I guess it's just supporting a standard pattern

"Pressing the G key types a G" :)

uckelman:
@lenzls Thanks for the PR. I've added you to the contributors list.

lenzls:
Awesome, thanks!

I have some other half-baked ideas around navigation, but I'll probably create an issue first to have a discussion upfront.

PR #12824, opened 2023-10-18
Ensure ColorManager and FontManager are created exactly once, before use

uckelman:


BrentEaston (approved):
Yep, That'll do it.

uckelman:
Emailed reporter to confirm the fix.

PR #12831, opened 2023-10-20
12830 Prevent Load Continuation from auto-forwarding to end of log

BrentEaston:
Don't ask me why this broke it, I have no idea.


riverwanderer:
> Don't ask me why this broke it, I have no idea.

If this problem arises in v3.7.1 then I was wondering if it might be due to something in #12667 or #12707. Mainly since those are the only changes I can see listed that are in the area of opening a game / log file.

BrentEaston:
Sorry, I wasn't clear. I have found and resolved the problem. I just don't understand how the change caused the issue! All good now with this PR

PR #12835, opened 2023-10-21
Piece definer line numbers

riverwanderer:
Complements search line cross-reference feature.

### DIFF
--- a/.github/ISSUE_TEMPLATE/module-bug-report.md
+++ b/.github/ISSUE_TEMPLATE/module-bug-report.md
@@ -1,0 +1,36 @@
+---
+name: Module bug report
+about: Create a report to help improve a specific module
+title: "[Module name] **enter actual module name within brackets and summarise the
+  issue here**"
+labels: bug
+assignees: ''
+
+---
+
+Vassal version: **enter version here**
+Module version: **enter version here**
+Reproduce the bug using the latest appropriate Vassal & module combination and enter their version numbers above. If this issue is reproducible in different modules, please use the Vassal bug report instead.
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior (include a link to the module and other files where required. Small files may be attached; use .zip if necessary):
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+- OS: [e.g. macOS Ventura 13.6 ]
+- Java version: [not required for Windows or macOS]
+
+**Additional context**
+Add any other context about the problem here.
--- a/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
@@ -23,6 +23,7 @@ import VASSAL.build.GpIdSupport;
 import VASSAL.build.module.KeyNamer;
 import VASSAL.build.module.documentation.HelpFile;
 import VASSAL.build.widget.PieceSlot;
+import VASSAL.configure.ConfigureTree;
 import VASSAL.configure.IntConfigurer;
 import VASSAL.i18n.Resources;
 import VASSAL.preferences.Prefs;
@@ -141,7 +142,7 @@ public class PieceDefiner extends JPanel {
   private JLabel scaleLabel;
   private final Prefs prefs;
 
-  private static String prototypeName; // If we're editing a prototype definition, this holds the name of it
+  private String prototypeName; // If we're editing a prototype definition, this holds the name of it
 
   // A Configurer to hold the users preferred maximum split size
   private static final int MINIMUM_SPLIT_SIZE = LabelUtils.noImageBoxImage().getWidth();
@@ -182,6 +183,10 @@ public class PieceDefiner extends JPanel {
     if (!this.prototypeName.equals(prototypeName)) {
       this.prototypeName = prototypeName;
     }
+  }
+
+  public String getPrototypeName() {
+    return this.prototypeName;
   }
 
   protected int getInUseSelectedIndex() {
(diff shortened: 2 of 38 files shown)


vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -141,7 +141,7 @@ public class PieceDefiner extends JPanel {
   private JLabel scaleLabel;
   private final Prefs prefs;
 
-  private String prototypeName; // If we're editing a prototype definition, this holds the name of it
+  private static String prototypeName; // If we're editing a prototype definition, this holds the name of it
```
BrentEaston, line 144: Why change this to static? It's possible to have multiple Piece Definers open at the same time, If you open a piece, then a prototype before closing the piece, the line numbers in the piece start acting like a prototype

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -1280,13 +1280,24 @@ public Component getListCellRendererComponent(
...
+      else {
+        lineNumber = (index == 0 ? "" : " ".repeat(index < 10 ? 2 : index < 100 ? 1 : 0) + index + ". ");
+      }
       super.getListCellRendererComponent(list, "", index, selected, hasFocus);
+
       if (value instanceof EditablePiece) {
-        setText(((EditablePiece) value).getDescription());
+        final String valueType = ((EditablePiece) value).getType();
+
+        setText((valueType.startsWith("cmt;") ? "" : lineNumber) + ((EditablePiece) value).getDescription());
```
BrentEaston, line 1296: Try this:

        final String description = ((EditablePiece) value).getDescription();
        if (valueType.startsWith("cmt;")) {
          setText("<html>" + lineNumber.replace(" ", "&nbsp;") + description.substring(6));
        }
        else {
          setText(lineNumber + description);
        }


Actually, a better solution is to remove the HTML from Comment.getDescription (just have it return comment) and add it here in the renderer. The HTML is just for the Piece Definer, so the piece Definer should add it in. That will make this code a little simpler and things like Search that also call getDescription will not see the html.

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -1280,13 +1280,24 @@ public Component getListCellRendererComponent(
...
       // It is incredibly inefficient for GamePieces and is not needed
       // since we overwrite the label text anyway.
+      String lineNumber;
+      if (PieceDefiner.prototypeName.isEmpty()) {
+        final int index1 = index + 1;
+        lineNumber = " ".repeat(index1 < 10 ? 2 : index1 < 100 ? 1 : 0) + index1 + ". ";
+      }
+      else {
+        lineNumber = (index == 0 ? "" : " ".repeat(index < 10 ? 2 : index < 100 ? 1 : 0) + index + ". ");
+      }
```
BrentEaston, line 1290: Given that the font used in the PieceDefiner is variable width, this code doesn't really have any effect, it does not cause the single digit and two digit numbers to line up. I would just removed all the leading spaces and just go with left justified numbers.

BrentEaston (changes requested):
Overall, I like the direction of this a lot. Just needs a few more tweaks I think

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
riverwanderer: > Why change this to static? It's possible to have multiple Piece Definers open at the same time, If you open a piece, then a prototype before closing the piece, the line numbers in the piece start acting like a prototype

Thanks Brent. I suspected there'd be an issue but lack of Java knowledge meant this was the quickest workaround for me. For the next build I am now using an independent means of determining if a Prototype is being processed.

PS I was unable to replicate problem behaviour to see what it looked like but I've just realised I didn't use the precise sequence of opening that you indicated above.

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
riverwanderer: Great suggestion, thanks @BrentEaston. I also took advantage of the comment handling now in the renderer to alter the formatting of a comment. See what you think in the next build.

.github/ISSUE_TEMPLATE/module-bug-report.md
```diff
@@ -0,0 +1,36 @@
+---
+name: Module bug report
+about: Create a report to help improve a specific module
```
BrentEaston, line 3: Are these issue templates supposed to be in this PR?

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -1281,15 +1282,47 @@ public Component getListCellRendererComponent(
...
+
       if (value instanceof EditablePiece) {
-        setText(((EditablePiece) value).getDescription());
+        final EditablePiece editableValue = (EditablePiece) value;
+        final String type = editableValue.getType().split(";")[0];
+        final String valueStr = editableValue.toString();
+        final String name = valueStr.substring(valueStr.indexOf("[name=")).split("=")[1].split(",")[0]; // extract name
+
+        // Prototype name field is null; skip first (dummy) item to avoid outputting a blank line
+        if (!name.isEmpty() || index > 0) {
```
BrentEaston, line 1293: This doesn't work. A Basic Piece is allowed to have an empty name and the piece name is defined in a Basic Name trait. 

You will need to pass the PieceDefiner into the Renderer via a new constructor so the renderer can query the PieceDefiner if it editing a Propertype or not

In the PieceDefiner:

Public PieceDefiner() {
   ...
   r = new Renderer(this);
   ...
}


public boolean isPrototype() {
  return prototypeName.isEmpty()
}

In the Renderer:

private PieceDefiner definer;

Renderer(PieceDefiner definer) {
   this.definer = definer;
}

...

if definer.isPrototype() {
...

BrentEaston (changes requested):
Looking great, Just the issue with recognizing prototypes properly.

.github/ISSUE_TEMPLATE/module-bug-report.md
riverwanderer: No. Depressing how they managed to sneak in, I thought I'd got rid. On second thoughts, maybe this PR was affected before I reverted that change from my master.

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
riverwanderer: Thanks. I couldn't work that out before. Next build incorporates these changes.

riverwanderer:
Superseded by #12861 which contains only the line-numbering commits.

PR #12855, opened 2023-10-26
12852 Prevent SumZone from generating Bad Data Errors

BrentEaston:


BrentEaston:
Tested successfully by Mark.

PR #12861, opened 2023-10-27
Number in-use traits in Piece Definer

riverwanderer:
Cleans up PR #12835.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
@@ -1303,10 +1303,10 @@ public class PieceDefiner extends JPanel {
           // For Pieces, bump index to 1
           // Special formatting for Comments
           if ((editableValue.getType().startsWith(Comment.ID))) {
-            setText(getLineNumber(index + lineOffset) + "<b>/* " + ConfigureTree.noHTML(((editableValue).getDescription())) + " */</b></html>");
+            setText(getLineNumber(index + lineOffset) + "<b>/* " + ConfigureTree.noHTML((editableValue.getDescription())) + " */</b></html>");
           }
           else {
-            setText(getLineNumber(index + lineOffset) + ConfigureTree.noHTML((editableValue).getDescription()) + "</html>");
+            setText(getLineNumber(index + lineOffset) + ConfigureTree.noHTML(editableValue.getDescription()) + "</html>");
           }
         }
       }
--- a/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
@@ -1297,17 +1297,16 @@ public class PieceDefiner extends JPanel {
 
       if (value instanceof EditablePiece) {
         final EditablePiece editableValue = (EditablePiece) value;
-        final String type = editableValue.getType().split(";")[0] + ";";
 
         // Prototype? Skip first (dummy) item to avoid outputting a blank line
         if (lineOffset == 1 || index > 0) {
           // For Pieces, bump index to 1
           // Special formatting for Comments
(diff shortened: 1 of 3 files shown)


vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -1281,15 +1292,36 @@ public Component getListCellRendererComponent(
...
       if (value instanceof EditablePiece) {
-        setText(((EditablePiece) value).getDescription());
+        final EditablePiece editableValue = (EditablePiece) value;
+        final String type = editableValue.getType().split(";")[0];
+
+        // Prototype? Skip first (dummy) item to avoid outputting a blank line
+        if (lineOffset == 1 || index > 0) {
+          // For Pieces, bump index to 1
+          // Special formatting for Comments
+          if (type.startsWith("cmt")) {
```
BrentEaston, line 1306: Better would be `if (type.startsWith(Commend.ID))`

BrentEaston (changes requested):
Just one minor change. Otherwise looks great.

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
riverwanderer: Done, accounting for the trailing separator in Comment.ID.

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -1281,15 +1292,36 @@ public Component getListCellRendererComponent(
       // It is incredibly inefficient for GamePieces and is not needed
       // since we overwrite the label text anyway.
       super.getListCellRendererComponent(list, "", index, selected, hasFocus);
+
+      final int lineOffset = definer.isPrototype() ? 1 : 0;
+
       if (value instanceof EditablePiece) {
-        setText(((EditablePiece) value).getDescription());
+        final EditablePiece editableValue = (EditablePiece) value;
+        final String type = editableValue.getType().split(";")[0] + ";";
```
BrentEaston, line 1300: Why is this needed? You can just check for if (editableValue.getType().startsWith(Comment.ID)). No need to break it up since the ; is already included in the ID.

BrentEaston (changes requested):

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
riverwanderer: Thanks - fixed in next build.

BrentEaston (approved):

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -1281,15 +1292,35 @@ public Component getListCellRendererComponent(
...
         final String s = value.getClass().getName();
-        setText(s.substring(s.lastIndexOf('.') + 1));
+        setText(getLineNumber(index + lineOffset) + s.substring(s.lastIndexOf('.') + 1) + "</html>");
       }
       return this;
     }
+
+    private static String getLineNumber(int lineNumber) {
+      return "<html><span style=color:#A0A0A0>" + lineNumber + ".</span> ";  // prep line number for output
+    }
```
uckelman, line 1323: I would not produce the `<html>` tags unbalanced like this. The function name says it produces the line number, so it should do only that. It would be better to have matching open and close tags produced in the same place.

uckelman: I suggest something like this:
```
private String cellText(int lineno, String s) {
  return "<html>" + getLineNumber(lineno) + s + </html>";
}
```
and then
```
setText(cellText(index + lineOffset, "whatever"));
```

PR #12886, opened 2023-11-08
Suppress piece palette boolean bad data errors

riverwanderer:
This PR suppresses the most common bad data errors that are generated by pieces on a palette. The issue arises when a boolean expression such as {x && y} contains properties that are neither true or false. This can happen for spurious reasons in that some properties aren't correctly initialised until instantiated onto a map ... e.g. `<layer>_Active`.

The PR checks whether the source of the error ("owner") is an Editable Piece and if so, the bad data report is not issued unless that piece is on a map (map name != null).

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/PropertyExpression.java
+++ b/vassal-app/src/main/java/VASSAL/configure/PropertyExpression.java
@@ -1,7 +1,6 @@
 package VASSAL.configure;
 
 import VASSAL.build.BadDataReport;
-import VASSAL.build.module.Map;
 import VASSAL.build.module.properties.PropertySource;
 import VASSAL.counters.EditablePiece;
 import VASSAL.counters.GamePiece;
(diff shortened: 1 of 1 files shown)


riverwanderer:
Tested with an expression in a Restrict Commands trait that fails to equate to a boolean. Trigger the trait with a right click on the piece
test1 - no error message on Piece Palette (new behaviour - passed).
test2 - bad data error message when piece is on a map (original behaviour, no change - passed).

BrentEaston (commented):
Is there a reason for just suppressing this particular subset of BadDataReports that may be generated in the palette, rather than all of them? I thought the idea was to suppress any BDR generated by a piece in the palette?

riverwanderer:
> Is there any more direct way we could check if a piece is in a palette?

Any thoughts on a method other than CurrentMap == null  @BrentEaston.  I thought maybe UniquePieceId will be null too, but that's just another proxy test, isn't it?

Issue #11999, opened 2023-02-13
Scope of Pre-Defined Scenario Refresh


riverwanderer:
From a forum [post](https://forum.vassalengine.org/t/refresh-predefined-setups-file-scope/72739).

Would be helpful if the PDS Refresh tool could be tweaked with the following targeting abilities:

a) Target a specific PreDefined setup folder only (workaround is to extract files into a temporary module for the refresh)

b) Target all .vsav files within a module - see C&C Ancients (v4 onwards) for an example of a module that has custom code resulting in save files within the module that do not appear in the standard predefined scenario structure.

Alternatively to (b), to avoid work in the Refresher, it would suffice if the Pre-Defined Setup component had a feature that allowed files to be listed internally in the module (such that the refresher will pick them up) but not be visible in the user menus.

riverwanderer:
#11506 implements alternative option (b).

riverwanderer:
To achieve the targetting part of this request in another way, a filter could be added to the PDS Refresh control panel.

PR #12893, opened 2023-11-09
Filter on the Predefined Setups control panel to speed refresh turnaround during module development

riverwanderer:
New field and associated functionality for PDS Refresh only:

image

Addresses an outstanding part of #11999.

Also adds a progress report to the PDS Refresh window and sets the GameModule dirty flag if a refresh is done.


riverwanderer:
Tested with literal, regex and no search filter. Progress banner displays and module save-prompt is now triggered.

riverwanderer:
Superceded by PR #12896.

PR #12889, opened 2023-11-08
Editor improvements (merged pre-release)

riverwanderer:
Merged queue of editor related changes practical for use & testing when developing under Vassal v3.7.x without compromising module compatibility.
- Refresh included Global Hotkey ability
- Search
- Side change messaging always sent to log / online players
- Suppress boolean errors on palette

PR #12898, opened 2023-11-11
PDS refresh improved controls with GHK feature

riverwanderer:
Expanded "orphan counter" functionality - reporting / ability to repair (sub-option).
Ability to do additional processing via a special post-refresh Global Hotkey, optionally without routine refresh reporting.
New filter field and associated functionality for PDS Refresh.
Other UI improvements.

New options & sub-options shown. Actual defaults are unchanged from vassal v3.7.5:
image

Addresses an outstanding part of #11999.

Other PDS Refresh improvements:
Control panel re-organised and window title upgraded to a Progress bar.
Removed faulty "test" option & non-functioning "delete when no map" option.
Alert option for end of refresh run (using default Vassal system wake-up sound).
Avoids duplicate file entries.
Error / warnings summary collated and reported at end of run.
Sets the GameModule dirty flag if a refresh is done.

Improvements to standard Refresh Counters:
Each of the warning messages is only issued (now highlighted in red) if 1 or instances of the error/anomaly are detected.
Control panel updated for consistency with PDS Refresh.


riverwanderer:
This PR has been robust for me in multiple use each day (testing builds via merge #12889).  Usage has included playing games that have been refreshed.

The core features (piece and deck refreshing) are mostly untouched - most changes are are in the Predefined refresh control routine, with just peripheral measurement and integration hooks touched in GameRefresher.  One exception - the PR's GameRefresher can repair bad GPIDs (obviously not a default, but I have used it successfully to repair CCN scenarios). Reporting is improved and now includes listing orphan pieces, so an informed decision on repair can be made.

Since it imposes no changes the Player, I am still hoping that this PR can be included in v3.7.6 so that the wider module dev community can benefit.

riverwanderer:
I am revising this PR via #12965 which addresses a dependency that the Global Hotkey feature has on Refresh. I will merge back to this PR once I've tested out my resolution to that issue.

vassal-app/src/main/java/VASSAL/build/GpIdChecker.java
```diff
@@ -51,6 +51,7 @@ public class GpIdChecker {
 
   protected GpIdSupport gpIdSupport;
   protected int maxId;
+  public int noGpIdMatch = 0; // shared to GameRefresher
```
uckelman, line 54: The only classes which should have public fields are ones which are _solely_ data classes.

vassal-app/src/main/java/VASSAL/build/GpIdChecker.java
```diff
@@ -258,11 +263,17 @@ public GamePiece createUpdatedPiece(GamePiece oldPiece) {
         if (oldPieceName.equals(gpName)) {
           newPiece = element.createPiece(oldPiece, this);
           copyState(oldPiece, newPiece);
+          if (fixGPID()) newPiece.setProperty(Properties.PIECE_ID, slotPiece.getProperty(Properties.PIECE_ID));
```
uckelman, line 266: Conditionals with single-line, unbraced bodies are hard to read and should be avoided.

vassal-app/src/main/java/VASSAL/configure/RefreshPredefinedSetupsDialog.java
```diff
@@ -188,75 +281,295 @@ public void log(String message) {
...
+    Pattern filterPattern2 = null;
+
+    if (isFilterMode()) {
+
+      try {
+        // matching, assuming Regex with no escape of the end string modifier, otherwise match all to end
+        filterPattern = Pattern.compile(pdsFilter, CASE_INSENSITIVE);
+        filterPattern2 = Pattern.compile(pdsFilter + ".*", CASE_INSENSITIVE);
+      }
+      catch (java.util.regex.PatternSyntaxException e) {
```
uckelman, line 310: Import this. We try to avoid using fully-qualified class names.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/build/GpIdChecker.java
riverwanderer: Corrected this one. I have more of these in recent work. I shall start using braced bodies only.

vassal-app/src/main/java/VASSAL/configure/RefreshPredefinedSetupsDialog.java
riverwanderer: Done.

claudiociardelli:
Hello Mark,
I am happy to see that you have been working on the PDS refresh tool. I have worked on this topic 2 or 3 yrs ago and am still receiving some github notifications. 
Hope you are getting on top of this complicated feature. I do not remember the details any more, but I remember that it was complex. 
Kr
Claudio

riverwanderer:
> I am happy to see that you have been working on the PDS refresh tool. I have worked on this topic 2 or 3 yrs ago and am still receiving some github notifications.

Thanks Claudio - I have had great benefit from that work over those years! More  recently, since I've started to dip my toe into Java and Vassal itself, through this PR I've added mainly peripheral features that I thought would be useful to make PDS-related workflow even more efficient.

In the core refresh, apart from an optional, non-default ability to "repair" orphan pieces (no matching GPiD), I've only tweaked the reporting output. I'd like to add ability to populate decks and at-start stacks but time to explore that eludes me at the moment.

vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -93,24 +93,38 @@ public final class GameRefresher implements CommandEncoder, GameComponent {
...
+  private int totalCount;
+  private int totalDecks;
+
+  int notFoundCount; // shared to PDS refresher
+  int noGpIdMatch; // shared to PDS refresher
+  int noStackCount; // shared to PDS refresher - not used!!!
+  int noMapCount; // shared to PDS refresher - not used!!!
+  int notOwnedCount; // shared to PDS refresher
+  int notVisibleCount; // shared to PDS refresher
+  int deckWarnings; // shared to PDS refresher
```
uckelman, line 124: What's the deal with these? They need a visibility modifier. Should they be private? protected?

riverwanderer: These variables and the 3 immediately preceding them are various types of error accumulated during a refresh run. Visibility must at least allow PreDefinedSetup.java to access them - it uses them to return an error count to RefreshPreDefinedSetupDialog. This is all so that the user gets a summary at the end rather than having to trawl through all the refresh logs.

Private doesn't work so I guess Protected ? IntelliJ doesn't approve of that and warns "Class member declared 'protected' in 'final' class"  hence I accepted it's suggestion, which removed the Protected modifier:

image

i can added the Protected modifiers back in  unless there's some other way I should be doing this.

uckelman: All see that you're doing with these variables outside of `GameRefresher` is summing them in one location. If that's all you need, then they should be made private and you can create a public method inside of which you compute the sum, and then call that method from outside of `GameRefresher`.

riverwanderer: Thanks. I've pushed the fix to this.

Whilst doing the fix, I've noticed that somehow in a commit on 7th Dec, I've managed to over-write the change history in GameRefresher. I presume it's happened whilst I was working around a glitch in IntelliJ.  I think I may need to unwind or re-do this PR to fix this.

image

uckelman (approved):

PR #12907, opened 2023-11-14
Deck Key Commands do not work on subsequent new games

rlament:
This fixes deck sorting not working on the next and each subsequent "File->New Game". The deck sort HotKey works the first instance of a game module. After a "File->Close" and then "New Game", the sort HotKey does not work. The workaround prior to this fix: user must completely exit Vassal and restart it.

The sortListener must be null to successfully assign a new HotKey during initialization.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/deck/DeckSendKeyCommand.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/deck/DeckSendKeyCommand.java
@@ -315,6 +315,7 @@ public class DeckSendKeyCommand extends AbstractDeckKeyCommand {
   public void deregisterListeners() {
     if (sendListener != null) {
       GameModule.getGameModule().removeKeyStrokeListener(sendListener);
+      sendListener = null;
     }
   }
 
--- a/vassal-app/src/main/java/VASSAL/build/module/map/deck/DeckSortKeyCommand.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/deck/DeckSortKeyCommand.java
@@ -121,6 +121,7 @@ public class DeckSortKeyCommand extends AbstractDeckKeyCommand {
   public void deregisterListeners() {
     if (sortListener != null) {
       GameModule.getGameModule().removeKeyStrokeListener(sortListener);
+      sortListener = null;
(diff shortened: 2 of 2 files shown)


BrentEaston (changes requested):
Excellent pick-up, thanks very much for the PR.

Looks like the DeckSendKeyCommand has the same problem, can you update the PR to include that one as well? Thanks

rlament:
> Excellent pick-up, thanks very much for the PR.
> 
> Looks like the DeckSendKeyCommand has the same problem, can you update the PR to include that one as well? Thanks

In my module, the deck has both a DeckSendKeyCommand and a DeckSortKeyCommand. DeckSendKeyCommand has been working for me; don't see an issue there. DeckSendKeyCommand has similar code, but it doesn't seem to be used at all in the player, perhaps only in the editor? I set numerous breakpoints in DeckSendKeyCommand.java and didn't hit them from the player. It seems that DeckSendKeyCommand's are registered in some other way, but at this point I'm not sure where that is. At least it's not via `DeckSendKeyCommand.registerListeners()`.

At this point, I don't see the need to make any changes other than changing the title of this pull request to specifically say "Deck Sort Key Commands..."

BrentEaston:
My testing indicates the Deck Send Key Command is failing in exactly the same way, for the same reason, as the Deck Sort Key Command.

rlament:
@BrentEaston Sorry, my confusion with Deck Send Key Command and Deck Global Key Command. Testing with correct command (Send not Global), I can reproduce the failure. You are right, Deck Send Key Command requires the same fix.

BrentEaston (approved):
Looks good to me!

uckelman:
@rlament Thanks for the contribution. I'd like to add your name to our list of contributers if you're willing. How should I list you?

rlament:
> Thanks for the contribution. I'd like to add your name to our list of contributers if you're willing. How should I list you?

@uckelman Glad to see this fix going into the main branch without delay. I'd be happy to be listed as a contributor. List me as: Ray Lament

uckelman:
I've added you to the list. You'll show up [here](https://vassalengine.org/about.html#team) the next time I deploy the site.

PR #12936, opened 2023-11-27
Labels have incorrect bounds when using italics

djberg96:
Currently a label font using italics can run off the edge of its bounding box, as you can see with the "0" here:

Screenshot 2023-11-26 at 9 20 29 PM

This is actually a quirk of italics, as the `getFontMetrics` function does logical vs visual calculations. Old but relevant conversation at https://bugs.openjdk.org/browse/JDK-4262130, from which I borrowed for the solution.

Original conversation starts at https://discord.com/channels/740127632586178572/911199889059360798/1178520800928735272 for those interested.


BrentEaston (approved):
Looks good to me

PR #12944, opened 2023-11-29
Add pieceCount to the map auto reporter

### LINKED ISSUE #12943: Add a piece count property to the map level auto-reporter
When moving pieces to a player's private window, such as a player hand, the chat window should not report the identity of the pieces. When moving pieces to a map, the auto-reporter generates one message regardless if moving one piece or a stack full of pieces. The only property in the auto-reporter that indirectly indicates the number of pieces, by revealing them all, is the pieceName property. A piece count property could keep players honest by reporting the number of units moved without revealing the pieces' identities.

Expanding on this very common scenario, consider a player drawing multiple cards from a facedown deck to the player hand. The pieces are not obscured, thus the hidden movement reporter is not invoked. When moving multiple pieces, the piece name property is a comma separated list revealing the identity of all pieces in the stack. Removing the pieceName property from the message or substituting a generic message also supresses the quantity information. 

Workarounds are possible. The module designer can turn off auto-reporting and add a Report Action to the pieces. Units could report their own movement, but this takes some knowledge and effort by the module designer. This technique generates one message per piece. A player has to manually count the number of lines of output to mentally compute the number of pieces moved. This is a workable solution that produces a chatty log while not communicating the number of pieces in an efficient manner.

The map auto-reporter is convenient labour saving feature for the module designer. The messages are defined by default. Meanwhile, decks have a built-in draw multiple function. The only missing component is the reporting of the quantity. The number of pieces is readily known and should be exposed to the module designer.

rlament:
Add `pieceCount` to the map auto reporter. Closes #12943

The label is also included in the FormattedExpressionConfigurer such that it appears in the Inserter dropdown.

Also included in this pull request are spelling and grammar corrections in the comments.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -612,7 +612,7 @@ public class Map extends AbstractToolbarItem implements GameComponent, MouseList
 
   /**
    * @return a String representation of the XML buildFile attribute with the given name. When initializing a module,
-   * this String value will loaded from the XML and passed to {@link #setAttribute}. It is also frequently used for
+   * this String value is loaded from the XML and passed to {@link #setAttribute}. It is also frequently used for
    * checking the current value of an attribute.
    *
    * @param key the name of the attribute. Will be one of those listed in {@link #getAttributeNames}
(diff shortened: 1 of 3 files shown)


BrentEaston (changes requested):
Couple of things.

1. I have set the milestone for this change to 3.8.0 as it is adding new functionality that does not exist in existing 3.7 releases. It will get merged into 3.8 when we get the 3.8 branch.
2. Can you please also update the reference guide (Map.adoc) in the Vassal-app Module

rlament:
> 1. I have set the milestone for this change to 3.8.0 as it is adding new functionality that does not exist in existing 3.7 releases. It will get merged into 3.8 when we get the 3.8 branch.

Maybe I can convince you to add it to 3.7. It not so much a new feature as an additional property to an existing feature. It is backward compatible as older versions will treat the property as unknown and therefore blank. The reported text may not make sense, but it won't break existing functionality.

I've updated the documentation and is ready for review.

BrentEaston (changes requested):
Looking good, Just minor issue with the doco.

BrentEaston (approved):
Looks good to me, passed testing.

PR #12965, opened 2023-12-05
PDS refresh improved controls - reverted to a single GHK

riverwanderer:
Re-works #12898 with the option of a single Global Hotkey only, available after the refresh. Satisfies a dependency that the GHK has on the pieces refresh.


riverwanderer:
Merged back into parent branch #12898 to be progressed there.

PR #12971, opened 2023-12-08
12971 - Refresh visible portions of Maps after each UI, log or network action

BrentEaston:
/**
   * Refresh the visible portions of all currently showing maps.
   * This is called in when the following are completed:
   *  - Any right-click menu option on a piece
   *  - Any Toolbar button action
   *  - Each Step replay from a log file
   *  - Each message received during on-line play
   *  These four actions can cause changes that result in text displaying Calculated properties and Beanshell expressions
   *  to change in places unrelated to where the action occurred, leaving these counters out of sync until a later click
   *  in their general area.
   */


riverwanderer:
@BrentEaston - Testing build VASSAL-3.7.6-SNAPSHOT-fc0e80a-remoteRefresh:  Partial refresh observed ("Infantry Forward Leader" cards x2 in the screen shot). The borders are supposed to refresh either fully pale (layer active) or bright red (layer inactive). These refreshes can triggered by Action Button trait or by right click menu commands. Fails observed both ways. Moving the cards to a different position in the hand window and repeating the test did make a difference - cards refreshed ok at other positions.

image

BrentEaston:
Can you send me a save file showing this happening, plus details of what to click? Thanks.

riverwanderer:
> Can you send me a save file showing this happening, plus details of what to click? Thanks.

to replicate - use the the Dev CCN module from [here](https://www.dropbox.com/scl/fi/ezkglknmamgygzjg68u6x/CCNapoleonics4_2_0DEV-Mark-Benson-s-conflicted-copy-2023-12-09.vmod?rlkey=4rulsj629ey6p53rz4qltlnu0&dl=0).
Open this save game from the module, joining as Solitaire: [test for refresh.zip](https://github.com/vassalengine/vassal/files/13627504/test.for.refresh.zip)

The bug seems sensitive to the position of cards within the window. So first position your screen along the lines shown in the screenshot here. Now right click on the leader / unit stack indicated and select the "Order" command. This should show the issue, on the two "Leader Infantry Forward" cards. Repeat the Order command to see the highlighting change again.

Pasted Graphic 4

BrentEaston:
Try the new version building now.

riverwanderer:
> Try the new version building now.

Seems fixed:  using Build VASSAL-3.7.6-SNAPSHOT-1c941de-remoteRefresh I can't reproduce the issue following my own instructions / test case above.

vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -2304,6 +2304,32 @@ else if (DRAWING_MOUSEOVER_INDEX_PROPERTY.equals(key)) {
...
+   *  These eight actions can cause changes that result in text displaying Calculated properties and Beanshell expressions
+   *  to change in places unrelated to where the action occurred, leaving these counters out of sync until a later click
+   *  in their general area.
+   */
+  public final void refreshVisibleMaps() {
+    for (final Map map : Map.getMapList()) {
+      final JComponent view = map.getView();
+      if (view.isShowing()) {
+        final Rectangle v = view.getVisibleRect();
+        view.repaint(v.x, v.y, v.width, v.height);
```
uckelman, line 2328: I'm not sure it makes a difference to specify the visible rectangle, as the nullary version of `repaint()` ought to only ever repaint the visible portion anyway. Have you tried calling `view.repaint()` instead and seeing if it makes a noticeable difference?

PR #12973, opened 2023-12-09
12765 - Cancelling Return To Deck no longer causes crash

BrentEaston:
When 'Prompt for Deck' option on a return to Deck trait is selected, then a list of available Decks is provided. You click one, then click the OK or Cancel button.

If you click OK or Cancel without selecting a Deck, you get a NPE.


vassal-app/src/main/java/VASSAL/counters/ReturnToDeck.java
```diff
@@ -151,6 +151,9 @@ public Command myKeyEvent(KeyStroke stroke) {
 
       if (deckSelect) {
         pile = promptForDrawPile();
+        if (pile != null) {
+          return null;  // No Deck selected
+        }
```
BrentEaston, line 156: This does not make sense

PR #12974, opened 2023-12-09
12931 - Prevent Zero length string error on Deck refresh

BrentEaston:


vassal-app/src/main/java/VASSAL/counters/Deck.java
```diff
@@ -405,7 +405,13 @@ private void updateCounts(GamePiece p, boolean increase) {
       if ("true".equals(evaluated) || (!"false".equals(evaluated) && PropertiesPieceFilter.parse(evaluated).accept(p))) { //NON-NLS
         final String mapProperty = prop.getPropertyValue();
         if (mapProperty != null) {
-          int newValue = Integer.decode(mapProperty);
+          int newValue;
+          try {
+            newValue = Integer.decode(mapProperty);
+          }
+          catch (NumberFormatException ignored) {
```
rlament, line 412: On a decrease, does it make sense to set the property to -1? What does a negative deck count mean? Seems to me that if the original value is invalid/unknown it should reset the property to zero and not decrement. On increment, setting the property to 1, as you have it, is the most sensible.

PR #12975, opened 2023-12-09
12894 Prevent crash if I/O Exception during CRC calculation

BrentEaston:


BrentEaston:
Also better reporting of what went wrong

vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -2466,16 +2466,9 @@ private void buildCrc() {
...
-      // Base CRC is of module file only
-      crc = CRCUtils.getCRC(List.of(files.get(0)));
-      // CombinedCrc includeds extensions as well
-      combinedCrc = CRCUtils.getCRC(files);
-    }
-    catch (IOException e) {
-      log.error("Error generating CRC", e); //NON-NLS
-      return;
-    }
+    crc = CRCUtils.getCRC(List.of(files.get(0)));
```
rlament, line 2469: Not sure if it is possible, but just to be paranoid, could check that `files` contains at least one file (i.e. not empty).

Otherwise, if it is ok to throw an `IndexOutOfBoundsException`, than that should be added to the documentation of this function. If it throws, `combinedCrc` won't be initialized.

vassal-app/src/main/java/VASSAL/tools/CRCUtils.java
```diff
@@ -44,11 +47,17 @@ public class CRCUtils {
    * @return CRC
    * @throws IOException oops
```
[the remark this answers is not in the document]
BrentEaston: Thanks, fixed.

vassal-app/src/main/java/VASSAL/build/GameModule.java
BrentEaston: No, the first file is guaranteed to be the module file.

Issue #12995, opened 2023-12-16
Very slow editing (or even scrolling past) pieces with lots of properties since 3.7.5


assaultdoor:
I've got a piece with roughly 1,000 properties. I know this isn't ideal, but it has generally worked fine. When I switched to version 3.7.5 (on OS X), VASSAL took a minute or more to open this piece. Scrolling through the piece was slow, and seemed to get slower as I moved down the list of properties. For that matter, just clicking on or moving over this piece in the editor (using the arrow keys) brings VASSAL to a halt.

I switched to OpenJDK 11, and the problem persisted.

I saved the module in 3.7.5 and restarted VASSAL, but the problem persisted.

I switched to 3.7.6 and the problem persisted.

I reverted to 3.7.4 and things are fine.

uckelman:
Where can we see the module involved?

assaultdoor:
https://drive.google.com/file/d/1gjYNCiUuZFgsmk5xlRj-Q6q0V1DdXQ7F/view?usp=drive_link

The worst offender that I noticed is under Hidden Palette > Pieces > Spirit setup components > spirit panel relay

assaultdoor:
A very basic module is here: https://drive.google.com/file/d/1INa58oVZIXmujaCtPjYDQ3ZjIiguul8z/view?usp=sharing

Try the piece "testing" in the at-start stack on the main map. It's got about 600 copies of the basic Delete property.

riverwanderer:
I suspect that PR #12861 - line numbering - is responsible.  It's working fine for me on pieces with up to 100 traits but the small overhead probably adds up. Modifying the demo, It's ok but starting to slow over 300 traits on my old MacBook. The example of 700+ traits is fairly unusable. Curiously, even selecting the piece takes forever at that number of traits.

Confirmed that the same operations are instantaneous with Vassal v3..7.4.

Paradoxically, the change actually helps with long trait lists, as Search can now give you line number references (just released in v3.7.6).  I wonder if there's a way to improve the efficiency of the line numbering but then again, a piece or prototype with 300+ traits may have other issues.

BrentEaston:
My suspicion would be that it is the use of html in the cell renderer. First test is to try just getting rid of the html and span tags and just have the line number in the default text.

riverwanderer:
> My suspicion would be that it is the use of html in the cell renderer. First test is to try just getting rid of the html and span tags and just have the line number in the default text.

[...]

PR #12997, opened 2023-12-18
Attempt performance fix.

riverwanderer:
Removes line number formatting, so that HTML is only used for Comment formatting (reverting to Vassal v3.7.4 in that respect).

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
@@ -1285,10 +1285,10 @@ public class PieceDefiner extends JPanel {
     }
 
     // This can be used from cellTextHTML() to format the line numbers - stopped using at Vassal v3.7.5 as performance was poor for very large trait lists
-    private static String getLineNumber(int lineNumber) {
-      // prep line number for output
-      return "<span style=color:#A0A0A0>" + lineNumber + ".</span>";
-    }
+//    private static String getLineNumber(int lineNumber) {
+//      // prep line number for output
+//      return "<span style=color:#A0A0A0>" + lineNumber + ".</span>";
+//    }
 
     private static String cellTextHTML(int lineno, String s) {
       return "<html>" + lineno + ". " + s + "</html>";
--- a/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
+++ b/vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
@@ -1284,49 +1284,40 @@ public class PieceDefiner extends JPanel {
       this.definer = definer;
     }
 
-    // This can be used from cellTextHTML() to format the line numbers - stopped using at Vassal v3.7.5 as performance was poor for very large trait lists
-//    private static String getLineNumber(int lineNumber) {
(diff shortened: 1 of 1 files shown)


riverwanderer:
This change - and more extensive reversion of the suspect PR - has made no impact on the issue #12995.

BrentEaston:
I commented on the original issue. You should be able to put all the HTML back if you change the call from EditableValue.getType() to Decorator.myGetType().

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -1295,33 +1295,29 @@ private static String cellText(int lineno, String s) {
...
-          // For Pieces, bump index to 1
-          // Special formatting for Comments
-          if ((editableValue.getType().startsWith(Comment.ID))) {
-            setText(cellText(lineno, "<b>/* " + ConfigureTree.noHTML((editableValue.getDescription())) + " */</b>"));
-          }
-          else {
-            setText(cellText(lineno, ConfigureTree.noHTML(editableValue.getDescription())));
-          }
+
+        if (value instanceof Decorator && ((Decorator) value).myGetType().startsWith(Comment.ID)) {
```
BrentEaston, line 1310: You should be able to replace this with just

`if (value instanceof Comment) {`

which will be more efficient.

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
```diff
@@ -1295,33 +1295,29 @@ private static String cellText(int lineno, String s) {
...
+        }
+        else if (lineno > 0) {
+          // Prototype? Skip first (dummy) item
+          setText(cellText(lineno, ConfigureTree.noHTML(((EditablePiece) value).getDescription())));
         }
+
       }
       else {
-        // FIXME: When does this get used ? Assumed that prototype might have to be skipped here also.
+        // Redundant ?
```
BrentEaston, line 1320: I really don't think anything coming through can not be an EditablePiece, , I think the `instaceof EditablePiece` above is just defensive programming to prevent un-guarded casts to EditablePiece. Does no harm being there, I would leave it in case there are other case we're not aware of. Ah, just found one, comment below.

BrentEaston (commented):
There is an issue in the MassPieceLoader which sub-classes the PieceDefiner.  Select 'Add multiple pieces' on a GamePiece palette widget and click on the 'Edit Template' button.

The MPL uses a subclass of the Prototype PieceDefiner.

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
riverwanderer: Change made in PR #13001.

vassal-app/src/main/java/VASSAL/counters/PieceDefiner.java
riverwanderer: PR #13001 modifies the comments to clarify the catch-all purpose of the EditablePiece test and related code.

riverwanderer:
Superceded by #13001.

PR #13001, opened 2023-12-19
Fix poor performance of Piece Definer; remove spurious blank line from Mass Piece Loader

riverwanderer:
(Updated) PR now reverts MPL to pre-v3.7.5 behaviour - no Basic Piece line, blank or otherwise.

This PR can supersede, or be merged back into, PR #12997.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/MassPieceLoader.java
+++ b/vassal-app/src/main/java/VASSAL/counters/MassPieceLoader.java
@@ -823,6 +823,11 @@ public class MassPieceLoader {
       refresh();
     }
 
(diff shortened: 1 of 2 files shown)


BrentEaston (commented):
To return the Mass Piece Loader behaviour to the way it was before, just add

`    @Override
    public boolean isPrototype() {
      return true;
    }
`
to MassPieceLoader.MassPieceDefiner to have the renderer treat the MPL the same as a prototype and hide the first entry

BrentEaston (approved):
Looks good to me and passes testing

PR #13005, opened 2023-12-20
13003 - Do not record Map Id's in saved Deck files so they will load on any Map

BrentEaston:


rlament:
The attached patch file adds support for loading malformed decks (i.e. decks with recorded map IDs). 

The assumption is that a piece belongs on the same map as the parent. A recent change in `setParent()` somewhat reinforces this assumption. This patch sets the piece's map to the parent's map, if a parent exists. This is quite a low-level modification and I'm not very familiar with this code base. If there is a situation where this assumption is invalidated, let me know. Any feedback is appreciated.

Tested this with decks created in 3.7.3 and 3.7.6 using the Imperial Assault Skirmish module as per the forum discussion.

This patch may also address the issue in pull request #12826. I was looking for an example of stack movement, but could find one.

[setstateparentmap.patch](https://github.com/vassalengine/vassal/files/13774603/setstateparentmap.patch)

BTW: by reverting `getstate()` as per this pull request (without the patch), the decks do load at the site where the user initiates the deck load command.

BrentEaston:
Hi Ray,

Thanks for the input.

Your fix is sort of working at odds with mine. I am trying to make sure that BasicPiece.map is always correct, so that BasicPiece.getState() doesn't have to worry about anything outside it. Your fix is saying don't trust BasicPiece.map because it may not be correct, so it is adding extra code to keep the current broken code going. 

The recent change to setParent() to use the parent map if the supplied map is null is really just a hack to catch any other bugs that might be out there.

The benefit your patch gives of being able to load the corrupted Decks gives I am not to worried about. That bug was present for only a few versions of Vassal and has a work-around. 

Regards.

BrentEaston:
From initial reporter:

Tried the “VASSAL-3.7.7-SNAPSHOT-85ec1ee-13003-SavedDeckWrongMap-windows-x86_64” and it’s working :slight_smile:

PR #13023, opened 2023-12-27
13021 - Allow wizard to load setups from extensions

BrentEaston:
Bug introduced into 3.7.6


BrentEaston:
Reporter confirms resolves issue

PR #13032, opened 2024-01-02
Draft: Game map progression animation

lenzls:
The general idea is to create a video of the map progression throughout a game. Ideally, we see the progress of the pieces over the course of a game, for example a shifting frontline.

## Current approach
The current approach is rather primitive and only semi-automated.
The MR adds another menu entry "Load series of files and create map animation" to the "tools" menu. If you click it, you need to select a folder containing all the vlog files of your game.

Then we'll automatically loop over those vlog files, iterate through every command in the logfile and take a map screenshot each time. At the very end of the process, the selected folder will contain a series of map screenshot, a screenshot for every command in every logfile.

After that, I'm (optionally) cropping the screenshots to the map section I'm interested in (for limited scenarios)
```
mogrify -format 'png' -crop 2173x1005+3543+938 +repage *-Map0.png
```
Note, I need to do that since Vassal's `SnapshotTask` currently takes x,y coordinates, but ignores them.

After that, I'm (optionally) watermarking the map screenshots with the name of the save game via
```
mogrify -format 'png' -font Liberation-Sans -fill white -undercolor '#00000080' -pointsize 26 -gravity NorthEast -annotate +10+10 %t *-Map0.png
```

As a last step, I'm combining all map screenshots into a video via
```
ffmpeg -f image2 -framerate 30 -pattern_type glob -i "*-Map0.png" output-30fps.webm
```

## Know shortcomings

- not fully integrated into the vassal java application (should/can it?)
- lots of temporary image files. The first two turns of my current small scenario game of Last Blitzkrieg resulted in 2042 screenshots and over 60GB of image data. The PNG screenshots are huge. Note that the final encoded video is only 13 MB.
- Currently I'm creating a screenshot for _every_ command, even when it was just a text message. I should filter that somehow to only screenshot the actual piece movements

## Questions

- Can we somehow easier get a video stream than doing expensive map screenshots?
- Can the ffmpeg part also be done in plain java?
- External dependencies and everything aside, would this kind of functionality be interesting to include in the vassal application?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ VNUM:=3.8.0
 # major-minor part of the version
 V_MAJ_MIN:=$(shell echo "$(VNUM)" | cut -f1,2 -d'.')
 
-MAVEN_VERSION:=$(VNUM)-SNAPSHOT
+MAVEN_VERSION:=$(VNUM)-EISMEER-SNAPSHOT-4
 #MAVEN_VERSION:=$(VNUM)-beta5
 #MAVEN_VERSION:=$(VNUM)
 
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
     <groupId>org.vassalengine</groupId>
     <artifactId>vassal-parent</artifactId>
-    <version>3.8.0-SNAPSHOT</version>
+    <version>3.8.0-EISMEER-SNAPSHOT-4</version>
     <packaging>pom</packaging>
 
     <name>Vassal Engine</name>
--- a/release-prepare/pom.xml
+++ b/release-prepare/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.vassalengine</groupId>
         <artifactId>vassal-parent</artifactId>
-        <version>3.8.0-SNAPSHOT</version>
+        <version>3.8.0-EISMEER-SNAPSHOT-4</version>
     </parent>
 
     <artifactId>release-prepare</artifactId>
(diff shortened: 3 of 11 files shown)


riverwanderer:
Certainly interests me. Even as a standalone app. Built-in would make it easier for everyone to find.

lenzls:
fyi: I just updated the code to newest master (luckily no conflicts whatsoever). 

I still don't have a plan or idea how to tackle the key shortcomings mention in the PR description. The current state is definitely not ready to me merged.

Still, besides myself, another community member is also using my SNAPSHOT build to create videos from their game: https://www.youtube.com/playlist?list=PLfAJ-jit45UM_jxG-R7FbDCeSLVoIgbew

lenzls:
> Perhaps a simpler option, which I've used (see [this youtube playlist](https://www.youtube.com/playlist?list=PL2L5yQfZrSeFVKfyPuUEE-Ji7wRfP32oL)) is to
> 
>     * Load a `.vlog` (e.g., tutorial)
> 
>     * Start screen-recording of the VASSAL window or entire screen
> 
>     * Step through the log, possibly bringing up windows when relevant.
> 
> 
> This approach has the benefits that
> 
>     * one can (almost) see the chat
> 
>     * show auxiliary windows (charts, OOBs, etc.)
> 
>     * and that one can in principle narrate on top of it.
> 
> 
> That said, it is in interesting idea.
> 
> Yours, _Christian_

Interesting approach! It does have some advantages.

One problem I can see is that it will only show things that fit on the screen. With the screenshot approach from the MR, we can create videos for larger games with multiple maps.

See for example this video: https://www.youtube.com/watch?v=P4fXirhIPqU

Kind regards

PR #13033, opened 2024-01-03
13022 - Prevent room non-owners starting new games

BrentEaston:
This solution pops up a dialog if a non-owner tries to start a new game in an online room and does not start the game.  Not ideal, but manipulating the enabled status of all the different menu options involved is too difficult.


rlament:
This change tightens things up but doesn't stop the owner from switching sides using this trick.

In regards to kicking a user from a slot, leaving this exploit available allows for it. Use this exploit to switch sides, do any cleanup and retire the user that so ungraciously left. Then sync on another player in the room.

PR #13038, opened 2024-01-03
13037 - Make Zoom component a default component of all maps

BrentEaston:


BrentEaston:
Updated, leaving Zoom as Mandatory

PR #13051, opened 2024-01-09
13049 - Compute correct _Degrees property for free rotations

### LINKED ISSUE #13049: Counter-clockwise free rotation computes the wrong angle
Take a counter named "Rotatable" with a "Can Rotate" trait with arbitrary rotations set.
From the default orientation, perform a small clockwise rotation.
Report the final orientation using the _Rotatable_Degrees_ property.
This should show a smallish positive number.
Reset the counters orientation to the default (typically zero degrees).
Perform a small counter-clockwise rotation.
Report the final orientation again using the _Rotatable_Degrees_ property.
This is also a smallish positive number, when it should be some number near, but not above 360 degrees.

Secondly, the _Degrees property is not constrained to the range 0 to 359 degrees. If one continues to rotate in the same direction, the angle exceeds 360 degrees, such that after two full rotations it is over 720.

### LINKED ISSUE #13050: During free rotation, dragging the mouse through the counter pivot point adds an extra 90 counter-clockwise rotation
When the mouse position and the counter `origin` are exactly equal (in both x and y), the `getRelativeAngle()` function returns 90 degrees when it should return zero degrees.

rlament:
On counter-clockwise free rotations compute the correct final orientation for the *<rotator_name>_Degrees* property. This fixes a bug where both clockwise and counter-clockwise rotations produce the same degrees despite the final facings being on either side of zero degrees.

When a free rotation passes a full rotation, the *<rotator_name>_Degrees* property keeps increasing potentially well past 360 degrees. This commit limits that property to the range 0 to 359 degrees.

The `getRelativeAngle()` function does not properly handle the case where the mouse position is the same as the counter's origin. This commit checks for that case.

The new unit tests demonstrate these issues and confirm the fixes.

Closes #13049
Closes #13050

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
+++ b/vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
@@ -227,7 +227,7 @@ public class FreeRotator extends Decorator
 
   public void setAngle(double angle) {
     if (isFreeRotation()) {
-      validAngles[angleIndex] = angle;
+      validAngles[angleIndex] = ((angle % 360) - 360) % 360;
     }
     else {
       // We (stupidly) store allowed angles in descending order from 0.
@@ -769,7 +769,7 @@ public class FreeRotator extends Decorator
       return String.valueOf(angleIndex + 1);
     }
     else if ((name + DEGREES).equals(key)) {
-      return String.valueOf((int) (Math.abs(validAngles[angleIndex])));
+      return String.valueOf((int) Math.round(Math.abs(validAngles[angleIndex])) % 360);
     }
     else {
       return super.getProperty(key);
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/counters/FreeRotator.java
```diff
@@ -227,7 +227,7 @@ public boolean isFreeRotation() {
 
   public void setAngle(double angle) {
     if (isFreeRotation()) {
-      validAngles[angleIndex] = angle;
+      validAngles[angleIndex] = ((angle % 360) - 360) % 360;
```
uckelman, line 230: Rather than repeating the angle normalization, it would be better to move the original lines 235-238 out ahead of the if block:
```
      // We (stupidly) store allowed angles in descending order from 0.
      // Normalize the angle to be in (-360, 0] to match that.
      angle = ((angle % 360) - 360) % 360;

      if (isFreeRotation()) { 
```

uckelman:
Looks good. Thank you for the thorough tests.

PR #13058, opened 2024-01-12
13057-Remove Unused Images does not find any images in use in an Extension

BrentEaston:
…nsion


BrentEaston:
tested successfully by original reporter.

PR #13082, opened 2024-01-21
13081 - Sleep function now sleeps correct number of milliseconds instead of 2000

BrentEaston:


BrentEaston:
Re-created based on 3.7.8

PR #13086, opened 2024-01-21
13073 - Attachments are lost after Undo

BrentEaston:


BrentEaston:
Re-create on 3.7.8 branch

PR #13090, opened 2024-01-22
13081 - Sleep function now sleeps correct number of milliseconds instead of 2000

BrentEaston:


dsmith71:
Adjusts speed according to the settings now. Some screen flicker at the slower speeds. Probably normal?
https://forum.vassalengine.org/t/possible-bug-mat-cargo-move-fixed-distance-trait/78721/25

PR #13093, opened 2024-01-22
13072 - Cargo now correctly follows Mat after Moved Fixed Distance

BrentEaston:


dsmith71:
Just got around to Testing this today and the Sleep timer can't find any issues. Commented on the Forum [https://forum.vassalengine.org/t/possible-bug-mat-cargo-move-fixed-distance-trait/78721/25](url)

BrentEaston:
Issue confirmed fixed by original reporter.

PR #13096, opened 2024-01-24
5813 Prevent Wizard crashing when using Prev/Next when no Password exists

BrentEaston:


vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
```diff
@@ -288,7 +289,7 @@ protected JComponent createPanel(WizardController controller, String id, Map set
       final JComponent c;
       if (NAME_STEP.equals(id)) {
         c = getNameControls(controller, settings);
-
+        // Creating Name/Pw panel for the first time. Make sure the Finish and Next buttons are sett correctly for the game type.
```
jonathanrwatts, line 292: *set

uckelman:
I've tested this. It works. Amazing to have a fix for a bug first reported against 3.1.15.

PR #13103, opened 2024-01-25
13102 - Current Location Fast Match was not recognizing pieces in same location but not same X, Y position

BrentEaston:


Cattlesquat:
Confirmed this seems to fix the behavior in PoG -- tested with trenches for both sides & control markers for both sides.

Cattlesquat (approved):
Looks right to me

PR #13104, opened 2024-01-25
13097 - CountMap() by Property gives Bad Data error

BrentEaston:


BrentEaston:
Fixed confirmed working.

PR #13105, opened 2024-01-25
13080 - Vassal crashes when copying a folder of Deck GKCS between Decks

BrentEaston:


BrentEaston:
Closing - Symptom of more general problem. See #13106

PR #13121, opened 2024-02-02
13115 Update Game Piece Layers updated correctly by Set Piece Property

BrentEaston:


BrentEaston:
Reporter confirms change fixes issue.

PR #13163, opened 2024-02-18
13163 - Initialize optional Global Property counters whenever player commands initiated

Cattlesquat:
Whenever the player initiates a command (key press w/ pieces selected, right-click piece context menu,, or drag-drop-move), if the following global properties exist they are updated:

**ClearedOncePerPlayerCommand** - set to zero
**PiecesReceivingPlayerCommand** - the total number of pieces in the command or drag, the maximum number which could theoretically receive the command, not that all of them would necessarily do anything

In theory I could put a GlobalOption attribute that the module designer would need to explicitly turn on to enable this, but I'm not sure I actually judge that to be necessary.

Since the first property is cleared to zero at the beginning of every player command, module logic can safely update the value (e.g. incrementing it as they process the command and take action), allowing further pieces in the list to assess e.g. if a maximum limit of desired responding pieces has been reached.

Two example use cases:
(1) PoG has big stacks of corps in the reserve box. Right click on the stack to send *one* to (some location on the right click menu), rather than presently where it would send the whole stack so you have to manually expand the stack and make sure you just right click a single piece.
(2) For the People rearranges units arriving at a particular location on the map (eg by dragging them there) into multiple different piles (e.g. the troops in a pile, the generals in a pile) for easier readability. Counting up until the final boundary value prevents needing to do the expensive "rearrange all the pieces in this space" until all pieces arriving in the space from a multi-piece drag have arrived -- only the "last" piece needs to activate this logic from its apply-on-move key command processing.


BrentEaston (commented):
My initial thought is that it's a bit clunky to force module designers to create Global Properties with specific names. These should really be Vassal Properties that Vassal automatically exposes.

My next thought is that I don't think they even have to have to be 'real' Global Properties being manipulated by Commands, they can just be values manipulated by the local client. They are only meaningful to the local client, remote clients don't need to see these, they will just get the resulting commands that are generated. That should significantly simplify the design.

Instead of the 'reset once per player Command' variable. I would make it a counter that increments by 1 as the command is applied to each top-level selected piece. It;s value will range from 1 to the value of the other value (Number of pieces).

BrentEaston (commented):
I had to do something very similar for the UI refresh fix I did - Refreshing the UI after each User UI gesture. This happens in the same places you want to intercept. Let me revisit that and see if I can generalise it to make this easier to add.

BrentEaston:
Have a look at my PR #13174. I missed the case in my original PR of the Apply-after-move key on Drag n Dropped counters. This fixes that, but also adds a framework that records the start, each piece being processed, and end of multi-piece processing initiated by a UI gesture on the local client.

All it needs is for GameModule to recognize a couple of new properties and return getUiPieceCount() and getUiPieceIndex() and I think that will satisfy this PR for 3.8.

Cattlesquat:
Oooh yeah that looks pretty good, and I like your approach a lot better. I will close this one and comment more on your PR

PR #13173, opened 2024-02-21
13152 - Prevent NPE when cut/paste List Scenario Property

BrentEaston:


vassal-app/src/main/java/VASSAL/build/AbstractConfigurable.java
```diff
@@ -213,6 +213,10 @@ public Configurer getConfigurer() {
     }
     else {
       ((AutoConfigurer) config).reset();
+      // Some Configurers clear their configurer during a reset to ensure a new one is generated to reset any internal components
+      if (config == null) {
```
uckelman, line 217: I don't understand how `config` can be `null` here. If `((AutoConfigurer) config).reset()` didn't throw, then `config` has to be nonnull, because a method was just called on it.

BrentEaston: Because config is actually this.config and this.config.reset() does a this.config = null (for some configurers). So before line 215 executes, config is non-null, but after it executes, it is.

PR #13174, opened 2024-02-21
13174 - Update Labels on remote maps after DragnDrop

BrentEaston:
Adds missing case of DnD after move key to PR #13121

Also includes a framework for reporting on multi-piece UI gesture processing in 3.8 (Not enabled in this PR)

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -227,6 +227,9 @@ public class GameModule extends AbstractConfigurable
 
   public static final String IS_VISIBLE = "_isVisible"; //NON-NLS
 
+  public static final String UI_PIECE_COUNT = "UiPieceCount";
+  public static final String UI_PIECE_INDEX = "UiPieceIndex";
+
   private static final char COMMAND_SEPARATOR = KeyEvent.VK_ESCAPE;
 
   public static final String RECENT_GAMES = "RecentGames"; //NON-NLS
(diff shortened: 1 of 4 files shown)


Cattlesquat:
This seems like it should handle all the cases I was thinking, and in a better & more efficient way. I especially like that you handle the counting-up process internally.

If you're putting this much of it into 3.7.9 anyway, could you do me a favor and put the properties themselves into 3.7.9 (without *documenting* them until 3.8)? That way I could release my next PoG module more or less immediately, and the feature in question would "magically work better" for people with 3.7.9 but wouldn't be broken for people with earlier 3.7. Did this once before between 3.5 and 3.6 (or was it 3.4 and 3.5?) with a thing for Almoravid.

BrentEaston:
Waiting for @Cattlesquat to be back from Hols and have a play with this.

Cattlesquat:
Found a problem -- when I picked something from the right-click menu the first time, and filtered with (UiPieceIndex <= 1) to make sure I only sent the first piece on the stack, everything worked fine. BUT, when I then right-clicked again on the same stack, and issued the same command again, this time no pieces responded. Seemed like something maybe isn't getting reset by the mere act of right-clicking?

BrentEaston:
Hi Brian,

I believe I have reproduced the issue you are reporting, but I think it is actually working correctly.

How do you know that no pieces responded?

I reproduced it by having a stack of counters with Send To Location traits and a Trigger with the UiPieceIndex test. The first right-click sends the first piece to a new hex BUT it is still selected, so when you right-click again, that first piece is still one of the selected units and is a candidate to receive the right-click command again. 

So the piece that had already moved is responding the second time, which looks likes none responded.

Cattlesquat:
Oh gosh I think you're right - zoinks!!!

PR #13179, opened 2024-02-24
13153 - Prevent Area of Effect getting stuck ON when migrating to 3.7

BrentEaston:


BrentEaston:
This definitely fixes the reported problem using the reportees supplied module and log files. He was unable to reproduce the fixed behaviour, almost certainly because he ran the wrong test version. He has now gone off the air, so I am calling this one fixed and tested.

PR #13185, opened 2024-02-28
13176 - Mouse-over viewer to respect Show Unrotated for Cargo on rotated Mats

BrentEaston:


BrentEaston:
User confirms issue solved

PR #13194, opened 2024-03-07
13194 - MUST have matching module version and reasonably matching (same minor version) Vassal in order to join Multiplayer

Cattlesquat:
…ersion


riverwanderer:
I think that this PR ought to treat module version numbers that use the same convention as Vassal for minor version checks in the same manner as for Vassal itself. It's reasonable to assume that someone using that convention is also attempting to ensure reasonable backward compatibility for the module.

PR #13336, opened 2024-05-18
13336 - Correct issue when auto-attach and allow attachement to self features both enabled

### LINKED ISSUE #13289: Auto-attach does not attach to self when piece is added to a stack
I noticed there is a bug in the attachment trait when configured to auto-attach and attach to self.
image

The first piece that is created has the attachment to self as expected.
image

A second piece created in an empty hex creates the expected attachments.
image

However if the second piece is created in a hex occupied by another unit, the new piece fails to attach to self.
image

This is true whether the piece has the same attachment or not.
image

**Workaround:**

If the piece missing the self-attachment is moved to an empty hex, the attachments are updated correctly.
image

marcfor:
Fixes #13289 
When "auto-attach" and "allow attachment to self"
are both enabled and a new piece is dragged onto
an existing stack, the new piece fails to
attach to self.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/AttachmentManager.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/AttachmentManager.java
@@ -63,7 +63,7 @@ public class AttachmentManager {
    * A piece has been added to the game.
    * 1. If it has as Auto-attach, attach all other Attachments with same name to it.
    * 2. Tell all other Auto-attach traits with same name to attach to it.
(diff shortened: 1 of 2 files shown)


Cattlesquat:
Looks right to me.

Cattlesquat (approved):

uckelman:
Thanks for the contribution. It will be in the next release, and I'll add you to the contributors list.

PR #13396, opened 2024-06-18
13360 - Coerce Regular Expression operator arguments to be strings

BrentEaston:
Prevents Bad Data in Module errors if Regex happens to look like an Integer.


BrentEaston:
Tested by Mark and found to resolve problem

PR #13399, opened 2024-06-19
13398 - Ensure a GKC is never applied to same piece twice if it moves

BrentEaston:


vassal-app/src/main/java/VASSAL/counters/GlobalCommandVisitor.java
```diff
@@ -39,6 +41,9 @@ public class GlobalCommandVisitor implements DeckVisitor {
   private int selectFromDeck;
   private GlobalCommand globalCommand;
 
+  // Keep track of the ID's of all pieces processed by this GKC.
+  private final Map<String, GamePiece> seen = new HashMap<>();
```
uckelman, line 45: Shouldn't this be Set and HashSet if you don't care about the values?

BrentEaston: Indeed. Update pushed.

PR #13402, opened 2024-06-19
Add a visual validator for Prototype name in Prototype trait

BrentEaston:
Similar to the Beanshell expression validator. Highlights when you prototype name doesn't match one in the module.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/UsePrototype.java
+++ b/vassal-app/src/main/java/VASSAL/counters/UsePrototype.java
@@ -30,17 +30,24 @@ import VASSAL.tools.RecursionLimitException;
 import VASSAL.tools.RecursionLimiter;
 import VASSAL.tools.RecursionLimiter.Loopable;
 import VASSAL.tools.SequenceEncoder;
+import VASSAL.tools.icon.IconFactory;
+import VASSAL.tools.icon.IconFamily;
 import VASSAL.tools.menu.MenuScroller;
 
+import javax.swing.Icon;
+import javax.swing.ImageIcon;
 import javax.swing.JButton;
+import javax.swing.JLabel;
 import javax.swing.JMenu;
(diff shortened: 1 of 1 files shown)


Cattlesquat:
Since this validator isn't optionally turn-offable, I wonder if it will create any strange new warning message cascades for modules that are presently "silently working in spite of prototypes that point at nothing"? It's just one of those "I Wonder If" moments...

BrentEaston:
It's not a 'Validator' validator. It doesn't generate any messages and is completely inactive unless you are editing a prototype in a module.

vassal-app/src/main/java/VASSAL/counters/UsePrototype.java
```diff
@@ -323,6 +336,86 @@ public String getState() {
     public String getType() {
       return ID + nameConfig.getValueString();
     }
+
+    private String getValueString() {
+      return nameConfig.getValueString();
+    }
+
+    private class Validator extends JLabel {
```
uckelman, line 344: Should this be `private static`?

BrentEaston: Re-organized and re-coded slightly. Ready to go.

vassal-app/src/main/java/VASSAL/counters/UsePrototype.java
```diff
@@ -323,6 +336,88 @@ public String getState() {
...
+          dirty = true;
+        }
+        else {
+          validating = true;
+          setStatus(UNKNOWN);
+          SwingUtilities.invokeLater(validationThread);
+        }
+      }
+
+      class ValidationThread implements Runnable {
```
uckelman, line 402: Should be private?

BrentEaston: Actually, everything in there should be private. Updated

PR #13423, opened 2024-06-26
Add parameterized type to JComboBox

djberg96:
Just experimenting with CoPilot, this was one of its suggestions.

Also, added .vscode to the .gitignore file.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/FontStyleConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/FontStyleConfigurer.java
@@ -33,7 +33,7 @@ public class FontStyleConfigurer extends Configurer {
 
   protected JPanel p;
   protected JPanel fontPanel;
-  protected JComboBox fonts;
+  protected JComboBox<String> fonts;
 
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/gamepieceimage/FontStyleConfigurer.java
```diff
@@ -89,7 +89,7 @@ protected void buildFonts() {
       fontPanel.remove(fonts);
     }
 
-    fonts = new JComboBox<>();
+    final JComboBox<String> fonts = new JComboBox<>();
```
uckelman, line 92: This is wrong. You're replacing assignment to a member with declaration of a new local variable with a name shadowing the member.

djberg96: I thought the double declaration looked weird, but that's what the AI suggested. See, our jobs are safe.

uckelman: I'm not worried.

PR #13439, opened 2024-07-02
13315 - Preserve all attachments across save/load

BrentEaston:
During game load, earlier loaded pieces where quietly dropping attachments to pieces that had not yet been loaded. This fix records those attachments in the AttachmentManager as 'Pending Attachments'. At the end of the game load, prior to any startup GKC's running, the AttachmentManager is called to re-present all Pending Attachements to their owning Attachment traits to re-try the attachment.


BrentEaston:
Mark confirms this resolves the issue.

PR #13451, opened 2024-07-07
13450 - Make Locked Scenario Options more legible

BrentEaston:
Scenario Options Tabs and pages now have Help.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/StringEnumConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/configure/StringEnumConfigurer.java
@@ -16,10 +16,13 @@
  */
 package VASSAL.configure;
 
+import java.awt.Color;
 import java.awt.Component;
 import java.awt.Dimension;
+import java.awt.Graphics;
 
 import javax.swing.DefaultComboBoxModel;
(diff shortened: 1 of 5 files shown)


BrentEaston:
Simple change, users agree it solves issue.

vassal-app/src/main/java/VASSAL/configure/StringEnumConfigurer.java
```diff
@@ -49,6 +52,14 @@ public Component getControls() {
       panel = new ConfigurerPanel(getName(), "[]", "[]rel[]"); // NON-NLS
 
       box = new JComboBox<>(validValues);
+      // Ensure the text stays visible when the combobox is disabled
+      box.setRenderer(new DefaultListCellRenderer() {
+        @Override
+        public void paint(Graphics g) {
+          setForeground(Color.BLACK);
```
uckelman, line 59: Should this be set to whatever the "enabled" color is? It won't necessarily be black in all cases.

BrentEaston: Good catch. Improved version committed that respects the existing UI colors.

PR #13453, opened 2024-07-08
13452 - Scenario option improvements

BrentEaston:
FULLY upwardly compatible and does not cause any error running on earlier 3.7 versions. Please include with 3.7.13.

1. Allow individual options to have the Description and Value position reversed. (Options)
2. Allow left and right columns to be aligned left, centre or right. (Tabs)
3. provide for a 'header' label on each tab, which can use HTML to display arbitrary styled text.
4. Allow option box to expand when window expands, allowing text fields to stretch.


riverwanderer:
Test using my implementation of Scenario Options in the C&C Nap v5 module.
Basic compatibility test - Scenario Options in existing module appear to work as before.
New features - i was apply to edit the module and apply the new features.

Observation -
The Title field states "May contain HTML" but it is still necessary to prefix the text with a `<html>` tag - this is no different from many other fields within Vassal, where HTML is supported. That includes the tab name on Scenario Options that also allows html with the same proviso. Seems that the mention of HTML in field description is redundant, though maybe it can be mentioned in the Reference Manual (also perhaps in a general section if it is not already mentioned that many fields accept HTML in this way).

BrentEaston:
Tested by several users. Fully backward and forward compatible with all 3.,7 versions.

vassal-app/src/main/java/VASSAL/build/module/properties/ScenarioPropertiesOptionTab.java
```diff
@@ -643,4 +690,22 @@ public void setEnabled(boolean enabled) {
       getConfigurer().setEnabled(enabled);
     }
   }
+
+  public static class AlignConfig implements ConfigurerFactory {
```
uckelman, line 694: Should this be private?

vassal-app/src/main/java/VASSAL/build/module/properties/ScenarioPropertiesOptionTab.java
```diff
@@ -643,4 +690,22 @@ public void setEnabled(boolean enabled) {
...
+    @Override
+    public Configurer getConfigurer(AutoConfigurable c, String key, String name) {
+      return new TranslatingStringEnumConfigurer(key, name,
+        new String[] {LEFT, CENTER, RIGHT},
+        new String[] {"Editor.left", "Editor.center", "Editor.right"});
+    }
+  }
+
+  public static class TextConfig implements ConfigurerFactory {
```
uckelman, line 703: Also this, private?

vassal-app/src/main/java/VASSAL/build/module/properties/ScenarioPropertiesOptionTab.java
BrentEaston: Yes

vassal-app/src/main/java/VASSAL/build/module/properties/ScenarioPropertiesOptionTab.java
BrentEaston: And Yes. New version pushed.

PR #13470, opened 2024-07-15
13447 - Maintain all attachments when refreshing Pre-defined scenarios

BrentEaston:


BrentEaston:
Mark confirms the fix resolves the problem.

PR #13474, opened 2024-07-15
11884 - Ensure OldMat is set when a piece leaves a Mat

BrentEaston:


BrentEaston:
Confirmed by user that it resolves the issue

PR #13476, opened 2024-07-15
13429 - Prevent crash when editing module with name shorter than 3 characters

BrentEaston:


vassal-app/src/main/java/VASSAL/launch/EditorWindow.java
```diff
@@ -423,7 +423,7 @@ public void clearListKeyCommands() {
   boolean isTempFile(String name) {
     return name == null ||
       name.isEmpty() ||
-      ("tmp".equals(name.substring(0, 3)) && name.contains(".zip"));  //NON-NLS
```
uckelman, line 426: Is there some reason not to use `name.startsWith("tmp")` here?

BrentEaston: True. Updated.

PR #13479, opened 2024-07-16
13393 - Prevent slowdown when editing multiple Prototypes in sequence.

BrentEaston:
Turned out the number of Property Change Listeners attached to Prototype Definitions (each firing of a rebuild of the prototype cross-refererence) increased geometrically with the number of prototype definitions that are edited and have their name changed, finally melting down around the 15th edit.

Re-factored the cross-reference rebuild code to not create additional property change listeners.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/PrototypesContainer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/PrototypesContainer.java
@@ -106,7 +106,6 @@ public class PrototypesContainer extends AbstractConfigurable {
     def.addPropertyChangeListener(evt -> {
       if (NAME_PROPERTY.equals(evt.getPropertyName())) {
         // When a prototype is renamed we need to rebuild the prototype map, so that if there was a duplicate of the same name it will re-establish its presence
-        definitions.clear();
         rebuildPrototypeMap();
       }
     });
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/build/module/PrototypesContainer.java
```diff
@@ -95,26 +95,19 @@ public static String getConfigureTypeName() {
...
+    GameModule.getGameModule().getAllDescendantComponentsOf(PrototypeDefinition.class).forEach((def) -> definitions.put(def.getConfigureName(), def));
   }
 
-
   public void addDefinition(PrototypeDefinition def) {
     definitions.put(def.getConfigureName(), def);
     def.addPropertyChangeListener(evt -> {
       if (NAME_PROPERTY.equals(evt.getPropertyName())) {
         // When a prototype is renamed we need to rebuild the prototype map, so that if there was a duplicate of the same name it will re-establish its presence
         definitions.clear();
```
jonathanrwatts, line 109: Isn't this line now redundant (`definitions.clear()`) is now the first command of `rebuildPrototypeMap()`)?

BrentEaston: Indeed. Thanks. Update pushed.

PR #13480, opened 2024-07-16
13359 - Report I/O Errors during Predefined Setup refresh in Refresher log, not as a Vassal bug.

BrentEaston:


riverwanderer:
@BrentEaston  - I can't find my comment on this PR, so maybe I didn't save it. I think the cause of this issue is in _RefreshPredefinedSetupsDialog.java_  - which already catching the i/o error. Unfortunately, at  line 429 it does so by calling `ErrorDialog.bug(e);`.  Will it work to replace that with an equivalent to the chat log call at line 309 that you've inserted into refreshWithStatus and remove the try...catch from refreshWithStatus (lines 303-311).

The key point is that the Refresh errors summary will then also be updated.

BrentEaston:
Not quite, the GameRefresher log doesn't exist at that level. I will change this so that I log the error in the refresher log, then re-throw the exception and just ignore the exception at the next level up.

PR #13502, opened 2024-07-23
13408 - Ensure pieces are indexed after Layer Change

BrentEaston:


BrentEaston:
Mark has confirmed this fix resolves the issue.

PR #13507, opened 2024-07-26
13505 - Prevent GameRefresher moving Decks with duplicate names

BrentEaston:


vassal-app/src/main/java/VASSAL/build/module/GameRefresher.java
```diff
@@ -611,6 +595,49 @@ else if (!decksToDelete.isEmpty()) {
...
+          // If the drawPile is on a map that doesn't have its current owning board active, then we
+          // cannot match that drawPile.
+          if (pile.getMap().getBoardByName(pile.getOwningBoardName()) == null) {
+            continue;
+          }
+        }
+
+        if (first == null) {
+          first = pile;
+          if (deck.getMap().equals(pile.getMap())) {
```
jonathanrwatts, line 631: Am I misunderstanding something here, or should this `if` statement be outside of the `(first == null)` check, instead of within it?  Is it not possible that you would find a match on the same map *after* finding a match on an earlier map?  As written, it looks to me like it stops checking for a matching map once it's found any matching name.

BrentEaston: Thanks Jonathon, my cut and paste hatchet job got out of hand. That code makes no sense at all. I have committed a new version.

PR #13508, opened 2024-07-27
13497 - Report HTML load errors as Bad Module Data, not Vassal bugs

BrentEaston:


vassal-app/src/main/java/VASSAL/build/widget/HtmlChart.java
```diff
@@ -87,7 +87,18 @@ private void setText(String text) {
   }
 
   private void setFile(String fname) {
-    setText(getFile(fname));
+    try {
+      setText(getFile(fname));
+    }
+    catch (RuntimeException e) {
```
uckelman, line 93: Can we be more specific about the exception type? There are a lot of things which are `RuntimeException` which could be bugs.

uckelman: Oh argh, I see from the bug report that it really is a `RuntimeException`, so this is the best we can do.

PR #13679, opened 2024-11-07
Add autosave functionality to BasicLogger

geezer09:
This pull request introduces an autosave feature to the `BasicLogger` class in the Vassal application. The most important changes include the addition of a scheduled executor service to handle periodic autosaving, the implementation of the autosave logic, and modifications to the `write` method to support writing to a specified file. We simply append the word autosave_ to the current file we are logging to and then save to that file every 5 minutes.

### Autosave Feature Implementation:

* Added `ScheduledExecutorService` to `BasicLogger` for scheduling periodic tasks. [[1]](diffhunk://#diff-767dfe244177a3b3bd7c0b28f6ded74c404f88c40087dc6dcadef5668554af5dR65-R67) [[2]](diffhunk://#diff-767dfe244177a3b3bd7c0b28f6ded74c404f88c40087dc6dcadef5668554af5dR111)
* Implemented `startAutosave` method to schedule the `autosave` method at fixed intervals.
* Added `autosave` method to handle the logic of saving the log file periodically.

### Modifications to Writing Logic:

* Modified `write` method to support writing to a specified file, used by the autosave feature.

### Localization:

* Added a new localization string for autosave log message in `VASSAL.properties`.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
@@ -122,7 +122,7 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
   }
 
   private void startAutosave() {
-    scheduler.scheduleAtFixedRate(this::autosave, 5, 5, TimeUnit.SECONDS);
+    scheduler.scheduleAtFixedRate(this::autosave, 1, 5, TimeUnit.MINUTES);
   }
 
   private void autosave() {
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -114,6 +118,22 @@ public BasicLogger() {
     newLogAction.setEnabled(false);
     logInput = new ArrayList<>();
     logOutput = new ArrayList<>();
+    startAutosave();
+  }
+
+  private void startAutosave() {
+    scheduler.scheduleAtFixedRate(this::autosave, 5, 5, TimeUnit.SECONDS);
```
uckelman, line 125: Five seconds is probably too short.

There's nothing to debounce the autosave if autosaving takes longer than the autosave interval.

geezer09: I changed that to start after a minute and save every 5 minutes.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -435,28 +455,28 @@ else if (result == 2) { // Turn Preference Off
...
         }
         metadata.save(zw);
       }
-
-      GameModule.getGameModule().getGameState().setModified(false);
-      undoAction.setEnabled(false);
-      ModuleManagerUpdateHelper.sendGameUpdate(outputFile);
     }
-
-    endLogAction.setEnabled(false);
```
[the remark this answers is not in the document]
geezer09: That was a mistake, I am now keeping the original write() function as it was and creating a new write(File file) function so that the autosaver can save to a different file and not trigger the lines below when saving.

 GameModule.getGameModule().getGameState().setModified(false);
 undoAction.setEnabled(false);
 ModuleManagerUpdateHelper.sendGameUpdate(outputFile);
 endLogAction.setEnabled(false);

PR #13686, opened 2024-11-11
Log autosave

geezer09:
This is the same PR as #13679 but I changed the name of the branch because I suspect it was causing issues when attempting to build because of the slash in my branch name.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
@@ -122,7 +122,7 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
   }
 
   private void startAutosave() {
-    scheduler.scheduleAtFixedRate(this::autosave, 5, 5, TimeUnit.SECONDS);
+    scheduler.scheduleAtFixedRate(this::autosave, 1, 5, TimeUnit.MINUTES);
   }
 
   private void autosave() {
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -114,6 +118,23 @@ public BasicLogger() {
     newLogAction.setEnabled(false);
     logInput = new ArrayList<>();
     logOutput = new ArrayList<>();
+    startAutosave();
+  }
+
+  private void startAutosave() {
```
cholmcc: Should be `protected` to allow override in custom code.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -114,6 +118,23 @@ public BasicLogger() {
...
+    scheduler.scheduleAtFixedRate(this::autosave, 1, 5, TimeUnit.MINUTES);
+  }
+
+  private void autosave() {
```
cholmcc: Should be protected to allow override in custom code.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -459,6 +480,27 @@ public void write() throws IOException {
     endLogAction.setEnabled(false);
   }
 
+  /**
+   * Write the logfile to a specified file. This is used by the autosave feature.
+   */
+  private void write(File file) throws IOException {
```
cholmcc: Should be protected to allow override in custom code.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -114,6 +118,23 @@ public BasicLogger() {
...
     logOutput = new ArrayList<>();
+    startAutosave();
+  }
+
+  private void startAutosave() {
+    scheduler.scheduleAtFixedRate(this::autosave, 1, 5, TimeUnit.MINUTES);
+  }
+
+  private void autosave() {
+    if (isLogging() && !isReplaying()) {
```
cholmcc: User should have the option to turn _off_ auto saving.   In some modules and plays, the auto save could take a long time, and users could prefer not to enable that but still do logging.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -114,6 +118,23 @@ public BasicLogger() {
     newLogAction.setEnabled(false);
     logInput = new ArrayList<>();
     logOutput = new ArrayList<>();
+    startAutosave();
+  }
+
+  private void startAutosave() {
+    scheduler.scheduleAtFixedRate(this::autosave, 1, 5, TimeUnit.MINUTES);
```
cholmcc: The rate of auto-saving should be configurable via user preferences.  That probably also means that starting the auto-save timer should be deferred until logging is actually started, so that the timer can take into account the user preference of how frequent the save is done.

vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -114,6 +118,23 @@ public BasicLogger() {
...
+  }
+
+  private void autosave() {
+    if (isLogging() && !isReplaying()) {
+      final File autosaveFile = new File(outputFile.getParent(), "autosave_" + outputFile.getName());  //$NON-NLS-1$
```
cholmcc: Perhaps a time-stamp should be appended to the auto-save file.  Of course, that would mean that one will get multiple files in a single run (which can take up quite a lot of disk space),  so maybe it should be a user configurable setting.

PR #13698, opened 2024-11-15
allow player to modify popup during game

geezer09:
I have implemented a function to allow the player to modify the popup menu during gameplay to deal with the problem of some menus being way too long. Basically, I created an "unwanted" menu and the player can move items to that submenu during gameplay to reduce clutter and customize as he likes. The code is currently functional but needs more work, I was just wondering if this is a feature that any other game actually needs. 

If not then I will implement in the VASL code which is probably doable if not ideal. What would be the best way to remove the following element added by the Map class

    addChild(new MenuDisplayer());

within the ASLMap class so that I can add my own ASLMenuDisplayer()?

Screenshot 2024-11-15 000842

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
@@ -35,27 +35,21 @@ import VASSAL.tools.NamedKeyManager;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 
-import javax.swing.Action;
-import javax.swing.JMenu;
-import javax.swing.JMenuItem;
-import javax.swing.JPopupMenu;
-import javax.swing.KeyStroke;
+import javax.swing.*;
 import javax.swing.event.PopupMenuEvent;
 import javax.swing.event.PopupMenuListener;
-import java.awt.Font;
-import java.awt.Point;
-import java.awt.Shape;
+import java.awt.*;
 import java.awt.event.ActionEvent;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
+import java.util.*;
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class MenuDisplayer extends MouseAdapter implements Buildable {
   @Deprecated(since = "2022-08-08", forRemoval = true)
   public static final Font POPUP_MENU_FONT = new Font(Font.DIALOG, Font.PLAIN, 10);
+  private static final java.util.Map<String, Boolean> keyCommandVisibilityMap = new ConcurrentHashMap<>();
 
   protected Map map;
   protected PieceFinder targetSelector;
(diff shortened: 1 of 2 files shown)


BrentEaston:
I think the concept of customizable menus has merit. 

Does this work at the counter level or the prototype level? I'm assuming at the counter level since prototypes have all been expanded at the point you working. Doesn't this mean you have to make the same changes to every single counter you touch? Sounds a bit painful?

> addChild(new MenuDisplayer());

To undo this, you will have to iterate through the children until you find an instance of MenuDisplayer and remove it. We could also modifiy the source in 3,.7,16 to something like this:

addChild(gerMenuDisplayer())

to make it easier to over-ride.

riverwanderer:
About any other module needing this… yes and no from me, I have modules with lots of menu items but I design in the nested menus in what I hope will be a useful structure, or at least it will be a consistent experience. I guess VASL demands a more personal approach or the existing menus can’t readily be changed?

A question on the PR, is the customised menu a local view or does it get replicated into the log?

geezer09:
I implemented successfully a few months ago in VASL, the operation does not have to be repeated for every counter and the state is saved so it keeps even after closing and reopening the module.

If you guys are interested I can show you the code or you can just look at the VASL repo.

vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
```diff
@@ -122,118 +116,144 @@ protected static JMenuItem makeMenuItem(KeyCommand keyCommand) {
...
+    public static JPopupMenu createPopup(GamePiece target, boolean global) {
+      final JPopupMenu popup = new JPopupMenu();
+      popup.putClientProperty("gamePiece", target);
+      final KeyCommand[] c = (KeyCommand[]) target.getProperty(Properties.KEY_COMMANDS);
+      if (c != null) {
+        final List<JMenuItem> commands = new ArrayList<>();
+        final List<KeyStroke> strokes = new ArrayList<>();
+        final java.util.Map<KeyCommandSubMenu, JMenu> subMenus = new java.util.HashMap<>();
+        final java.util.Map<String, List<JMenuItem>> commandNames = new java.util.HashMap<>();
+        final JMenu unwantedMenu = new JMenu("Unwanted");
```
riverwanderer, line 185: "Unwanted" - should be translatable with an entry in Properties. Better still, user-configurable ; e.g. by a checkbox on the submenu trait that makes the submenu a possible target for user customisation (complex?), which would free the feature from depending on a hard-coded submenu name.

vassal-app/src/main/java/VASSAL/build/module/map/MenuDisplayer.java
```diff
@@ -111,7 +105,7 @@ protected static String getMenuText(KeyCommand keyCommand) {
   // This both eliminates duplicate code AND makes this critical menu-building functionality able to "play well with others".
   // Menu text & behavior can now be custom-classed without needing to override the monster that is MenuDisplayer#createPopup.
   protected static JMenuItem makeMenuItem(KeyCommand keyCommand) {
-    final JMenuItem item = new JMenuItem(keyCommand.isMenuSeparator() ? MenuSeparator.SEPARATOR_NAME : getMenuText(keyCommand));
+    final CustomMenuItem item = new CustomMenuItem(keyCommand.isMenuSeparator() ? MenuSeparator.SEPARATOR_NAME : getMenuText(keyCommand), keyCommand);
```
riverwanderer, line 108: From my limited knowledge of MenuDisplayer etc, I read this PR as effecting a Custom Menu that, set by any player, will be visible to all other players. Is that desirable as a generic feature?

riverwanderer (changes requested):

PR #13710, opened 2024-11-21
Improvement to BeanShell Sleep()

riverwanderer:
A further change building on #13707 makes BeanShell Sleep() less obtrusive by disabling the OS decoration on Sleep's JDialog box.

This PR also corrects some minor doc errors in parameter headings.


riverwanderer:
This PR has tested OK on MacOS & Windows in a full game using CCN module v5.1.0beta9, which makes extensive use of Sleep(0).  In this test no adverse affects were noted and the other user perceived improved performance.

riverwanderer:
Retry on 3.7.

Issue #13665, opened 2024-10-30
New functionality in gamepiece editor layer trait 


geezer09:
I think it should be possible to have a layer that is not always active but starts out as active. Right now the only options (as far as I understand) is to have a layer be always active or off when initialized. I would like to add a checkbox in the gamepiece editor below the always active? checkbox  called start active? 

This checkbox will be disabled when the always active is enabled but displayed when not and will give the option of creating a layer that is not always active yet starts the game active. Is there an easier way of acheiving the desired outcome?

jonathanrwatts:
For pieces in At-Start Stacks (i.e., starting on the board), you can use a Startup Global Key Command to trigger the layer activation at the start of the game.  For pieces added from palettes during the game, you can use the "send keystroke to pieces ending movement on this map" option of the map board(s) to activate the layer when they are placed (you can check if `oldBoard` is blank to determine that the piece was just newly placed, IIRC).

geezer09:
That solution will not always work. For example in my particular case i have two layers i need to be able to alternate, therefore they both cannot always be active. If neither is set to “always active” then both layers are off and the piece is invisible. One of the layers has to be active when it is in the palette.

jonathanrwatts:
I guess I don't understand your use case.  If, as you say, the layers need to alternate and one needs to be in the palette, then it seems like the perfect use case for a single, always on, layer instead of 2 separate layers.

PR #13773, opened 2025-01-12
Minor changes to Hideable.java

geezer09:
The purpose of this PR is to fix a bug in VASL counters where a hidden counter will display all of its layers. This is a problem with most half-squads and other random counters that contain more that one visible trait.

My solution attempts to fix this by first drawing the counter off-screen with no transparency, and then apply transparency to the entire counter (as opposed to each of the layers) before drawing the piece in it's final location.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Hideable.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Hideable.java
@@ -42,6 +42,7 @@ import java.awt.Graphics2D;
 import java.awt.Rectangle;
 import java.awt.Shape;
 import java.awt.geom.AffineTransform;
+import java.awt.image.BufferedImage;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -210,14 +211,41 @@ public class Hideable extends Decorator implements TranslatablePiece {
         g2d.fill(t.createTransformedShape(piece.getShape()));
       }
 
-      final Composite oldComposite = g2d.getComposite();
-      g2d.setComposite(
-        AlphaComposite.getInstance(AlphaComposite.SRC_OVER, transparency)
-      );
-      piece.draw(g, x, y, obs, zoom);
-      g2d.setComposite(oldComposite);
+      // 1) Determine piece bounds at current zoom
+      final Rectangle bounds = piece.getShape().getBounds();
+      final int w = (int) Math.ceil(bounds.width * zoom);
+      final int h = (int) Math.ceil(bounds.height * zoom);
+
+      // If there's nothing visible at this zoom, skip
+      if (w <= 0 || h <= 0) {
+        return;
+      }
+
+      // 2) Create offscreen image and draw piece fully opaque
+      final BufferedImage offscreen = new BufferedImage(w, h, BufferedImage.TYPE_INT_ARGB);
+      final Graphics2D offG = offscreen.createGraphics();
+      try {
+        offG.scale(zoom, zoom);
+        offG.translate(-bounds.x, -bounds.y);
+
+        piece.draw(offG, 0, 0, obs, 1.0); // Fully opaque
+      }
+      finally {
+        offG.dispose();
+      }
+
+      // 3) Draw offscreen image with overall transparency
+      final Composite oldComp = g2d.getComposite();
+      g2d.setComposite(AlphaComposite.getInstance(AlphaComposite.SRC_OVER, transparency));
+
+      final int drawX = x + (int) Math.round(bounds.x * zoom);
+      final int drawY = y + (int) Math.round(bounds.y * zoom);
+      g2d.drawImage(offscreen, drawX, drawY, obs);
+
+      g2d.setComposite(oldComp);
     }
     else {
+      // Normal draw at the end
       piece.draw(g, x, y, obs, zoom);
     }
   }


vassal-app/src/main/java/VASSAL/counters/Hideable.java
```diff
@@ -210,14 +211,41 @@ public void draw(Graphics g, int x, int y, Component obs, double zoom) {
...
+      final Graphics2D offG = offscreen.createGraphics();
+      try {
+        offG.scale(zoom, zoom);
+        offG.translate(-bounds.x, -bounds.y);
+
+        piece.draw(offG, 0, 0, obs, 1.0); // Fully opaque
+      }
+      finally {
+        offG.dispose();
+      }
```
uckelman, line 235: This is going to cause a `BufferedImage` to be created for each invisible-to-others `Hideable` on each piece on each repaint. I am concerned about the performance impact of this.

uckelman: The `BufferedImage` you're creating here is also not guaranteed to be the preferred image type for the graphics device it will be painted to.

uckelman:
> The purpose of this PR is to fix a bug in VASL counters where a hidden counter will display all of its layers.

Why do hidden counters display all of their layers? What's the root cause of the bug?

geezer09:
It is not only hidden counters that display all their layers, all counters do it. Normally this is not apparent because the top layer hides all the others.

However it is noticeable with counters that have layers with reduced opacity (ie hidden counters) where the bottom layers can bleed through because of this reduced opacity. 

This is not really a bug in the sense that the program is working as intended, changing the program so that only the top layer gets drawn would require a lot more code and might not have the desired effect. There are probably counters that rely on this to layer images.

vassal-app/src/main/java/VASSAL/counters/Hideable.java
uckelman: You can use `VASSAL.tools.image.ImageUtils.createCompatibleTranslucentImage()` to make a `BufferedImage` with an alpha channel of the right image type for the frame buffer. To get the image into the image cache so that it's not recreated on each draw might require using `GamePieceOp`.

But I'm not convinced this is the right approach to begin with.

jonathanrwatts:
> However it is noticeable with counters that have layers with reduced opacity (ie hidden counters) where the bottom layers can bleed through because of this reduced opacity.

This statement has me confused.  Are you saying that some `Layer` traits get drawn even when the `Invisible` trait is used?  Is this all `Layer` traits, or only those that are below the `Invisible` trait in the stack?  If the former, that sounds like an actual bug.  If the latter, that is VASSAL working as designed, and I'm pretty sure some modules *rely* on that behavior to work properly (to hide upper layers that have private information without hiding the entire piece).

geezer09:
> > However it is noticeable with counters that have layers with reduced opacity (ie hidden counters) where the bottom layers can bleed through because of this reduced opacity.
> 
> This statement has me confused. Are you saying that some `Layer` traits get drawn even when the `Invisible` trait is used? Is this all `Layer` traits, or only those that are below the `Invisible` trait in the stack? If the former, that sounds like an actual bug. If the latter, that is VASSAL working as designed, and I'm pretty sure some modules _rely_ on that behavior to work properly (to hide upper layers that have private information without hiding the entire piece).

This issue might exclusively happen in vasl with its somewhat complicated counters. Here is a link to a discussion about it

http://www.gamesquad.com/forums/index.php?threads/problem-07-jan-2025-hip-us-halfsquads-show-full-squad-outline.201850/

My understanding os that all layers beneather the invisible layer get drawn with reduced opacity, allowing them to bleed through.

I’ll probably just attempt to implement a workaround in VASL although this is clearly a vassal issue.

uckelman:
Maybe someone could make a minimal test module that demonstrates the problem. If the problem really is in Vassal, it should be possible to show it with a single piece.

vassal-app/src/main/java/VASSAL/counters/Hideable.java
Cattlesquat: Would it make more sense to keep the BufferedImage around ("cache it") after it gets created the first time, so that it isn't getting re-generated every frame? So the first time this draw path comes up it makes the BufferedImage and then after that it just keeps it around?

jonathanrwatts:
> My understanding os that all layers beneather the invisible layer get drawn with reduced opacity, allowing them to bleed through.

This statement also makes no sense to me.  Surely the layers *must* be drawn from the bottom up, not from the top down.  If a lower layer had "reduced opacity", that would cause layers below it to bleed through, not make it bleed through to layers above it!

BrentEaston:
Has this issue been observed in ANY other module? I am almost 100% certain that the described behaviour is caused by bugs in the VASL custom code which over-rides many Vassal traits. My suspicion would be that one of the other VASL modified traits is doing something it shouldn't be.

I don't see how this problem could possibly exist in base Vassal without impacting multiple other modules.

geezer09:
No this is 100% a vassal issue. The thing is very few modules (if any other than VASL) use counters that have 2 layers with images and use the invisible trait as well. I was able to easily reproduce this issue by opening up a module and adding a a layer with an image and the invisible trait as well. I used BCS Arracourt but any module will do, you can see the bug from the piece editor just by doing as I said and then setting the piece to invisible.

BrentEaston:
@geezer09 Overlapping updates Joseph :)

geezer09:
@BrentEaston, i must not really understand the solution you are proposing because it sounds like exactly what I have done in my code. I draw the counter offscreen and then apply the transparency to the overall image instead of applying to each individual layer in the counter.

Note that this solution would not hide the base layer if it is normally seen. In fact it is the more “natural” solution in that it draws the counter as the player sees it, and then applies the transparency to that overall image. In other words the counter will look exactly as it did before being set to invisible, except with some transparency. 

Obviously this was the intent when the invisible trait was first implemented (i presume) but this edge case of have 2 visible layers was not considered.

BrentEaston:
Gack,

Sorry Joseph, I read offscreen as offmap and spent too long reading the comments and not long enough reading your code. Mea Culpa.

Yes, I believe your solution is the correct approach.

I think we just need to add a couple of things to meet Joel's concerns.

As he mentioned, use `VASSAL.tools.image.ImageUtils.createCompatibleTranslucentImage()` rather than BufferedImage.createImage(). Vassal will work out the correct image type for you. Theoretically, the image type may be different on different devices..

I agree with Brian it is worth doing some caching here, it should be fairly straight forward. While GamePieceOp probably provides this, no-one understands how it works except Joel, and we need him focused on other things. 

I would just cache the image in Hideable and re-generate it if zoom or piece.getProperty(Properties.VISIBLE_STATE) changes value.

Cheers,
Brent.

Cattlesquat:
There are lotsssss of vassal pieces with more than one active “layer” (aka embellishment) at a time. Very common use case.

geezer09:
The short answer is because it is made difficult to implement a counter with two layers that alternate. I opened up an issue about this a while back #13665.

I don’t understand why, but the only two options for a layer are to be always active or start the game as off. So the easiest way to implement a counter that alternates between two layers was to have a layer always on and another layer on top of it that starts deactivated. As long as there is no transparency, you can then alternate between the two layers by turning the top later on and off, but the bottom one remains always active.

The problem with looking for a solution in that direction is that even if changes are made to the piece editor, it would involve going through every counter in VASL to implement the fix which is a pretty hefty task and error prone.

PR #13793, opened 2025-01-25
13791 - Prevent Checkbox Scenario Option switch description side option being forced on.

BrentEaston:


BrentEaston:
User confirms fix solves issue.

PR #13801, opened 2025-01-28

riverwanderer:
PR #13720 rebased onto v3.7.


riverwanderer:
Incorrect target.

PR #13802, opened 2025-01-28

riverwanderer:
PR #13720 rebased onto v3.7.


riverwanderer:
Updated demo file module: 
[PlaceMarkerStack Test.zip](https://github.com/user-attachments/files/18580026/PlaceMarkerStack.Test.zip)

@BrentEaston - versioning seems to be working:
Existing piece (no refresh) will not stack away (pre 3.7.16 behaviour).
New piece or after a refresh - pieces do stack away when offset is used.

BrentEaston (approved):

PR #13804, opened 2025-01-28
Sleep() performance improvement

riverwanderer:
#13710 rebuilt on 3.7.

Testing indicates a performance improvement where Sleep(0) is used to force a screen refresh, at least on a Mac where the 3.7.15 will display a dummy window on the physical screen. This artefact is removed by the revised approach in this PR. The improvement is particularly noticeable on a 2016 MacBook with the C&C Napoleonics module v5.1.


riverwanderer:
Superceded by #13845.

PR #13833, opened 2025-02-10
Do not subject Continuation file to online restrictions.

riverwanderer:
Skips online check if we're only doing a Load Continuation. New Game & loading a Log file remain barred to room guests.


riverwanderer:
User confirmed that this fixes his group's issue.

BrentEaston (approved):

PR #13835, opened 2025-02-10
Disable Toolbar whilst a file Dialog is active

riverwanderer:
Intended to prevent menu shortcuts from intercepting shortcuts within the File Dialog boxes.


riverwanderer:
The toolbar is not disabled by commit 
[2296b8c](https://github.com/vassalengine/vassal/pull/13835/commits/2296b8cf3edb778904969250ae5dc3f2f24d2b9c). Initial tests appeared successful because - so it seems - Delete key works in the file dialog **until** the Edit menu is clicked on or (perhaps) a Delete is pressed when the main window is in focus. Thereafter delete is trapped by the Toolbar, causing the problem.

riverwanderer:
Abandoning this attempt in favour of a more basic approach in #13855.

PR #13850, opened 2025-02-18
Editor Search - play nice when a Regex pattern match exception has occurred

riverwanderer:
Fixes the exception that occurs after a Regex Pattern match error has been trapped. The subsequent null pattern needs to be checked for, bypassing the rest of the search attempt. 

Also removes deprecated searchAdvanced.


riverwanderer:
After continued use in my own editing, this change seems stable and ready to merge.

BrentEaston (approved):

PR #13716, opened 2024-11-25
Workaround to height zoom issue.

riverwanderer:
Repeating set height action works around the issue.

PR #13854, opened 2025-02-19
Workaround to height zoom issue

riverwanderer:
Fixing to Zoom to Height on first user attempt. Re-done against 3.7 branch.

Clean build on 3.7 to supercede #13716.


BrentEaston (approved):
Happy with this. Windows does not seem to be affected and solution for Mac has no impact on Windows users. Rarely used feature, no performance implications.

PR #13856, opened 2025-02-21
13856 Multi-Location Command includes the option to sort the order of the commands generated

Cattlesquat:
Would replace #13168 with a 3.7 version -- basically this is just a polish item


uckelman:
Merged #13168 for 3.8.0 instead.

Issue #12734, opened 2023-09-21
Reference manual anchors not working on Help buttons


riverwanderer:
References to anchors in the Reference Manual do not seem to work. The links always position at the head of the page.

e.g.  Help button on Definition of Player Sides.

riverwanderer:
Possibly a known issue with the Java URL() function that anchor points are not actioned on local files.

PR #13865, opened 2025-02-24
Fix a bad reference manual link

riverwanderer:
Startup GKC help link needs to go via GlobalKeyCommands.html not via Map.html.


riverwanderer:
More work needed - help directs to the top of the correct page. It should go direct to the SKGC sub-section.

riverwanderer:
Positioning at the top of the section is the best that I can do. Anchors don't work despite the apparent support in the HelpFile() methods. Previous attempts to address the issue have not succeeded. Ref issue #12734.

BrentEaston:
Thanks Mark. I have also tries and failed to get the anchors working. It just isn't supported in the underlying library.

BrentEaston (approved):

PR #13872, opened 2025-02-27
13871 - Show layers in hidden pieces (to owner) correctly.

BrentEaston:


geezer09:
I've done a quick test using VASL module and this seems to be working fine. No notable performance issues with 100+ hidden counters.

Issue #12480, opened 2023-07-20
Pieces drop through to window beneath


riverwanderer:
Observed in Vassal since v3.2.17 (at least). Demo on MacOS but occurs on other platforms, I believe.

May relate to #9517 but has different symptoms from the ones described in the associated replies on the original forum [post](https://forum.vassalengine.org/t/query-on-pieces-dropping-through-to-window-beneath/11229/12).

The attached video demonstrates the issue as experienced in the case of dragging a card from a deck into a private window. The fault is seen in any of the Commands & Colors modules. To replicate the video, use the Commands & Colours: Medieval module. Open a scenario as Solitaire. Open a hand window and position it over the board, as shown in the video. Now drag cards from the draw deck into the window, attempting similar drop positions to those where the fault is trigger in the video.

_To a point in the forum thread - this fault can be reproduced regardless of whether the top left hand corner of the private window is over a piece or not._  

Fault as generally experienced by C&C module users (card draws 2 & 5 show the fault occurring):
https://github.com/vassalengine/vassal/assets/58135975/e8541cdd-2348-447d-9e4f-8bf5fba0b293

Further demo to follow.

riverwanderer:
A minimal demo module for Vassal v3.7: [CCNdragBugDemo.zip](https://github.com/vassalengine/vassal/files/12110321/CCNdragBugDemo.zip)
To start, open "Drag & Drop Bug" from the File menu and follow the instructions in the Chat log.

The demo manifests the issue in two modes, depending on the role you choose: "`<observer>`" or "`Player`", shown in the attached videos.

1. Video of `<observer>` mode: 
https://github.com/vassalengine/vassal/assets/58135975/a48c54db-86fc-418c-a6c1-cab5c8b562bd

2. Video of `Player` mode: 
https://github.com/vassalengine/vassal/assets/58135975/5a7e878b-4a20-44a3-a5c1-a634d6c9993b

uckelman:
Also reported here: https://forum.vassalengine.org/t/counters-disappearing-on-linux/77662

riverwanderer:
Examining the manifestation of this issue in Commands & Colors: Napoleonics within PieceMove.java, I observe two particular cases that occur when a card is dragged from a deck on the Main Map to an overlying Player Hand window ("Allies Hand").

In the first form, the underlying map ("Main Map") is pointed to by "map" and that is where the piece ends up.

[...]

PR #13875, opened 2025-03-01
No clicks, no drop!

riverwanderer:
Ensures zero getClickCount() is ignored for mouseReleased(). Investigation found that a zero getClickCount() is consistently associated with issue #12480 - the first such event being zero in a "double" event.

See this [post](https://forum.vassalengine.org/t/drag-drop-bug-piece-drops-through-to-an-underlying-window-advice-needed/84754/4?u=marktb1961) for evidence and some discussion of the issue.  Note that the latest evidence belies that the root cause is overlying windows, since a double event may be observed even when only one window is on screen.

The question of why a zero-click event is sometimes generated remains open. However, initial testing in modules where the fault is easily replicated demonstrates a successful workaround even if this is that case.


BrentEaston:
This looks like a good solution to me. It seems to be a known issue that rapid clicking can generate zero click events.

PR #13883, opened 2025-03-02
Checkpoint feature

### LINKED ISSUE #13726: Module-accessible undo checkpoint
Create a new hotkey driven component, essentially a simplified Action button or an option on the Action Button, that will disable UnDo beyond the point at which that action is taken. This feature would be of use in module design at certain checkpoints where accidental undos irrevocably compromise the gameflow - typically, when a card is drawn.

A code fragment that might achieve this (taken from GameRefresher.java is:
```
 {
        final BasicLogger log = GameModule.getGameModule().getBasicLogger();
        if (log != null) {
          log.blockUndo(1);
        }
```

riverwanderer:
Global Option to set a hotkey that will block undo, using the mechanism that Vassal already uses to block undo of Startup GKCs and in one or two other places.


riverwanderer:
Works from an Action Button (via a keyboard hotkey or toolbar) but not it seems in a chain of commands (whether Action Button or traits).

PR #13904, opened 2025-03-19
Add Startup GKC for when a player joins, retires, or changes sides V2

### LINKED ISSUE #13903: Startup GKC that fires when player joins or changes sides V2
The most recent addition to the Startup GKC added the firing of a GKC at the start of a fresh game and when a player joins or retires. Combining these two distinct events (game start and player join/retire) into a single GKC masks the initial event(s) of a player joining via the startup wizard. As it should, the existing codebase fires only one Startup GKC at the start of a fresh game regardless of the number of players joining in.

There are cases where acting on a player joining/retiring can be beneficial. For example, modifying the button text with the player's name for a 'Player Hand' window.

In consideration of backward compatibility, this proposal is to add a new entry to the Startup GKC list that fires only on player join, retire or change sides. Hence that same issue name with the appended "V2".

Reference to forum discussion: [Startup GKC not working on Player Join](https://forum.vassalengine.org/t/startup-gkc-not-working-on-player-join/84799)

rlament:
This is different from the existing Startup GKC in that it does not fire on start of fresh game. This GKC fires as players join at startup via the wizard and anytime during the game when any player(s) retires, joins or changes side. Closes #13903

I neglected to add a reference to the original PR for Startup GKC on fresh game or player join/retire #11918

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/PlayerRoster.java
@@ -506,6 +506,7 @@ public class PlayerRoster extends AbstractToolbarItem implements CommandEncoder,
 
       final Add a = new Add(this, GameModule.getActiveUserId(), GlobalOptions.getInstance().getPlayerId(), newSide);
       a.execute();
+      fireSideChange(null, newSide);
       gm.sendAndLog(a);
 
       pickedSide = true;
--- a/vassal-app/src/main/java/VASSAL/build/module/StartupGlobalKeyCommand.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/StartupGlobalKeyCommand.java
@@ -32,6 +32,7 @@ import VASSAL.i18n.Resources;
 import VASSAL.tools.NamedKeyStroke;
 import VASSAL.tools.SequenceEncoder;
 import VASSAL.tools.UniqueIdManager;
+import org.apache.commons.lang3.StringUtils;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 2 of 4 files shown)


rlament:
Good points. I had to rework things particularly for point 3.

> Does this seem to work when you test it in multiplayer and in log-based play?

Yes, I tested both scenarios. I created a test module that reports on all 5 Startup GKCs. Tested modules with and without sides. When there are no sides, this Startup GKC never fires. The latest change adds a `fireSideChange` event to the finish step of the wizard. I looked at all the listeners and didn't spot any nasty side effects.

> (1) At beginning of the game (does it send one GKC per player? Does it correctly do this AFTER sending any for the game itself being started?)

Yes, one GKC per player at the beginning. It can be before or after any other startup GKC. That depends on the relative placement within the module editor. Startup GKCs are fired in order from top down, when the corresponding condition is satisfied.

> (2) When someone joins in multiplayer is it sending the GKC? 

Yes

>  (3) When someone is added into a log-based game is it sending the GKC?

Yes, it sends the GKC and logs it if logging is active.

> (4) And one different type of question -- do you really WANT the GKC when someone joins as "observer"...

This GKC is modelled after APPLY_START_GAME_OR_SIDE_CHANGE. The exception being one GKC per player at start instead of just one GKC for just the first player (or game start). This GKC works the same as APPLY_START_GAME_OR_SIDE_CHANGE which also reports switching to "observer". A module can check {PlayerSide} to ignore specific "sides". Although upon testing, sometimes it would be nice to know the side from which the player left {PreviousSide?}. Perhaps a TBD?

rlament:
> is this different from the new feature that you are developing here ?

Great tip on handling the pre-retirement to clear out the previous side. Tried it and it works really well.

This feature deals with side selection via the wizard. The retire button functionality just falls into place. This feature fires a Startup GKC as each player joins using the wizard user interface.

The existing "At Start of Fresh Game or Player join/Side-Change" Startup GKC only fires once from the wizard regardless of how many players join the game. So the first player gets preferential treatment and the others are left out.

@riverwanderer I'm wondering which Startup GKC you use and how you deal with side selection from the wizard for the second and each subsequent player?

riverwanderer:
>  I'm wondering which Startup GKC you use and how you deal with side selection from the wizard for the second and each subsequent player?

The module from which the example was derived uses this SGKC to trigger, which sounds like it is problematic. So I could try the "Every Game Launch/Load of Session" instead, that's what I was using for this before V3.7:

image

PR #13916, opened 2025-03-27
A prototype for clickable maps

riverwanderer:
Pilot implementation of [this idea](https://forum.vassalengine.org/t/clickable-maps/84901).

Includes console logging code.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/Map.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/Map.java
@@ -303,7 +303,7 @@ public class Map extends AbstractToolbarItem implements GameComponent, MouseList
   private boolean anyMouseoverDrawn = false;
 
   public static String getClickMap() {
-   return clickMap; }
+    return clickMap; }
 
   public static String getClickZone() {
     return clickZone; }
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/Map.java
```diff
@@ -2021,6 +2041,24 @@ else if (multicaster != null) {
     clearFirst = true;
     theMap.repaint();
     activeMap = this;
+
+    // Look for VassalActiveMapGHK in zone or otherwise in map / module
+    // FIXME - incorrect zone found at some points (non-rectangular shape?). Also, why not finding some zones.
+
+    final Point mp = scale(p, 1 / activeMap.getZoom());
```
rlament, line 2048: In addition to scaling you need to apply a translation for the top left visible corner of the map. It works fine when the map is fully zoomed out, but once zoomed in it reports the wrong ones.

`mouseReleased` may not be the best event to hook. This event fires on control-click, shift-click, dragging a stack into the zone, click drag and release of the mouse and maybe others. The user might try to multiselect a number of counters with a bounding box, which should not generate a zone click event. Untested and just guessing that `mouseClick` might be a better option.

For a clickable zone there should be some feedback to the user that the zone is the active control much like pressing on a button control. That would mean adding some effect to the zone on `mousePressed`. But then possibly cancelling that on a mouse drag?

In my module there are both convex and concave zones. As long as the map was fully zoomed out the zone selection was pretty good. Because zones may not perfectly align with the underlying bitmap, a dead band around the zone edges would be nice, but not easy to implement. Although if there is a visual effect, like shading the zone on mouse press, it would be obvious which is the selected zone.

Something else to consider: if the user inadvertently clicks on a zone, how do they cancel out of it?

This is an interesting start but there's a lot more involved to make it a workable feature.

PR #13926, opened 2025-03-30
Release 3.7 - disable empty SubMenu

riverwanderer:
If a sub-menu is empty or contains only disabled items, then it will still be displayed but it will itself be disabled (greyed out).


riverwanderer:
> Does this make it impossible to descend into the submenu?

That's my intention but the submenu should contain no visible items at this point (or I have misunderstood). It has tested out OK in a simple case but the PR would bear some testing from me on multiple levels of menu nesting.

riverwanderer:
I appear to have accidentally reverted a code change from PR #13882, I will correct that.

riverwanderer:
Update: my mistake. Looks like the PR "clash" was handled ok. Both changes in in the Release.3-7 merge as far as I can tell.

riverwanderer:
Tests OK with nested submenus. Note - a submenu containing a submenu is accessible, regardless of the state of the contents of the lower level submenu.

PR #13929, opened 2025-03-31
Fixing the fix for null sender to chat socket handler

riverwanderer:
Revision of PR #13876 modelled on PR #11056 which addresses a similar issue elsewhere - the test needs to follow synchronisation.

No reporting as it seems from test that this code path may be traversed when a user is intentionally offline.

This PR includes some code cleanup. The actual logic change is in send().


riverwanderer:
This fix is critical as the original merged PR attempts a fix in the wrong place and generates a spurious "Disconnected" chat message.

PR #13955, opened 2025-04-03
13682 Prevent Hotkey from restarting an open log file

### LINKED ISSUE #13682: Hotkey open log file restarts an existing open log session (losing data)
The hotkey to open a logfile should silently do nothing if a logfile is already open (equivalent to the menu item "Begin Logfile..." being greyed out).

However, the hotkey in fact offers to open a new logfile. When this is done, the currently open-for-write logfile is lost.

rlament:
The menu system prevents the user from clicking 'Begin Logfile' when a log file is already open. Hotkeys were not restricted and could restart the log file. Closes #13682

This change only affects the 'Begin Logfile' command. We could instead replace this check with one in `KeyStrokeListener.keyPressed()` to  prevent the actions of any and all disabled KeyStrokeListeners. Not sure if that is a breaking change and/or desirable. Any thoughts are appreciated.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
@@ -673,7 +673,9 @@ public class BasicLogger implements Logger, Buildable, GameComponent, CommandEnc
 
     @Override
     public void actionPerformed(ActionEvent e) {
-      beginOutput();
+      if (isEnabled()) {
+        beginOutput();
+      }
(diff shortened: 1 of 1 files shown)


riverwanderer:
Confirmed for me - tested [build](https://vassalengine.org/builds/build/2872382019) with C&C Nap module  and it fixes the issue with the open logfile hotkey. Also tested that Alt+W will still trigger other actions (even though this seems unlikely to be required).

PR #13957, opened 2025-04-03
Add dice & random text button summary output property

riverwanderer:
Also, fix an issue whereby Random Text Button does not implement results sorting. Caution: AI-mediated work.

PR #13963, opened 2025-04-04
Revert "Add dice & random text button summary output property"

riverwanderer:
Reverts vassalengine/vassal#13957

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
@@ -326,7 +326,7 @@ public class DiceButton extends AbstractToolbarItem {
     int currentResult = -1;
     int count = 0;
 
-    for (final int result : keepDice) {
+    for (int result : keepDice) {
       if (result == currentResult) {
         count++;
       }
--- a/vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
@@ -326,21 +326,20 @@ public class DiceButton extends AbstractToolbarItem {
     int currentResult = -1;
     int count = 0;
 
-    for (int result : keepDice) {
-      if (result == currentResult) {
-        count++;
-      }
-      else {
-        if (currentResult != -1) {
-          if (summaryVal.length() > 0) {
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
```diff
@@ -286,12 +285,14 @@ else if (KEEP_EQUAL.equals(keepOption) && roll != keepValue) {
...
-      // tempDice has been sorted, so can just pull the required number from the front or back of the array
+      // tempDice has been sorted, so can just pull the required number from the the front or back of the array
       if (KEEP_SMALLEST.equals(keepOption)) {
         System.arraycopy(tempDice, 0, keepDice, 0, keepCount);
       }
       else {
-        System.arraycopy(tempDice, tempDice.length - keepCount, keepDice, 0, keepCount);
+        for (int i = 0; i < keepCount; i++) {
+          keepDice[i] = tempDice[tempDice.length - keepCount + i];
+        }
```
rlament, line 295: Any reason for replacing the arraycopy with a loop? arraycopy is still used a couple lines above. I'm not seeing the reason for this change resulting in the loss of consistency with the other half of the 'if' statement.

riverwanderer: This reversion either wan't applied or has been overridden since. I raised it as I thought I'd accidentally applied a change to the actual 3.7 release, in my amateurish confusion during attempts to move my PR #13957 from the master (3.8) to queue it for the current release. Now that I've replicating with the upstream 3.7 repository, I believe that my intended changes (including using arrayCopy here) have made it to the pre-release 3.7.16. 

A bit more about the arrayCopy / for loop: the change to arraycopy was an improvement proposed by IntelliJ's code checking which I accepted.

Thanks for your scrutiny (and in advance, for the future). I appreciate it.

uckelman: `Systsem.arraycopy` will be faster than copying consecutive elements one at at time in pretty much all cases.

uckelman:
What's the state of this PR? Is it still wanted?

riverwanderer:
No longer required.

Issue #13846, opened 2025-02-16
Feature request: set width of text label


assaultdoor:
Could we please have the option to set the width of a text label and wrap text to that width? It would be so much better than having to use the <br> tag dozens of times.

Unless I've missed this feature somewhere, and have spent hours and hours mucking about with the placement of <br> tags.

riverwanderer:
That would be a useful feature. I've had to do the same myself. Would also be useful in mouse-over.

rlament:
If all you want is multi-line text labels use the html line break element `<br>`. This works on mouse-over too. I also use this in the toolbar buttons. The text has to be wrapped in `<html>` tags, for example:

```<html>First Line<br>Second Line</html>```

**Comment by rlament** [comment]
Since I replied with such a boneheaded response, I looked into this further and created a pull request.

Just one thing of note. The `<pre width=32>` tag supports a width attribute in pixels, 32 in this example. It doesn't wrap however, just truncates. Not that useful.

**Comment by rlament** [comment]
Well, this is making me look even more foolish. Can't say for sure what I tried before jumping off to write some code. I must have got the syntax wrong. Came across this GameSquad post [https://www.gamesquad.com/forums/index.php?threads/can-anyone-help.187829/](https://www.gamesquad.com/forums/index.php?threads/can-anyone-help.187829/) which got me back on this subject.

Here's a brief summation in case the above link breaks. The tags `<div>` `<table>` and `<p>` do support the width attribute. It seems that the width is always in pixels. For example for a 80 pixel max width:
`<div width=80>Some long text that is not Latin</div>`

Within tables the column widths can be set in the first row using a width attribute on either `<th>` or `<td>` elements.

The "Label text" field does support BeanShell expressions even though there is no expression builder in the editor. By using properties there's plenty of flexibility for custom styles and colors. If the "Label text" is html, remove the `$label$` from the "Name format" field since it does not interpret the html.

All this works with mouse-over too.

Here's a more elaborate example nicely formatted for readability.
```
<html>
<table width=80 border=4 cellspacing=1 style="border-color:rgba(255, 0, 0, 0.5);">
    <caption style="color:green;background:#ffffed">Caption</caption>
    <tr>
        <td>Top-left Box</td>
        <td border=0>Right Side</td>
    </tr>
    <tr>
        <td colspan=2>$Marker$</td>
    </tr>
</table>
</html>
```

[...]

PR #13971, opened 2025-04-06
13846 Fix the width of a text label and wrap text

rlament:
This is a proposal for fixing the width of a text label, wrapping the text and permitting the height to grow adding as many lines as needed to fit all the text.

Since html in the Label text field has some control over the layout of the text, it made sense to add a width specification here. The proposal is a fixed width specifier, in pixels, using the 'max-width' style property. For example, to set a 40 pixel width:
```
<html><span style="max-width:40">This is an exceedingly long label for some silly counter</span></html>
```
The underlying Swing html parser ignores unknown/unsupported style properties, so the extra styling has no effect. However, the width value can be extracted by Vassal which can fix the one dimension and then query the Swing component for the other dimension. The final dimensions are set as the preferred dimensions.

Line breaks after paragraphs `<p>` and line breaks `<br>` are still honoured. The `max-width` style can appear in any element and applies to the entire label. Somewhat of a break from the typical where the style applies only to the containing element. Purist may object.

Since the Label's text field is a Beanshell expression, the width can be specified as the result of some calculation.

This pull request adds the same feature to mouse-over. There are so many configuration options that this is not fully tested.

If it is possible to set a fixed width, then so can the height. The module designer could specify the `height` style fixing the height and making the width variable. Not sure how useful this is. Already included but it could be dropped.

For the width specifier, I used `max-width` since the `<pre>` tag supports a width attribute. But I don't see this as a conflict anymore. `max-width` could change to `width` if that is the consensus.

I have not updated the help files yet. Waiting if/until this is considered a workable and worthwhile addition.
Closes: #13846

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Labeler.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Labeler.java
@@ -79,6 +79,9 @@ import java.util.List;
 import java.util.Objects;
 import java.util.function.DoubleConsumer;
 
+import static VASSAL.tools.image.LabelUtils.findHtmlStyle;
+import static VASSAL.tools.image.LabelUtils.getPreferredSize;
+
 /**
  * d/b/a "Text Label"
  *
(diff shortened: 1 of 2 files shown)


riverwanderer:
I'm interested in this, as I have had to grapple with the issue a good few times. Hope I can get some time to look at it at some point soon.

rlament:
Discovered that various tags do support a _width_ attribute which does wrap long text at word boundaries. See my comment in issue #13846 for more information. Will close this merge request as Vassal already supports the requested functionality.

In any event, this was an interesting exercise in adding new styling to Vassal which may not be supported by the Java libraries.

PR #13979, opened 2025-04-07
12924 Report format drop down fix 

### LINKED ISSUE #12924: Report format insert feature selects all text causing over-write
This problem is observed in Change-Property, Action Button and Report Action traits. Presumably whether the report format field appears. Bug has crept in with v3.7 - om V3.6.19, the field is selected only after the insert has been done, and is de-selected immediately before an insert.

Example in Vassal 3.7.5:
image

image

rlament:
Noticed this was being looked at and one of those pull requests sparked an idea. Added a check to prevent 'select all' when the focus remains on the same control. Fixed the tab key so that focus actually goes to the next control on the page. Added 'insert key' to expand the dropdown.

Closes #12924

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/FormattedStringConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/configure/FormattedStringConfigurer.java
@@ -29,6 +29,8 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.FocusEvent;
 import java.awt.event.FocusListener;
+import java.awt.event.KeyListener;
+import java.awt.event.KeyEvent;
 
 import javax.swing.DefaultComboBoxModel;
 import javax.swing.JComboBox;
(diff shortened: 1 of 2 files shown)


riverwanderer:
Tests out great for me, nice one! Hope this can go into 3.7.16 @uckelman.

rlament:
It is also possible to delete the implementation of the `FocusListener` interface. It appears that this class nor any derived class use it. Let me know if it should go.

I did quick tests of derived classes: DeckSelectionConfigurer, FlareFormattedStringConfigurer, FormattedStringConfigurer, PlayerIdFormattedStringConfigurer and PropertyExpression.

riverwanderer:
> Fixed the tab key so that focus actually goes to the next control on the page

Are you sure that wasn't working already? Seems to be working for me.

In any open window, fields preserve current contents section/caret state, I notice. If none, then tab will select-all.  I suspect this is standard.

rlament:
> > Fixed the tab key so that focus actually goes to the next control on the page
> 
> Are you sure that wasn't working already? Seems to be working for me.
> 
> In any open window, fields preserve current contents section/caret state, I notice. If none, then tab will select-all. I suspect this is standard.

Just to be clear it is the 'tab out' that's broken, tabbing into is fine, you just can't leave. Retested on 3.6.7 (before 11557) and tab out does not work. Definitely does not work with 3.7.15. The fix is this one line: `dropList.setFocusable(false);`

In your revert branch this line has been added. So it had a bit more than just a revert and thus tabbing worked.

PR #13983, opened 2025-04-09
Release 3.7 - revert to standard text select

riverwanderer:
Reverting to standard text select (per vassal 3.6.7). After this reversion, focus does not interfere. Triple click to select all text.

This PR will impact PR #12924 which may be revised to focus on tab and insert key actions.

Includes minor mods to satisfy intellij advisories.


Cattlesquat:
This is the wrong way to go, per discord discussion.

Issue #13036, opened 2024-01-03
Show image file name for button icons


riverwanderer:
It can be useful to know the image name of button icons when editing.
For example, to make it easier to find the same image for another button or html tag in the module.

The image file name could perhaps be display in subdued text next to the default button (see below for the button I am referring to). Or perhaps instead, display the file name as a tooltip when the user hovers a cursor over the icon image.

<img width="618" alt="image" src="https://github.com/vassalengine/vassal/assets/58135975/07fe2890-a5bb-4ff1-8b64-35c36ebdb55d">

PR #13998, opened 2025-04-16
Fix13036 show image file name for button icons 3 7b

Niedowidek:
Added a tooltip to button icon showing image filename as suggested in https://github.com/vassalengine/vassal/issues/13036

Branch name fixed.
Checkstyle formatting applied.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/AutoConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/configure/AutoConfigurer.java
@@ -20,6 +20,7 @@ package VASSAL.configure;
 import VASSAL.build.AutoConfigurable;
 import VASSAL.build.Configurable;
 import VASSAL.build.GameModule;
+import VASSAL.configure.button.ToolTipConfigurer;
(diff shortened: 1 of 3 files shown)


riverwanderer:
Tested by mouse over of an existing map button image.
1. Mouse over shows image file name OK
2. Removed image... mouse over still shows old image name.
3. Closed the component (map) panel and re-opened - mouseover still appeares and the old image name is still there.
4. Added a new image. Mouseover still shows old image name.

Niedowidek:
@riverwanderer 
Mark, 
I made modifications after your tests. 
Could you check it once more ?

riverwanderer:
Tested (sorry for delay, been busy).  Working ok now.
I notice that when a file is added, until the module is closed and reopened, the full file path is shown. Also, when an in-built default is chosen the /images file path is always displayed. I don't think these a major issues. Possibly, in fact helpful.

I am happy that the PR addresses the request. Thank you @Niedowidek.

Niedowidek:
Thank you very much !

Issue #12772, opened 2023-10-02
Add a toNumber() String function 


BrentEaston:
Currently, if you want to take substring and then do arithmetic on them you have to use Integer.parseInt() to convert them to a number which is a) not documented b) Painful and not obvious and c) prone to throwing an Exception.

Add a .toNumber() string function  that will attempt o convert its argument to a number, or 0 if invalid

rlament:
Are either `.toInteger()` or `.asInteger()` better names? Consider adding a similar conversion for floating point numbers. The term 'Number' is too generic creating confusion if supporting two numeric types.

BrentEaston:
Good point. toInteger() and toFloat() would be better.

Niedowidek:
Using **NumberUtils** from org.apache.commons.lang3.math should be sufficient to achieve this goal, doesn't it ?
It has toInt and toFloat methods.
They return 0 (zero) if conversion fails.

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html#toInt(java.lang.String)

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html#toFloat(java.lang.String)

riverwanderer:
PR #[11330](https://github.com/vassalengine/vassal/pull/11330/files) allowed .toString() in Beanshell. Perhaps it provides a clue as to were these additional methods can be added.

Niedowidek:
My point was that new methods aren't needed as one can use Apache Commons for this.
Or maybe I don't understand the issue?

riverwanderer:
Just to level-set - this issue is about adding  a new feature into Vassal's Beanshell interpreter. For example, if you edit a module, you could add the Action Button below with an expression as shown. At the moment, that Action Button will generate an error thus:
- Bad Data in Module: Action Button Action Button Source: {TEST_result.toInteger().toString() + 7} Error: Expression evaluation error.

Note: At the moment would report `{TEST_Result + 7} `fine, as long as test_Result contained an Integer.

<img width="796" alt="Image" src="https://github.com/user-attachments/assets/21d8f2ff-e626-43f9-a442-2229f559275e" />

Niedowidek:
@riverwanderer Is there a specific module and Action Button where I could reproduce this bug? What steps should I take ?

I could try to fix this issue if you don't anything against it.

riverwanderer:
[...]

PR #14045, opened 2025-05-09
Fix12772 ToNumber function to convert strings to numbers

Niedowidek:
https://github.com/vassalengine/vassal/issues/12772
- added functions to expression interpreter and to String menu
- updated manual


vassal-doc/src/main/readme-referencemanual/ReferenceManual/ExpressionString.adoc
```diff
@@ -37,6 +37,8 @@ So the first character in a string is always character 0, not character 1
 | <<#toUpperCase,toUpperCase>> | Convert lowercase characters in a string to uppercase
 | <<#toLowerCase,toLowerCase>> | Convert uppercase characters in a string to lowercase
 | <<#toString,toString>> | Convert a non-string to a String
+| <<#toInteger,toInteger>> | Convert a non-string to a String
+| <<#toFloat,toFloat>> | Convert a non-string to a String
```
jonathanrwatts, line 41: Aren't these descriptions wrong? (lines 40-41)

Niedowidek: You are right. I'll correct it.

Niedowidek: @jonathanrwatts  I corrected it. Thanks for pointing this out.

jonathanrwatts (commented):
Looks good to me, but I'm not a VASSAL code expert, so probably need a more experienced reviewer to look it over.

Niedowidek:
@jonathanrwatts Thank you !
@BrentEaston What do you think about this PR?

vassal-doc/src/main/readme-referencemanual/ReferenceManual/ExpressionString.adoc
```diff
@@ -37,6 +37,8 @@ So the first character in a string is always character 0, not character 1
 | <<#toUpperCase,toUpperCase>> | Convert lowercase characters in a string to uppercase
 | <<#toLowerCase,toLowerCase>> | Convert uppercase characters in a string to lowercase
 | <<#toString,toString>> | Convert a non-string to a String
+| <<#toInteger,toInteger>> | Convert a string to an integer number
+| <<#toFloat,toFloat>> | Convert a string to a float number
```
rlament, line 41: These functions take a string parameter but they are not string member functions. This section of the documentation refers to functions of the form: <string>.function(parameters). So these functions don't really belong here. They should be in a separate sub-section here or under a completely different category.

Is "Math" a better spot for these? If converting to a number the reason is likely for some mathematical function. Maybe there should be a new "Number" section as a complement "String"? Just some thoughts.

Since these are Top-level functions and not member functions, as Brent suggests they should be capitalized. Member functions begin with a lowercase letter, Top-level uppercase.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -1700,13 +1700,44 @@ private int parseInt(Object src, String function, Object value, int dflt) {
       result = Integer.parseInt(value.toString());
```
rlament, line 1712: Quoting the original request:

> Add a .toInteger() string function that will attempt to convert its argument to a whole number

From that I assume that the function should accept a string in floating point format. Currently this throws an exception if the input is in floating point format. The original request however doesn't specify it the number should be truncated or rounded.

Earlier in this file, around lines 313-328, is example code that attempts both integer and floating point conversions. Not exactly what's required, but close.

vassal-doc/src/main/readme-referencemanual/ReferenceManual/ExpressionString.adoc
Niedowidek: Thank you.

I added "Number" menu and moved the functions there.
I capitalized functions' names.

image

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -1700,13 +1700,44 @@ private int parseInt(Object src, String function, Object value, int dflt) {
...
-      ErrorDialog.dataWarning(new BadDataReport(message, data, e));
+      reportIllegalNumber(src, function, value, e);
     }
     return result;
   }
 
+  private float parseFloat(Object src, String function, Object value, float dflt) {
+    float result = dflt;
+    try {
+      result = NumberUtils.toFloat(value.toString(), 0f);
```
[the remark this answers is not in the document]
Niedowidek: I simplified both methods to use just NumberUtils.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: That's what I'm not sure about.

Currently ToInteger returns an int when it is given an integer as a string, otherwise zero.
ToFloat return a float when it gets a float or an integer as a string, otherwise zero.

I added some unit tests to VASSAL.script.ExpressionInterpreterTest to illustrate an expected behaviour.

Should ToInteger return something when given a float number? Should it be truncated or rounded?

@BrentEaston Could you clarify this, please?

Niedowidek:
@rlament Could you take a look, please?

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
```diff
@@ -376,4 +377,43 @@ public void sum_and_count() throws ExpressionException {
     }
   }
 
+  @Test
+  void givenWholeNumberAsString_whenToInteger_thenReturnInt() throws ExpressionException {
+    assertEquals(123, new ExpressionInterpreter("ToInteger(\"123\") + 7").toInteger(null, "123"));
```
rlament: These tests don't exercise the ExpressionInterpreter. A more complete test should be calling the `evaluate()` function and testing the return value. Something like:
>>>
    ExpressionInterpreter interpreter = new ExpressionInterpreter("ToInteger(\"123\") + 7");
    String result = interpreter.evaluate();
    assertEquals("130", result);
>>>

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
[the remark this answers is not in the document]
rlament: You have taken this in the opposite direction from my intent. I think we need to report an error to the user. What you have done is fail silently. This makes it much harder to debug because the error is now hidden. The `ToInteger` or `ToFloat` could be a part (or multiple parts) of a complex calculation. In some cases, zero may be a valid result, so testing may not reveal the parse failure.

I'm wondering if it is possible to catch the NumberFormatException in the ExpressionInterpreter instead of here. These functions then become plain wrappers for `parseInt` and `parseFloat`. This should be possible, but not sure.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
rlament: I am considering taking back my suggestion for separate integer and floating point parsers. A single function that accepts either integer or floating point strings is simpler from the user point of view, especially if the data type is unimportant. The parser could be wrapped with `Math.round` if the output must be an integer. For truncation there is the `.intValue()` function.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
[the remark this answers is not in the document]
Niedowidek: Original request said to return "0 if invalid" but it seems right that the function should report an incorrect value. I'll change it.

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
Niedowidek: Thank you :) I corrected it.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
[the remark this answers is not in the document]
Niedowidek: Now it reports an error (as you suggested) and returns 0 (as originally requested).

image

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: I replaced ToInteger and ToFloat with a single ToNumber function.
Under the hood it's NumberUtils.createNumber which throws NumberFormatException if given not a number.

Niedowidek:
@rlament I implemented your suggestions. Can you take a look please ?

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
```diff
@@ -375,5 +380,44 @@ public void sum_and_count() throws ExpressionException {
...
+  }
+
+  @Test
+  void givenFloatingPointNumberAsStringWithQualifier_whenToInteger_thenReturnZero() throws ExpressionException {
+    assertEquals("123.7", evaluate("ToNumber(\"123.7f\")"));
+  }
 
+  @Test
+  void givenOctalPointNumberAsStringWithQualifier_whenToInteger_thenReturnZero() throws ExpressionException {
+    assertEquals("452", evaluate("ToNumber(\"0704\")"));
```
rlament: This is likely to be an unfortunate surprise to a module designer. Most won't be using octal numbers nor perhaps even aware of them. It is not uncommon for grids to be numbered with leading zeros. These should be treated as decimal numbers.

See the note in the docs under [Preserve leading zeros in Integers](https://vassalengine.org/doc/latest/ReferenceManual/GlobalOptions.html#leadingzeros) This is one of the most likely use cases for `ToNumber`.

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
```diff
@@ -375,5 +380,44 @@ public void sum_and_count() throws ExpressionException {
...
+  }
+
+  @Test
+  void givenWord_whenToInteger_thenReturnZero() throws ExpressionException {
+    assertEquals("0", evaluate("ToNumber(\"NaN\")"));
+  }
+
+  @Test
+  void givenEmpty_whenToInteger_thenReturnZero() throws ExpressionException {
+    assertEquals("0", evaluate("ToNumber(\"\")"));
```
rlament: The test for null argument is missing. Use `evaluate("ToNumber(null);")`.

Niedowidek: Thank you. I added it.

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
Niedowidek: Thank you, I corrected it.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -1720,7 +1721,11 @@ private void reportIllegalNumber(final Object src, final String function, final
 
   public Object toNumber(Object src, Object stringToConvert) {
     try {
-      return NumberUtils.createNumber((String) stringToConvert);
+      String toConvert = StringUtils.startsWith((String) stringToConvert, "0x")
```
jonathanrwatts: I know this was added to avoid accidentally treating a decimal number with a leading zero as an octal number, but that may not be clear to future viewers/editors--I think a code comment explaining what you're doing and why would be warranted.

Niedowidek: I thought about it myself. Thank you. Fixed it by moving code into a new method.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -1700,13 +1700,43 @@ private int parseInt(Object src, String function, Object value, int dflt) {
...
-      final String data = "Data=[" + value.toString() + "]."; //NON-NLS
-      ErrorDialog.dataWarning(new BadDataReport(message, data, e));
+      reportIllegalNumber(src, function, value, e);
     }
     return result;
   }
 
+  private void reportIllegalNumber(final Object src, final String function, final Object value, final Exception e) {
+    final String message = "Illegal number in call to Beanshell function " + function + ". " + ((src instanceof Decorator) ? "Piece= [" + ((Decorator) src).getProperty(BasicPiece.BASIC_NAME) + "]. " : ""); //NON-NLS
+    String reportedValue = value != null ? value.toString() : "null";
```
rlament: Declare `reportedValue` as final to fix the PMD check that fails the build.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -1700,13 +1700,43 @@ private int parseInt(Object src, String function, Object value, int dflt) {
...
+    catch (NumberFormatException e) {
+      reportIllegalNumber(src, "ToNumber", stringToConvert, e);
+    }
+    return 0;
+  }
+
+  private static String convertOctalToDecimal(final String stringToConvert) {
+    return StringUtils.startsWith(stringToConvert, "0x")
+            ? stringToConvert
+            : StringUtils.stripStart(stringToConvert, "0");
```
rlament: The unit tests are missing the most obvious test case: `ToNumber(\"0\")`. It would return the correct value, but incorrectly generate a warning message. Should these inputs (0f, 0d, 0e0) return 0 or 0.0? Also, negative octal values are not handled (who would ever use one?).

Unit tests can capture warning message calls with a mock. This example counts calls to the `warn()` function.
```
try (MockedStatic<GameModule> staticGm = Mockito.mockStatic(GameModule.class)) {
  GameModule gm = mock(GameModule.class);
  staticGm.when(GameModule::getGameModule).thenReturn(gm);
  assertEquals("0", evaluate("ToNumber(\"0\");"));
  // No warnings.
  verify(gm, times(0)).warn(anyString());
}
```

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -1700,13 +1700,43 @@ private int parseInt(Object src, String function, Object value, int dflt) {
...
+   *
+   * toInteger(stringToConvert)   - convert given String to integer or to zero if conversion is impossible
+   * toFloat(stringToConvert)     - convert given String to float or to zero if conversion is impossible
+   */
+
+  public Object toNumber(Object src, Object stringToConvert) {
+    try {
+      String toConvert = convertOctalToDecimal((String) stringToConvert);
+      if (toConvert == null) toConvert = "null";
+      return NumberUtils.createNumber(toConvert);
```
rlament: If the input string has to be massaged prior to calling a library function, maybe this is the wrong library function. Do the conversion much like the ExpressionInterpreter already does with cascading converters. Start with `Integer.parseInt` which rejects hex and octal numbers. There is no indication that these need to be supported. If that parse fails try BigInteger. If that fails parseDouble or createDouble. If that fails return 0 and report the error.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: Done

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: I tried something similar. Worked for a single test but when I run all the tests in a class, then the same single test fails. I don't understand why.

rlament: This happens when one of the earlier test(s) changes the state of the system and doesn't reset it. It can be either a global variable or static data. Sorry, I did not run the full suite of tests. It is the `new ExpressionInterpreter("Random(\"NaN\")")` in test Random that conflicts with `givenWord_whenToNumber_thenReturnZero`. Both generate the same error because the input parameters are the same. `ErrorDialog` stores warning messages in static variable `reportedDataWarnings`. Since the data is the same, there is no warning from the second test. This is intentional to avoid polluting the message log with multiples of the same warning message.

So a different approach is necessary. Instead of counting warning messages, count `BadDataReport` objects. These are created before the check for duplicates.
```
private String evaluateWithWarning(String expression) throws ExpressionException{
  try (MockedConstruction<BadDataReport> report = Mockito.mockConstruction(BadDataReport.class)) {
    ExpressionInterpreter interpreter = new ExpressionInterpreter(expression);
    final String result = interpreter.evaluate();
    // Exactly one warning
    assertEquals(1,report.constructed().size());
    return result;
  }
}
```

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: parseDouble returns NaN when given string "NaN"
I assumed that it would be wrong according to this issue's requirements so I throw exception. Maybe I should change it ?

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: Thank you ! That worked :)

Niedowidek:
@rlament Could you review it again?

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
```diff
@@ -375,5 +381,85 @@ public void sum_and_count() throws ExpressionException {
       assertThat("Count on named map", result, is(equalTo("2")));
     }
   }
+  
+  private String evaluate(String expression) throws ExpressionException{
+    ExpressionInterpreter interpreter = new ExpressionInterpreter(expression);
+    return interpreter.evaluate();
+  }
```
rlament: If you are not using this function anymore it can be deleted.

My thought was to use to function for successful conversions and another function what you called  `verifyEvaluatedWithWarning` for failures. Something like `evaluateWithSuccess` and `evaluateWithWarning` in my opinion is more obvious without a need to actually having to inspect the functions and therefore no need for the `warningsExpected` parameter. The success function would check for zero warnings. But, what you have is fine. These are just unit tests. If this function is not being used, just delete it.

vassal-app/src/main/java/VASSAL/configure/BeanShellFunctionMenu.java
```diff
@@ -82,6 +82,7 @@ public class BeanShellFunctionMenu extends JPopupMenu {
     Resources.getString("Editor.BeanShell.warning2"),
     Resources.getString("Editor.BeanShell.warning3")
   };
+  public static final String TO_NUMBER = "ToNumber";
```
rlament: A minor point, move this public declaration up to where `COMMENT` is declared. Moves this public declaration out of this section of protected members.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -1700,13 +1722,39 @@ private int parseInt(Object src, String function, Object value, int dflt) {
...
+    final String data = "Data=[" + reportedValue + "]."; //NON-NLS
+    ErrorDialog.dataWarning(new BadDataReport(message, data, e));
+  }
+
+  /*
+   * toInteger & toFloat
+   *
+   * toInteger(stringToConvert)   - convert given String to integer or to zero if conversion is impossible
+   * toFloat(stringToConvert)     - convert given String to float or to zero if conversion is impossible
+   */
```
rlament: Comment is still referring to the old names.

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
```diff
@@ -375,5 +381,85 @@ public void sum_and_count() throws ExpressionException {
...
+    verifyEvaluatedWithWarning("ToNumber(\"123.7f\")", "123.7", 0);
+  }
 
+  @Test
+  void givenOctalPointNumberAsStringWithQualifier_whenToNumber_thenReturnAsDecimal() throws ExpressionException {   
+    verifyEvaluatedWithWarning("ToNumber(\"0704\")", "704", 0);
+  }
+
+  @Test
+  void givenHexadecimalPointNumberAsStringWithQualifier_whenToNumber_thenReturnNumber() throws ExpressionException {
```
rlament: Change suffix to ...thenReturnZero

vassal-doc/src/main/readme-referencemanual/ReferenceManual/ExpressionNumber.adoc
```diff
@@ -0,0 +1,34 @@
...
+==== Number functions
+
+
+[#Range]
+[width=100%,cols="33%a,67%a"]
+|===
+|*Number* |Convert a string to a number. +
+
+|ToNumber(string) |
+[#ToInteger]
```
rlament: Is `#ToInteger` reference being used? Can't find one. It is the old function name.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
rlament: When Beanshell accesses a variable that is NaN, it does not warn. I think NaN should be treated as valid input. Just call `parseDouble` directly, keep it simple and stick with the same behaviour. It is less likely to break things. Besides if this function throws on NaN, it should also handle +/-Infinity. Maybe the user wants to set a variable to NaN, although not sure why. `x=ToNumber("NaN")` is one way to do just that.

On an unrelated note, there should be at least one unit test for BigInteger, in particular to confirm parseInt fails and BigInteger succeeds.

rlament: I must have inadvertently deleted my response to the question about "NaN". My apologies if you are seeing duplicates.

When Beanshell accesses a variable that is NaN, it does not warn. I think NaN should be treated as valid input. Just call parseDouble directly, keep it simple and stick with the same behaviour. It is less likely to break things. Besides if this function throws on NaN, it should also handle +/-Infinity. Maybe the user wants to set a variable to NaN, although not sure why. x=ToNumber("NaN") is one way to do just that.

On an unrelated note, there should be at least one unit test for BigInteger.

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
Niedowidek: OK, thank you. I deleted it.

vassal-app/src/main/java/VASSAL/configure/BeanShellFunctionMenu.java
Niedowidek: Right, I moved it.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: I changed it.

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
Niedowidek: Thank you. I corrected it.

vassal-doc/src/main/readme-referencemanual/ReferenceManual/ExpressionNumber.adoc
Niedowidek: Thank you. I missed it.

Niedowidek:
@rlament I pushed the commits. I hope I haven't missed anything this time.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -96,6 +100,24 @@ public class ExpressionInterpreter extends AbstractInterpreter implements Loopab
...
+  private static final List<Function<String, Object>> toNumberConverters = new ArrayList<>();
+  static {
+    toNumberConverters.add(Integer::parseInt);
+    toNumberConverters.add(BigInteger::new);
+    toNumberConverters.add(getDoubleConverter());
+  }
+
+  private static Function<String, Object> getDoubleConverter() {
+    return value -> {
+      double result = Double.parseDouble(value);
```
rlament: You did have a question about dealing with NaN. My response was to treat NaN as valid input. I don't see a response to my comment and this function is still here so I have to assume you do not agree. The original request just says "0 if invalid". Is NaN valid or invalid?

My suggestion is keep it simple. Delete `getDoubleConverter` and just add `Double.parseDouble` to `toNumberConverters`.

If keeping `getDoubleConverter`, it should also handle +/-Infinity. Currently +/-Infinity are valid while NaN is invalid.

There is a PMD error on line 113: `result` should be declared `final`.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -1700,13 +1722,38 @@ private int parseInt(Object src, String function, Object value, int dflt) {
...
+
+  /*
+   * toNumber
+   *
+   * toNumber(stringToConvert)   - convert given String to a number or to zero if conversion is impossible
+   */
+
+  public Object toNumber(Object src, Object stringToConvert) {
+    Exception lastException = null;
+    for (Function<String, Object> converter : toNumberConverters) {
```
rlament: PMD error: `converter` should be declared final.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
```diff
@@ -1700,13 +1722,38 @@ private int parseInt(Object src, String function, Object value, int dflt) {
...
-      final String data = "Data=[" + value.toString() + "]."; //NON-NLS
-      ErrorDialog.dataWarning(new BadDataReport(message, data, e));
+      reportIllegalNumber(src, function, value, e);
     }
     return result;
   }
 
+  private void reportIllegalNumber(final Object src, final String function, final Object value, final Exception e) {
+    final String message = "Illegal number in call to Beanshell function " + function + ". " + ((src instanceof Decorator) ? "Piece= [" + ((Decorator) src).getProperty(BasicPiece.BASIC_NAME) + "]. " : ""); //NON-NLS
+    final String reportedValue = value != null ? value.toString() : "null";
```
rlament, line 1722: Should this line have a `//NON-NLS`?

Niedowidek:
> There should be at least one unit test for BigInteger.
> 
> Update the title of this pull request to reflect the name of the function. My suggestion: ToNumber function to convert strings to numbers.

Title updated.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: Well, strange as it seems I just missed that comment. That's why there's no BigInteger test either.

I would like to discuss what this function is for.
I thought that Vassal editors would just do a simple arithmetic. Do we really assume that NaN, BigInteger etc. will be used in board game module ?

Niedowidek: Anyway, I replaced getDoubleConverter as you wished.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: Thanks

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: I checked history. It was added in 2020. I just refactored this lines putting them into a method.

[Internationalization (i18n) Guidelines for Vassal Developers](https://forum.vassalengine.org/t/internationalization-i18n-guidelines-for-vassal-developers/84965) state that `Mark lines containing strings not to be translated with // NON-NLS`.

So for me it looks legit. However I can delete it if you are sure that it doesn't belong here.

rlament: You are right. Looks fine.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
rlament: Take a look at ExpressionInterpreter.java lines 325 to 340. This is were Vassal variables are converted to numbers. To remain consistent with the existing behaviour, the `ToNumber` function should produce the same output. Since this code does a BigInteger conversion so should `ToNumber`.

I'd be surprised if anyone actually uses BigIntegers. After a little testing, it looks like BigIntegers are just treated as long strings of digits. Addition operators do string concatenation not arithmetic addition; multiplication throws an exception.

Someone might use NaN to represent invalid so it may not be an error in some cases. If `ToNumber` blocks use of NaN, there are workarounds. Having `ToNumber("NaN")` return zero may be more confusing than returning NaN. A zero return value certainly would be surprising unless well documented.

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
```diff
@@ -376,4 +382,94 @@ public void sum_and_count() throws ExpressionException {
...
+  }
+  
+  @Test
+  void givenNaN_whenToNumber_thenReturnNaN() throws ExpressionException {
+    verifyEvaluatedWithWarning("ToNumber(\"NaN\")", "NaN", 0);
+  }
+
+  @Test
+  void givenBigInteger_whenToNumber_thenReturnNumber() throws ExpressionException {
+    verifyEvaluatedWithWarning("ToNumber(\"9223372036854775808 \")", "9.223372036854776E18", 0);
```
rlament: You certainly have a knack for writing unit tests that expose latent bugs. I would not have spotted this otherwise.

Because of the trailing whitespace in the `ToNumber` input parameter, the output does not match the input. Both parseInt and BigInteger::new do not accept whitespace while parseDouble does. In this case parseDouble does the conversion hence the exponentiation. If the trailing space is removed, this test fails.

The `ToNumber` function should trim the input string to remove leading and trailing whitespace.

Similarily, adding a space to the input parameter fails the `givenWholeNumberAsString_whenToNumber_thenReturnInt` test because the output becomes a double.

vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
Niedowidek: Thank you for claification.

vassal-app/src/test/java/VASSAL/script/ExpressionInterpreterTest.java
Niedowidek: That's an incredibly polite way to say that I made a mistake leaving that unwanted space there :D
However you are right that it exposes a bug. I added 3 tests and trimming of input.

rlament: I don't have any say in what happens next with this pull request. It can be marked as "Ready for Review". It has been a very educational exercise for me.

Niedowidek: Thank you. Do you know if I should notify someone now ?

I think I can't edit the issue to change its status.

rlament: I think only maintainers can add labels. You would have to contact someone with write access.

This feature is new functionality so I suspect it would go into version 3.8, which I assume is low priority.

riverwanderer: I was able to add the label. It's up to @uckelman which release this gets into. There is not strictly a compatibility issue  (as far as I know), so perhaps it will go into 3.7.

rlament (approved):
Until today I didn't realize I could submit my approval. So here it is.

Issue #13980, opened 2025-04-08
Dice button - property prefix anomaly


riverwanderer:
The "Description" field in the Dice Button and similar components, is somewhat misleading as the Dice Button and others form properties which derive their names from this field. This not indicated by hint text and the reference manual is more misleading still (see next).

History: During developments of v3.3 or later, when the Description field was added to DiceButton name it appears that the new field became the property prefix, whereas in 3.2.17 the prefix was derived from the text that formed the name of the button. Documentation aligns to this previous behaviour.

So, in this example from 3.2.15, the property desc_result is set to the contents of the report variable "result" following the button action:
<img width="799" alt="Image" src="https://github.com/user-attachments/assets/a20a8e73-8b77-4be9-b3d7-d5102521ed4f" />

It is good that button name and properties are separate, as some button names (html for example) will form undesirable property names. Also, attempting to change that will have to address backward compatibility.

The issue affects similar property-setting components but does not affect traits such as layer because these have a distinct "name" field - explicitly used to form property names, eg:
<img width="629" alt="Image" src="https://github.com/user-attachments/assets/97642d00-b2cc-4e0d-9d46-fa4f8802c939" />

A comprehensive solution would be to re-purpose the existing "Description" field as a "Name" field, as done for the traits, and add a new "true" description field that is only a comment field. Likely this would be classed as part of a major release.

A lighter touch or interim solution would be to update documentation and hint text fields.

Issue #14065, opened 2025-05-22
Layer change does not update on screens of peers in an online game


riverwanderer:
Version of Vassal affected: v3.7.16

Last release working: v3.7.15

The issue is observed in the module [The Great War](https://vassalengine.org/wiki/Module:The_Great_War). 

In this module the problem is, when a single die rolled in an online game, the result does not display on the screen of the other player.

<img width="193" alt="Image" src="https://github.com/user-attachments/assets/d85d6e3f-48ed-45f3-a704-73e1b3cf38a0" />

The die shown here ("Burst") did not show on the screen of the other player.

When an existing result is in place, a single die roll does clear the old result even though the new result is not displayed (on the remote screen).

Multiple dice rolls always show - included the first result (top left of a series of layers on the dice result piece).
A single die result always shows on replay of a log file.

The problem is not observed in C&C Ancients, which also displays dice results within a piece with several layers.

In a mixed Vassal version (v3.7.15 plays v..3.7.16), the problem occurs only when the v.3.7.16 rolls 1 die.

riverwanderer:
See the Dicecache prototype in The Great War module. In particular, this is the layer that is failing to display on the remote screen:

<img width="181" alt="Image" src="https://github.com/user-attachments/assets/86b80605-c610-4c38-afea-3d9277b02ad4" />

riverwanderer:
> See my comment on line [#65](https://github.com/vassalengine/vassal/pull/65) of PR [#13965](https://github.com/vassalengine/vassal/pull/13965) -- I think it's the cause. Apologies; I noticed the error when that PR was first added, but was busy and forgot to comment on it.

Thanks for the hint. Presumably there is more to the issues around PR #13965 than I've understood from the point raised there (a flaw in a comment). Could you help me further with more specifics ?

jonathanrwatts:
Sorry, looks like that was a false alarm; I just checked, and that change never made it into the final code, so it can't be the cause of this new bug.

jonathanrwatts:
Did you test the new DiceButton code in PR #13972 with a single die roll?  I'm not seeing anything obviously wrong, but I don't have a Java debugging environment set up to actually test things....

riverwanderer:
@jonathanrwatts - yes, I tested with single die rolls and moveover, the same single die roll mechanism is working in other modules. However, I wouldn't rule anything out at this stage. Single user testing won't uncover the problem - it is something to do with replication of the game state to a remote user.

[...]

Issue #4427, opened 2021-06-28
Summary Option for Random Text Dice Button


uckelman:
| Bugzilla |  |
|----------|----------|
| ID | BZ1827 |
| Reported | 2005-12-10 03:05:00 +0100 |
| Modified | 2020-06-22 00:12:51 +0200 |
| Product | VASSAL |
| Component | Unknown |
| Version | unspecified |
| Platform | All |
| OS | All |
| Status | NEW |
| Resolution | None |
| Priority | medium |
| Severity | enhancement |
| Milestone | --- |
| Reporter | [Brent Easton](mailto:b.easton@exemail.com.au) |
| Assignee | [Bugs](mailto:bugs@vassalengine.org) |

#### [Joel Uckelman](mailto:uckelman@nomic.net) 2010-08-25 14:35:37 +0200
<pre>---- Reported by b.easton@exemail.com.au 2005-12-10 03:05 ----

Easiest way to implement this would probably be to add 
a $summary$ option to the Report Format in the 
DiceButton class, then it will work for DiceButton and 
Symbolic Dice Button also. 

User has option to select $result$ or $summary$ in the 
Report Format.

Original Request from Discussion Group:
---------------------------------------
It would be nice if Random Text dice could do simple 
aggregation reporting for
rolling multiple dice. IE. if i have a die defined 
with sides A,B,C and have
a button set to roll 20 of these, it would be nice to 
have the option not to
see the results as

A, A, B, B, A, A, C, A, B, C, A, B, C, A, B, B, A, C, 
A, B

but instead something like

A (x9),  B (x7), C (x4)"

---- Additional Comments From uckelman@nomic.net 2006-11-27 23:55:41 ----

Logged In: YES 
user_id=245140
Originator: NO

This is something that I'll most likely look into when I get to the dice rolling stuff I promised I'd do.

--- Bug imported by uckelman@nomic.net 2010-08-25 14:35 MST  ---

This bug was previously known as _bug_ 1377513 at http://www.vassalengine.org/trackershow_bug.cgi?id=1377513

Unknown Component 
   Using default product and component set in Parameters 
The original assignee of this bug does not have
   an account here. Reassigning to the default assignee
   for the component, uckelman@nomic.net.
   Previous assignee is unknown.
Missing Resolution. Setting status to NEW
Unknown resolution "".
   Setting resolution to MOVED</pre>

Niedowidek:
[...]

PR #14170, opened 2025-07-24
14065 Remove whitespace characters from internationalized Dice Summary strings

rlament:
Removed the linefeed characters from the Dice Summary delimiters in the i18n file. Without digging into command parsing, it looks like the linefeed characters confused the parser. This then blocks the update of the die roll result on the other player's screen.

Leading and trailing whitespace in i18n properties files is problematic. These strings can be modified inadvertently. Some editors can automatically remove trailing plain whitespace depending on their configuration. Whitespace can be escaped but some editors may interpret and replace these. Leading or trailing whitespace may not be obvious to translators. Best to avoid leading and trailing whitespace in the i18n files.

As an aside, the adoc should indicate that the summary is internationalized and parsing should be discouraged.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
@@ -335,9 +335,11 @@ public class DiceButton extends AbstractToolbarItem {
 
     for (int i = 0; i < sortedKeys.length; i++) {
       if (i > 0) {
-        summaryVal.append(Resources.getString("Dice.summary_separator"));
+        summaryVal.append(Resources.getString("Dice.summary_separator"))
+                .append(' ');
       }
       summaryVal.append(sortedKeys[i])
+              .append(' ')
               .append(Resources.getString("Dice.summary_times"))
               .append(resultCounts.get(sortedKeys[i]));
     }
--- a/vassal-app/src/main/java/VASSAL/build/module/RandomTextButton.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/RandomTextButton.java
@@ -164,8 +164,12 @@ public class RandomTextButton extends DiceButton {
     }
 
     for (int i = 0; i < summaryKeys.length; i++) {
-      if (i > 0) summaryResult.append(Resources.getString("Dice.summary_separator"));
+      if (i > 0) {
+        summaryResult.append(Resources.getString("Dice.summary_separator"))
+                .append(' ');
+      }
       summaryResult.append(summaryKeys[i])
+              .append(' ')
               .append(Resources.getString("Dice.summary_times"))
               .append(resultCounts.get(summaryKeys[i]));
     }
--- a/vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
+++ b/vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
@@ -642,8 +642,8 @@ Dice.add_to_each_side=Add to each die
 Dice.add_to_total=Add to overall total
 Dice.random_text_non_numeric=Numeric Random Text Button [%1$s] has a non-numeric face
 Dice.random_text_too_few_faces=Random Text Button [%1$s] has too few faces for roll
-Dice.summary_separator=,\n
-Dice.summary_times=\nx
+Dice.summary_separator=,
+Dice.summary_times=x
 
 # Documentation
 Documentation.about_module=About Module


riverwanderer:
Thank you @rlament! I've no idea how linefeed characters got in there (at least to my recollection) but I have come across trailing space characters. I will avoid them myself in future.

rlament:
I modified the DiceButton documentation to note that the summary is localized text and fixed some typos. Added a note to the effect that the name property is the Description field. I'll leave it to a reviewer to decide whether or not this closes #13980.

rlament:
> There's an issue with the test builds...

Well, this is the first time I've tried to use a "test build". I'm seeing the same issue, a pop-up saying "The application requires a Java Runtime Environment". I'm focusing on just the Windows x86_64 builds. So I went back in time and tried the older 3.7.16 builds. I can't get any of them to run; same error message. Then I tried the last 3.8.0 build; also a no go. In desperation, copied the jre/bin and jre/lib folders from 3.7.16 to the build folder. That launches the module manager. But then trying to open or edit a module produces the "Possible Anti-Virus Program Interference" pop-up. At this point not sure why none of the builds run. I don't have a working baseline to test against. With my lack of experience I don't know where things have gone astray.

riverwanderer:
When I am working on Vassal it is usually through IntelliJ and pushing against the 3.7.x build generally works for me. I am not in a position to try replicate your builds myself at the moment. Perhaps @uckelman can  confirm that the build target for this PR was correct and/or what might have gone awry with including Java.

rlament:
> Original reporter confirmed that this PR fixes #[14065](https://github.com/vassalengine/vassal/issues/14065). Solution to #[13980](https://github.com/vassalengine/vassal/issues/13980) also accepted.

@riverwanderer Did you also confirm that the summary text is still in the same sort order and format as before? I did test with a custom module and it did look ok to me.

vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
```diff
@@ -50,10 +50,9 @@
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
-
+import java.util.SortedMap;
+import java.util.TreeMap;
```
uckelman: What is the purpose of the changes to `DiceButton`?

vassal-app/src/main/java/VASSAL/build/module/RandomTextButton.java
```diff
@@ -164,8 +164,12 @@ protected void DR() {
...
 
     for (int i = 0; i < summaryKeys.length; i++) {
-      if (i > 0) summaryResult.append(Resources.getString("Dice.summary_separator"));
+      if (i > 0) {
+        summaryResult.append(Resources.getString("Dice.summary_separator"))
+                .append(' ');
+      }
       summaryResult.append(summaryKeys[i])
+              .append(' ')
               .append(Resources.getString("Dice.summary_times"))
```
uckelman, line 173: What is the purpose of the changes to `RandomTextButton`?

vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
riverwanderer: The overall purpose (released with 3.7.16) was to provide as "summary" property that contains the roll results in the format requested in #[4427](https://github.com/vassalengine/vassal/issues/4427).

vassal-app/src/main/java/VASSAL/build/module/RandomTextButton.java
riverwanderer: As with DiceButton, the overall purpose (released with 3.7.16) was to provide as "summary" property that contains the roll results in the format requested in #[4427](https://github.com/vassalengine/vassal/issues/4427).

vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
uckelman: Yes, but what do the changes to `DiceButton` do?

vassal-app/src/main/java/VASSAL/build/module/RandomTextButton.java
uckelman: That doesn't answer my question. What I want to know is what these particular changes are for. What would happen were they not here at all?

vassal-app/src/main/java/VASSAL/build/module/DiceButton.java
uckelman: I can see from the code that it does what the commit message says: `Replace Map with SortedMap and delete sortedKeys array`. But why is this being done? It doesn't seem related to the title of the PR.

uckelman: > The overall purpose (released with 3.7.16) was to provide as "summary" property that contains the roll results in the format requested in #[4427](https://github.com/vassalengine/vassal/issues/4427).

Those changes should be in a different PR from this one, as they're not relevant to fixing the whitespace in the summary strings.

vassal-app/src/main/java/VASSAL/build/module/RandomTextButton.java
rlament: This adds a space character after the comma separator and a space character before the multiplication character (comma and 'x' respectively for English). The whitespace character was in the translation file, in one case leading and the other trailing whitespace. In this case linefeed (\n) where in fact it appears that a space is preferable. Instead of the prepended/appended trailing whitespaces in the translation file, the space is hardcoded here.

Without this change, the other players in a multiplayer online game do not see the full results of a multi-die result. It seems that the parser on the receiving end gets confused by the linefeed characters.

In regards to the SortedMap/TreeMap: It is late in this project's lifetime to be optimizing working code. It's an opportunity to introduce bugs. The HashMap implementation is new for version 3.7.16. "New" in that this code was copied from another function replacing the prior sort algorithm here. I debated whether or not to include the SortedMap change and did expect some pushback. The SortMap implementation removes an array, the array copy and a sort operation. The SortedMap change is in a separate commit that can easily be reverted.

Going forward we have three different sort algorithms. There are four options:
1. ~~Revert back to 3.7.15 and forego the translation strings.~~
2. ~~Use the 3.7.15 sort and add the translation strings.~~
3. Use the 3.7.16 sort (HashMap) with the translation string fix (revert commit df8f534 of this merge request).
4. Use this merge request as is with SortedMap.

Which one is the overall preference?

Edit: Going back to 3.7.15 is not an option. That would remove the summary results field. Removed the sort such that this pull request only addresses the reported bug (option 3). Ready for re-review.

Issue #12051, opened 2023-02-27
Enhance Movement Trail to (optionally) remove last hop if Location Name does not change


BrentEaston:
See discussion https://forum.vassalengine.org/t/movement-trails-in-area-movement-games/76618/3

To maintain compatibility, we would add an option to the Movement Trail to ‘Only keep last position if location name does not change? Y/N’.

The Movement Trail trait keeps track of the trail for it’s piece, so it should be relatively simple to implement this completely within the Movement Trail trait itself. When a move completes, then if the option is set and the LocationName has not changed, then eat the last hop of the trail before adding the new one.

djberg96:
+1 for this, would definitely be nice for area impulse games!

djberg96:
@Niedowidek I don't think you need to ask, just submit a PR. :)

Niedowidek:
OK, thanks @djberg96  :)

Niedowidek:
Points are added to trail in
VASSAL.counters.Footprint#addPoint

Niedowidek:
Manuals to be updated:
game-pieces.adoc
MovementTrail.adoc

PR #14231, opened 2025-09-07
Fix12051 enhance movement trail

Niedowidek:
https://github.com/vassalengine/vassal/issues/12051
- added option "Only keep last position if location name does not change" to "Movement Trail" trait
- when the option is selected a movement trail would always show just a single stop in each area, regardless of how often the piece may have been moved within any of the areas that it moved through
- updated manual MovementTrail.adoc 
- updated a unit test

Should I also update game-pieces.adoc ? There is a different window presented (image132.png) which I don't know how to access?


vassal-app/src/main/java/VASSAL/counters/Footprint.java
```diff
@@ -89,12 +90,14 @@ public class Footprint extends MovementMarkable {
   protected int edgePointBuffer;               // How far Off-map to draw trail points (pixels)?
   protected int edgeDisplayBuffer;             // How far Off-map to draw trail lines (pixels)?
   protected String description;                // Description for this movement trail
+  protected boolean keepLastPosition = false;   // Only keep last position if location name does not change
```
jonathanrwatts: Your naming is very confusing; you seem to have set things up so that if `keepLastPosition` is true, you will throw out positions before the last position if the location name doesn't change.  While I understand what the code is doing *after* thinking about it for a minute, Just reading through the code (lines 346-347 in particular) it looks contradictory.  Can I suggest `keepLastPositionOnly` or `keepOnlyLastPosition` instead?

Niedowidek: Yes, right. Thank you. I corrected the name as you suggested to `keepLastPositionOnly`

rlament:
It would be better not to have imports from other folders in the project. `map/PieceMover.java` imports from `counters/Footprint.java`. Something like this is already done elsewhere, but this is not a good practise. Was wondering if this is absolutely necessary and investigated further, I came up with an alternate solution. See #14342 and let me know what you think. I separated the handling of the MOVED and MAYBE_MOVED properties.

Some issues with this implementation:
1. Off-board movement: A movement trail should include all off-board movement. Consecutive off-board locations should not be removed from the movement trail.
2. Movement between mats, which are in the same area, should be treated as a move to a new location.
3. Question: how should one treat the initial position? For the starting area should the movement trail record the initial (first) position or the last position before leaving the area? Based on the request in the Vassal Forum, I think it should be the latter. Quoting the post: "Ideally, no trails would appear until a piece is moved out of its starting area"

Niedowidek:
Hmm, on Build/Debian I get
```
vassal-app/src/main/java/VASSAL/counters/footprint/LastPositionRemover.java:19: error: package Strings does not exist
        && Strings.CS.equals(map.locationName((Point) dropTarget), map.locationName(actualPoint))) {
                  ^
vassal-app/src/main/java/VASSAL/counters/footprint/LastPositionRemover.java:27: error: package Strings does not exist
          && Strings.CS.equals(map.locationName(previousPoint), map.locationName(actualPoint))
                    ^
```

Seems weird. Why is it available in Windows build and not in Debian?

rlament:
> Why is it available in Windows build and not in Debian?

The Debian build doesn't use Maven. Dependencies are managed manually. Not sure what is missing. I got around it by not using Strings.CS.

uckelman:
Closed in favor of #14342.

PR #14244, opened 2025-09-22
Build a Debian package

cholmcc:
This commit adds the `debian` directory to the repository, and sets up the packaging workflow to also produce a Debian package.  The Debian package is architecture independent, and should be installable on most Debian derivatives.

The code is _not_ build by Maven, because many of the used Maven packages and plugins are not available in Debian and its derivatives. Debian packages _must_ build cleanly on a Debian system.  Thus, the code and documentation is compiled and build directly in the `debian/rules` file.

The Debian package integrates Vassal in to the desktop, and includes an enhanced launcher script, as well as a `man(6)` page, and indexed documentation for more system integrations.

When a new Vassal release is made, then the `debian/changelog` file should be updated too.  This can be done with the utility `dch` available on all Debian and derivative systems.  Perhaps we can find a way to automatise this in the future.

### DIFF
--- a/debian/vassal.6
+++ b/debian/vassal.6
@@ -78,11 +78,11 @@ Edit the module in the file \fImodule\fP (typically ending in
 \fB.vmod\fP). 
 .TP
 .BI \-\-edit\-extension \ extension
-Edit the extension \fIextension\fP (typically ending in \fB.vmdx\fP or
-\fB.vext\fP).  Note that the corresponding module (typically ending in
-\fB.vmod\fP) \fImust\fP be known to Vassal.  That is, you \fImust\fP
-have the module in your module library by opening the module at least
-once, before opening its extension.
+Edit the extension \fIextension\fP (typically ending in \fB.vmdx\fP).
+Note that the corresponding module (typically ending in \fB.vmod\fP)
+\fImust\fP be known to Vassal.  That is, you \fImust\fP have the
+module in your module library by opening the module at least once,
+before opening its extension.
 .TP
 .BI \-i,\-\-import \ adc2\-gamebox
 Imports the \fIAide de Champ\fP (version 2) gamebox
@@ -155,9 +155,6 @@ exchange in the PBEM play-mode.
 A module extension.  Module extensions are similar to modules, but
 augment a module with more elements, such as new counters, boards,
 cards, and so on.
-.TP
-.IB extension .vext
-Same as \fIextension\fB.vmdx\fP above.
 .PP
 All of these file types are really \fIZip\fP (\fBzip\fP(1)) archives.
 Modules and extensions contain an XML file that describes the content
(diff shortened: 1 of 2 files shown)


debian/vassal.6
```diff
@@ -0,0 +1,206 @@
...
+.B \-c,\-\-chatlog
+Echo error log to chat log.
+.TP
+.BI \-e,\-\-edit \ module
+Edit the module in the file \fImodule\fP (typically ending in
+\fB.vmod\fP). 
+.TP
+.BI \-\-edit\-extension \ extension
+Edit the extension \fIextension\fP (typically ending in \fB.vmdx\fP or
+\fB.vext\fP).  Note that the corresponding module (typically ending in
```
uckelman: `.vext` should not be listed here. It is not a supported extension extension.

debian/vassal.6
```diff
@@ -0,0 +1,206 @@
...
+Play log.  These contain the \fIfull\fP log (since the time of
+recording) of a play of a module.  This is typically what one would
+exchange in the PBEM play-mode.
+.TP
+.IB extension .vmdx
+A module extension.  Module extensions are similar to modules, but
+augment a module with more elements, such as new counters, boards,
+cards, and so on.
+.TP
+.IB extension .vext
```
uckelman: Ditto. Do not use or even mention `.vext`.

debian/vassal.sharedmimeinfo
```diff
@@ -0,0 +1,28 @@
...
+   <mime-type type="application/x-vassal-save">
+     <comment>VASSAL save file</comment>
+     <glob pattern="*.vsav"/>
+     <generic-icon name="application-x-vassal"/>
+     <sub-class-of type="application/zip"/>
+   </mime-type>
+   <mime-type type="application/x-vassal-extension">
+     <comment>VASSAL module extension</comment>
+     <glob pattern="*.vmdx"/>
+     <glob pattern="*.vext"/>
```
uckelman: Remove `.vext`.

debian/vassal.6
cholmcc: Fixed

debian/vassal.6
cholmcc: Fixed

debian/vassal.sharedmimeinfo
cholmcc: Fixed

.gitignore
```diff
@@ -28,3 +29,22 @@ snapshots/
...
+.doc
+.jar
+.manifest
+.stage
+debian/.debhelper
+debian/debhelper-build-stamp
+debian/files
+debian/vassal.substvars
+debian/vassal
+deprecated
```
uckelman, line 50: Where are these files outside of `debian`? I'm not seeing them as untracked in local builds.

cholmcc: These are generated during the build of the Debian package, and should therefore be ignored.

PR #14259, opened 2025-10-01
Load Continuation... moved down the Tools menu order

riverwanderer:
Simple change to move Load Continuation to a more logical menu position and reduce accidental hits in place of "Refresh Counters" which is the top item in the menu.

Current (3.7.18) menu:
image

This PR moves Load Continuation to below the Load Log options, thus:
image


rlament:
If the problem is accidental hits of "Refresh Counters" then it's the one that should be moved down perhaps with separators both above and below it. I find that I use "Load Continuation" quite a bit; the other menu items almost not at all.

The refresh and load menu items open dialog boxes. For consistency, all of them should have the triple dot suffix.

Is it time to remove "Load Continuation" from the File menu?

PR #14260, opened 2025-10-02
Reorder items on Player Tools menu to put all Loads together

riverwanderer:
Simple change to move Load Continuation to a more logical menu position and reduce accidental hits in place of "Refresh Counters" which is the top item in the menu. Supercedes #14259.

Current (3.7.18) menu:
image

This PR moves Load Continuation to below the Load Log options and Refresh into the "developer" section along with Debug, thus:
image

In addition, ellipses are now added consistently to the Load... options and the deprecated Load Continuation File menu item and associated actions is removed.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/GameState.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameState.java
@@ -192,7 +192,6 @@ public class GameState implements CommandEncoder {
         }
       }
     };
-    loadContinuation.setEnabled(false);
 
     loadAndFastForward = new AbstractAction(Resources.getString("GameState.load_and_fast_forward")) {
       private static final long serialVersionUID = 1L;


riverwanderer:
@rlament - revised PR, covering your suggestions (ref. PR#14259), including removing the deprecated Load Continuation.

vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
```diff
@@ -143,14 +142,14 @@ private void setupMenubarAndActions() {
       new MenuProxy(Resources.getString("General.tools"));
     toolsMenu.setMnemonic(Resources.getString("General.tools.shortcut").charAt(0));
 
-    toolsMenu.add(mm.addKey("GameRefresher.refresh_counters"));
```
rlament, line 146: Github doesn't seem to allow commenting on unchanged lines, so adding a comment here. Earlier in this file there is a FIXME around line 123. The FIXME can be deleted. The mnemonic lookup is on a string in the resource file. There should not be any collisions unless the person doing the translation picks unwisely. There is nothing here in this java code to fix.

There's a similar set (5 by my count) of FIXME in GameState.java, in function addTo(). These can be deleted for the same reason.

vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
```diff
@@ -143,14 +142,14 @@ private void setupMenubarAndActions() {
...
-
-    toolsMenu.addSeparator();
     toolsMenu.add(mm.addKey("GameState.load_and_fast_forward"));
     toolsMenu.add(mm.addKey("GameState.load_and_append"));
 
+    toolsMenu.add(mm.addKey("GameState.load_continuation"));
+
     toolsMenu.addSeparator();
+
+    toolsMenu.add(mm.addKey("GameRefresher.refresh_counters"));
```
rlament, line 151: Refresh Counters also opens a dialog. It should have ellipsis suffix too.

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
```diff
@@ -781,8 +781,8 @@ GameState.dont_prompt_again=Don't Prompt Again
 GameState.old_continuation_title=Load Continuation Has Moved
 GameState.old_continuation_heading=Load Continuation Has Moved
 GameState.old_continuation_warning=Loading vlogs directly (with "Load Game or Log") is usually the safest way to replay them.\n\nLegacy continuation loading often creates confusion and is very sensitive to the slightest player mistake or accidental procedural misstep. Only very special use cases require legacy loading.\n\nLoad Continuation is now found on the Tools menu.
-GameState.load_and_fast_forward=Load Log and Fast Forward
-GameState.load_and_append=Load Log, Fast Forward, and Append
+GameState.load_and_fast_forward=Load Log & Fast Forward...
+GameState.load_and_append=Load Log, Fast Forward & Append...
```
rlament, line 783: Debatable whether shortening the 'and' is necessary. FYI: The text for the Polish translation is longer.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -181,24 +181,6 @@ public void actionPerformed(ActionEvent e) {
...
-      private static final long serialVersionUID = 1L;
-
-      @Override
-      public void actionPerformed(ActionEvent e) {
-        ProblemDialog.show(
-          JOptionPane.INFORMATION_MESSAGE,
-          GameModule.getGameModule().getPlayerWindow(),
-          null,
-          Resources.getString("GameState.old_continuation_title"),
-          Resources.getString("GameState.old_continuation_heading"),
```
rlament, line 195: The `old_continuation_title` and `old_continuation_heading` resource strings can be deleted from the resources files. These are the main resource file and the Spanish and Polish translations. My local build succeeded. Not sure if the build server will complain.

rlament:
> I had to leave _loadGameOld_ defined, in order to get past a build error.

I believe the semantic versioning won't let you delete class members. If the version number ever gets to 3.8 then this variable could be deleted. Add a FIXME at the declaration as a reminder.

vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
riverwanderer: >  Earlier in this file there is a FIXME around line 123. The FIXME can be deleted. The mnemonic lookup is on a string in the resource file. There should not be any collisions unless the person doing the translation picks unwisely. There is nothing here in this java code to fix.
> 
> There's a similar set (5 by my count) of FIXME in GameState.java, in function addTo(). These can be deleted for the same reason.

How about I change the FIXMEs to a comment warning translators? Should that go in the Properties, however? And anyway, what is an example of a "collision" here and its effect?

vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
riverwanderer: I was trying to limit the scope of these adjustments (there are more), but I will do this one and any more I find in PlayerWindow. I am planning to try using the ellipsis character (…) as this is what I think system menus use.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
riverwanderer: I will give this a go, initially in a separate PR  as that will be easier for me to rollback from. I think it will be ok.

vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
riverwanderer: I have removed these FIXMEs.

vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
riverwanderer: Some of these functions do not **_require_** further input beyond an "OK". If we start giving those ellipsis suffixes, they clutter up the menus. I did want to add one to "Remove unused images" but I could not find a property entry for that (the one that is there is a window heading). There is a literal in the code but I didn't investigate that yet.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
riverwanderer: I just removed them all (plus in the Danish properties file). It built ok on the server.

vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
rlament: Ok, fair enough, no ellipsis.

BTW Unicode for ellipsis is \u2026. Although all current menus use three dots,.it might be better to just stay consistent.

"Remove unused images" is an editor menu. I would consider that change out of scope for this PR since this is mainly a Vassal Player PR.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -177,27 +178,8 @@ public void actionPerformed(ActionEvent e) {
...
-          Resources.getString("GameState.old_continuation_warning")
-        );
-      }
-    };
-    loadGameOld.setEnabled(false);
+    loadGame.putValue(Action.MNEMONIC_KEY, (int)Resources.getString("GameState.load_game.shortcut").charAt(0));
```
rlament, line 182: Another comment on an unmodified line: The `loadContinuation.setEnabled(false)` on about line 195 is redundant and can be deleted. The enable state is set at the end of the function on around line 288. It's not necessary to do it twice.

vassal-app/src/main/java/VASSAL/launch/PlayerWindow.java
rlament: Collision as in two or more menu items using the same shortcut key. One, likely the first one, will mask the others preventing some commands from being launched from a keyboard key sequence.

rlament (approved):
Looks good to me.

PR #14266, opened 2025-10-03
Special dice button sequencing / undo fix

### LINKED ISSUE #14237: Symbolic Dice Button breaks atomic undo
A Symbolic Die roll is made and an action is generated from a hotkey configured on the Symbolic Dice component.

Example of such a component, configured with an action Hotkey and a report:

Image

Now attempt an undo... the die roll report is undone but **not** the action (here _fmtFRBattleRoll_)  The action in this example involves moving cards between maps/decks.

If the dice roll in bundled into an action button on a PIECE, then undo works as it should - one undo action undoes the entire die roll & action sequence.

riverwanderer:
AI (Gemini)-guided fix for [14237]( https://github.com/vassalengine/vassal/issues/14237).

Summary from the AI sessions; changes primarily involve:
1. Adding the lastRollResults field to track the visual state for undo.
2. Updating the DR() method to implement the new roll/undo logic, including capturing the previousRollResults, using sendAndLog(finalCommand), and updating lastRollResults. Essentially integrating functionality from DoActionButton including loop control and pre/post Global Hotkeys.
3. Modifying reportResults(int[] results) to only update the visual state (as the chat/property update is now handled elsewhere in DR()).
4. Updating addTo(Buildable parent) to initialize lastRollResults.
5. Refactoring the ShowResults command to accept and manage both rolls (new state) and previousRolls (state to restore on undo) to support the new undo functionality.
6. Adding a two-argument constructor to ShowResults for binary compatibility.
7. Updating encode and decode in the main class to handle the new ShowResults constructor signature


riverwanderer:
PR tested OK against the Undo and Remote sequencing issues reported in the linked issues.

riverwanderer:
Unfortunately, the change nerfs Looping and the associated Pre & Post action buttons. Doubly unfortunate, since these functions are really redundant and it would have been better if Logging/UnDo worked correctly, even without looping (which a GHK to a normal action button can accomplish).

More work required.

Issue #13018, opened 2023-12-26
All Deck GKC's to be Global GKC's


BrentEaston:
Add a checkbox to Deck GKC's to affect pieces outside the Deck.

PR #14300, opened 2025-10-17
Add restrictRange attribute to DeckGlobalKeyCommand

gantoin:
Related to https://github.com/vassalengine/vassal/issues/13018

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/DeckGlobalKeyCommand.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/DeckGlobalKeyCommand.java
@@ -74,12 +74,16 @@ import java.util.List;
  */
 public class DeckGlobalKeyCommand extends MassKeyCommand implements RecursionLimiter.Loopable {
 
+  public static final String RESTRICT_RANGE = "restrictRange"; //$NON-NLS-1$
+  protected boolean restrictRange = true;
+
   public DeckGlobalKeyCommand() {
     globalCommand = new DeckGlobalCommand(this);
     globalCommand.setReportSingle(true);
     globalCommand.setSuppressSounds(false);
     setConfigureName("");
     setShowDisabledOptions(false);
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/DeckGlobalKeyCommand.java
```diff
@@ -278,6 +318,20 @@ public String getComponentTypeName() {
...
+
+      // If restrict range is false, apply to all pieces on the map
+      if (!dgkc.isRestrictRange()) {
+        final Map map = d.getMap();
+        if (map != null) {
+          return apply(new Map[]{map}, filter, null);
+        }
+        else {
+          return new NullCommand();
+        }
```
rlament, line 331: Should this 'else' block be deleted? It seems that decks may not necessarily have an associated map. In that case shouldn't this function continue with the original behavior and apply only to pieces in the deck. In other words, not return here.

vassal-app/src/main/resources/VASSAL/i18n/Editor.properties
```diff
@@ -667,6 +667,7 @@ Editor.Deck.load_deck=Load Deck
 Editor.DeckGlobalKeyCommand.command=Global Command
 Editor.DeckGlobalKeyCommand.matching_properties=Matching expression
 Editor.DeckGlobalKeyCommand.affects=Affects
+Editor.DeckGlobalKeyCommand.restrict_range=Restrict to pieces in Deck
```
rlament, line 670: "Apply" is the word associated with Global Key Commands. "Restrict" could be confused with the "Restrict..." traits. Take a look at `Editor.MassKey.deck_content`. Something like `Apply only to contents of Decks` may be more explanatory and inline with the existing terminology. However this doesn't indicate to the user that the Global Key Command applies to all pieces on the map when unchecked. Naming is hard.

PR #14305, opened 2025-10-20
Fix for invalid filenames when extracting archive

cholmcc:
When extracting archive members, fix up the file name so that it is
valid on the target platform.   Files are extracted from the archive
when the embedded HTML viewer needs a file - e.g., an image.

This adds the method `santizeFileName` to `FileUtils`, and uses that in `DataArchiveHTMLEditorKit#DataArchiveImageView`.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,20 +5,21 @@
 
 3.7.19 - ???
 
+* 14318: Upgraded spotbugs-annotations to 4.9.8
+* 14311: Upgraded logback-classic to 1.5.20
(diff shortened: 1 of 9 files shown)


uckelman:
There's a good supply of examples where this will fail [here](https://github.com/vassalengine/game-library-service/blob/main/src/upload.rs#L50).

E.g., `COM` and `PRN` will both be turned into `___` and now you'll have a name collision.

It strikes me as more time efficient to rename the problematic file than to change the code.

uckelman:
Fixed in #14323 instead.

PR #14342, opened 2025-11-08
12051 Add Movement Trail option to keep only last position per location

### LINKED ISSUE #12051: Enhance Movement Trail to (optionally) remove last hop if Location Name does not change

rlament:
This is an alternate implementation to pull request #14231 which addresses and closes #12051

Add an option to keep the last movement trail position only when location name or mat change.
All off-board movement is always appended to the movement trail.
Check for mat to mat movement within a location.
Add unit tests for movement trails.
Update movement trails documentation adding the new option.

Closes #12051

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -61,7 +61,6 @@ import VASSAL.tools.image.ImageUtils;
 import VASSAL.tools.imageop.Op;
 import VASSAL.tools.swing.SwingUtils;
 
-import org.apache.commons.lang3.Strings;
 import org.apache.commons.lang3.SystemUtils;
 
 import javax.swing.ImageIcon;
@@ -778,7 +777,7 @@ public class PieceMover extends AbstractBuildable
         else {
           // Compare location names
           final String previousLocation = map.locationName(p.getPosition());
-          locDefinitelyChanged = !Strings.CS.equals(previousLocation, map.locationName(loc));
+          locDefinitelyChanged = (previousLocation == null) || !(previousLocation.equals(map.locationName(loc)));
           if (!locDefinitelyChanged && GameModule.getGameModule().isMatSupport()) {
             final String mat = (String) p.getProperty(MatCargo.CURRENT_MAT_ID);
             final String oldMat = (String) p.getProperty(BasicPiece.OLD_MAT_ID);
--- a/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
@@ -53,6 +53,7 @@ import VASSAL.counters.PieceVisitorDispatcher;
 import VASSAL.counters.Properties;
 import VASSAL.counters.PropertyExporter;
 import VASSAL.counters.Stack;
+import VASSAL.i18n.Resources;
 import VASSAL.tools.DebugControls;
 import VASSAL.tools.FormattedString;
 import VASSAL.tools.LaunchButton;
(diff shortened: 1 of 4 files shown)


vassal-app/src/main/java/VASSAL/build/module/map/PieceMover.java
```diff
@@ -766,7 +766,28 @@ protected Command movedPiece(GamePiece p, Point loc) {
     Command c = new NullCommand();
     c = c.append(setOldLocations(p));
     if (!loc.equals(p.getPosition())) {
-      c = c.append(markMoved(p, true, false));
+      boolean locDefinitelyChanged = false;
```
Niedowidek, line 770: suggestion (non-blocking): I think whole locDefinitelyChanged computation should go to a separate method and then to a new class (I prefer not to bloat classes that already have hundreds of lines).

vassal-app/src/main/java/VASSAL/counters/Footprint.java
```diff
@@ -243,7 +252,16 @@ public List<String> getMenuTextList() {
 
   @Override
   public void setProperty(Object key, Object val) {
-    if (Properties.MOVED.equals(key) || Properties.MAYBE_MOVED.equals(key)) {
+    if (keepLastPositionOnly && Properties.MAYBE_MOVED.equals(key)) {
```
Niedowidek: todo (blocking): Unfortunately it doesn't work correctly. See https://github.com/vassalengine/vassal/pull/14342#issuecomment-3753155098

Niedowidek (changes requested):

Niedowidek:
@rlament Regarding "off board moves should always be recorded by movement trails" - how can I reproduce that? Can I do this in Turning Point Stalingrad or should I use another mod?

rlament:
Without any code changes, I got the  movement trails working in the Stalingrad. It took the longest time to figure out what module change "fixed" it. It turns out there is a `trueMovedSupport` flag that must be set when ignoring moves within the same location. `MovementMarkable` shares functionality with `Footprint`. When I checked the 'Mark When Moved' checkbox `Ignore moves which don't change either Location Name or Mat` that unknowingly set the flag. Added setting of the flag in `Footprint` and that seems to fix it. I didn't see this earlier because my test module had the `Ignore moves` checkbox checked. In the latest commit I did some renaming to better align with the `MovementMarkable` terminology.

Niedowidek:
@rlament Great ! Does it mean we can go with your solution ?

rlament:
@Niedowidek If you have confirmed the latest change fixes the behaviour in your module then please mark your approval for this PR. Thanks for the review and leading the way in highlighting the relevant parts of the code. If you had not initiated a PR for this feature request, it would not have happened. Next time pick an easier "good first issue". 😄

vassal-app/src/main/java/VASSAL/counters/Footprint.java
```diff
@@ -150,7 +154,12 @@ public void mySetState(String newState) {
       }
     }
     everInitialized = ss.nextBoolean(false);
+    lastPointListSize = ss.nextInt(items);
     requireNoOuterRotate();
+    if (ignoreSameLocation) {
+      GameModule.getGameModule().setTrueMovedSupport(true);
```
Niedowidek, line 160: praise: this is the line we needed :) Nice one.

vassal-app/src/main/java/VASSAL/counters/Footprint.java
```diff
@@ -243,7 +256,16 @@ public List<String> getMenuTextList() {
 
   @Override
   public void setProperty(Object key, Object val) {
-    if (Properties.MOVED.equals(key) || Properties.MAYBE_MOVED.equals(key)) {
+    if (ignoreSameLocation && Properties.MAYBE_MOVED.equals(key)) {
```
Niedowidek, line 259: question (non-blocking): Can you explain why it works ? What's the difference between MOVED and MAYBE_MOVED, especially in the context of this issue? I'm curious.

Niedowidek (changes requested):
"every position on the trail in the Volga should be visible" still needs to be implemented.

Niedowidek:
Volga river movement looks fine. Off-board movement too. I can see that it registers every position on the trail off-board even though they are not visible.

image

With the option turned off it works fine too:

image

Niedowidek:
@rlament I think it works OK now.

Niedowidek (approved):

vassal-app/src/main/java/VASSAL/counters/Footprint.java
rlament: One can't just look at this one line in isolation. This 'if' block here is inserted before the existing code. Notice that the 'else' is the old code and handles the usual case where it adds every position to the trail.

`PieceMover.movedPiece()` does the pre-move region check. You did mention making the added code a separate module/function. That would be good especially if it is used in multiple places. `EditablePiece.checkTrueMoved()` does the post-move region check which is similar but a little different.

Getting back to `PieceMover.movedPiece()`, if the region is different, `locDefinitelyChanged` is true. This flag passes to the `markMoved` command. If the flag is set, it calls `setProperty` on MOVED, if false MAYBE_MOVED. So it sets one property or the other, never both. MOVED adds a point to the trail. MAYBE_MOVED deletes the last point in the trail. The `lastPointListSize` variable ensures it only deletes one point per region.

`EditablePiece.checkTrueMoved()` calls `setProperty` on MOVED only if the region changed. Therefore when the unit moves to a new region MOVED is set twice but the second one is effectively a no-operation. When the unit moves within the same region MAYBE_MOVED is set once.

Hopefully that makes some sense. It took a while to clue into what is going on here.

Niedowidek: Thank you very much for your explanation.

PR #14347, opened 2025-11-13
12033 Log more info for the piece not on map exception

rlament:
Report the piece name and the maps that cause the infrequent but chronic `IllegalArgumentException: Piece is not on this map`. Some additional information may help resolve the root cause of this exception. Sample output using a fabricated name and locations:

java.lang.IllegalArgumentException: Piece is not on this map: Main Map. Piece 'Superb Infantry Training' is on Player Mat.


rlament:
> could the cause of OldMap being null be related to the same underlying issue as the Dice Properties one?

It is possible that the issue with the OldXXX properties is similar. I'd had the same thought and ran a quick test with the dice properties fix. That didn't work. Will have to run it through the debugger and see what it is doing. It is possible that a prior undo operation some time earlier sets up the conditions for the failure. Have to look at which command changes the Map property.

Issue #14371, opened 2025-11-26
Window placement: ensure monitor-aware window and dialog positioning; prevent snap-back to primary display


marangonico:
**Why this issue**
This issue addresses problems related to multi-monitor configurations, including cases with non-aligned desktop offsets, where windows and dialogs may open partially or entirely off-screen.
The issue was reproduced on Linux + KDE; the PR has been tested on Linux + KDE and Windows.

**Summary**
Windows and dialogs should open on the most appropriate monitor and stay fully visible across multi‑monitor configurations. The accompanying PR makes placement consistent and monitor‑aware, avoiding the frequent “snap‑back” to the primary monitor. Dialogs are centered on the same monitor as their parent window when no explicit bounds are provided.

**Current behavior (before PR)**
- Some dialogs open on the primary monitor (screen 0), even if the parent window is on a secondary display.
- If the parent/owner window is not yet visible, final clamping can reassign dialogs to the primary monitor.
- Main window placement in single‑window mode did not uniformly use the common persistence/placement logic.

**Expected behavior**
- Main windows and dialogs open on the most appropriate monitor:
  - Restore previous bounds when available.
  - Otherwise use sensible defaults, select the best monitor, and clamp within its usable area.
  - Never open off‑screen.
- Dialogs without saved bounds are centered on the same monitor as their parent window.

**What the PR changes (high level)**
- Consistent, monitor‑aware placement across the app:
  - GameModule main window uses PositionOption in all modes; bounds persisted under `"BoundsOfGameModule"` with sensible defaults.
  - Dialog creation prefers the module’s PlayerWindow (or otherwise a suitable owner), computes the target monitor from the intended owner location even if not yet visible, then centers the dialog on that monitor.
  - SwingUtils.ensureOnScreen clamps using the window’s current bounds to the screen with the largest intersection (or nearest if none), preventing snap‑back to the primary display.

**Verification steps**
1. Move the main window (PlayerWindow) to a secondary monitor and open a dialog: it should appear centered on that monitor.
2. Resize a dialog beyond the target monitor size: it should be reduced and remain fully visible.
3. Disconnect/reconnect the secondary monitor and relaunch: the main window and dialogs should appear on an active monitor with sensible placement.
4. Open a dialog while the owner window is not yet visible: it should still open on the intended monitor (not screen 0).

**Edge cases considered**
- Different monitor resolutions and offsets (non‑aligned desktops).
- Owner window not yet showing (no reliable GraphicsConfiguration).
[...]

PR #14372, opened 2025-11-27
Feature/window placement engine

marangonico:
### Summary

This PR improves window and dialog placement in multi-monitor configurations, including setups with non-aligned desktop offsets. Windows and dialogs now open on the most appropriate monitor, remain fully visible, and no longer “snap back” to the primary display.

The issue was reproduced on Linux + KDE; the changes have been tested on Linux + KDE and Windows.

### What changes

* Window placement is now consistently **monitor-aware** across the application.
* The GameModule main window uses `PositionOption` in all modes, with bounds persisted under `"BoundsOfGameModule"` and sensible defaults.
* Dialogs prefer the module’s `PlayerWindow` (or another suitable owner) and are centered on the same monitor as their parent when no saved bounds exist.
* Target monitor selection works even when the owner window is not yet visible.
* `SwingUtils.ensureOnScreen` clamps windows to the screen with the largest intersection (or nearest one), preventing reassignment to the primary monitor.

### Result

* Main windows and dialogs restore previous bounds when available, otherwise use safe defaults.
* Windows never open off-screen and always stay within the usable area of the selected monitor.
* Behavior is robust across different resolutions, offsets, DPI/insets, and maximized state restoration.

### Adresses:
- #14371

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/documentation/HelpWindow.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/documentation/HelpWindow.java
@@ -25,6 +25,8 @@ import javax.swing.event.HyperlinkListener;
 
 import VASSAL.tools.menu.MenuManager;
 import VASSAL.tools.swing.HTMLWindowHelper;
+import VASSAL.preferences.PositionOption;
+import VASSAL.preferences.Prefs;
 
 /**
  * A Window that displays HTML content, with navigation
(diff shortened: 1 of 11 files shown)


vassal-app/src/main/java/VASSAL/build/module/documentation/HelpWindow.java
```diff
@@ -38,7 +40,13 @@ public HelpWindow(String title, URL contents) {
     super(title);
     setDefaultCloseOperation(HIDE_ON_CLOSE);
     setJMenuBar(MenuManager.getInstance().getMenuBarFor(this));
+
+    // Persist and restore window bounds using the common layout engine
+    final PositionOption option = new PositionOption(PositionOption.key + "HelpWindow", this); //NON-NLS
+    Prefs.getGlobalPrefs().addOption(option);
+
     helper.setup(this, contents);
+
```
rlament, line 49: `addOption()` eventually invokes `setFrameBounds()` and in there it uses the position and size of the frame from V_Global. `helper.setup()` sets the window size based on the screen bounds since the `pack()` function doesn't produce reasonable dimensions. It is checking one frame for monitor determination but using another for the actual display.

PR #14386, opened 2025-12-08
14237 Fix Symbolic Dice Button atomic undo

### LINKED ISSUE #14237: Symbolic Dice Button breaks atomic undo

rlament:
Expanded ShowResults command to include the previous roll(s) using the ChangePropertyCommand as an example. Despite the addition, ShowResults remains backward compatible. Including the previous roll(s) in the command allows remote undo during online play. The parameter list and member variables of the ShowResults command follow the ChangePropertyCommand convention.

Made the construction of SpecialDiceButton class a litle bit more object oriented no longer requiring a special constructor to by-pass base class initialization. Added the prepareActionsCommand() function as a precursor to doActions() permitting child classes to perform custom setup. Deprecated the DR() function at the same time making it an empty function.

Tested pre/post commands with looping for online, standalone and log playback.

For proper update of dice properties during online play, PR #14382 is required.

Closes #14237

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/SpecialDiceButton.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/SpecialDiceButton.java
@@ -41,8 +41,6 @@ import VASSAL.search.HTMLImageFinder;
 import VASSAL.tools.FormattedString;
 import VASSAL.tools.KeyStrokeListener;
 import VASSAL.tools.LaunchButton;
-import VASSAL.tools.RecursionLimitException;
-import VASSAL.tools.RecursionLimiter;
 import VASSAL.tools.SequenceEncoder;
 import VASSAL.tools.UniqueIdManager;
 import VASSAL.tools.imageop.ImageOp;
@@ -52,6 +50,7 @@ import net.miginfocom.swing.MigLayout;
 import org.apache.commons.lang3.ArrayUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.w3c.dom.Element;
 
 import javax.swing.Icon;
 import javax.swing.ImageIcon;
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/SpecialDiceButton.java
```diff
@@ -150,33 +157,40 @@ protected String getReportPrefix() {
     return " *** " + getConfigureName() + " = "; //$NON-NLS-1$ //$NON-NLS-2$
   }
 
+  @Deprecated(since = "2025-12-04", forRemoval = true)
+  protected void DR() {
+  }
```
uckelman, line 162: Won't this break anything that had been calling `DR` and expecting it to do something?

rlament: > Won't this break anything that had been calling DR and expecting it to do something?

Do you know of any module with custom code overriding SpecialDiceButton?

I've added the call to doActions from DR, but not convinced that this is right. DR was wired to the button click as the target of an ActionEvent. DR shouldn't really be called directly. The ActionEvent (in DoActionButton) is wired to doActions that calls prepareActionsCommands which does what DR used to do.

uckelman: > Do you know of any module with custom code overriding SpecialDiceButton?

I don't. Extracting all the custom code from the module library is a lot of work, and even then isn't conclusive as there could be modules which extend SpecialDiceButton which aren't there.

PR #14412, opened 2025-12-30
Look and Feel for the Module Manager using FlatLaf and user selectable themes

### LINKED ISSUE #12821: Fix text colour in Module column of Module Manager
Should be same colour as the other columns, not black.

See discord discussion [here](https://discord.com/channels/740127632586178572/911006365626036294/1163374790179033209) for background and screenshots.

Analysis from @uckelman : 
image

rlament:
This PR addresses the color contrast in the Module Manager and adds user selectable themes. The selected theme is stored in the global preferences.

Not tested on the MacOS. Need someone to give it a try.

Select the preferred theme from the Module Manager: File->Preferences->Look and Feel->Save->Quit and restart. Restart is required to refresh all visible windows. New windows, such as from launching a game module, will follow the newly selected theme without a restart.

This PR creates some downstream visibility issues. Some of the default buttons (server, camera, zoom) are hard to see in dark mode. Messages in the chat log are hard to read in dark mode. However, the player window and chart windows are still in classic white and shouldn't be in dark mode. Some more work to do if this PR has traction.

Closes #12821

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -211,6 +211,10 @@ jobs:
             sed -e '1{N;s/\n/ /;}' -e 's/ *([^)]*)//g' -e 's/,//g' )
           sudo apt install -y build-essential devscripts fakeroot ${dep}
 
+      - name: Install flatlaf
+        run: |
+          wget -O /usr/share/java/flatlaf.jar https://repo1.maven.org/maven2/com/formdev/flatlaf/3.7/flatlaf-3.7.jar
+
       - name: Make orig archive
         run: |
           package=$(dpkg-parsechangelog --show-field Source)


riverwanderer:
Confirmed fixes #12821 when tested on MacOS 12.7.6 (Monterey).

Also tested "Dark" mode; it works though it is obvious that colour schemes developed without this feature in mind - including the default chat colour - will present readability issues (just an observation):
e.g.
image

riverwanderer:
Continued working through all modes on the Mac, running up the Here I Stand module each time. All look ok to me; I perceived the differences between some modes but didn't look closely enough to see how Light varies from IntelliJ and Dark from Darcula.

riverwanderer:
> However, the player window and chart windows are still in classic white and shouldn't be in dark mode.

I'm seeing these windows in whatever mode is set - including in Dark mode. Aren't you expecting that, or have I misunderstood you here @rlament ?

rlament:
Thanks for the in depth review. Glad to hear that Mac look and feel works with similar issues.

> I perceived the differences between some modes but didn't look closely enough to see how Light varies from IntelliJ and Dark from Darcula.

The differences are subtle. Little things such as check box, radio button and arrow styles just to name a few. Perhaps not very obvious on the Mac.

> > However, the player window and chart windows are still in classic white and shouldn't be in dark mode.
> 
> I'm seeing these windows in whatever mode is set - including in Dark mode. Aren't you expecting that, or have I misunderstood you?

For Map, Private and Player Windows (an incomplete list) the background color can be specified. In the modules I happened to look at modules with backgrounds set to white, which overrides the Look and Feel. If there is a Map within a Chart, the background color can be set there too. For these windows, the toolbar, window border and decorations follow the look and feel, but not the background color.

PR #14427, opened 2026-01-06
Check files in FileConfigurer are not the empty string, due to a change in Java 25

uckelman:


vassal-app/src/main/java/VASSAL/configure/FileConfigurer.java
```diff
@@ -93,7 +93,7 @@ public String getValueString() {
   public void setValue(Object o) {
 // FIXME: this creates a problem when the referenced file is in the JAR
     final File f = (File) o;
-    if (f != null && f.exists()) {
+    if (f != null && !f.getPath().isEmpty() && f.exists()) {
```
rlament, line 96: For forward and backward compatibility, the code should refrain from using an empty string to denote no file selection. With that consideration, the change should be in the other setValue function. Change the test to check for null or empty, if so then set the value to null.
```
  public void setValue(String s) {
    if (StringUtils.isEmpty(s))
      setValue((Object) null);
    else {
      setValue(new File(s));
    }
  }
```
This one line change by itself fixes the problem. The change in `setValue(Object o)` could remain but I'd revert it in case it masks some other issue.

PR #14433, opened 2026-01-07
Fix for change in `java.io.File` API in Java 25

cholmcc:
In particular the test if a file exists `new File("").exists()` now
returns `true` rather than false.  And `File("")` points to the run-time
default directory (`user.dir`) rather than to nothing. See also

    https://www.oracle.com/java/technologies/javase/25all-relnotes.html#JDK-8024695

The fix changes the `FileChooser` so that files that are opened for
reading are checked for existence (using `isFile`) before returning to
the caller.  If the file isn't found, then return `ERROR_OPTION`, so
that the client code can simply assume a return value of
`ACCCEPT_OPTION` means the chosen file is valid and can be used.

Some other similar fixes around in the code.

These all simplify the client code.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -697,7 +697,7 @@ public class GameModule extends AbstractConfigurable
     final DataArchive darch = getDataArchive();
 
     final File f = new File(darch.getName());
-    if (!f.exists() || f.length() == 0) {
+    if (!f.isFile() || f.length() == 0) {
       // new module, no buildFile
       build(null);
     }
--- a/vassal-app/src/main/java/VASSAL/build/module/GameState.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameState.java
@@ -801,7 +801,7 @@ public class GameState implements CommandEncoder {
     final GameModule g = GameModule.getGameModule();
 
     try {
-      if (!f.exists()) throw new FileNotFoundException("Unable to locate " + f.getPath());
+      if (!f.isFile()) throw new FileNotFoundException("Unable to locate \"" + f.getAbsolutePath() + "\"");
 
       // Check the Save game for validity
       if (!isSaveMetaDataValid(f)) {
@@ -1016,7 +1016,7 @@ public class GameState implements CommandEncoder {
   }
 
   protected boolean checkForOldSaveFile(File f) {
-    if (f.exists()) {
+    if (f.isFile()) {
       // warn user if overwriting a save from an old version
       final AbstractMetaData md = MetaDataFactory.buildMetaData(f);
       if (md instanceof SaveMetaData) {
--- a/vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
@@ -795,7 +795,7 @@ public class WizardSupport {
           @Override
           public void propertyChange(PropertyChangeEvent evt) {
             final File f = (File) evt.getNewValue();
-            if (f == null || !f.exists()) {
+            if (f == null || !f.isFile()) {
               controller.setProblem(Resources.getString("WizardSupport.NoSuchFile")); //$NON-NLS-1$
             }
             else if (f.isDirectory()) {
--- a/vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
@@ -252,12 +252,9 @@ public class ImageSaver extends AbstractToolbarItem {
       files.add(f);
 
       // make sure that we can write the file before proceeding
-      if (f.exists()) {
-        if (!f.canWrite()) {
-          throw new IOException(
-            "Cannot write to the file \"" + f.getAbsolutePath() + "\""
-          );
-        }
+      if (f.exists() && (f.getPath().isEmpty() || !f.canWrite())) {
+        throw new IOException(
+          "Cannot write to the file \"" + f.getAbsolutePath() + "\"");
       }
       else {
         final File p = f.getParentFile();
--- a/vassal-app/src/main/java/VASSAL/build/module/metadata/MetaDataFactory.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/metadata/MetaDataFactory.java
@@ -54,7 +54,7 @@ public class MetaDataFactory {
   public static AbstractMetaData buildMetaData(File file) {
 
     // Check the file exists and is a file
-    if (file == null || !file.exists() || !file.isFile())
+    if (file == null || !file.isFile())
       return null;
 
     try (ZipFile zip = new ZipFile(file)) {
--- a/vassal-app/src/main/java/VASSAL/configure/IconConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/configure/IconConfigurer.java
@@ -145,15 +145,12 @@ public class IconConfigurer extends Configurer {
     if (fc.showOpenDialog(getControls()) != FileChooser.APPROVE_OPTION) {
       setValue(null);
       setToolTipText(null);
+      return;
     }
-    else {
-      final File f = fc.getSelectedFile();
-      if (f != null && f.exists()) {
-        GameModule.getGameModule().getArchiveWriter()
-                                  .addImage(f.getPath(), f.getName());
-        setValue(f.getName());
-        setToolTipText(f.getPath());
-      }
-    }
+    final File f = fc.getSelectedFile();
+    GameModule.getGameModule().getArchiveWriter()
+        .addImage(f.getPath(), f.getName());
+    setValue(f.getName());
+    setToolTipText(f.getPath());
   }
 }
--- a/vassal-app/src/main/java/VASSAL/configure/ImageSelector.java
+++ b/vassal-app/src/main/java/VASSAL/configure/ImageSelector.java
@@ -191,7 +191,7 @@ public class ImageSelector extends Configurer implements ItemListener {
     final FileChooser fc = gm.getEditorImageChooser();
     fc.setFileFilter(new ImageFileFilter());
 
-    if (fc.showOpenDialog(gm.getPlayerWindow()) == FileChooser.APPROVE_OPTION && fc.getSelectedFile().exists()) {
+    if (fc.showOpenDialog(gm.getPlayerWindow()) == FileChooser.APPROVE_OPTION) {
       final String name = fc.getSelectedFile().getName();
       gm.getArchiveWriter().addImage(fc.getSelectedFile().getPath(), name);
       select.setModel(new DefaultComboBoxModel<>(ArrayUtils.addFirst(gm.getDataArchive().getImageNames(), NO_IMAGE)));
--- a/vassal-app/src/main/java/VASSAL/counters/ImagePicker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/ImagePicker.java
@@ -167,8 +167,7 @@ public class ImagePicker extends JPanel
     final FileChooser fc = gm.getEditorImageChooser();
     fc.setFileFilter(new ImageFileFilter());
 
-    if (fc.showOpenDialog(this) == FileChooser.APPROVE_OPTION
-         && fc.getSelectedFile().exists()) {
+    if (fc.showOpenDialog(this) == FileChooser.APPROVE_OPTION) {
       final String name = fc.getSelectedFile().getName();
       gm.getArchiveWriter().addImage(fc.getSelectedFile().getPath(), name);
       select.setModel(new DefaultComboBoxModel<>(
--- a/vassal-app/src/main/java/VASSAL/launch/AbstractLaunchAction.java
+++ b/vassal-app/src/main/java/VASSAL/launch/AbstractLaunchAction.java
@@ -170,28 +170,17 @@ public abstract class AbstractLaunchAction extends AbstractAction {
     // loop until cancellation or we get an existing file
     if (fc.showOpenDialog() == FileChooser.APPROVE_OPTION) {
       lr.module = fc.getSelectedFile();
-      if (lr.module != null) {
-        if (lr.module.exists()) {
-          final AbstractMetaData metadata =
-            MetaDataFactory.buildMetaData(lr.module);
-          if (!(metadata instanceof ModuleMetaData)) {
-            ErrorDialog.show(
-              "Error.invalid_vassal_module", lr.module.getAbsolutePath()); //NON-NLS
-            logger.error(
-              "-- Load of {} failed: Not a Vassal module", //NON-NLS
-              lr.module.getAbsolutePath()
-            );
-            lr.module = null;
-          }
-        }
-        else {
-          lr.module = null;
-        }
-// FIXME: do something to warn about nonexistent file
-//        FileNotFoundDialog.warning(window, lr.module);
+      final AbstractMetaData metadata =
+        MetaDataFactory.buildMetaData(lr.module);
+      if (!(metadata instanceof ModuleMetaData)) {
+        ErrorDialog.show(
+          "Error.invalid_vassal_module", lr.module.getAbsolutePath()); //NON-NLS
+        logger.error(
+          "-- Load of {} failed: Not a Vassal module", //NON-NLS
+          lr.module.getAbsolutePath());
+        lr.module = null;
       }
     }
-
     return lr.module;
   }
 
--- a/vassal-app/src/main/java/VASSAL/launch/Editor.java
+++ b/vassal-app/src/main/java/VASSAL/launch/Editor.java
@@ -154,19 +154,12 @@ public class Editor extends Launcher {
       if (fc.showOpenDialog() == FileChooser.APPROVE_OPTION) {
         lr.importFile = fc.getSelectedFile();
 
-        if (lr.importFile != null) {
-          if (lr.importFile.exists()) {
-            final AbstractMetaData metadata =
-              MetaDataFactory.buildMetaData(lr.importFile);
-            if (!(metadata instanceof ImportMetaData)) {
-              ErrorDialog.show("Error.invalid_import_file", lr.importFile.getAbsolutePath());  //NON-NLS
-              logger.error("Import of " + lr.importFile.getAbsolutePath() + " failed: unrecognized import type");  //NON-NLS
-              lr.importFile = null;
-            }
-          }
-          else {
-            lr.importFile = null;
-          }
+        final AbstractMetaData metadata =
+          MetaDataFactory.buildMetaData(lr.importFile);
+        if (!(metadata instanceof ImportMetaData)) {
+          ErrorDialog.show("Error.invalid_import_file", lr.importFile.getAbsolutePath());  //NON-NLS
+          logger.error("Import of " + lr.importFile.getAbsolutePath() + " failed: unrecognized import type");  //NON-NLS
+          lr.importFile = null;
         }
       }
 
--- a/vassal-app/src/main/java/VASSAL/launch/LoadModuleAction.java
+++ b/vassal-app/src/main/java/VASSAL/launch/LoadModuleAction.java
@@ -64,10 +64,7 @@ public class LoadModuleAction extends GameModuleAction {
       }
 
       if (fc.showOpenDialog() == FileChooser.APPROVE_OPTION) {
-        final File f = fc.getSelectedFile();
-        if (f != null && f.exists()) {
-          target = f;
-        }
+        target = fc.getSelectedFile();
       }
 
       // bail out if still no target
--- a/vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
+++ b/vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
@@ -566,7 +566,7 @@ public class ModuleManagerWindow extends JFrame {
       Arrays.stream(moduleConfig.getStringArray())
     ).sorted().distinct().forEach(s -> {
       final ModuleInfo module = new ModuleInfo(s);
-      if (module.getFile().exists() && module.isValid()) {
+      if (module.getFile().isFile() && module.isValid()) {
         moduleList.add(module);
       }
       else {
--- a/vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
+++ b/vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
@@ -212,7 +212,15 @@ public abstract class FileChooser {
 
     @Override
     public int showOpenDialog(Component parent) {
-      final int value = fc.showOpenDialog(parent);
+      int value = fc.showOpenDialog(parent);
+      final int mode = fc.getFileSelectionMode();
+      final File selected = getSelectedFile();
+      if (value == APPROVE_OPTION &&
+          ((mode == DIRECTORIES_ONLY && !selected.isDirectory()) ||
+           (mode == FILES_ONLY       && !selected.isFile()))) {
+        value = ERROR_OPTION;
+        fc.setSelectedFile(null);
+      }
       updateDirectoryPreference();
       return value;
     }
@@ -356,7 +364,7 @@ public abstract class FileChooser {
 
       fd.setModal(true);
       fd.setFilenameFilter(filter);
-      if (cur != null) {
+      if (cur != null && !cur.getPath().isEmpty()) {
         if (cur.isDirectory())
           fd.setDirectory(cur.getPath());
         else {
@@ -376,10 +384,20 @@ public abstract class FileChooser {
       final int value;
       if (fd.getFile() != null) {
         cur = new File(fd.getDirectory(), fd.getFile());
-        value = APPROVE_OPTION;
+        if (cur.isFile() && mode == FILES_ONLY) {
+          value = APPROVE_OPTION;
+        }
+        else if (cur.isDirectory() && mode == DIRECTORIES_ONLY) {
+          value = APPROVE_OPTION;
+        }
+        else {
+          value = ERROR_OPTION;
+          cur = null;
+        }
       }
       else {
         value = CANCEL_OPTION;
+        cur = null;
       }
       updateDirectoryPreference();
       return value;
--- a/vassal-app/src/main/java/VASSAL/tools/icon/IconFamily.java
+++ b/vassal-app/src/main/java/VASSAL/tools/icon/IconFamily.java
@@ -644,34 +644,28 @@ public class IconFamily extends AbstractConfigurable {
       final FileChooser fc = GameModule.getGameModule().getEditorImageChooser();
       fc.setFileFilter(new FamilyImageFilter(family.getName()));
       fc.setSelectedFile(new File(family.getName() + ".*")); //$NON-NLS-1$
-      if (fc.showOpenDialog(getControls()) != FileChooser.APPROVE_OPTION) {
-        setWarning(""); //$NON-NLS-1$
-        setValue(null);
-      }
-      else {
+      if (fc.showOpenDialog(getControls()) == FileChooser.APPROVE_OPTION) {
         final File f = fc.getSelectedFile();
-        if (f != null && f.exists()) {
-          final String name = f.getName();
-          if (name.split("\\.").length != 2) { //$NON-NLS-1$
-            showError(Resources.getString("Editor.IconFamily.illegal_icon_name")); //$NON-NLS-1$
-          }
-          else if (!name.startsWith(family.getName())) {
-            showError(Resources.getString("Editor.IconFamily.bad_icon_name", family.getName())); //$NON-NLS-1$
-          }
-          else if (!ImageUtils.hasImageSuffix(name)) {
-            showError(Resources.getString("Editor.IconFamily.bad_icon_file")); //$NON-NLS-1$
-          }
-          else {
-            GameModule.getGameModule().getArchiveWriter().addImage(f.getPath(),
-                buildPath(f.getName()));
-            setWarning(""); //$NON-NLS-1$
-            setValue(name);
-          }
+        final String name = f.getName();
+        if (name.split("\\.").length != 2) { //$NON-NLS-1$
+          showError(Resources.getString("Editor.IconFamily.illegal_icon_name")); //$NON-NLS-1$
+        }
+        else if (!name.startsWith(family.getName())) {
+          showError(Resources.getString("Editor.IconFamily.bad_icon_name", family.getName())); //$NON-NLS-1$
+        }
+        else if (!ImageUtils.hasImageSuffix(name)) {
+          showError(Resources.getString("Editor.IconFamily.bad_icon_file")); //$NON-NLS-1$
         }
         else {
+          GameModule.getGameModule().getArchiveWriter().addImage(f.getPath(),
+              buildPath(f.getName()));
           setWarning(""); //$NON-NLS-1$
-          setValue(null);
+          setValue(name);
         }
+      }
+      else {
+        setWarning(""); //$NON-NLS-1$
+        setValue(null);
       }
     }
 
--- a/vassal-app/src/main/java/VASSAL/tools/image/tilecache/ImageTileDiskCache.java
+++ b/vassal-app/src/main/java/VASSAL/tools/image/tilecache/ImageTileDiskCache.java
@@ -96,7 +96,7 @@ public class ImageTileDiskCache implements ImageTileSource, FileStore {
     double scale) throws ImageIOException {
 
     final File f = new File(tileNameFor(name, tileX, tileY, scale));
-    return f.exists() && f.isFile();
+    return f.isFile();
   }
 
   /** {@inheritDoc} */
--- a/vassal-app/src/main/java/VASSAL/tools/imports/ImportAction.java
+++ b/vassal-app/src/main/java/VASSAL/tools/imports/ImportAction.java
@@ -149,10 +149,8 @@ public final class ImportAction extends EditModuleAction {
 
     if (fc.showOpenDialog() == FileChooser.APPROVE_OPTION) {
       final File f = fc.getSelectedFile();
-      if (f != null && f.exists()) {
-        loadModule(f);
-        actionCancelled = false;
-      }
+      loadModule(f);
+      actionCancelled = false;
     }
   }
 
@@ -263,8 +261,10 @@ public final class ImportAction extends EditModuleAction {
    */
   public File getCaseInsensitiveFile(File f, File base,
       boolean queryIfNotFound, FileFilter filter) {
+    if (f.getPath().isEmpty())
+      return null;
     // Easy case
-    if (f.exists())
+    if (f.isFile())
       return f;
 
     final String name = Importer.getFileName(f.getName());
@@ -307,8 +307,7 @@ public final class ImportAction extends EditModuleAction {
       fc.setSelectedFile(new File(f.getName()));
 
       if (fc.showOpenDialog() == FileChooser.APPROVE_OPTION) {
-        final File p = fc.getSelectedFile();
-        if (p.exists()) return p;
+        return fc.getSelectedFile();
       }
     }
 
--- a/vassal-app/src/main/java/VASSAL/tools/io/Tailer.java
+++ b/vassal-app/src/main/java/VASSAL/tools/io/Tailer.java
@@ -71,6 +71,7 @@ public class Tailer {
    */
   public Tailer(File file, long poll_interval) {
     if (file == null) throw new IllegalArgumentException("file == null");
+    if (file.getPath().isEmpty()) throw new IllegalArgumentException("Empty file name");
 
     this.file = file;
     this.poll_interval = poll_interval;
--- a/vassal-app/src/main/java/VASSAL/configure/FileConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/configure/FileConfigurer.java
@@ -93,7 +93,7 @@ public class FileConfigurer extends Configurer {
   public void setValue(Object o) {
 // FIXME: this creates a problem when the referenced file is in the JAR
     final File f = (File) o;
-    if (f != null && f.exists()) {
+    if (f != null && !f.getPath().isEmpty() && f.exists()) {
       if (archive != null) {
         addToArchive(f);
       }
@@ -110,7 +110,7 @@ public class FileConfigurer extends Configurer {
 
   @Override
   public void setValue(String s) {
-    if (s == null)
+    if (s == null || s.isEmpty())
       setValue((Object) null);
     else {
       setValue(new File(s));


rlament:
I think this is overkill. In places where the `File` object is the result of a file chooser, on cancellation it will be null not an empty string. No need to test the path in these cases.

The place to apply a fix is where the File object is assigned an empty string, changing that assignment to a null instead. https://github.com/vassalengine/vassal/pull/14427#discussion_r2667065869

vassal-app/src/main/java/VASSAL/counters/MassPieceLoader.java
```diff
@@ -558,7 +558,7 @@ public Component getTableCellEditorComponent(JTable table, Object value,
      */
     protected void loadImageNames(File dir) {
       imageNames.clear();
-      if (dir != null && dir.isDirectory()) {
+      if (dir != null && !dir.getPath().isEmpty() && dir.isDirectory()) {
```
rlament: Why would the current directory (i.e. empty string) be invalid here? Seems reasonable to load images from the current directory, although I have not tested that this actually works.

cholmcc: Basically, if you want the test to be the same as before, then we'd better check for an empty file name.  But, I don't think this is a case where it is truly necessary.  But note, the empty string as a file name refers to the `user.dir` directory - which isn't necessarily the same as the current directory.   Also, Vassal explicitly changes directory to its installation directory, and in so far as that is considered the same as the `""` directory, I do not think it makes much sense to load images from there.  

But, as I said, I don't necessarily believe this is strictly needed, and it can be removed.  What do you think?

rlament: As a proponent for minimal changes my suggestion is to revert. The goal shouldn't be making the test to be the same as before, it should be avoiding the corner case of empty string.

cholmcc: OK, I removed the check above in `MassPieceLoader`.   Do you see another other change that you think are not warranted?   Or do you believe that the rest of the proposed changes are sound?  

Note that the previous force-push to this PR made it so that the `FileChooser` does most of the checking, and the client code only need to worry about the return value of `#showOpenDialog` or `#showSaveDialog`, and can then rest assured that the `File` object returned is indeed valid. 

Yours,
_Christian_

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -801,6 +801,7 @@ public boolean loadGame(File f, boolean continuation, boolean forceForeground) {
     final GameModule g = GameModule.getGameModule();
 
     try {
+      if (f.getPath().isEmpty()) throw new FileNotFoundException("No save specified");
       if (!f.exists()) throw new FileNotFoundException("Unable to locate " + f.getPath());
```
rlament: This is loading a module so the exception message is inaccurate.

If the group consensus is to touch many or most of the file.exists() call sites, substituting `isFile()` for `exists()` is much more readable. It makes the intent clear.

Could eliminate the new `if` statement and add literal quotes around the `f.getPath`. For an empty path the exception message would be `Unable to locate ""`. Clearly indicates that there is no file name.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -1016,7 +1017,7 @@ protected String saveString() {
   }
 
   protected boolean checkForOldSaveFile(File f) {
-    if (f.exists()) {
+    if (!f.getPath().isEmpty() && f.exists()) {
```
rlament: Not needed since `exists()` returns true for files or directories. This change is in fact changing the behaviour.

vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
```diff
@@ -252,7 +252,7 @@ private void writeImage(final File f, BufferedImage img, Rectangle r)
       files.add(f);
 
       // make sure that we can write the file before proceeding
-      if (f.exists()) {
+      if (!f.getPath().isEmpty() && f.exists()) {
```
rlament: Why does it do `files.add(f)` before the file checks? The add should be after the check.

In the 'else' below, `getParentFile()` can return null. Should check `p` for null afterwards.

Notice here how the filenames are quoted in the error message. Should strive to follow that format.

vassal-app/src/main/java/VASSAL/build/module/metadata/MetaDataFactory.java
```diff
@@ -54,7 +54,7 @@ public class MetaDataFactory {
   public static AbstractMetaData buildMetaData(File file) {
 
     // Check the file exists and is a file
-    if (file == null || !file.exists() || !file.isFile())
+    if (file == null || file.getPath().isEmpty() || !file.exists() || !file.isFile())
```
rlament: Already checks for `isFile`, the ` file.getPath().isEmpty() || !file.exists()` is unnecessary.

vassal-app/src/main/java/VASSAL/launch/LoadModuleAction.java
```diff
@@ -65,9 +65,7 @@ public void performAction(ActionEvent e) throws IOException {
 
       if (fc.showOpenDialog() == FileChooser.APPROVE_OPTION) {
         final File f = fc.getSelectedFile();
-        if (f != null && f.exists()) {
-          target = f;
-        }
+        target = f;
```
rlament: Could assign directly to `target`. Don't need the `File f` anymore.

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
```diff
@@ -566,7 +566,7 @@ protected void buildTree() {
       Arrays.stream(moduleConfig.getStringArray())
     ).sorted().distinct().forEach(s -> {
       final ModuleInfo module = new ModuleInfo(s);
-      if (module.getFile().exists() && module.isValid()) {
+      if (!module.getFile().getPath().isEmpty() && module.getFile().exists() && module.isValid()) {
```
rlament: The `ModuleInfo` constructor calls `loadMetaData` which calls `MetaDataFactory.buildMetaData` which checks for file existence storing the result in the `valid` flag.

The only check required here is the `module.isValid()`.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
```diff
@@ -212,7 +212,13 @@ public void setDialogTitle(String title) {
...
     @Override
     public int showOpenDialog(Component parent) {
-      final int value = fc.showOpenDialog(parent);
+      int value = fc.showOpenDialog(parent);
+      if (value == APPROVE_OPTION 
+          && (getSelectedFile().getPath().isEmpty() ||
+              !getSelectedFile().exists())) {
+        value = CANCEL_OPTION;
+        fc.setSelectedFile(null);
+      }
```
rlament, line 223: I really think that it is not possible to select an empty filename. It will be so disconcerting to the user who presses 'OK' and nothing happens.

If this is staying, I'd only check for empty path and not check for existence.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
```diff
@@ -227,6 +233,7 @@ && getSelectedFile().exists()
                                                                     Resources.getString("Editor.FileChooser.exists"),
                                                                     JOptionPane.YES_NO_OPTION)) {
         value = CANCEL_OPTION;
+        fc.setSelectedFile(null);
```
rlament: Not sure if this is a good thing. Consider the scenario where a user forgot a small edit, cancelled and then repeats the save. It might be nice to already have the same file selected.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
```diff
@@ -356,7 +363,7 @@ else if (parent instanceof Frame) {
 
       fd.setModal(true);
       fd.setFilenameFilter(filter);
-      if (cur != null) {
+      if (cur != null && !cur.getPath().isEmpty()) {
```
rlament, line 367: Not necessary to change this line. If cur is an empty string, it is treated as a directory.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
```diff
@@ -374,12 +381,20 @@ public int showOpenDialog(Component parent) {
...
+          value = APPROVE_OPTION;
+        }
+        else {
+          value = CANCEL_OPTION;
+          cur = null;
+        }
       }
       else {
         value = CANCEL_OPTION;
+        cur = null;
```
rlament: My suggestion is to revert all of `showOpenDialog`. Cancelling a chosen action without informing the user is problematic. How is anyone going to debug this if there ever is a report of selecting 'OK' and nothing happening?

vassal-app/src/main/java/VASSAL/tools/icon/IconFamily.java
```diff
@@ -650,27 +650,21 @@ protected void selectImage() {
...
-          else if (!ImageUtils.hasImageSuffix(name)) {
-            showError(Resources.getString("Editor.IconFamily.bad_icon_file")); //$NON-NLS-1$
-          }
-          else {
-            GameModule.getGameModule().getArchiveWriter().addImage(f.getPath(),
-                buildPath(f.getName()));
-            setWarning(""); //$NON-NLS-1$
-            setValue(name);
-          }
+        final String name = f.getName();
```
rlament, line 649: The logic in `if (fc.showOpenDialog(getControls()) != FileChooser.APPROVE_OPTION)` above should be flipped. Check should be for `== FileChooser.APPROVE_OPTION` with correspond if/else blocks swapped. `showOpenDialog` can also return `ERROR_OPTION` which should be treated like a cancel.

vassal-app/src/main/java/VASSAL/tools/image/tilecache/ImageTileDiskCache.java
```diff
@@ -96,7 +96,7 @@ public boolean tileExists(
     double scale) throws ImageIOException {
 
     final File f = new File(tileNameFor(name, tileX, tileY, scale));
-    return f.exists() && f.isFile();
+    return !f.getPath().isEmpty() && f.exists() && f.isFile();
```
rlament: The `f.isFile()` is sufficient to check for existence.

vassal-app/src/main/java/VASSAL/tools/imports/ImportAction.java
```diff
@@ -264,7 +262,7 @@ public void loadModule(File f) throws IOException {
   public File getCaseInsensitiveFile(File f, File base,
       boolean queryIfNotFound, FileFilter filter) {
     // Easy case
-    if (f.exists())
+    if (!f.getPath().isEmpty() && f.exists())
```
rlament: If the path is empty string, we are better off returning here. Revert this line.

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
```diff
@@ -719,6 +724,9 @@ public static class DirClassSource extends ClassSource
 
 		public static byte [] readBytesFromFile( File base, String className ) 
 		{
+                        if (className.isEmpty())
+                                return null;
+
```
rlament: Since ".class" gets appended to the filename the `exists()` a few lines down will never get an empty string. It will most likely return null there unless there happens to be a ".class" file. Could revert this part.

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
```diff
@@ -361,6 +361,11 @@ synchronized void map( URL url )
 		throws IOException 
 	{ 
 		String name = url.getFile();
+                if ( name.isEmpty()) {
+			String s = "Empty a classpath component";
+			errorWhileMapping( s );
+		}
+
```
rlament: Should we really mess with this? Revert this bit and let the isDirectory handle the empty paths.

This file uses tabs for indenting, unfortunately. Better to reconfigure your editor to insert tabs to keep it consistent. It's a pain. It's easier just to back out the changes.

vassal-app/src/main/java/VASSAL/launch/LoadModuleAction.java
cholmcc: Right.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
cholmcc: > If cur is an empty string, it is treated as a directory.

No, that's not the change in 25.  The change is, if the file name is empty, it is interpreted as `user.dir`, which in the case of Vassal on Linux is the the installation directory of Vassal (because that's explicitly set), and on other OSs is the users home directory.   That is really not the same thing, so it makes sense to check that the file name isn't empty. 

If a user is asked to pick a directory, then the file chooser will give the actual directory (file) name, if one is chosen, and it will not be empty.  

It is far better to catch this early than at some later point in the client code.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
cholmcc: > It might be nice to already have the same file selected. 


Most places where the `FileChooser` is used, it is made a-new every time, and therefore the selected file is set anew, so I don't think there's a use case like the one you propose.   The point of setting the selected file to `null` is really to make sure that the calling code doesn't mis-use the selected file if the user didn't want to use that file.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
cholmcc: Well, it depends on the client code.   If the client code really needs a file to be selected, it could raise an issue.  If it doesn't, then it wont.   The point of moving the check into the `FileChooser` is to catch problems as soon as possible.    So if you think the `FileChooser` should raise an issue if an invalid file was chosen here, then it could do so.   If you move the responsibility of validating the file to the client code, then you will make multiple points of failure rather than one.

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
cholmcc: Right, but the check for an empty string is still needed _before_ `isValid` - otherwise the `loadMetaData`, `buildMetaData`, etc. may fail in a funny way.   The point is, that the pre-25, the `exists` check did 

```
IF file name not empty AND file system entry exists THEN 
   IF module is valid THEN 
      ...
   END IF
END IF
```

Now, it is 
```
IF file name is empty THEN 
  SET file to be the directory pointed to by user.dir
END IF
IF file system entry exists THEN
  IF module is valid THEN
     ...
  END IF
END IF
```
To get the pre-25 behaviour, you need 
```
IF file name is not empty THEN
  IF file name is empty THEN 
    SET file to be the directory pointed to by user.dir
  END IF
  IF file system entry exists THEN
    IF module is valid THEN
       ...
    END IF
  END IF
END IF
```

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
cholmcc: > I really think that it is not possible to select an empty filename.

Unless you are absolutely certain, I think you will need to check in the code.   It is clear that empty string file names causes problems in the code, so where better to remove the problem than at its root. 

>  It will be so disconcerting to the user who presses 'OK' and nothing happens.

Well, if that's a problem, either add logging or even an error dialog.   I guess that could be achieved by changing the return value to `ERROR_OPTION` - then the client code will know that an error occurred and can take appropriate measures. 

> If this is staying, I'd only check for empty path and not check for existence.

That makes absolutely no sense.

vassal-app/src/main/java/VASSAL/tools/image/tilecache/ImageTileDiskCache.java
cholmcc: If `f.exists()` was deemed necessary in the past, then it `!f.getPath().isEmpty()` would be need now.  Otherwise, `f.isFile` should have been enough in the past too.

vassal-app/src/main/java/VASSAL/tools/imports/ImportAction.java
cholmcc: Then it shouldn't be a revert, but rather 
```
if (f.getPath().isEmpty()) return null;
```

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
cholmcc: > Should we really mess with this?

The BS library used by Vassal is already a patched and outdated version of BS, so no reason to not change it.

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
cholmcc: But better to catch the problem sooner rather than later.  You don't want a file name  like `.class`, since that is likely not a valid Java class file.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
cholmcc: > This is loading a module so the exception message is inaccurate.

I'm fairly certain this is loading saves - not modules. 

> `Unable to locate ""`. Clearly indicates that there is no file name. 

I think the above error message is confusing to most users, while `No save specified` is more meaningful.

vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
cholmcc: > Why does it do files.add(f) before the file checks? The add should be after the check.

That's an older bug that hasn't been corrected I guess.

vassal-app/src/main/java/VASSAL/build/module/metadata/MetaDataFactory.java
cholmcc: OK

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
cholmcc: If the `FileChooser` returns `APPROVE_OPTION`, then the client code should be confident that a _valid_ file was chosen by the `FileChooser`.   It makes no sense that the client code would need to check that the chosen file is not `null`, the empty string, or doesn't exists (in case of `Open`). 

If the `FileChooser` returns `CANCEL_OPTION`, then the client code should behave as if never prompted the user for a file name, and no file name should be set in the `FileChooser`. 

If the `FileChooser` returns `ERROR_OPTION`, then the client code should raise an issue if appropriate, or otherwise behave as if it never prompted the user for a file name, and no file name should be set in the `FileChooser`. 

That will make the division of labour and responsibilities very clear.

vassal-app/src/main/java/VASSAL/tools/icon/IconFamily.java
cholmcc: If you say that `showOpenDialog` has _three_ return values 
- `APPROVE_OPTION`,
- `CANCEL_OPTION`, and
- `ERROR_OPTION` 
then I think the logic is best encoded as 
```
if (fc.showOpenDialog(...) != FileChooser.APPROVE_OPTION) {
   // Error handling 
  return;
}
// Good case 
```
since a single `if` will catch all non-good states.

vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
```diff
@@ -795,7 +795,7 @@ protected JComponent createPanel(final WizardController controller, String id, f
           @Override
           public void propertyChange(PropertyChangeEvent evt) {
             final File f = (File) evt.getNewValue();
-            if (f == null || !f.exists()) {
+            if (!f.isFile()) {
```
uckelman: This will thrown an NPE now if `f` is null.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -1016,7 +1016,7 @@ protected String saveString() {
   }
 
   protected boolean checkForOldSaveFile(File f) {
-    if (f.exists()) {
+    if (f.isFile()) {
```
uckelman, line 1019: This changes the test conditions. What if `f` exists but is not a file?

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -801,7 +801,7 @@ public boolean loadGame(File f, boolean continuation, boolean forceForeground) {
     final GameModule g = GameModule.getGameModule();
 
     try {
-      if (!f.exists()) throw new FileNotFoundException("Unable to locate " + f.getPath());
+      if (!f.isFile()) throw new FileNotFoundException("Unable to locate \"" + f.getPath() + "\"");
```
uckelman: This changes the test conditions. What if f exists but is not a file?

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
rlament: Well it only catches the problem a couple lines earlier. If '.class' doesn't exist it already returns null. If it does exist likely then the RuntimeException.

Please keep the indenting consistent with the original file. This file uses the tab for indenting. And avoid adding the blank lines of whitespace (line 729).

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
rlament: Any change to the code should be tested. That's a reason not to change it.

vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
rlament: Regardless of the position of `files.add(f)`, by adding the `isEmpty` there now has to be a check for `p == null` in the else block (line 264). This wasn't an issue before.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
rlament: This is the `loadGame` function. It is looking for a file and nothing else. If it's not a file it should throw.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
rlament: I agree. This line should be reverted. Better to have the more general check over the specific one.

vassal-app/src/main/java/VASSAL/build/module/WizardSupport.java
rlament: Looks like that was always the case. It did check for null but then used it in the 'else' anyway.

vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
uckelman: It doesn't matter in practice if file as added to the files list before for after those exceptions are thrown, but you could move it after if you want.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
rlament: > I really think that it is not possible to select an empty filename.

If there does exist an issue with FileChooser and empty filenames, there would be a lot of noise about this in the wider Java community. Consider how this is to be tested. If it can't be tested, do not check for it.

> If this is staying, I'd only check for empty path and not check for existence.

The reason for this exercise is to detect empty filenames. This function never checked for file existence, so if absolutely necessary to include, it should only check for the empty path. However, I'm still of the opinion this FileChooser change is entirely unnecessary.

vassal-app/src/main/java/VASSAL/configure/IconConfigurer.java
```diff
@@ -145,15 +145,12 @@ private void selectImage() {
...
-                                  .addImage(f.getPath(), f.getName());
-        setValue(f.getName());
-        setToolTipText(f.getPath());
-      }
-    }
+    final File f = fc.getSelectedFile();
+    GameModule.getGameModule().getArchiveWriter()
+        .addImage(f.getPath(), f.getName());
+    setValue(f.getName());
+    setToolTipText(f.getPath());
```
uckelman, line 154: What is the reason for this change?

vassal-app/src/main/java/VASSAL/configure/ImageSelector.java
```diff
@@ -191,7 +191,7 @@ private void pickImage() {
     final FileChooser fc = gm.getEditorImageChooser();
     fc.setFileFilter(new ImageFileFilter());
 
-    if (fc.showOpenDialog(gm.getPlayerWindow()) == FileChooser.APPROVE_OPTION && fc.getSelectedFile().exists()) {
+    if (fc.showOpenDialog(gm.getPlayerWindow()) == FileChooser.APPROVE_OPTION) {
```
uckelman, line 194: What is the purpose of this change?

vassal-app/src/main/java/VASSAL/counters/ImagePicker.java
```diff
@@ -167,8 +167,7 @@ public void pickImage() {
     final FileChooser fc = gm.getEditorImageChooser();
     fc.setFileFilter(new ImageFileFilter());
 
-    if (fc.showOpenDialog(this) == FileChooser.APPROVE_OPTION
-         && fc.getSelectedFile().exists()) {
+    if (fc.showOpenDialog(this) == FileChooser.APPROVE_OPTION) {
```
uckelman, line 170: Why?

vassal-app/src/main/java/VASSAL/launch/Editor.java
```diff
@@ -154,19 +154,12 @@ protected File promptForFile() {
...
-          }
-          else {
-            lr.importFile = null;
-          }
+        final AbstractMetaData metadata =
+          MetaDataFactory.buildMetaData(lr.importFile);
+        if (!(metadata instanceof ImportMetaData)) {
+          ErrorDialog.show("Error.invalid_import_file", lr.importFile.getAbsolutePath());  //NON-NLS
+          logger.error("Import of " + lr.importFile.getAbsolutePath() + " failed: unrecognized import type");  //NON-NLS
+          lr.importFile = null;
```
uckelman, line 162: Why?

vassal-app/src/main/java/VASSAL/launch/AbstractLaunchAction.java
```diff
@@ -170,27 +170,19 @@ protected File promptForFile() {
...
+        ErrorDialog.show(
+          "Error.invalid_vassal_module", lr.module.getAbsolutePath()); //NON-NLS
+        logger.error(
+          "-- Load of {} failed: Not a Vassal module", //NON-NLS
+          lr.module.getAbsolutePath());
+        lr.module = null;
       }
     }
+// FIXME: do something to warn about nonexistent file
+//        FileNotFoundDialog.warning(window, lr.module);
```
uckelman: Why?

vassal-app/src/main/java/VASSAL/launch/LoadModuleAction.java
```diff
@@ -64,10 +64,7 @@ public void performAction(ActionEvent e) throws IOException {
       }
 
       if (fc.showOpenDialog() == FileChooser.APPROVE_OPTION) {
-        final File f = fc.getSelectedFile();
-        if (f != null && f.exists()) {
-          target = f;
-        }
+        target = fc.getSelectedFile();
```
uckelman, line 67: Why?

vassal-app/src/main/java/VASSAL/tools/imports/ImportAction.java
```diff
@@ -149,10 +149,8 @@ public void performAction(ActionEvent e) throws IOException {
 
     if (fc.showOpenDialog() == FileChooser.APPROVE_OPTION) {
       final File f = fc.getSelectedFile();
-      if (f != null && f.exists()) {
-        loadModule(f);
-        actionCancelled = false;
-      }
+      loadModule(f);
```
uckelman, line 152: Why are the checks on `f` being removed here?

vassal-app/src/main/java/VASSAL/tools/io/Tailer.java
```diff
@@ -101,6 +101,10 @@ public synchronized void start() throws IOException {
     // NB: This method is synchronized to ensure that there is never more
     // than one tailer thread at at time.
     if (!tailing) {
+      if (file.getPath().isEmpty()) {
+        throw new IOException("Empty file name");
+      }
+
```
uckelman: The check for an empty filename should be in the constructor.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
uckelman: Concur, this is wrong. Checking `isFile()` will make it impossible to select directories successfully.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
cholmcc: Right.  Should perhaps also check for a directory because a directory with same name - although unlikely - might exist.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
cholmcc: At this point, we are looking for a file - not a directory - so if we get a directory that would be an error.

vassal-app/src/main/java/VASSAL/tools/imports/ImportAction.java
```diff
@@ -307,8 +307,7 @@ public File getCaseInsensitiveFile(File f, File base,
       fc.setSelectedFile(new File(f.getName()));
 
       if (fc.showOpenDialog() == FileChooser.APPROVE_OPTION) {
-        final File p = fc.getSelectedFile();
-        if (p.exists()) return p;
+        return fc.getSelectedFile();
```
uckelman, line 310: Why this change?

vassal-app/src/main/java/VASSAL/configure/IconConfigurer.java
cholmcc: Since the `FileChooser` now verifies that the chosen file _actually_ exists, there's no reason to check for it here.

vassal-app/src/main/java/VASSAL/configure/ImageSelector.java
cholmcc: As above - `FileChooser` checks that the file _actually_ exists and there's no reason to check again if the `FileChooser` returns `FileChooser.APPROVE_OPTION`.

vassal-app/src/main/java/VASSAL/counters/ImagePicker.java
cholmcc: Same as above.

vassal-app/src/main/java/VASSAL/launch/Editor.java
cholmcc: Same as above

vassal-app/src/main/java/VASSAL/tools/imports/ImportAction.java
cholmcc: Same reason as above.

vassal-app/src/main/java/VASSAL/launch/LoadModuleAction.java
cholmcc: As above.

vassal-app/src/main/java/VASSAL/launch/AbstractLaunchAction.java
cholmcc: The comment is carried over from before.  The code change above it is for the same reason as above.  When `FileChooser` returns `FileChooser.APPROVE_OPTION` the file has already been checked and there's no need to check again.

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
uckelman: I wouldn't touch anything in BeanShell unless there is no choice.

vassal-app/src/main/java/VASSAL/tools/io/Tailer.java
cholmcc: OK, fair enough

vassal-app/src/main/java/VASSAL/tools/imports/ImportAction.java
cholmcc: Same reason as above - if `FileChooser` returns `FileChooser.APPROVE_OPTION` then the file _actually_ exists and there's no need for further checks.

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
uckelman: A directory, which is what you'll get with an empty string on Windows, will throw an `IOException` at line 730 as-is. This doesn't need changing.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
cholmcc: But the following code actually assumes that the passed file path is a file, and will likely throw some odd error if it isn't.    So I think it is correct to check for a file here, but another error should likely be raised if the file path points to a directory.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
rlament: This isn't catching anything early. It's either setting the directory for the FileChooser or not setting it. The directory to which it is set doesn't matter so much. Maybe just a slight annoyance for the end user in that it presents the wrong (in the user's mind) directory. Inconsequential and unnecessary to make a change here.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
cholmcc: Also, looking at how the method is called, I don't think you will ever get a directory, so the test for a file seems entirely valid.

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
cholmcc: The BeanShell code used in Vassal is somewhat outdated as it is, _and_ patches have been applied, which means it is no longer a question of third-party code.   If it was, then I would agree with you.  Any case, I don't think this particular check is all that important, so I took it out.

vassal-app/src/main/java/bsh/classpath/BshClassPath.java
cholmcc: > ...  empty string on Windows ...

I think it applies to _all_ platforms as the change of 25 is really at the API level.

vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
rlament: After the `final File p = f.getParentFile();` in the 'else' block, check that p is not null. `new File("").getParentFile()` returns null. Prior to this change this wasn't an issue. With the change to the `if` the null check is needed.

Second thought: change the location of the `isEmpty` check. Null check of `p` not required.
```
if (f.exists()) {
  if (f.getPath().isEmpty() || !f.canWrite()) {
```

vassal-app/src/main/java/VASSAL/build/GameModule.java
```diff
@@ -697,7 +697,7 @@ void build() throws IOException {
     final DataArchive darch = getDataArchive();
 
     final File f = new File(darch.getName());
-    if (!f.exists() || f.length() == 0) {
+    if (f.getPath().isEmpty() || !f.exists() || f.length() == 0) {
```
rlament: Might as well just simplify this to `!f.isFile() || f.length() == 0`. It expresses the intent so much better. The first check in the 'else' block is `file.isFile`, so why not improve readability and be consistent?

vassal-app/src/main/java/VASSAL/build/module/GameState.java
rlament: It would be more informative to report `getAbsolutePath()` in the error message.

vassal-app/src/main/java/VASSAL/launch/AbstractLaunchAction.java
rlament: If `FileChooser` guarantees selection of an existing file then you have addressed the FIXME and it can be deleted. Yay!

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
rlament: Maybe the 'setDirectory' should be passed `cur.getAbsolutePath()` instead. That would set the O/S appropriate directory if ever given a File(""). Still thinking it's best to remove the `isEmpty` check.

vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
cholmcc: So you want to check if the file exists _and_ _either_ the it is the empty path or not write-able?  That can be shortened to 

```
if ((f.getPath().isEmpty() || !f.canWrite()) && f.exists()) {
```
or 
```
if (f.exists() && (f.getPath().isEmpty() || !f.canWrite())) {
```

vassal-app/src/main/java/VASSAL/build/GameModule.java
cholmcc: Which `else` block are you talking about?  Line number.  The `else` that follows does not check `file.isFile` but proceeds to retrieve the `buildFile.xml` from the module Zip archive.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
cholmcc: I don't know what the original rational for setting with `cur.getPath` rather than `cur.getAbsolutePath`.  In any case, I think it is a little beyond the scope of this PR, and should perhaps be addressed in a different PR.

vassal-app/src/main/java/VASSAL/build/module/map/ImageSaver.java
rlament: Ok, that works.

vassal-app/src/main/java/VASSAL/build/GameModule.java
rlament: Sorry, I should have been more specific. Looks like you figured it out. In the first else where it calls `MetaDataFactory.buildMetaData(f)`, the first thing that function does is check if it is a file.

vassal-app/src/main/java/VASSAL/build/module/GameState.java
```diff
@@ -801,7 +801,7 @@ public boolean loadGame(File f, boolean continuation, boolean forceForeground) {
     final GameModule g = GameModule.getGameModule();
 
     try {
-      if (!f.exists()) throw new FileNotFoundException("Unable to locate " + f.getPath());
+      if (!f.isFile()) throw new FileNotFoundException("Unable to locate \"" + f.getAbsolutePath() + "\"");
```
uckelman, line 804: This message is no longer accurate if the condition changes. What if `f` is an existing directory?

rlament: Perhaps it should follow the exception and simply state "File not found". Although this is somewhat of a moot point. The `ReadErrorDialog()` in the catch doesn't even use this message. It pulls text from the resource file.

If the path is a directory, it is still an error. Not sure that it is even possible to get here with a directory. The FileChooser won't let me pick a directory. While from the command line, the parser detects the load problem before ever getting here.

vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
```diff
@@ -212,7 +212,11 @@ public void setDialogTitle(String title) {
 
     @Override
     public int showOpenDialog(Component parent) {
-      final int value = fc.showOpenDialog(parent);
+      int value = fc.showOpenDialog(parent);
+      if (value == APPROVE_OPTION && !getSelectedFile().isFile()) {
```
uckelman: This will prevent open dialogs from selecting directories. Do we know this won't break some modules?

cholmcc: See latest commit - check according to request mode.

rlament: Please demonstrate how the existing FileChooser fails such that it injects invalid data into Vassal. Let's not fix a problem that doesn't exist. FileChooser is configured in advance for the requested type (file/directory). Can it actually return a directory when requesting a file?

PR #14435, opened 2026-01-10
Module Manager Look and Feel

### LINKED ISSUE #12821: Fix text colour in Module column of Module Manager

rlament:
Module Manager honors the look and feel as set by the user on Windows/Mac/Linux systems. Select look and feel using the `swing.defaultlaf` system property.

- Module Manager gets the color from look and feel.
- Maintains the same look and feel for child processes (player and editor).
- Dynamically loads a custom Look and Feel.
- Supports loading of Intellij theme.json files using the FlatLaf extension.

For dark modes there are some downstream visibility issues. Some of the default buttons (server, camera, zoom) are hard to see in dark mode. Messages in the chat log are hard to read.

Custom Look and Feel can be set from the command line. Use the `swing.defaultlaf` system property to specify the class name. Define the `look.feel` property to specify a jar or json file. It can be a relative or absolute path. As a quick example of using Flatlaf on Windows, download and copy the jar to the Vassal.exe folder. Create or edit the vassal.l4j.ini file in the same folder, adding:
```
-Dswing.defaultlaf=com.formdev.flatlaf.FlatDarkLaf
-Dlook.feel=flatlaf-3.7.jar
```
Vassal should now start in dark mode. Other operating system have slightly different mechanisms for injecting command line parameters.

To load a json Look and Feel file, place the _theme.json_ file and the FlatLaf jar files in the Vassal installation folder. For example using the Salmon theme on Windows, add the following lines to the vassal.l4j.ini file. Loading a json file requires both FlatLaf jars.
```
-Dlook.feel=flatlaf-3.7.jar;flatlaf-intellij-themes-3.7.jar
-Dswing.defaultlaf=Salmon.theme.json
```
The _-Dlook.feel_ parameter follows the classpath syntax. The separator on Windows is the semicolon and colon for all other operating systems.

Closes #12821

### DIFF
--- a/vassal-app/src/main/java/VASSAL/launch/StartUp.java
+++ b/vassal-app/src/main/java/VASSAL/launch/StartUp.java
@@ -22,10 +22,12 @@ import java.awt.AWTError;
 import java.awt.Font;
 import java.awt.Toolkit;
 import java.lang.reflect.InvocationTargetException;
+import javax.swing.JOptionPane;
 import javax.swing.UIManager;
 import javax.swing.SwingUtilities;
 import javax.swing.UnsupportedLookAndFeelException;
 
+import VASSAL.i18n.Resources;
 import VASSAL.preferences.Prefs;
 import VASSAL.preferences.ReadOnlyPrefs;
 import org.apache.commons.lang3.SystemUtils;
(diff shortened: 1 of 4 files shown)


vassal-app/src/main/java/VASSAL/launch/StartUp.java
```diff
@@ -117,23 +141,31 @@ protected void initUIProperties() {
...
             // Check that the look and feel is in the class path.
             Class.forName(defaultLaf, false, getClass().getClassLoader());
           }
           catch (ClassNotFoundException e) {
-            // The desired Look and Feel is not available.
-            // Record the error and report it later once the UIManager is up.
-            defaultLaf = null;
-            System.clearProperty("swing.defaultlaf");
+            // Attempt to dynamically load the look and feel class.
+            final String lookAndFeelPath = System.getProperty("look.feel"); // NON-NLS
```
cholmcc: Uh!? why the change to `look.feel` property from `swing.defaultlaf`?  Is that property set when `swing.defaultlaf` is set? It doesn't look like it from the [Java tutorial](https://docs.oracle.com/javase/tutorial/uiswing/lookandfeel/plaf.html).  Or is it an introduction of a new property - specific to Vassal (in which case it should at least be prefixed by `vassal`)?  If so, then why introduce a new property when there's already one - well-known and established - property for the exact same purpose?

PR #14483, opened 2026-02-07
Skip empty Turn Lists in the Turn Counter

### LINKED ISSUE #14471: [Third Reich 4th edition] ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
| ABR | |
|-----|-----|
| Version | 3.7.19 |
| Reporter | elvilalta@gmail.com |
| Error Log | [97373 bytes](https://vassalengine.org/files/gh/fd4f500a3d834cbbda95eaf949e66c4f974bbab4) |

<pre>Turn is not working. 

Third Reich 4th edition v2.2b03 3.7.19

java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
 at VASSAL.build.module.turn.ListTurnLevel.getSetControl(ListTurnLevel.java:259)
 at VASSAL.build.module.turn.TurnLevel.getSetControls(TurnLevel.java:228)
 at VASSAL.build.module.turn.TurnTracker$SetDialog.addChildControls(TurnTracker.java:1273)
 at VASSAL.build.module.turn.TurnTracker$SetDialog.setControls(TurnTracker.java:1261)
</pre>

rlament:
Module designers can unintentionally or intentionally add an empty Turn List sub-component to a Turn Counter button. A new Turn List initializes as empty by default with no levels.

This PR adds handling of empty turn lists. The turn tracker should not throw exceptions on empty lists when advancing to the next or previous list item. For any list without defined levels or sub-levels, the turn counter skips over those empty turn lists.

Closes #14471

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/turn/ListTurnLevel.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/turn/ListTurnLevel.java
@@ -80,8 +80,10 @@ public class ListTurnLevel extends TurnLevel implements ActionListener {
   protected void setLow() {
     current = first;
     rolledOver = false;
-    if (current < active.length && !active[current]) {
-      // first is not active. Find the next active index.
+    if ((current < active.length && !active[current])
+            || (list.length == 0 && getTurnLevelCount() == 0)) {
+      // Either not active or the item is an empty leaf node.
+      // Advance to the next active index.
       advance();
     }
     super.setLow();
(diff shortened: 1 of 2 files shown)


uckelman:
I worry that this change masks a bug in module design.

jonathanrwatts:
Perhaps send a warning to the log and the chat?

rlament:
Maybe not so much a bug but rather masking an incomplete module. It is quite possible that a module designer adds a list and forgets to populate it. Meanwhile running the module give no indication of potentially missing functionality. But this could be said of other components.

It is valid to have an empty list with sub-lists. Any warning would have to be solely on leaf nodes.

Consider this unconventional Turn Tracker design. There is a desire to fire a Global Hot Key when the Turn Tracker advances. The turn level itself may be managed elsewhere in another component, perhaps a counter/label on the map. Or alternatively the Turn Level Format references a Global Property and the Hotkey updates that property. The empty list can define what text the Turn Counter displays, including Global Properties. Without an empty Turn List the Global Hot Key does not fire. There are situations where an empty Turn List is valid design such as this. It will be hard to detect bad design without generating false positives.

PR #14490, opened 2026-02-11
Add keyboard shortcuts to menus

rlament:
Add keyboard shortcuts to the Module Manager, Editor and the Player windows. The shortcut keys are language specific and are defined in the language resource files. The English and all European languages have shortcut keys assigned. This was an interesting exercise in increasing my gaming vocabulary.

I welcome native language speakers to chime in on the language specific shortcut key assignments. Almost undoubtedly there are predefined shortcut keys for the typical operating system commands that I am not aware of. My attempt here is to make shortcut keys unique per sub-menu.

In the Editor, hotkeys are not assigned since most of the editor commands have accelerator keys.

Also in the editor, 'File|Save As' had been given the control-A accelerator key. This is the 'Select All' command which is intercepted by the edit controls. Only if the menu has focus does the control-A accelerator key work as a Save-As. This somewhat defeats the purpose of having an accelerator key. Changing Save-As to `shift-control-S` makes Save-As available just like control-S for File|Save.

Toolbar buttons have precedence over menus. For example if a module uses Alt-F for an Action Button, then the File menu for the English language (and some others) will not open. F10 would if it is not also reassigned. Shortcut keys can be useful but they can be masked by modules.


rlament:
> defining menu short-cuts like Ctrl-S, Ctrl-A, Ctrl-F, and so on ...
> Will the key combination be sent to the menu or to the module?

We have to frame this discussion solely around the Player window. That window specifically has the potential conflicts with modules. The Player window has three top level menu items, in English _File_, _Tools_, and _Help_. That's Alt-F, Alt-T and Alt-H respectively. Any of these move focus to the menu. BTW, F10 is another key that sends focus to the menu. Once there the user can press the underscored letter corresponding to desired command. As an example, for <ins>N</ins>ew Game it's Alt-F -> N. Or the user can traverse the menu with arrow keys and press the _Enter_ key. The potential conflict is only with the three Alt codes.

It does get interesting because the three Alt codes change from language to language. The German word for _Tools_ is _Werkzeuge_, so Alt-W would make a lot of sense. But Alt-W is the _Begin log file_ keycode unless a module designer explicitly changes it. This means that certain letters are off limits and annoyed German users have to use some less obvious letter. Alt-S is the server control window putting it off limits too. Both of these are configurable in the Global Options, but it probably is best to avoid the default settings.

Buttons on the toolbar take precedence over the menu. You get solely the toolbar action and not both (i.e. not the menu). If a counter(s) has focus, it takes precedence over the toolbar buttons.

In regards to control codes, the Module Editor has always had these defined. There is nothing new here for control codes. I just pointed out the conflict with _Save As_ and the resolution.

riverwanderer:
On MacOS I see nothing different. Is the change restricted to Windows ?

rlament:
> On MacOS I see nothing different. Is the change restricted to Windows ?

No, should work with Mac too. On the Module Manager, the first character of every File menu entry should be underlined. On Windows it depends on how the menu is opened. Mac might be the same. Use the keyboard to open the menu, Alt-F or whatever is the equivalent for Alt on the Mac. When clicking on 'File' with the mouse the underscores are not shown. Does Alt-F Q close the Module Manager? If so then it is working.

On Linux the underscores are present regardless. The appearance on each O/S may be affected by the look and feel? I'm not sure.

riverwanderer:
> No, should work with Mac too.

I tried Alt-F (Option-F on Mac) etc and it does nothing. I have a feeling that MacOS does not deal with shortcuts in this way. My "[research](https://share.google/aimode/56ZxA3akLwT2NUZER)" seems to confirm that.

rlament:
My understanding of the dynamic nature of the underscore is basically a compromise between asethetics and accessibility. Mouse users don't care to see them. They do assist keyboard users with speedier navigation than with arrow keys alone. The behaviour is not limited to Swing or any particular look and feel, VSCode on Windows works the same way.

Since MacOS completely disregards the shortcut keys instead using the first letters of the menu item, for greater cross-platform commonality the best shortcut key would be the first letter of each menu item. But, any pre-existing shortcuts should remain unchanged as much as possible. I have compared the latest release with this PR and reverted some of my earlier changes.

@cholmcc Are you considering any modifications to the Danish shortcuts? I can update shortcuts as you see fit.

cholmcc:
> @cholmcc Are you considering any modifications to the Danish shortcuts? I can update shortcuts as you see fit.

Note, my initial objection was to _hotkey_ short-cuts - e.g., a key like `Ctrl-A` for _Quit_.   However, what you are defining are _mnemonic_ keys - e.g., `Alt-F+A` for _Quit_.   For mnemonic keys, it makes sense to translate them according to the menu entry.   E.g., 

 **Module mananger**

| File menu | Key | Danish | Key | 
|--------------|------|------|---|
| Open Module ...| O | Åben modul ... | Å |
| Edit Module ... | E | Rediger modul ... | R |
| New Module ...| N | Nyt modul ... | N |
| Preferences | P | Præferencer | P |
| Quit | Q | Afslut | A |

| Tools menu | Key | Danish | Key | 
|-------------------|------|-----------|---|
| Server status | S | Server status | S |
| Convert Module ... | C | Konverter modul ... | K |
| Clear Tile Cache | X | Ryd flise lager |  R |

| Help menu | Key | Danish | Key |
|------------------|-------|-------------|----|
| Help | H | Hjælp | H |
| Take the tour | T | Tag rundturen | T |
| Vassal's users guide | V | Vassel brugermanual | V |
| Check for updates ... | U | Tjek for opdateringer ... | O |
| Show Error Log | L | Vis fejl rapporter |F |
| About Vassal | A | Om Vassal | M |

**Player**

| File menu | key | Danish | Key |
|----------------|------|-------------|-----|
| New game ... | N | Nyt spil ... | N |
| Load Game or Log ... | O | Åben spil eller optagelse ...| Å |
| Open recent ... | R | Åben seneste ... | S | 
| Save Game ... | S | Gem spil ...| G | 
| Save Game as ... | A | Gem spil som ... | E |
| Close Game | X | Afslut spil | F |
| Start Log ...| L | Start optagelse ...| O | 
| End Log | E | Afslut optagelse | L |
| Preferences | P | Præferencer | P |
| Scenario Options | C | Scenarie valg | V |
| Quit | Q | Afslut | A |

| Tools menu | Key | Danish | Key |
|------------------|-------|------------|------|
| Load Log & Fast-forward | F | Åben optagelse og spol fremad | Å |
| Load Log, Fast-forward, and append | A | Åben optagelse, spol fremad, og tilføj | T |
| Load continuation | C| Åben forsættelse | F |
| Refresh Counters | R| Opdater brikker | O |
| Show Debug Window | D | Vis afluser vindue | V | 

However, I still find it problematic to add these mnemonic keys when 
- they are not shown by default with certain Java Look'n'Feels, but the still work - users may get confused that pressing `Alt-F` followed by `P` opens the preference dialog when there's no indication in the UI of such a mnemonic key 
- they do not work on certain platforms - e.g., MacOS - at all - in fact, `Alt-F` doesn't even show the _File_ menu (`Ctrl-F2` does). 

Yours,
_Christian_

rlament:
Updated the Danish translation file. These are the outstanding issues as far as I can tell.

1. What is the preferred keycode for the top level Værktøjer menu? Alt-T to match the English Tools, or Alt-V?
2. Typo in Åben forsættelse? Should be fortsættelse, correct?
3. For the help menu, there is little value in adding mneumonic keys. I only added one for "About". Should _Om VASSAL_ stay as _M_ or is _O_ better since there are no other conflicts?
4. I left Scenarie indstillinger unchanged and used the letter _I_. Too many other messages use the word indstillinger to change the menu text. Better to be consistent with the terminology.
5. I took a stab at updating _GameState.old_continuation_warning_ since it refers to altered menu items. I think the message is outdated still referring to the old location of _Load Continuation_ under the files menu. Pointing this out in case you have a chance to review it. Although I'm not trying to turn this PR into a translation exercise.

The Vassal resource files use the term _shortcut_, so for source code consistency I kept with the same wording.

I will edit my earlier post and update the table.

PR #14493, opened 2026-02-13
Build RPM package and FlatPak recipe on release

cholmcc:
This patch will set-up the release workflow to also build a RedHat
package (RPM) and a FlatPak recipe (YAML) when the repository is tagged
for a release.

Both the RPM and the YAML will use the already build
`VASSAL-<release>-linux.tar.bz2` as the source of the RPM and the
installation image with FlatPak.  This saves a bunch of rebuilds and
size of installations.

Both the RPM and the YAML will provide full system integration, meaning
desktop menu entries and file associations.

The RPM is installed by downloading and double-clicking the file.
Alternatively one can do
```
sudo dnf install vassal-3.7.21-1.noarch.rpm
```

The YAML is installed by downloading and then running
```
sudo flatpak install org.freedesktop.Platform
sudo flatpak install org.freedesktop.Sdk
sudo flatpak install org.freedesktop.Sdk.Extension.openjdk25
sudo flatpak-builder --install build VASSAL-3.7.21-flatpak.yaml
```
(maybe one can do something from GUI - not too familiar with FlatPak).


NN708:
Hi @cholmcc, I just wanted to let you know that the Flatpak is already built on [Flathub](https://flathub.org/en/apps/org.vassalengine.vassal). The manifest and MetaInfo files can be found here:
+ https://github.com/flathub/org.vassalengine.vassal/blob/master/org.vassalengine.vassal.yml
+ https://github.com/vassalengine/vassal/blob/master/dist/linux/org.vassalengine.vassal.metainfo.xml

cholmcc:
> @cholmcc I understand your point now. However, Flathub has strict publishing requirements: apps must be built from source whenever possible ([ref](https://docs.flathub.org/docs/for-app-authors/requirements#building-from-source)), 

That's funny, because elsewhere - can't remember where - they suggest building bundles from existing `.deb` or `.rpm` packages of binaries. 

Also, since the Maven build of Vassal pulls down binary blobs of third party JARs, that seems a funny requirement.  For Debian builds, that would be entirely disabled, and Maven builds can only use sanctioned JARs.  It is not clear to me if `flathub` does something similar. 

> and metadata files should be distributed in the source code rather than included in the Flatpak manifest ([ref](https://docs.flathub.org/docs/for-app-authors/requirements#required-metadata)).

I think that's a weird read.   Do they really want to force up-stream to include a file?   That makes no sense.   Also the text itself seems contradictory - it says the file is "typically" included, and then later that "must be integrated" - so what is it?   I find it hard to believe that all the applications on FlatPak include a `.metainfo.xml` file in their source directories. 

> The difficulty in maintaining dependency sources stems from the lack of Maven support here: [flatpak/flatpak-builder-tools#487](https://github.com/flatpak/flatpak-builder-tools/issues/487). If someone were to implement this, the process could be automated.

Then why not by-pass Maven altogether.   Vassal really doesn't need to be build with Maven - a simple `Makefile` will suffice, or it can even be encoded directly into the manifest.  For example, for the Debian package, where not all dependencies are available in the Debian Maven repository, I write the build "by-hand" in `debian/rules`.  

Of course, you still need to get the dependency JARs somehow.   

Could the script linked to in [https://github.com/flatpak/flatpak-builder-tools/issues/487](https://github.com/flatpak/flatpak-builder-tools/issues/487) - i.e., [`fetch-maven-deps-for-flatpak.sh`](https://github.com/lenucksi/SieveEditor/blob/master/scripts/fetch-maven-deps-for-flatpak.sh) not be run as part of the Vassal release process?   Then, the generated `maven-dependencies.yaml` and a generated `org.vassalengine.vassal.yaml` be pushed to `flathub/org.vassalengine.org`. 

Another option, omitting `flathub` altogether, would be to generate the manifest as done in the patch at hand, and then either have people build from that, or make your own distribution server of the bundle.  Probably less attractive. 

I will close this PR, since to get things to work with `flathub` will require more work and should be done in a separate branch.   I'll leave the branch for future reference. 

Yours,
_Christian_

PR #14533, opened 2026-03-04
Configure mnemonics for non-English yes/no buttons

rlament:
This pull request addresses two related issues.

1. English users can use mnemonic keys (Alt-Y/Alt-N) to answer yes/no prompts. For non-English locales the mnemonic keys have not been translated. This PR applies the translation.
2. For a Yes-No-Cancel OptionPane where the buttons are assigned custom text, the mnemonic keys did not work at all for any locale. This PR enables the Yes/No mnemonics.

This image shows the before and after appearance of queries in German.
option_pane

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
@@ -50,8 +50,12 @@ import javax.swing.AbstractAction;
 import javax.swing.Action;
 import javax.swing.ImageIcon;
 import javax.swing.JButton;
+import javax.swing.JDialog;
 import javax.swing.JOptionPane;
+import javax.swing.UIManager;
 import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.awt.Dimension;
 import java.awt.event.InputEvent;
 import java.awt.event.KeyEvent;
 import java.io.BufferedOutputStream;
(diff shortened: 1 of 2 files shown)


vassal-app/src/main/java/VASSAL/build/module/BasicLogger.java
```diff
@@ -405,29 +409,54 @@ public void queryNewLogFile(boolean atStart, boolean fastForwarded) {
...
+      cancel.addActionListener(new ActionListener() {
+        @Override
+        public void actionPerformed(ActionEvent event) {
+          pane.setValue(JOptionPane.CANCEL_OPTION);
+          g.getPrefs().setValue(prefName, Boolean.FALSE);
+        }
+      });
+
+      yes.setPreferredSize(new Dimension(100, yes.getPreferredSize().height));
+      no.setPreferredSize(new Dimension(100, no.getPreferredSize().height));
```
uckelman: I suspect this is not right for the preferred width. That should depend on the font size and text length.

Isn't the way to do all of this setting the locale on `JOptionPane`?

https://stackoverflow.com/a/18189543/181106

rlament: It seems that when an `options` array of buttons is provided to `JOptionPane` the buttons are not formatted in the same way as buttons created internally by `JOptionPane`. Without resizing the buttons, the custom yes/no buttons appear smaller in width than in the existing dialog. Hard coding the dimension was my brute force attempt to make the buttons appear as before.

The buttons wrap the yes/no text, which is typically short words. Instead of hard coding a dimension, I changed this to double the left/right insets. This seems simpler than dealing with fonts and text lengths.

> Isn't the way to do all of this setting the locale on JOptionPane?

I never got `JOptionPane.setDefaultLocale()` to work with custom buttons (i.e. dialogs that don't use the typical yes/no/cancel text). Where is did work, it didn't work for all supported languages. I don't recall the details as this was some time ago.

uckelman: I tried `JOptionPane.setDefaultLocale()` and it appears to work: https://github.com/vassalengine/vassal/pull/15040

Do you see anywhere that it doesn't?

rlament: `JOptionPane.setDefaultLocale()` does not support Chinese, Dutch, Polish, and Danish locales. Commented out the `UIManager.put` lines in Resources.java and tested this by quitting a game and reviewing the "Save Game?" prompt.

The intent of this PR is to add mnemonics to the non-English languages. The translations are already done through the resource files. In English I can Alt-F -> Q to quit followed by Alt-N to not save. It would be nice to do this in the other languages too using the locale specific mnemonic.

PR #14535, opened 2026-03-04
New logfile cleanup

bacota:


vassal-app/src/main/java/VASSAL/build/module/LogCleaner.java
```diff
@@ -0,0 +1,198 @@
...
+    Command cleanedRoot = logCommand;
+    
+    // Clear the existing subcommands by creating a new root of the same type
+    // For simplicity, we'll just append to a NullCommand which will work for encoding
+    cleanedRoot = new NullCommand();
+    
+    // Append only the commands we want to keep
+    for (int i = 0; i < subCommands.length; i++) {
+      if (!toRemove[i]) {
+        cleanedRoot.append(subCommands[i]);
```
rlament, line 172: Use assignment to get rid of the top level NullCommand.
`cleanedRoot = cleanedRoot.append(subCommands[i]);`

Copilot's solution is far too complicated.

riverwanderer: I wonder whether this PR is desirable from the viewpoint of facilitating cheating in PBEM games, or to put it more kindly, presenting a temptation to those who having a run of bad luck. 

I know that this is a controversial topic but it is not the only example of Vassal's design relying on obscurity to keep players on the straight and narrow.

Assuming the legitimate use-case warrants the PR, perhaps some mitigation can be considered.  Eg insert a warning into the log when undone actions are removed. Such as a summary at the end of each block of undos.

rlament: Fair PBEM involves considerable faith and trust in opponents' honesty. I had the similar thought about making it easier to redo moves. As it is there are other ways to redo for a better outcome particularily in the PBEM setting.

How often are there legitimate corrections in log files? It seems that this would only be an occasional occurrence. Although in situations where there are multiple overlapping undo/redo steps it can be difficult to follow.

Perhaps a better approach is not writing the output to a file. Instead, filter out the undo commands on playback. Effectively moving the filter from the producer to the consumer. A preference setting could enable this feature. The chat log could output a "skipping undo" message in place of the do/undo block.

PR #14538, opened 2026-03-05
14537 Preserve the user directory preference after canceling from a file load or save dialog

### LINKED ISSUE #14537: The save game directory changes after canceling a save
When choosing to cancel the saving of a game, the save game directory (_saveGameDir_ in the preferences file) changes. This is the directory initially presented by the FileChooser. It is very handy for it to present the directory of the last save. It is disconcerting when it starts in an unexpected directory.

To reproduce, first start a game and save it to a known location. Then follow these steps:
- Start a New Game
- Move some pieces to change the game state
- File -> Close Game
- Save Game? -> yes
- Note the initial directory in the FileChooser. It should be at the location of the saved game.
- FileChooser -> cancel

Repeat those steps, but notice this time the FileChooser opens in a different directory. It seems to be going up in the folder hierarchy to the parent directory.

rlament:
In FileChooser, the _setCurrentDirectory()_ and _getCurrentDirectory()_ were not complementary functions. The _get_ should return the _set_ value. When _set_ was given a directory, the _get_ returned the parent. _getCurrentDirectory()_ now checks if the stored path is a directory and returns it.

Closes #14537

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
+++ b/vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
@@ -394,12 +394,11 @@ public abstract class FileChooser {
           value = ERROR_OPTION;
           cur = null;
         }
+        updateDirectoryPreference();
       }
       else {
         value = CANCEL_OPTION;
-        cur = null;
       }
-      updateDirectoryPreference();
       return value;
     }
 


vassal-app/src/main/java/VASSAL/tools/filechooser/FileChooser.java
```diff
@@ -394,12 +394,11 @@ else if (cur.isDirectory() && mode == DIRECTORIES_ONLY) {
           value = ERROR_OPTION;
           cur = null;
         }
+        updateDirectoryPreference();
```
uckelman, line 397: Is this the correct place to move `updateDirectoryPreference()`? This also catches the `ERROR_OPTION` case. Should `updateDirectoryPreference` be called only in the `APPROVE_OPTION` cases?

rlament: Yes, this place is fine because `updateDirectoryPreference` does nothing in the case of `ERROR_OPTION`. Preferences are not updated within `updateDirectoryPreference` when `cur` is null.

uckelman: It's true that `updateDirectoryPreference()` does nothing now on `ERROR_OPTION`, but only because `cur` is set to null first, which need not remain the case. I'm going to move `updateDirectoryPreference()` out of the `ERROR_OPTION` case to avoid this.

PR #14547, opened 2026-03-09
13702 Optimize sum and count expressions

### LINKED ISSUE #13702: Optimise CountZone() evaluation order
Based on observation of errors from this expression, it seems to me that the Beanshell expression is evaluated on pieces that don't possess the named property:

`CountZone(cBoard,cMain,"cmdCard","{!ObscuredToOthers && FromSide == gActiveSide}") `

This would imply that pieces are scanned for a match on the expression first, rather than looking for pieces with property "cmdCard". If that's right then this type of function might give improved performance if the search were first on the property.

Perhaps it's worse than that, and both conditions are evaluated (i.e. a redundant check is performed).

If this is an issue with CountZone(), then all forms of Sum/Count functions should be checked where both Property and Expression are accepted.

rlament:
Wondering if these changes to the sum/count functions make any noticeable impact on modules. Doing some very basic profiling with _System.nanoTime_ shows an improvement depending on game piece design and filtering (match expression) options.

The existing code relies on _updateTotal()_ for computing both sums and counts. These actions are similar yet different enough to warrant separate functions. They do handle properties differently. Since these functions are invoked from within loops, any reduction in conditional testing and branching can have an impact.

The _updateTotal()_ function applies the match expression (if defined) to all pieces first, whether or not those pieces have the property of interest. The BeanShell processing is computationally expensive and it is best to avoid it when possible. The new _updateCount()_ and _updateSum()_ functions compute the match expression as a matter of last resort. Take the sum function, if the property is non-existent or is not a number or is zero, then checking the match expression is unnecessary and thus never invoked.

The greatest benefit would be in modules that subset properties. With differing units having different properties, summing on something like _alliedPower_ or _axisPower_ could eliminate half the counters from a calculation before applying any match expressions. In cases where there are common properties with a reliance on match expressions, this code change likely won't make much of a difference.

Profiling with unit tests shows a marked improvement. Whether this translates to more responsive modules is to be seen. The matching expression is too powerful not to use, but at least it won't be invoked on pieces unrelated to the property in question.

Closes #13702

### DIFF
--- a/vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
+++ b/vassal-app/src/main/java/VASSAL/script/ExpressionInterpreter.java
@@ -651,11 +651,11 @@ public class ExpressionInterpreter extends AbstractInterpreter implements Loopab
       final GamePiece piece = (GamePiece) ps;
       final Stack s = ((GamePiece) ps).getParent();
       if (s == null) {
-        result = updateTotal(result, piece, property, filter, true);
+        result = updateSum(result, piece, property, filter);
       }
       else {
         for (final GamePiece gamePiece : s.asList()) {
-          result = updateTotal(result, gamePiece, property, filter, true);
+          result = updateSum(result, gamePiece, property, filter);
         }
       }
     }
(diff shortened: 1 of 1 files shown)


uckelman:
If you really want this to be faster, the way to do it is to hoist the null and emptiness checks for `propertyName` and the null check for `filter` out of the loops, which you can do because they are loop invariant.

uckelman:
For `updateSum` and `updateCount`, pass in an iterable rather than have the loop on the outside. Then you can hoist the null and emptiness checks outside of the loop.

PR #14556, opened 2026-03-15
14479 Support Multi-Location commands within Sub-Menus

### LINKED ISSUE #14479: Sub-Menu not usable with Multi-Location Command
Sub-Menus collect menu items by the menu command name. However Multi-Location Commands may have unpredictable names. Is it possible for Sub-Menus to contain menu items without using exact matches of menu command names? Would something like a regex would be possible?

rlament:
When combining the Multi-Location command with other commands the pop-up menu can get unwieldly. Moving the Multi-Location command to a sub-menu would logically separate location commands from other commands.

This one possible solution adds a new property to the Multi-Location command. Each Multi-Location command receives a copy of this string property making it common to all commands. Adding this same string to the Sub-Command list of a Sub-Menu links all the Multi-Location commands to the Sub-Menu.

Closes #14479 

MultiLocationCommand_submenu


vassal-app/src/main/java/VASSAL/counters/MultiLocationCommand.java
```diff
@@ -206,6 +212,7 @@ public void mySetType(String type) {
     menuText.setFormat(st.nextToken(Resources.getString("Editor.MultiLocationCommand.loc_default_command")));
     key = st.nextNamedKeyStroke();
     curMapOnly = st.nextBoolean(true);
+    subMenu = st.nextToken("");
```
uckelman: Won't this break things if not added at the end?

vassal-app/src/main/java/VASSAL/counters/MultiLocationCommand.java
```diff
@@ -553,6 +567,7 @@ public String getType() {
         .append(menuTextInput.getValueString())
         .append(keyInput.getValueString())
         .append(curMapOnlyInput.getValueString())
+        .append(subMenuInput.getValueString())
```
uckelman: Similarly, shouldn't any new fields be stored at the end of the sequence?

vassal-app/src/main/java/VASSAL/counters/MultiLocationCommand.java
rlament: Yes, but at the time of fixing the merge conflict master was not released. At this point the only release that would break is 3.8 Beta 1. Things are in a state of flux. If you deem this as a major risk, I'll move the field to the end. Let me know one way or the other.

uckelman: No, I mean that inserting new items into the sequence in places other than the end will prevent 3.8 from correctly loading modules using MultiLocationCommand written by anything before 3.8.

rlament: This PR is targeting master not 3.7. The 3.7 release reads up to field `curMapOnly`. New to 3.8 are `menuSort`, `ascending` and `numeric`. Therefore technically anything after `curMapOnly` is at the end. Anything before 3.8 is unaffected. This comes down to the sequencing of PR merges. The other one (#13168) happened to be first. If breaking 3.8 Beta 1 is a big deal, I can move this field to the very end.

uckelman: > If breaking 3.8 Beta 1 is a big deal, I can move this field to the very end.

Hardly anyone has tried 3.8.0-beta1 at this point, but I'd rather we not gratuitously break it.

uckelman: If you fix this, then I'll merge the PR.

rlament: Ok, should be ready to merge now. Submenu is last in the definition. Added a description to the reference manual.

PR #14579, opened 2026-03-28
Make charts zoomable

### LINKED ISSUE #13035: Universal Zoom
It seems that some windows don't have zoom controls (not sure if this is "by default" or are completely unable to have zoom controls).

Either way, Zoom should be a universal function for every window handled by VASSAL (not handled by individual modules).

This is especially important for users with HiDPI screens.

cholmcc:
This PR will make charts zoomable, so as to allow users to zoom in or out on an (image) chart display.  This is useful in case the chart has small sized text or is very large, increasing the overall accessibility of modules.

The zoom feature can be disabled if a module designer prefers the old
behaviour.   If zoom is enabled, then the chart display will listen for
Ctrl-MouseWheel and configurable key events (`+` and `-`) by default.

The added configurables are
- whether to enable zoom all together (enabled by default)
- the least zoom factor (30% by default)
- the largest zoom factor (300% by default)
- the zoom in hotkey (`+` by default)
- the zoom out hotkey (`-` by default)

One could consider to add a tool-bar and zoom in and out buttons.

This partially closes #13035
Closes #14343

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/widget/Chart.java
+++ b/vassal-app/src/main/java/VASSAL/build/widget/Chart.java
@@ -17,52 +17,159 @@
  */
 package VASSAL.build.widget;
 
+import java.awt.Graphics;
+import java.awt.Graphics2D;
 import java.awt.Component;
 import java.awt.Dimension;
 import java.awt.Image;
+import java.awt.event.ActionEvent;
+import java.awt.event.InputEvent;
+import java.awt.event.MouseWheelEvent;
+import java.awt.event.MouseWheelListener;
+import java.awt.geom.AffineTransform;
 import java.io.File;
 import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
 
 import javax.swing.JLabel;
 import javax.swing.JScrollPane;
+import javax.swing.SwingUtilities;
 
 import VASSAL.build.Buildable;
 import VASSAL.build.Widget;
+import VASSAL.build.GameModule;
 import VASSAL.build.module.documentation.HelpFile;
+import VASSAL.configure.NamedHotKeyConfigurer;
+import VASSAL.configure.VisibilityCondition;
 import VASSAL.i18n.Resources;
 import VASSAL.tools.AdjustableSpeedScrollPane;
 import VASSAL.tools.DataArchive;
+import VASSAL.tools.ErrorDialog;
 import VASSAL.tools.imageop.Op;
 import VASSAL.tools.imageop.OpIcon;
 import VASSAL.tools.imageop.SourceOp;
+import VASSAL.tools.swing.SwingUtils;
+import VASSAL.tools.NamedKeyStroke;
+import VASSAL.tools.NamedKeyStrokeListener;
 
 /**
  * A Chart is used for displaying charts and tables for the module.
  * The charts are loaded as images stored in the DataArchive. As a subclass
  * of Widget, a Chart may be added to any Widget, but it may not contain
  * children of its own.
+ *
+ * Scrollbars and zoom capability is only added if the embedded image
+ * is larger than 300x300 pixels.
+ * 
+ * The widget allows zooming in and out on the embedded image either
+ * using the mouse scroll key, with the control key held down, or by
+ * configurable keyboard short cuts, defaults to '+' and '-'.
+ *
+ * TODO:
+ * - Possibly make a toolbar with zoom in/out buttons, in which case
+ *   their icons should be configurable too.
  */
 public class Chart extends Widget {
   public static final String NAME = "chartName"; //NON-NLS
   public static final String FILE = "fileName"; //NON-NLS
+  public static final String MIN_ZOOM = "minZoom"; //NON-NLS
+  public static final String MAX_ZOOM = "maxZoom"; //NON-NLS
+  public static final String ENABLE_ZOOM = "enableZoom"; //NON-NLS
+  public static final String ZOOM_IN_KEY = "zoomInKey"; //NON-NLS
+  public static final String ZOOM_OUT_KEY = "zoomOutKey"; //NON-NLS
   private Component chart;
   private String fileName;
   private SourceOp srcOp;
   private JLabel label;
+  private double currentScale = 1;
+  private Boolean enableZoom = true;
+  private Double minZoom = .3;
+  private Double maxZoom = 3.;
+  private NamedKeyStroke zoomInKey = NamedKeyStroke.of('=', InputEvent.SHIFT_MASK);
+  private NamedKeyStroke zoomOutKey = NamedKeyStroke.of('-');
 
+  /**
+   * Construct the swing component.
+   */
   @Override
   public Component getComponent() {
     if (chart == null) {
       label = new JLabel();
       srcOp = (fileName == null || fileName.isBlank()) ? null : Op.load(fileName);
       if (srcOp != null) {
-        label.setIcon(new OpIcon(srcOp));
+        label.setIcon(new OpIcon(srcOp) {
+            /**
+             * This version _does not_ reset the scaling done in the
+             * graphics context.
+             *
+             * It seems that OpIcon rescales the images so that it
+             * fits with the expectations of the toolbar UI.  However,
+             * when we are dealing with charts, we are not constrained
+             * to the toolbar UI, and we should _not_ reset to some
+             * default, system dependent, scale.
+             *
+             * Instead, we use the current scaling to draw the chart.
+             */
+            @Override 
+            public void paintIcon(Component c, Graphics g, int x, int y) {
+              final Graphics2D g2d = (Graphics2D) g;
+              final AffineTransform orig_t = g2d.getTransform();
+              
+              try {
+                g.drawImage(Op.scale(sop, currentScale).getImage(null), 0, 0, c);
+              }
+              catch (CancellationException | InterruptedException e) {
+                ErrorDialog.bug(e);
+              }
+              catch (ExecutionException e) {
+                if (!Op.handleException(e)) ErrorDialog.bug(e);
+              }
+              
+              g2d.setTransform(orig_t);
+            }
+          });
       }
       final Dimension d = label.getPreferredSize();
-      if (d.width > 300 || d.height > 300) {
+      if (d.width > 300 || d.height > 300 || enableZoom) {
         final JScrollPane scroll = new AdjustableSpeedScrollPane(label);
         scroll.getViewport().setPreferredSize(label.getPreferredSize());
         scroll.getViewport().setAlignmentY(0.0F);
+
+        if (enableZoom) {
+          // Key handling.  This should probably be configurable,
+          // Note, we filter on whether the event originated in this
+          // window (actually it's parent).
+          final NamedKeyStrokeListener zoomIn  = new NamedKeyStrokeListener(e -> {
+            if (!fromThisWindow(e)) return;
+            zoom(+1); });
+          final NamedKeyStrokeListener zoomOut = new NamedKeyStrokeListener(e -> {
+            if (!fromThisWindow(e)) return;
+            zoom(-1); });
+          zoomIn.setKeyStroke(zoomInKey);
+          zoomOut.setKeyStroke(zoomOutKey);
+
+          GameModule.getGameModule().addKeyStrokeListener(zoomIn);
+          GameModule.getGameModule().addKeyStrokeListener(zoomOut);
+        
+          // Listen for control-mouse-wheel events. 
+          scroll.getViewport().addMouseWheelListener(new MouseWheelListener() {
+              /**
+               * Zoom view of image when mouse wheel is rolled up or
+               * down
+               */
+              @Override
+              public void mouseWheelMoved(MouseWheelEvent e) {
+                if (!SwingUtils.isSelectionToggle(e)) return;
+                
+                if (e.getScrollType() != MouseWheelEvent.WHEEL_UNIT_SCROLL)
+                  return;
+                
+                zoom(e.getUnitsToScroll() < 0 ? 1 : -1);
+              }
+            });
+        }
         chart = scroll;
       }
       else {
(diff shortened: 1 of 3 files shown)


cholmcc:
This PR _could_ go into 3.7.22 as it will not introduce a backward-incompatibility (the attributes to enable/disable zoom will simply be ignored in older Vassals - thus retaining old behaviour).

vassal-app/src/main/java/VASSAL/build/widget/Chart.java
```diff
@@ -17,52 +17,159 @@
...
+   * Construct the swing component.
+   */
   @Override
   public Component getComponent() {
     if (chart == null) {
       label = new JLabel();
       srcOp = (fileName == null || fileName.isBlank()) ? null : Op.load(fileName);
       if (srcOp != null) {
-        label.setIcon(new OpIcon(srcOp));
+        label.setIcon(new OpIcon(srcOp) {
```
uckelman, line 102: Maybe this shouldn't be done with OpIcon. Maps already have scaling. Would it make sense for Charts to simply become Maps?

cholmcc: Often people use maps for charts to make them zoomable - as a wor around. However charts and maps are conceptually very different. Pieces can be placed on maps, and the may have associated key commands. Charts simply display information. I would keep Charts but facilitate zooming for accessibility. I think `OpIcon` - locally specialised - is quite ok.  Still want to be able to cache and so on.

delatbabel:
LGTM. 2 of the modules that I am maintaining would like to have this feature.

cholmcc:
Text for Release Notes:

## Charts are zoom-able 

Charts are now by default zoom-able.   A user may zoom in and out on a chart by holding down `Ctrl` and moving up and down with the mouse scroll-wheel (or touch-pad emulation). 

The chart window configuration also allows a module designer to set some parameters on zooming: 
* whether to enable zoom all together (enabled by default)
* the least zoom factor (30% by default)
* the largest zoom factor (300% by default)
* the zoom in hotkey (`+` by default)
* the zoom out hotkey (`-` by default)

Chart

uckelman:
Consider that a better solution would be tabbed Maps. That would make Charts redundant and we could advise people not to use Charts at all.

delatbabel:
There are frequently things that need to be dragged between a chart and a map, often in each direction. So having the ability to have tabbed charts that are zoomable in a separate window to a zoomable map would be preferable.

Can we have tabbed charts|maps where one tab can be torn off and moved to a new window? Like the way that most browsers work these days.

uckelman:
> There are frequently things that need to be dragged between a chart and a map

You can't drag anything to or from a Chart. They don't support having pieces on them. If you need a chart that supports having pieces on it, what you really have is a display, not a chart, and you should make that a Map. 

This is one of the reasons why I suggested tabbed Maps, because tabbed Maps would make Charts obsolete.

riverwanderer:
> > There are frequently things that need to be dragged between a chart and a map
> 
> You can't drag anything to or from a Chart. They don't support having pieces on them. If you need a chart that supports having pieces on it, what you really have is a display, not a chart, and you should make that a Map.
> 
> This is one of the reasons why I suggested tabbed Maps, because tabbed Maps would make Charts obsolete.

I don't know if I've missed something and you already know this but  a map can be defined within a chart window - thereby making tabbed Maps possible. It's a fairly recent innovation, I believe. Works nicely though.

uckelman:
> I don't know if I've missed something and you already know this but a map can be defined within a chart window

I had forgotten about that. In that case, the approach I suggest is:

* Deprecate Chart.
* Remove the ability to add Chart in the Editor.
* Convert any Chart seen in a buildFile.xml to a Map on load.
* Look into making the tabs in a Chart Window draggable.

riverwanderer:
> > I don't know if I've missed something and you already know this but a map can be defined within a chart window
> 
> I had forgotten about that. In that case, the approach I suggest is:
> 
> * Deprecate Chart.
> * Remove the ability to add Chart in the Editor.
> * Convert any Chart seen in a buildFile.xml to a Map on load.
> * Look into making the tabs in a Chart Window draggable.

I'd like to add: provide a mechanism to protect "chart maps" against pieces being dragged onto the chart. The obvious way would be to enable Private Maps to be added to a chart window. In fact, to support backward coompatibility, converted charts should be "read-only" in that way.

uckelman:
> provide a mechanism to protect "chart maps" against pieces being dragged onto the chart.

Why? Is there some problem that could cause?

cholmcc:
> There are frequently things that need to be dragged between a chart and a map, often in each direction. So having the ability to have tabbed charts that are zoomable in a separate window to a zoomable map would be preferable.

If you need to have pieces on a "Chart" of some kind, then that chart should be a `Map`. 

If the chart is only there to show information - which is often the case - then it should be a `Chart`. 

Currently, `Chart` elements cannot be zoomed in or out, leaving people with a visual impairment at a disadvantage.  This PR aims to make informational (only) charts zoom-able.  

"Charts" that need to contain pieces or the like are _not_ affected because they already have the the zoom capability via the `Map` interface. 

Yours,
_Christian_

cholmcc:
> > provide a mechanism to protect "chart maps" against pieces being dragged onto the chart.
> 
> Why? Is there some problem that could cause?

If the chart is really only there to give information - say a TEC or CRT - then a user should _not_ be able to drag a piece onto that map.  It can create all sorts of havoc in the module logic if that is the case. 

I think it is a really _really_ bad idea to deprecate `Chart`.   `Chart` serves a very particular purpose - information only - that `Map` is rather ill suited to handle - precisely because you can place pieces there.    

Also, it can potentially create havoc if `Chart` are automatically changed to `Map`, and it actually defies the purpose of what module developers put into a `Chart`.   Now, some module developers have used `Map` to show information only images - precisely because `Chart` does not allow for zoom.   The prevalence of `Map` for charts is a result of the short-coming of `Chart ` to provide zoom - _not_ because module developers _need_ the `Map` capabilities (although some do). 

Point of example [Advanced Squad Leader: Starter Kit #1](https://vassalengine.org/library/projects/advanced_squad_leader_starter_kit_1_cholmcc): 

- I use `Chart` elements to show TEC, etc. in a tabbed window.   It would be wrong if I could drag _any_ piece on to these charts.
- I use a `Map` element to show the IFT - in the same tabbed window - so that I can put in pieces to highlight the column and row. 
- I a `Map` element - with multiple `Board` elements to show the Scenario Card, so that pieces can be placed there - ready to be dragged on to the main board. 

Yours,
_Christian_

delatbabel:
> If you need to have pieces on a "Chart" of some kind, then that chart should be a Map.
> 
> If the chart is only there to show information - which is often the case - then it should be a Chart.

Understood, but in this case I was using the "common law" definition of the word "Chart" rather than the Vassal definition. For example in WiF we have a "map" which is the main map and an "air combat chart" which is actually a different map. Air units are moved from the main map from the "air combat chart" map in order to conduct an air combat, and then later returned to the main map when the air combat is concluded.  Furthermore there are other "maps" which are really maps and can also contain air units that might be involved in combat.

It's a useful distinction that the main map is in one window and the "air combat chart" map is a different window because it's easier to drag and drop pieces between two different windows than it is between two tabs of the same window.

My comment was aimed at this suggestion:

> Consider that a better solution would be tabbed Maps. That would make Charts redundant and we could advise people not to use Charts at all.

Which I believe is a bad idea. I mean we can already have a tabbed "Charts" window in which maps can also be tabs, but deprecating all charts and making them all tabs of the main map would be a bad idea.

uckelman:
> deprecating all charts and making them all tabs of the main map would be a bad idea.

I agree---that's not what I was suggesting.

The component that is currently called a Chart Window---the thing which contains the tabs---seems fine to me as-is. At no point was I suggesting that we _move_ any components from where they are.

uckelman:
> If the chart is really only there to give information - say a TEC or CRT - then a user should not be able to drag a piece onto that map. It can create all sorts of havoc in the module logic if that is the case.

Do you know of any existing modules where dropping a piece onto a chart would cause a problem?

cholmcc:
> > If the chart is really only there to give information - say a TEC or CRT - then a user should not be able to drag a piece onto that map. It can create all sorts of havoc in the module logic if that is the case.
> 
> Do you know of any existing modules where dropping a piece onto a chart would cause a problem?

Yes - basically all the modules I did, and many others.   The problem is, that the piece is no longer in a predictable place.   That means that any logic the module may implement to take of aspects of the game _cannot_ assume that the piece is in a known map (it can basically be anywhere) including places that are not meaningful for the game, and so may break that logic.

I think the reverse is really the question:  Why would you want to drop a piece on an information-only chart?   If there was a point to dropping a piece (or some specific pieces) on to a chart, then that chart should have been implemented as a `Map` - not a `Chart`. 

Honestly, I'm not sure why we are spending this much energy on a relatively small change (enhancement I believe).  Adding the zoom capability to `Chart` has no impact on other elements, and it does not aim to substitute key aspects of a `Map` - namely that `Map`s can contain grids, zones, and pieces.    All the PR really does is make it possible to zoom in on a `Chart` in those modules that uses such an element for information-only displays.  It seems to me that the discussion is turning into a philosophical discussion of what it means to be a `Chart` ("what is it like to be a bat"), which is beside the point. 

Even if `Chart`s are to be deprecated and `Map`s are to replace them - which is an entirely different discussion, it does not really affect what this PR is about, which is to provide greater accessibility for users. 

Yours,
_Christian_

uckelman:
> Yes - basically all the modules I did, 

Would you give a specific example?

cholmcc:
> Would you give a specific example?

For example: 

[Napoleon at Waterloo (2nd & 3rd Ed)](https://vassalengine.org/library/projects/napoleon_at_waterloo_cholmcc)

If a unit is moved off-board to some "chart" map, then it will be wrongly counted for victory points.  If units are moved somewhere unpredictable - where they do not belong by any means - then reinforcements may not arrive as expected. 

[Panzerarmee Afrika: Rommel in the Desert, April 1941 - November 1942](https://vassalengine.org/library/projects/panzerarmee_afrika_cholmcc) 

Similarly, the count of Allied units in Libya can be wrong, affecting, for example, the Axis Malta option. 

These are only _some_ examples. 

The point is, pieces has _no business_ being on charts in these modules, and the modules works on the assumption that `Chart` does not accept pieces, and I'm sure you can find many other modules that works on the same assumption.  `Map` would break that in a way that is not fixable.   One cannot prevent a user from dragging a piece to `Map` window (Undoing a drag in a `Movement ends here` key command does not work). 

Conceptually, a `Chart` is _very_ different from a `Map`.   `Chart`s hold information, `Map` holds pieces.  Two very different things, and it is great that the modules can make that distinction.   For some users it is annoying that one cannot zoom in or out on a `Chart`, and this is what this PR fixes. 

Honestly, I really don't understand why this is even a discussion in this context.   The PR does not try to address any of these issues, nor does it enforce any kind of design with regards to these issues.   All it does it make it easier for people to see what is in a `Chart` by allowing them to zoom in or out.   Please explain why the larger - and somewhat beside-the-point - discussion is needed in this context.  

Yours,
_Christian_

uckelman:
> I really don't understand why this is even a discussion in this context.

Two reasons, one of which is procedural, one of which is specific to Charts:

* The reason we're having a design discussion after the PR was made is that we didn't have a design discussion before the PR was made. I would have raised these same issues prior to any work being done had I been aware anyone was planning to do the work.

* I have considered Charts a mistake the entire time I've been with the project. I would have objected to them at the time they were under consideration had I been there. They fundamentally are maps, but with their own worse rendering pipeline. If Charts had never been created, years of people wishing for scalable Charts, running out of memory due to large chart images, or using Charts to make displays but then finding out they can't put pieces on them could have been avoided. If Charts had never existed, module designers would have designed differently---but because Charts were added, we're now stuck with modules that depend on their characteristics. I had been hoping that with someone looking into Charts now, we would have an opportunity to retire Charts, but that seems not to be possible. Realizing that retiring Charts would be impossible for V3 was a result of this conversation for me, not something I knew to be so at the outset.

So, we can't remove Charts without breakage. Charts have the same problems with rendering that Boards did prior to the introduction of tiling, e.g., that the images can be very large---especially for Charts, which are much more likely to be zoomed in than out. The rendering pipeline used by Board is the same as what Chart needs. That rendering pipeline has had extensive real-world testing and should be reused. Similarly, the Zoomer used by Maps already does everything that a Chart could need. For consistency and to avoid duplication, the Zoomer should be used for Charts as well.

cholmcc:
> Two reasons, one of which is procedural, one of which is specific to Charts:

What I take from what you say, is that `Chart`s are not a _design_ flaw, but rather a flaw in _implementation_.   `Chart` rendering _should_ have been implemented as how boards are rendered - i.e., tilled and so on.   And then there's clearly different uses of that rendering implementation: 
* one that is used in conjunction with zones, grids, and pieces, and 
* one that is used to display information. 

I appreciate that it could be better if `Chart`s had a similar or even the same rendering algorithm as `Board`.  However, that's a decision way in the past.  In the mean time, this PR fixes a different implementation - and in some sense design - flaw with the _current_ `Chart` implementation. 

In fact ... 

> If Charts had never been created, years of people wishing for scalable Charts, running out of memory due to large chart images

... this PR may alleviate that issue.   Because you can zoom in on an image, you can now make due with an image of _lower_ resolution and _still_ be able to read the information.  Currently, there's a tendency to add very high resolution  images for charts, because that's the only way to make them legible.   With the PR, you can add a lower resolution image - say half - and zoom in twice and still be able to read the image. 

> or using Charts to make displays but then finding out they can't put pieces on them could have been avoided. I

In some sense, this is a documentation problem (or people not reading the documentation).   If it is obvious enough to people that charts cannot contain pieces or the like, then they surely wouldn't try to use them for that. 

> If Charts had never existed, module designers would have designed differently

Well, one would likely need to recreate the `Chart` component - i.e., a component that is purely to show information in and will never accept a drop of pieces or the like, so I don't think that assertion is entirely correct, and the problem would be a different - but equally difficult - one.  

What is actually a bit surprising - if true - is that the Java `Graphics` engine doesn't deal with all this.  Most modern desktop GUI renders can deal with rather high resolution images in a relatively painless way.  

Yours,
_Christian_

uckelman:
> I appreciate that it could be better if Charts had a similar or even the same rendering algorithm as Board. However, that's a decision way in the past. 

I extracted the renderer from Board this evening and have it working with Charts as well. I'll make a PR tomorrow.

Update: https://github.com/vassalengine/vassal/pull/15083

cholmcc:
> > I appreciate that it could be better if Charts had a similar or even the same rendering algorithm as Board. However, that's a decision way in the past.
> 
> I extracted the renderer from Board this evening and have it working with Charts as well. I'll make a PR tomorrow.
> 
> Update: #15083

OK.  So how does that affect this PR in terms of the Zoom?  Should `Chart`s rather have a `Zoomer` component?

uckelman:
> Should Charts rather have a Zoomer component?

Possibly. I'm looking into that now.

cholmcc:
> Possibly. I'm looking into that now.

OK.  

In the mean time I've updated this PR so that it applies cleanly to the `release-3.7` branch. 

I'm not so sure `Zoomer` is the best option.  As it is coded now, it is very `Map` specific.   Also, it requires a Tool-bar, which `Chart` does not have (unless, of course, it is added).    In that sense, I think this PR, for all its simplicity, is probably a sane choice. 

Yours,
_Christian_

uckelman:
See #15097, which incorporates some of this PR but provides fixed zoom levels.

cholmcc:
Looks good, both in the code and in execution. 

I was wondering if there was a point in binding a key - default say "1" - to reset the zoom level to the initial value? 

_Edit_: Meant to leave this on https://github.com/vassalengine/vassal/pull/15097 :smile: 

Yours,
_Christian_

uckelman:
Could add that. Ctrl-0 is typically the default for 100%. (We have a potentially odd situation here in that the default can be something other than 100%.)

uckelman:
Closing in favor of #15097.

PR #14580, opened 2026-03-28
Consolidate and streamline Linux builds

cholmcc:
This PR will consolidate and streamline build for Linux.   As such,
this PR is _mostly_ reorganisation of data into a single place so that
it is easier to maintain and build upon.

- Data common to various Linux packages, such as the Debian, RPM, and FlatPak packages, as well as the tar-ball, is put in `dist/linux` and distributed from there.

  This data includes
  - A `.desktop` file for desktop integration
  - A `.mime.xml` file for mime-types of Vassal modules, saves, logs, and extension.
  - a `man(1)` page on `vassal`
  - An improved `VASSAL.sh` script which will allow
    - Defining properties via `-Dname=value`
    - Run Vassal in the Java Debugger, including option to point to the Vassal sources.
    - Direct execution of player, editor, etc. by-passing the module manager
  - Note that these files are _also_ distributed with the tar-ball, so that a user may use those for desktop integration.
  - The script `dist/linux/integration.sh` will also be distributed with the tar-ball.  If a user executes that, it will set-up or remove desktop integration.
- The descriptions in the Debian, RPM, and FlatPak packages have been made the same - for consistency.
- The PR also removes the `vassal-maven` repository because it is no longer used.

Most of these files have been renamed to `org.vassalengine.vassal` to conform with the AppStream requirements and practices.

The main point of the updated launch script `VASSAL.sh`, are

- To allow users to pass property definitions to the JVM, for example

  ``` -Dswing.systemlaf=javax.swing.plaf.metal.MetalLookAndFeel ``` to set a different Look'n'Feel.
- To allow users to load modules, saves, logs, etc. from the current directory _without_ giving the full path to the file. For example ``` VASSAL.sh -l BattleForMoscow.vmod ``` (currently, one would need to do ``` VASSAL.sh -l `pwd`/BattleForMoscow.vmod ``` which feels awkward).
- To run Vassal in the Java debugger `jdb` to hunt down problems.

A follow-up PR to this PR will then set-up the release workflow to generate the files `org.vassalengine.vassal.yaml` and `maven-dependencies.json` for the FlatPak repository at [`flathub/org.vassalengine.org`](https://github.com/flathub/org.vassalengine.org). Thus, after that workflow has run, someone can pick up the artefacts of that workflow and commit them to `flathub/org.vassalengine.org`.

Closes #14514

Yours,
_Christian_


dist/linux/integration.sh
```diff
@@ -0,0 +1,249 @@
...
+        
+        del_file $wrap_file
+        
+        return 0
+    fi
+
+    launch=${VASSAL_DIR}/VASSAL.sh
+    
+    msg "${BY}Making wrapper script in ${UG}${wrap_file}${NC}"
+    mkdir -p `dirname ${wrap_file}`
```
uckelman: Use `$()` instead of backticks. Backticks are deprecated.

dist/linux/integration.sh
```diff
@@ -0,0 +1,249 @@
...
+
+	By default, a desktop short-cut is installed.
+	EOF
+}
+
+# --- Delete a file --------------------------------------------------
+del_file() {
+    file=$1
+    msg "${BY}Removing ${UC}$file${NC}"
+    if test ! -f $file ; then
```
uckelman: Paths need to be quoted when they are arguments. What if `$file` contains a space? (This applies throughout the script.)

dist/linux/integration.sh
cholmcc: OK.  Would still like to see a reference for the claim that backticks ``` ` ``` are deprecated.

dist/linux/integration.sh
cholmcc: OK.  Not too likely, but OK.

dist/linux/integration.sh
uckelman: https://www.gnu.org/software/bash/manual/bash.html#Command-Substitution-1

uckelman:
> To allow users to load modules, saves, logs, etc. from the current directory without giving the full path to the file. For example VASSAL.sh -l BattleForMoscow.vmod (currently, one would need to do VASSAL.sh -l `pwd`/BattleForMoscow.vmod which feels awkward).

Which part of the PR does this?

PR #14306, opened 2025-10-20
Added animated symbolic dice button

cholmcc:
This adds the possibility to add one or more animated dice to a module. The dice are animated using the defined images with configurable number of frame and frames per second.

One class `AnimatedDiceButton` has been added, as well as some translatable strings, and the documentation has been updated.

Dice like these can be added by `Add Symbolic Dice (animated)` in the module context menu.

Solve https://forum.vassalengine.org/t/animated-dice/1509/16

PR #14581, opened 2026-03-28
Added animated symbolic dice button

cholmcc:
This adds the possibility to add one or more animated dice to a module. The dice are animated using the defined images with configurable number of frame and frames per second.

One class `AnimatedDiceButton` has been added, as well as some translatable strings, and the documentation has been updated.

Dice like these can be added by `Add Symbolic Dice (animated)` in the module context menu.

Closes #14306


uckelman:
Is there a test module for this?

cholmcc:
> Is there a test module for this?

Of course there is :smile:  See [RollingDiePR.vmod](https://gitlab.com/wargames_tex/vassal_my_contrib/-/blob/master/animdice/RollingDiePR.vmod?ref_type=heads).

cholmcc:
I just pushed, to update to current `release-3.7`.

uckelman:
* The dimensions of the rolling dialog change during rolling, which makes it jumpy.
* The result appears on the button and in the chat well before the rolling animation is finished.
* It's likely simpler to use the same Animator that [Flare](https://github.com/vassalengine/vassal/blob/master/vassal-app/src/main/java/VASSAL/build/module/map/Flare.java) uses for animation. All the scheduling can be handled by the Animator.
* Image rotation and scaling should be handled via image ops.

cholmcc:
The above concerns fixed, except using the `Animator` - dealt with within the thread (similar solution).

uckelman:
It appears that you removed AnimatedDiceButton.java in the last push.

cholmcc:
> removed AnimatedDiceButton.java i

I accidentally messed up my branches - fixed here and in `cholmcc_look_and_feel`

uckelman:
You'll get better quality results if you use `Op.rotateScale` instead of doing the rotation and scaling separately.

cholmcc:
> You'll get better quality results if you use `Op.rotateScale` instead of doing the rotation and scaling separately.

* Face images that are shown as part of the animation are _only_ rotated. 
* Face images that are shown as part of the result on the button are _only_ scaled.  
* There's no case where images are _both_ scaled _and_ rotated.

uckelman:
It's not great UX for the dice to disappear from the button while a roll is happening. Could the same rolling animation appear in the button?

The animation code could be simplified a great deal by using `Animator`. I encourage you to have a look at that.

cholmcc:
> It's not great UX for the dice to disappear from the button while a roll is happening.

It's an unfortunate side effect.   I'll see if I can do something about it. 

> Could the same rolling animation appear in the button?

I don't think that will look very good.   The rotated face images may exceed the size of the button, and the button may have to change size, all of which I think will look pretty bad.  One could of course scale down the faces while animating, but that will also look a bit weird.   

I think the best outcome is to keep the old faces on the button, and then only change them when the animation is done. 

> The animation code could be simplified a great deal by using `Animator`. I encourage you to have a look at that.

The use of `org.jdesktop.timing.Animator` isn't straight forward to me, nor did I manage to find any real documentation on that.     Perhaps you will have some input, given that
- I only want to take a _specific_ number of steps 
- I want the time between frame to be fixed 
- I want to know when the animation is done. 

As far as I can tell, this should be accomplished by 

```
Animator animator = new Animator(steps, 1000 / fps, LOOP, new TimingAdapter {
  int step = 0;
  @Override
  void repeat() {
     animResultsIcon.setStep(step++);
     dialogLabel.repaint();
  }
  @Override
  void end() {
     if (reportResultInButton) {
        buttonResultsIcon.setWaiting(false);
        getLaunchButton().setIcon(buttonResultsIcon);
        getLaunchButton().repaint();
      }
      if (reportResultAsText) {
        command = command.append(reportAnimTextResults());
      }
      actionsAndSend(command);
  }
});
```
I don't think that really simplifies the code a whole lot.  Now, from the [source code I could find](https://sources.debian.org/src/timingframework/1.0-2/org/jdesktop/animation/timing/Animator.java), it seems that `org.jdesktop.timing.Animator` uses a `SwingTimer` to create the events.  If there's some benefit to that over using a `Thread`, then that may be a reason to switch, but other than that, I fail to see any significant benefit.

uckelman:
> nor did I manage to find any real documentation on that

Sadly Debian doesn't package the current, maintained version, for which there is documentation: https://javadoc.io/doc/net.java.timingframework/timingframework-core/latest/index.html

uckelman:
> it seems that org.jdesktop.timing.Animator uses a SwingTimer to create the events. If there's some benefit to that over using a Thread, then that may be a reason to switch, but other than that, I fail to see any significant benefit.

`Animator` ensures, via `SwingTimer`, that UI-modifying functions which must be called on the EDT are.
E.g., at present, `getLaunchButton().setIcon(buttonResultsIcon)` will be called on the created `Thread`, not on the EDT, as is required by Swing.

cholmcc:
> Sadly Debian doesn't package the current, maintained version, for which there is documentation: https://javadoc.io/doc/net.java.timingframework/timingframework-core/latest/index.html

Sorry, but that is not the same code at all.  The code you reference above is not what you are using in Vassal - that's pretty evident from the package name, the API, and so on.  What you use in Vassal is almost 20 years old and does not seem to be maintained anymore.

uckelman:
> Sorry, but that is not the same code at all

It is, twenty years further on. The explanation is found in the README of the [project's repo](https://github.com/akuhtz/timingframework):

> The 1.1 "classic" version of this library is fully documented in chapters 14-15 of the book Filthy Rich Clients.

The classic version referred to there is the one we're using.

cholmcc:
> as is required by Swing.

One should think that Swing would throw in that case - if it is truly _required_.  

So an alternative would be to use a `javax.swing.Timer` directly:

```
javax.swing.Timer timer = new Timer(fps,e -> {
  step++;
  animResultsIcon.setStep(step++);
  dialogLabel.repaint();

  if (step < steps) 
    return;

  if (reportResultInButton) {
        buttonResultsIcon.setWaiting(false);
        getLaunchButton().setIcon(buttonResultsIcon);
        getLaunchButton().repaint();
      }
      if (reportResultAsText) {
        command = command.append(reportAnimTextResults());
      }
      actionsAndSend(command);

      timer.stop();
  }
} );
```
That should work out nicely too, and require no third-party library.

uckelman:
> One should think that Swing would throw in that case - if it is truly required.

All but a few Swing methods are not thread-safe, which is in the Swing docs somewhere. It would be too slow to check on every method call of any UI object that it's being called from the expected thread; rather, they rely on the caller to do the right thing. Sometimes you will get weird exceptions from calling UI methods off the EDT, but it's not predictable because it depends on the scheduling of one thread against another.

cholmcc:
> The classic version referred to there is the one we're using.

Hence _not the same_ as what you pointed to.  You can't even find the code there.  

The `org.jdesktop.timinig.Animator` code is used in three places in Vassal - The Flare, Board, and Map.  In  Flare, it should be fairly easy replace.  In Board, it also seems fairly simple.  Finally, in Map, it seems to me that the JVM should already take care of auto scrolling - maybe not when Map was written, but by now that should be fairly standard.

cholmcc:
Test for [Release Notes](https://forum.vassalengine.org/t/vassal-3-8-release-notes/90445):

----

## New Features 

### New component: Animated symbolic dice 

The new component _Symbolic Dice Button (animated)_ adds a button to the toolbar of the module.  Pressing the button will, if so configured, pop up a window in which the dice roll is visualised as a short animation.

AnimatedDice

The module developer may configure the various aspects of the animation, such as 
- Number of frames per second (FPS)
- Number of faces to show before the final result
- Scale of face images on button

AnimatedDiceConfigurer

The new component is implemented on top if the existing [Symbolic Dice Button](https://vassalengine.org/doc/latest/ReferenceManual/SpecialDiceButton.html#top), and as such has all the features of that component, including playing an audio clip, execute additional hotkey commands, and so on.  Please refer to the [Reference Manual](https://vassalengine.org/doc/latest/ReferenceManual/SpecialDiceButton.html#top) for more information. 

----

PR #14593, opened 2026-03-31
Upgraded bundled Java to (Windows x86_64, MacOS), 25.0.2+10 (Windows aarch64), 21.0.10+10 (Windows x86)

uckelman:


cholmcc:
I really don't get why you didn't take #14612 - this patch _will_ fail when you update the build-time to 26, because Temurin `jlink` will refuse to link Bellsoft `jmod`s.   #14612 takes care of that _and_ makes the CI build and deployed JREs be the same - thus ensuring that everything is in sync.

uckelman:
I didn't merge #14612 for 3.7.21 because I haven't had time to review it yet. This PR works with the Java versions it's for. #14612 is still open; I'll review it for 3.7.22.

cholmcc:
> I didn't merge #14612 for 3.7.21 because I haven't had time to review it yet.

Given that you are linking 26 `jmod`s to 25 `class` objects, it could become a problem.   Perhaps it would be wise to review #14612 now and get that over with.

uckelman:
> Given that you are linking 26 jmods to 25 class objects

What makes you think that's what's happening?

cholmcc:
> What makes you think that's what's happening?

My bad - I miss-read the title  - I though you had upgraded to 26.   If you stay on 25 - fine.  And then, #14612 must be considered _soon_ to accommodate a imminent upgrade to 26.   So why not do it now?

uckelman:
> And then, https://github.com/vassalengine/vassal/pull/14612 must be considered soon to accommodate a imminent upgrade to 26. So why not do it now?

There's no rush to sort out upgrading to 26, since 25 is maintained and we'll be shipping the current 25 release at the point when 3.7.21 goes out. I'm not doing it now because other things are ahead of it in priority.

cholmcc:
> There's no rush to sort out upgrading to 26

you _did seem rather eager to update earlier.  But, I may have misunderstood you.  Any case, I volunteered some time to figure things out and came up with a solution - take it or leave it. 

Yours,
_Christian_

PR #14599, opened 2026-04-01
Fix use of deprecated logback attributes

### LINKED ISSUE #14520: logback config uses deprecated condition attribute
```
11:24:37,642 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Found logback-core version 1.5.27
11:24:37,648 |-INFO in ch.qos.logback.classic.util.ContextInitializer@4310d43 - No custom configurators were discovered as a service.
11:24:37,652 |-INFO in ch.qos.logback.classic.util.ContextInitializer@4310d43 - Trying to configure with ch.qos.logback.classic.util.DefaultJoranConfigurator
11:24:37,656 |-INFO in ch.qos.logback.classic.util.ContextInitializer@4310d43 - Constructed configurator of type class ch.qos.logback.classic.util.DefaultJoranConfigurator
11:24:37,659 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback-test.xml]
11:24:37,671 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Found resource [logback.xml] at [jar:file:/home/uckelman/projects/VASSAL/trunk/release-prepare/target/lib/Vengine.jar!/logback.xml]
11:24:38,205 |-INFO in ch.qos.logback.classic.model.processor.ConfigurationModelHandlerFull - Scan attribute not set or set to unrecognized value.
11:24:38,224 |-INFO in ch.qos.logback.core.model.processor.ModelInterpretationContext@54a7079e - value "errorLog-3.7.21-SNAPSHOT-f0d2a9e23" substituted for "errorLog-${git.version}"
11:24:38,225 |-WARN in ch.qos.logback.core.model.processor.conditional.IfModelHandler - The 'condition' attribute in <if> element is deprecated and slated for removal. Use <condition> element instead.
11:24:38,225 |-WARN in ch.qos.logback.core.model.processor.conditional.IfModelHandler - See also https://logback.qos.ch/codes.html#conditionAttributeDeprecation
11:24:38,826 |-INFO in ch.qos.logback.core.model.processor.conditional.IfModelHandler - Condition [isDefined("VASSAL.conf")] evaluated to false on line 5
11:24:38,826 |-WARN in ch.qos.logback.core.model.processor.conditional.IfModelHandler - The 'condition' attribute in <if> element is deprecated and slated for removal. Use <condition> element instead.
11:24:38,826 |-WARN in ch.qos.logback.core.model.processor.conditional.IfModelHandler - See also https://logback.qos.ch/codes.html#conditionAttributeDeprecation
11:24:38,831 |-INFO in ch.qos.logback.core.model.processor.conditional.IfModelHandler - Condition [property("os.name").toLowerCase().startsWith("windows")] evaluated to false on line 10
11:24:38,831 |-WARN in ch.qos.logback.core.model.processor.conditional.IfModelHandler - The 'condition' attribute in <if> element is deprecated and slated for removal. Use <condition> element instead.
11:24:38,831 |-WARN in ch.qos.logback.core.model.processor.conditional.IfModelHandler - See also https://logback.qos.ch/codes.html#conditionAttributeDeprecation
11:24:38,833 |-INFO in ch.qos.logback.core.model.processor.conditional.IfModelHandler - Condition [property("os.name").toLowerCase().startsWith("mac")] evaluated to false on line 15
11:24:38,833 |-INFO in ch.qos.logback.core.model.processor.ModelInterpretationContext@54a7079e - value "/home/uckelman/.VASSAL" substituted for "${user.home}/.VASSAL"
11:24:38,833 |-INFO in ch.qos.logback.core.model.processor.ConversionRuleModelHandler - registering conversion word pid with class [VASSAL.tools.logging.ProcessIDConverter]
11:24:38,839 |-INFO in ch.qos.logback.core.model.processor.ModelInterpretationContext@54a7079e - value "/home/uckelman/.VASSAL/errorLog-3.7.21-SNAPSHOT-f0d2a9e23" substituted for "${CONF_DIR}/${LOG_FILE}"
11:24:38,841 |-INFO in ch.qos.logback.core.model.processor.AppenderModelHandler - Processing appender named [FILE]
11:24:38,841 |-INFO in ch.qos.logback.core.model.processor.AppenderModelHandler - About to instantiate appender of type [ch.qos.logback.core.FileAppender]
11:24:38,847 |-INFO in ch.qos.logback.core.model.processor.ModelInterpretationContext@54a7079e - value "/home/uckelman/.VASSAL/errorLog-3.7.21-SNAPSHOT-f0d2a9e23" substituted for "${CONF_DIR}/${LOG_FILE}"
11:24:38,917 |-INFO in ch.qos.logback.core.FileAppender[FILE] - File property is set to [/home/uckelman/.VASSAL/errorLog-3.7.21-SNAPSHOT-f0d2a9e23]
11:24:38,918 |-INFO in ch.qos.logback.core.model.processor.AppenderModelHandler - Processing appender named [CHATTER]
11:24:38,918 |-INFO in ch.qos.logback.core.model.processor.AppenderModelHandler - About to instantiate appender of type [VASSAL.build.module.ConsoleAppender]
11:24:38,920 |-INFO in ch.qos.logback.classic.model.processor.RootLoggerModelHandler - Setting level of ROOT logger to INFO
11:24:38,920 |-INFO in ch.qos.logback.core.model.processor.AppenderRefModelHandler - Attaching appender named [FILE] to Logger[ROOT]
11:24:38,920 |-INFO in ch.qos.logback.core.model.processor.AppenderRefModelHandler - Attaching appender named [CHATTER] to Logger[ROOT]
11:24:38,920 |-INFO in ch.qos.logback.core.model.processor.DefaultProcessor@26e356f0 - End of configuration.
11:24:38,921 |-INFO in ch.qos.logback.classic.joran.JoranConfigurator@47d9a273 - Registering current configuration as safe fallback point
11:24:38,925 |-INFO in ch.qos.logback.classic.util.ContextInitializer@4310d43 - ch.qos.logback.classic.util.DefaultJoranConfigurator.configure() call lasted 1265 milliseconds. ExecutionStatus=DO_NOT_INVOKE_NEXT_IF_ANY
```

uckelman:
Fixes #14520.


cholmcc:
> I've set `VASSAL.conf` in all of the launchers, here: #14602.

OK, our developments have crossed :smile: 

I made another PR: #14603 - this fixes it in the code rather than in the start-up scripts.  I'll look at #14602

cholmcc:
I think it's a bad idea to put the deduction of the configuration directory into the `logback.xml` file - especially because the deduction is already done in the Java code - and it can be set on the command line.  Having the same deduction done _twice_ can lead to inconsistencies.   

I would go with either #14062 (with `logback-1.2.xml` for Debian) or #14603. 

Yours,
_Christian_

debian/rules
```diff
@@ -23,7 +23,7 @@ APP_JSRCDIR	:= $(APP_SRCDIR)/java
 APP_SRC	    	:= $(shell find $(APP_JSRCDIR)/bsh          -name "*.java") \
 	           $(shell find $(APP_JSRCDIR)/org/netbeans -name "*.java") \
 		   $(shell find $(APP_JSRCDIR)/org/litesoft -name "*.java") \
-		   $(shell find $(APP_JSRCDIR)/VASSAL       -name "*.java")
+		   $(shell find $(APP_JSRCDIR)/VASSAL       -name "*.java" | grep -v vassal-app/src/main/java/VASSAL/tools/logging/PropertyNoCaseStartsWithCondition.java)
```
cholmcc, line 26: I would _strongly_ advice against doing this - there's no good reason why the configuration directory should be deduced in the `logback` XML when it is already either set on the command line or is deduced by `StandardConfig`.  Those two options are clearly superior to this. 

Filtering out specific pieces of code will quickly become a maintenance nightmare.  Another reason why the approach of the two other PRs are better than this one.  

Using either of the two other PRs will also mean that you have one less piece of code to maintain - which is always good :smile:

uckelman: > I would strongly advice against doing this

This PR isn't _doing_. This PR is maintaining the status quo we've had with how logback determines the configuration directory for at least a decade. It is well-tested. I do not want to change the setup on our side at this late date.

uckelman: Having seen the options on offer, this is the one which makes the fewest changes to the existing flow, and it pushes the complexity of the solution into the build which is causing the complexity. I don't expect there to be enough future changes for this to become a maintenance problem.

cholmcc: > This PR isn't _doing_. 

I'm not sure I understand your comment. 

> Having seen the options on offer, this is the one which makes the fewest changes to the existing flow, and it pushes the complexity of the solution into the build which is causing the complexity. 

I don't agree with that assessment at all - you keep the duplicate complexity - once in `StandardConfig` and once in `logback.xml` _and_ you've added extra complexity via `PropertyNoCaseStartsWithCondition` _as well as_ its exclusion from the Debian build.    Out of the three options, I think this is the _worst_ possible fix. 

My PR that pushed the initialisation of the logger to _after_ the `StandardConfig` initialisation was pretty simple and didn't cause any unforeseen issues.  In fact, the logger should be instantised until it's configuration is known.  

Your PR that pushed the setting of `VASSAL.conf` didn't change the flow at all, but made _sure_ that preconditions was know.  In that sense, it was a simplification. 

> This PR is maintaining the status quo we've had with how logback determines the configuration directory for at least a decade. 

That hardly seems like an argument, given that it is problematic to have the same deduction done twice.   And, it is by far _not_  maintaining the _status quo_, because it adds complexity. 

I find it hard to follow your logic sometimes: It seems like you in some cases find it more important to "maintaining the status quo" and in other cases are more willing to fix long-standing issues (whether reported or not).  

>  I don't expect there to be enough future changes for this to become a maintenance problem.

Famous last words. 

I would recommend to roll-back the merge and pick one of the other PRs instead. 

Yours,
_Christian_

uckelman: > I don't agree with that assessment at all - you keep the duplicate complexity - once in StandardConfig and once in logback.xml and you've added extra complexity via PropertyNoCaseStartsWithCondition as well as its exclusion from the Debian build. Out of the three options, I think this is the worst possible fix.

"keep" is the important word here. What is kept is _not changed_. My goal here was to change as little as possible with the way `VASSAL.conf` is set, because we know that the way is has been works.

> I find it hard to follow your logic sometimes: It seems like you in some cases find it more important to "maintaining the status quo" and in other cases are more willing to fix long-standing issues (whether reported or not).

My goal with V3 maintenance has long been to minimize the effort required to keep things running. When we fix long-standing issues, I'm balancing risk of causing new problems and the debugging and support time we would spend addressing new problems with the support time we already spend addressing existing issues.

> >  I don't expect there to be enough future changes for this to become a maintenance problem.
> Famous last words.

We last made changes because of a logback upgrade in 2023, but we've used dozens of successive logback releases in that time. There's a good chance the issue will not arise again during the lifetime of V3.

If we find that special-casing the Debian build causes too much maintenance burden, we have several options---revisiting one of the closed PRs, convincing Debian to upgrade logback, or ceasing to do Debian builds.

cholmcc: > with the way VASSAL.conf is set, because we know that the way is has been works. 

_Even_ if there's a clear duplication of the deduction which - all things equal - _doubles_ the chance that it will fail.   I fail to see how keeping this as they were is any argument in this case. 

>> I find it hard to follow your logic sometimes: 
> My goal with V3 maintenance has long been to minimize the effort required to keep things running. 

Right.  which is why it would make sense to fix potential pitfalls to prevent it from ever being a problem again.  I think this PR is creating potentially more problems than it is actually solving.  And given there were _less_ intrusive options, I don't see the logic. 

>  There's a good chance the issue will not arise again during the lifetime of V3.

Which would be a reason to leave things as they were _before_ and then live with the deprecation warnings.  If V3 is supposed to be short-lived, then why bother introducing complexity to prevent a set of _warnings_? 

> because of a logback upgrade in 2023, but we've used dozens of successive logback releases in that time.

`logback` seems to have been superseded by [`org.apache.commons.logging`](https://commons.apache.org/proper/commons-logging/) and [`org.apache.logging.log4j` (version 2)](https://logging.apache.org/log4j/2.x/manual/api.html#loggers).  Also, it really does seem to me that `logback` is a legacy library - despite it being maintained.   In a sense, the whole point of using [`org.slf4j`](https://www.slf4j.org/) is so that one can replace the logger "backend".

> If we find that special-casing the Debian build ... or ceasing to do Debian builds.

The Debian packages have been relatively popular with the users, so there's that to consider too.  

Yours,
_Christian_

uckelman: > Which would be a reason to leave things as they were before and then live with the deprecation warnings. 

No. The warnings print _to the console_, and this has resulted in users asking about them because they think the warnings indicate a problem. Leaving them increases our support burden.

> logback seems to have been superseded by [org.apache.commons.logging](https://commons.apache.org/proper/commons-logging/) Also, it really does seem to me that logback is a legacy library - despite it being maintained.

On what evidence?

> In a sense, the whole point of using [org.slf4j](https://www.slf4j.org/) is so that one can replace the logger "backend".

slf4j doesn't provide a logging backend. It's a unified interface for whatever logging backend you choose, not an alternative to logback. We still must provide a logging backend.

> The Debian packages have been relatively popular with the users, so there's that to consider too.

The Debian packages are less than 1% of the total downloads of 3.7.20. They're less popular than the SHA26 hash file.

cholmcc: > The Debian packages are less than 1% of the total downloads

Out of the Linux downloads it is >10% (167 out of 1418).  It is kinda ridiculous to compare Debian to all - especially when the vast majority of downloads are for Windows.    Out of the total, yes < 1% (167 out of 19213).  But then again `other` is only 688, and MacOS only 3116, so maybe we can do away with those too?  After all, MacOS is maintenance burden from hell. 

For 3.7.21 Debian package sits at 20% :smile:  Give people some time to get used to it.  In fact, I think it is kinda remarkable that the Debian packages reached more than 10% of the Linux downloads on the first release it became available.  

> slf4j doesn't provide a logging backend. 

I know - read the comment.   

> On what evidence? 

Single developer, decline is sponsors, ... 

Yours,
_Christian_

PR #14602, opened 2026-04-02
Set VASSAL.conf from the launch scripts so that logback.xml has access to it without needing conditionals

uckelman:
This works around current logback having deprecated conditional attributes on if elements and Debian shipping an EOL version of logback that doesn't support conditional elements.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,7 +222,7 @@ $(TMPDIR)/windows-%-$(VERSION)-build:
 
 JREOPTS:=
 
-$(TMPDIR)/windows-noinst-$(VERSION)-build/VASSAL.l4j.xml: JREOPTS:=<opt>-DVASSAL.conf="%EXEDIR%\\..\\VASSAL"</opt>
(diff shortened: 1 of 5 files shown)


cholmcc:
I think this should work.  One concern, which may not be a thing at all, is if a user should be able to override the  `VASSAL.conf` setting.   I doubt it will be an issue though.  Not even [`VASSALLauncer`](https://www.gamesquad.com/forums/index.php?resources/vassal-launcher.160/) uses that - even though it would simplify things for them :slightly_smiling_face: 

Which `-D` property definition wins on the command line depends on the order.  I believe latter definitions will win, so that if a user does 
```
VASSAL.sh -DVASSAL.conf=/some/directory
```
then `VASSAL.conf` will point to `/some/directory`. 

I think this PR is the simpler of this and #14603 and should probably be used. 

Yours,
_Christian_

uckelman:
Using #14599 instead.

PR #14603, opened 2026-04-02
Patch to fix deprecated logback XML elements

cholmcc:
Patch to fix deprecated logback XML elements

This patch fixes the `logback.xml` configuration so as to not use deprecated elements.

This is done by slightly alterning the initialisation ordering of things. The code is changed so that `VASSAL.launch.StandardConfig` is  instantised before calls to `LoggerFactory.getLogger`, so that when the logger starts up, the system property `VASSAL.conf` is already set, and so we do not need to deduce it in `logback.xml`.

Ideally, `VASSAL.launch.StandardConfig` would propagate its data member `errorLogPath` as a system property - say `VASSAL.log`. However, to set that member, we need to read the Git version from `git.properties` first, which is done by `VASSAL.tools.version.GitProperties`.   That class instantised the log with `LoggerFactory.getLogger` to handle problems reading the `git.properties` file.  That means that we cannot set - say `VASSAL.log` - _before_ the first call to `LoggerFactory.getLogger`, and we have a race-condition problem. 

One _could_ argue, that if `VASSAL.tools.version.GitProperties` fails to read the `git.properties` file and therefore cannot set system property `git.version`, then any error reported will be put into `<confdir>/errorLog-git.version_IS_UNDEFINED` (where `<confdir>` is the configuration directory - say `~/.VASSAL`, and that therefore `VASSAL.tools.version.GitProperties` should _not_ use the `logback` logging facility, but rather throw a regular exception to be handled by the launched application. 

If that was the case, then we wouldn't have the chicken-egg problem outlined above.  In this patch, however, a more conservative approach is taken. 

This patch also adds the attribute `converterClass` to the `conversionRule` tag in `logback.xml` for handling the template `%pid`.  This is because older `logback` expects this attribute rather than the attribute `class`.   It should be OK to specify both `converterClass` and `class` because a: the XML parser will not fail on unknown attributes, and b: the `logback` code will only use the attribute that it expects. 

Alternative to #14599

### DIFF
--- a/vassal-app/src/main/java/VASSAL/launch/StandardConfig.java
+++ b/vassal-app/src/main/java/VASSAL/launch/StandardConfig.java
@@ -39,11 +39,6 @@ public class StandardConfig implements Config {
   private final String reportableVersion;
 
   public StandardConfig() throws IOException {
-    // Set the version, reportable version
-    final GitProperties gitProperties = new GitProperties();
-    version = gitProperties.getVersion();
-    reportableVersion = version.contains("-") && !version.matches(".*-beta\\d+") ? version.substring(0, version.indexOf('-')) : version;
-
     baseDir = Path.of(System.getProperty("user.dir"));
 
     docDir = baseDir.resolve(
@@ -66,6 +61,7 @@ public class StandardConfig implements Config {
     else {
       confDir = Path.of(System.getProperty("user.home"), ".VASSAL"); //NON-NLS
     }
+    System.setProperty("VASSAL.conf", confDir.toString());
 
     final String cacheProp = System.getProperty("VASSAL.cache");
     if (cacheProp != null) {
@@ -79,6 +75,11 @@ public class StandardConfig implements Config {
     }
 
     Files.createDirectories(confDir);
+
+    // Set the version, reportable version
+    final GitProperties gitProperties = new GitProperties();
+    version = gitProperties.getVersion();
+    reportableVersion = version.contains("-") && !version.matches(".*-beta\\d+") ? version.substring(0, version.indexOf('-')) : version;
 
     prefsDir = confDir.resolve("prefs");
     errorLogPath = confDir.resolve("errorLog-" + getVersion());
--- a/debian/rules
+++ b/debian/rules
@@ -32,7 +32,6 @@ APP_I18N	:= $(shell find $(APP_SRCDIR)/resources/VASSAL/i18n/ \
 APP_BSH		:= $(shell find $(APP_SRCDIR)/resources/VASSAL/script/\
 			-name "*.bsh")
 APP_MISC	:= $(APP_SRCDIR)/resources/grayscale.icc	\
-		   $(APP_SRCDIR)/resources/logback.xml		\
 		   $(APP_SRCDIR)/resources/removed		\
 		   $(APP_SRCDIR)/resources/vassal.vmoptions
 APP_SPLASH	:= $(APP_JSRCDIR)/org/netbeans/modules/wizard/defaultWizard.png
@@ -132,6 +131,8 @@ deprecated: .compile
 	cp -a $(APP_ICON)                        $(OUT_DIR)/
 	cp -a $(APP_HELP)                        $(OUT_DIR)/
 	cp -a $(APP_MISC)			 $(OUT_DIR)/
+	cp -a $(APP_SRCDIR)/resources/logback-1.2.xml		\
+		   				 $(OUT_DIR)/logback.xml
 	cp -a git.properties                     $(OUT_DIR)/
 	cp -a deprecated			 $(OUT_DIR)/
 	cp -a $(APP_SPLASH)                      $(OUT_DIR)$(subst $(APP_JSRCDIR),,$(APP_SPLASH))
--- a/vassal-app/src/main/java/VASSAL/launch/Editor.java
+++ b/vassal-app/src/main/java/VASSAL/launch/Editor.java
@@ -50,10 +50,11 @@ import VASSAL.tools.version.VersionUtils;
 public class Editor extends Launcher {
   public static void main(String[] args) throws IOException {
     Info.setConfig(new StandardConfig());
+    logger = LoggerFactory.getLogger(Editor.class);
     new Editor(args);
   }
 
-  private static final Logger logger = LoggerFactory.getLogger(Editor.class);
+  private static Logger logger = null;
 
   protected Editor(String[] args) {
     // the ctor is protected to enforce that it's called via main()
--- a/vassal-app/src/main/java/VASSAL/launch/Launcher.java
+++ b/vassal-app/src/main/java/VASSAL/launch/Launcher.java
@@ -49,7 +49,7 @@ import java.util.concurrent.ExecutionException;
  * @since 3.1.0
  */
 public abstract class Launcher {
-  private static final Logger logger = LoggerFactory.getLogger(Launcher.class);
+  private static Logger logger = null;
 
   protected final LaunchRequest lr;
 
@@ -62,6 +62,8 @@ public abstract class Launcher {
   protected Launcher(String[] args) {
     if (instance != null) throw new IllegalStateException();
     instance = this;
+
+    if (logger == null) logger = LoggerFactory.getLogger(Launcher.class);
 
     LaunchRequest lreq = null;
     try {
--- a/vassal-app/src/main/java/VASSAL/launch/ModuleManager.java
+++ b/vassal-app/src/main/java/VASSAL/launch/ModuleManager.java
@@ -63,8 +63,7 @@ import VASSAL.tools.menu.MacOSXMenuManager;
  * @since 3.1.0
  */
 public class ModuleManager {
-  private static final Logger logger =
-    LoggerFactory.getLogger(ModuleManager.class);
+  private static Logger logger = null;
 
   private static final String NEXT_VERSION_CHECK = "nextVersionCheck"; //NON-NLS
   private static final String AUTO_VERSION_CHECK = "autoVersionCheck"; //NON-NLS
@@ -109,7 +108,9 @@ public class ModuleManager {
       e.printStackTrace();
       System.exit(1);
     }
-
+    // Initialize logger after standard configuration
+    logger = LoggerFactory.getLogger(ModuleManager.class);
+    
 // FIXME: We need to catch more exceptions in main() and then exit in
 // order to avoid situations where the main thread ends due to an uncaught
 // exception, but there are other threads still running, and so VASSAL
--- a/vassal-app/src/main/resources/logback-1.2.xml
+++ b/vassal-app/src/main/resources/logback-1.2.xml
@@ -4,7 +4,7 @@
   <property name="CONF_DIR" value="${VASSAL.conf}" />
 
   <conversionRule conversionWord="pid"
-                  class="VASSAL.tools.logging.ProcessIDConverter" />
+                  converterClass="VASSAL.tools.logging.ProcessIDConverter" />
 
   <appender name="FILE" class="ch.qos.logback.core.FileAppender">
     <file>${CONF_DIR}/${LOG_FILE}</file>
--- a/vassal-app/src/test/resources/logback.xml
+++ b/vassal-app/src/test/resources/logback.xml
@@ -1,44 +1,16 @@
 <configuration>
   <conversionRule conversionWord="pid"
-                  converterClass="VASSAL.tools.logging.ProcessIDConverter" />
+                  class="VASSAL.tools.logging.ProcessIDConverter" />
 
-  <if condition='property("os.name").toLowerCase().startsWith("windows")'>
-    <then>
-      <appender name="FILE" class="ch.qos.logback.core.FileAppender">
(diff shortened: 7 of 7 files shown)


vassal-app/src/main/java/VASSAL/launch/StandardConfig.java
```diff
@@ -80,8 +76,14 @@ else if (SystemUtils.IS_OS_WINDOWS) {
...
     Files.createDirectories(confDir);
 
+    // Set the version, reportable version
+    final GitProperties gitProperties = new GitProperties();
+    version = gitProperties.getVersion();
+    reportableVersion = version.contains("-") && !version.matches(".*-beta\\d+") ? version.substring(0, version.indexOf('-')) : version;
+
     prefsDir = confDir.resolve("prefs");
     errorLogPath = confDir.resolve("errorLog-" + getVersion());
+    System.setProperty("VASSAL.log", errorLogPath.toString());
```
uckelman: What uses this?

uckelman (changes requested):
`vassal-app/src/test/resources/logback.xml` should also be updated.

vassal-app/src/main/java/VASSAL/launch/StandardConfig.java
cholmcc: Nothing - I thought it could be used in the `logback.xml` file - but it wont work correctly I think.

uckelman: Would you remove it, as it's unused?

cholmcc: Sure.  I'm about to force-push an update to this PR.   I will update the description of the PR too - please check the considerations regarding `GitAttributes` initialisation in that.  Will be there in a few.

cholmcc: New push fixes this

cholmcc:
> `vassal-app/src/test/resources/logback.xml` should also be updated.

How do you see that file be updated?  The tests does not instantise `VASSAL.launch.StandardConfig`, and thus we cannot ensure that `VASSAL.conf` is set before hand, let alone `git.version`. 

What I would suggest, is to write the test log to `./errorLog-test` or similar - i.e., write in current directory of where the tests are performed.   End-users won't run these tests in any case, so there's no need to store the log in some user directory.   The tests are really only run by developers. 

What do you think? 

Yours,
_Christian_

uckelman:
I had thought `vassal-app/src/test/resources/logback.xml`  could be updated as in #14602.

cholmcc:
> I had thought `vassal-app/src/test/resources/logback.xml` could be updated as in #14602.

OK - write in current directory as in `vassal-test.log` - great - makes sense.  Done.

uckelman:
I've tried this PR now and found it doesn't solve the problem. Because the `class` attribute is specified, we still get the deprecation warning for it:
```
19:38:18,185 |-WARN in ch.qos.logback.core.joran.action.ConversionRuleAction - Both [converterClass] attribute and [class] attribute specified. 
```

cholmcc:
> I've tried this PR now and found it doesn't solve the problem. Because the `class` attribute is specified, we still get the deprecation warning for it:

OK.   the `conversionClass` tag works for `logback` 1.2, and it also seems to work with 1.3 - both without warnings, so I will remove the `class` attribute.   Note that the `class` attribute does not work with `logback` 1.2.    This will give the spurious warning 
```
20:51:06,296 |-WARN in ch.qos.logback.core.joran.action.ConversionRuleAction - [converterClass] attribute is deprecated and replaced by [class]. See element [conversionRule] near line 7
...
20:51:06,336 |-INFO in ch.qos.logback.core.model.processor.ConversionRuleModelHandler - registering conversion word pid with class [VASSAL.tools.logging.ProcessIDConverter]
```

The `logback` code should really ignore superfluous or unsupported tags in the XML configuration, and certainly not raise "deprecation" warnings against it.   In fact, the code should handle both tags, and use the most meaningful tag.  If only an older tag is given, it should fall back on that for compatibility reasons.  Just like VASSAL does it :smile: 

Yours,
_Christian_

uckelman:
Ah, sorry, it's the other way around: The `converterClass` attribute is the deprecated one. That's the one which has to be removed.

cholmcc:
> Ah, sorry, it's the other way around: The `converterClass` attribute is the deprecated one. That's the one which has to be removed.

If we remove `converterClass`, then the conversion rules won't work with `logback` version 1.2.  But if we leave `converterClass` it will still work with `logback` 1.3 - albeit with a warning.  I've made a bug report against `logback` :smile:

I would prefer to have the code work on _both_ 1.2 and 1.3, and then live with the superfluous warning in 1.3, rather than the output being wrong with 1.2.   What do you think? 

Yours,
_Christian_

uckelman:
What escape hatch does Debian provide for avoiding the outdated packages they ship?

cholmcc:
> What escape hatch does Debian provide for avoiding the outdated packages they ship?

Uh!? I don't understand your question. 

Are you asking how come Debian (and by extension a bunch of other Linux distributions - including Ubuntu) does not have a more recent version of - say - `logback`?  

As I eluded to earlier, the reasons can be legion.    

Specifically for `logback`, I think it is because that library isn't in much use any more and there isn't a high demand for updates.    You can look at the [`liblogback-java`](https://tracker.debian.org/pkg/logback) information and see that no-one did a manual request. 

Another reason may be - as you can see from  [`liblogback-java`](https://tracker.debian.org/pkg/logback)  - is that `logback` seems to have security holes in it.   Debian being used in many critical applications is pretty mindful of security issues. 

But, I'm speculating.   If you want to know the real reason, I think an email to [`pkg-java-maintainers@lists.alioth.debian.org`](https://qa.debian.org/developer.php?login=pkg-java-maintainers%40lists.alioth.debian.org) could clear it up.     BTW, those are the same guys that would probably need to maintain Vassal in Debian - if Vassal is to be distributed with Debian (and derivatives). 

If I misunderstood your question, please elaborate.   Thanks. 

Yours,
_Christian_

uckelman:
What I am asking is: How do we use the current version of logback on Debian?

uckelman:
> Another reason may be - as you can see from [liblogback-java](https://tracker.debian.org/pkg/logback) - is that logback seems to have security holes in it. Debian being used in many critical applications is pretty mindful of security issues.

The security issues listed there are ones which are fixed in the current release.

cholmcc:
> The security issues listed there are ones which are fixed in the current release.

Sure.  However, Debian has not audited it yet, so that could be reason.   The point of pointing  out - no pun intended - 
the security issues, was mostly to say that `logback` is probably not the safest piece of code out there.  And for the past decade it seems like it's a one man operation.   

> What I am asking is: How do we use the current version of logback on Debian?

Uh! That depends a whole lot on how much effort you want to put in to it and what the end goal is. 

Starting with the goal:
- Is the goal to make a distribution-grade package that can distributed with Debian (and all of its derivatives)? 
- Or is the (lesser) goal to provide an easier way for users to install Vassal? 

Let us take the first goal.    As mentioned at another time, to have a package go into Debian proper, it _has_ to build cleanly against 3rd-party code _already_ in Debian.    That is, a package 
- Cannot require software, or software version, _not_ already in Debian 
- It _must_ build without downloads, etc. 
- Embedded 3rd party libraries are generally _not_ allowed.   Debian will typically rather back-port code - if need be - then to embed 3rd party code into a package. 
So when it comes to `logback`, what that means is, to build a Debian distribution-grade package against a newer `logback`, then that would require that `logback` be upgraded in Debian proper. 

Incidentally, the reason why I didn't use the Maven build system of Vassal for the Debian package is precisely because of the above.  The Vassal Maven build pulls in a lot of 3rd party code that isn't in Debian, and so the build would fail when using Debian's Maven wrappers. 

For the second goal, there are a number of options.   One could package a newer `liblogback-java` one self, and then build against that.   One could even set up a repository for that package - and other packages, such as `vassal` it self, and pull it from there.    One could also embed the code directly, though that's a maintenance headache.  Or one could make an effort to get the Maven build to work reasonably for Debian.   Finally, and least attractive, one could also take the avenue that I did for the RPM: Pull down the binary build and simply put that into a package. 

I think the goal should be to get Vassal into Debian proper.  In that way, Vassal would be instantaneously available to a much larger audience.   I'm aware that you are very much looking toward Vassal 4, and that this may not have a direct impact on that development.  However, in the long run, I think having Vassal 3 in Debian will make the likeliness of Vassal 4 also be available there, much higher. 

Finally, I think we are using a lot of time for relatively little gain.   `logback` 1.2 works perfectly fine for the purposes of Vassal.   And, as it is unlikely that Vassal 3 will undergo major developments that would necessitate a later `logback`, I don't see why Vassal can't live with 1.2.    If you want to prioritise the regular builds and have `class` but not `converterClass` in the configuration files, then so be it - not a major problem.   If you think we should put both tags in and live with the deprecation warning, then that would be great too.  You decide.  And with that said, I think going for Debian proper is still a good goal. 

Yours,
_Christian_

uckelman:
I agree that this has consumed far to much time for the gain.

It occurred to me this morning that we could simply exclude the changes for the current version of logback from the Debian package. I've done that now in #14599. Please take a look.

cholmcc:
> It occurred to me this morning that we could simply exclude the changes for the current version of logback from the Debian package. 

That is indeed a solution. However, ...

> I've done that now in #14599. Please take a look.

... there's no need to retain the `<if ...>` switches in that XML if `VASSAL.conf` is guaranteed to be set - either on the `java` command line, or in the code. 

I will update this PR to push a specific `logback.xml` for Debian.  I think it would make sense to keep it out of the `debian` sub-directory and rather have it as `vassal-app/src/main/resources/logback-1.2.xml` - next to the default `logback.xml`. 

Yours,
_Christian_

uckelman:
> I think it would make sense to keep it out of the debian sub-directory and rather have it as vassal-app/src/main/resources/logback-1.2.xml - next to the default logback.xml.

Why?

cholmcc:
> > I think it would make sense to keep it out of the debian sub-directory and rather have it as vassal-app/src/main/resources/logback-1.2.xml - next to the default logback.xml.
> 
> Why?

Several reasons - Debian isn't too happy with files dumped into `debian`.  Secondly, the `logback-2.1.xml` isn't _specific_ to Debian but could be used elsewhere too (it may not, but it could).  Thirdly, keeping the two next to each other makes is more likely that they will be maintained concurrently. And finally, why not? 

Yours,
_Christian_

uckelman:
> Debian isn't too happy with files dumped into debian.

Ok, I'll find another place for it.

If we put it with the other logback.xml, it will be built into Vengine.jar unnecessarily.

uckelman:
Using #14599 instead.

cholmcc:
Something I just thought of, which may make you reconsider to use this approach rather than https://github.com/vassalengine/vassal/pull/14599: 

- With #14599 the `errorLog` will _always_ end up in one of 
  ```
  ~/.VASSAL/errorLog-<version>    # Linux
  ~/Library/Application Support/VASSAL/errorLog-<version>  # MacOS
  %APP_DATA%\VASSAL\errorLog-<version>  # Windows
  ```
- With _this_ PR, the `errorLog` will end up in 
  ```
   ${VASSAL.conf}/errorLog-<version>
  ```
  where `VASSAL.conf` is the system property and `${VASSAL.conf}` is what ever that property is set to. 

So why does that matter?   Well, a user could define the `VASSAL.conf` property to point to something entirely different from `~/VASSAL`, `~/Library/Application Support/VASSAL`, or `%APP_DATA%\VASSAL` - say `~/VASL`.  With #14599, the `errorLog` will _not_ be put in `~/VASL` but in `~/.VASSAL`, which I think is wholly unexpected.  

The use case is something like [VASSALLauncher](https://www.gamesquad.com/forums/index.php?resources/vassal-launcher.160/).  They use different configuration directories to accommodate different set-ups (different user, extensions, etc.).  They _could_ (but they don't) accomplish that simply by defining `VASSAL.conf` to point to some other directory.  

For example, you may be playing ASL with one group of friends and with a specific set of extensions and a specific password.   At the same time you are playing ASL with a different group of friends with a different set of extensions and another password.   That could easily be accommodated by defining `VASSAL.conf` differently in the two cases 
```
$ VASSAL.sh -DVASSAL.conf=~/.vasl-group-a   # First group of friends 
$ VASSAL.sh -DVASSAL.conf=~/.vasl-group-b   # Second group of friends
```
When playing with group a, the `errorLog` ends up in `~/.vasl-group-a/errorLog-<version>`, while playing with group b, the error log is `~/.vasl-group-b/errorLog-<version>`.   With #14599 there's no such clear division, and messages can get muddled up.   That also means that Vassal will try to use an invalid `errorLog` location for example when doing ABRs, report in the chat, and so on.   That is why it is _much_ better to define the `errorLog` location in _one_ place. 

On Linux, it is easy to set `VASSAL.conf` - simply pass `-DVASSAL.conf=<path>` on the command line.  For MacOS, a similar thing could be accomplished by changes to `dist/macos/VASSAL.sh`.  For Windows, one would need a `VASSAL.l4j.ini` file with content like `-DVASSAL.conf=<path>` 

Yours,
_Christian_

PR #14612, opened 2026-04-08
Upgraded bundled Java to 26+35 (Windows x86_64, MacOS), 26+37 (Windows aarch64)

cholmcc:
This patch will
- Modify the `boostrap.sh` script to make it more flexible.  This is used in the workflows

  Note, when `bootstrap.sh` is invoked from the GitHub workflows, we pass in the select Java API version.  As the version we get also contains the build, the script will automatically extract that build too.  E.g., if the GitHub action `setup-java` figures out that the latest Temurin JDK with API=26 is `26.0.0+35`, then `bootstrap` will use API=26 and BUILD=35.  In that way, we can ensure that the JDK we are using to build the packages with is also the same we are shipping with the final installers.

- Split the Windows ARM package off from the main build of Windows 64, MacOS, Linux, and other packages, as it needs a different JDK. This is because in OpenJDK, `jlink` from one build of OpenJDK cannot link `jmods` from another build of OpenJDK.  Thus, we need to use the JDK - and therefore `jlink` - from the same build that we want to package into the installation package.

- Both the package and release workflows have been updated to use the new flexibility of the `bootstrap.sh` script.  Also, the JDK API version used (except for Windows 32 which is fixed at 21) is set at the top of the workflow file and then propagated to the jobs.  This makes it easier to change the API to use.
- In the `Makefile`, the `release` and `release-windows` targets do not build `release-windows-aarch64` target.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master, release-3.7 ]
 
+env:
+  JDK_API: 26
+  
 jobs:
   package:
     name: Test & Package
@@ -17,7 +20,7 @@ jobs:
           fetch-depth: 1
 
       - name: Set branch
-        if: github.ref != 'refs/heads/master' && github.ref != 'refs/heads/release-3.7' 
+        if: github.ref != 'refs/heads/master' && github.ref != 'refs/heads/release-3.7'
         # Checking out a detached head for PR means we have to set the branch
         # name for it to show up later; shallow checkout means the branch
         # won't exist yet so there's no conflict.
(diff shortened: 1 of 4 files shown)


uckelman:
@cholmcc

https://github.com/vassalengine/vassal/actions/runs/25102311082/job/73554487603

The build is now failing because
```
./bootstrap.sh -j ${{ steps.setup-java.outputs.version }} windows-aarch64
```
has a version number that bellsoft doesn't have available for download.

cholmcc:
The problem is fixed in https://github.com/vassalengine/vassal/pull/14687

PR #14656, opened 2026-04-18
Editor search matches Marker names as Property names

### LINKED ISSUE #13052: Editor Search - consider matching Marker names as Property Names
In this example search... 

image

A marker name such as below is missed. It is found if Trait Descriptions is checked:
image

Conversely, a Global Property with the same name is matched not on Trait Description but on Property Name.

This behaviour is seen on Vassal 3.6.19 and more recent versions, so not due to the search updates in Vassal 3.7.6.

It would be more user-intuitive if Marker names matchesd as Property names.

rlament:
An Editor search advanced filters change. Changes the filter selection for searching for Marker properties to be consistent with other properties.

On the _Marker_ edit pane, the name field is labelled "Property name". It seems logical to search for a Marker in the Property name search category. With this change, _Marker_ search becomes consistent with searches on _Calculated Property_ and _Dynamic Property_.

Closes #13052

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Marker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Marker.java
@@ -267,6 +267,6 @@ public class Marker extends Decorator implements EditablePiece {
    */
   @Override
   public List<String> getPropertyList() {
-    return Arrays.asList(values);
+    return Arrays.asList(keys);
   }
 }
--- a/vassal-app/src/main/java/VASSAL/counters/Marker.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Marker.java
@@ -263,6 +263,8 @@ public class Marker extends Decorator implements EditablePiece {
   }
 
   /**
+   * This implementation differs from the typical as the Marker's property names
+   * are the list of {@link #keys}.
    * @return a list of any Property Names referenced in the Decorator, if any (for search)
    */
   @Override


vassal-app/src/main/java/VASSAL/counters/Marker.java
```diff
@@ -267,6 +267,6 @@ public List<String> getExpressionList() {
    */
   @Override
   public List<String> getPropertyList() {
-    return Arrays.asList(values);
+    return Arrays.asList(keys);
```
uckelman, line 272: I fundamentally do not understand this change. Why is replacing `values` with `keys` correct here? Everything else uses `values`.

rlament: I suppose the programmer perspective is different from that of the users. Using the advanced editor search function with the `Property names` checkbox, should it match the marker's `Property name` or `Property value`? Vassal 3.7.22 matches the value not the name.
Marker_search

uckelman: Would you add a comment to that effect? Otherwise the next time someone looks at this, they'll think it's wrong.

rlament: Added a Javadoc comment. Was tempted to simplify `getPropertyNames()` and have `getPropertyList()` defer to it. Due to the override declarations, I suppose these functions could diverge.

PR #14660, opened 2026-04-19
Fix for showing obscured, rotated peek in mouse-over

### LINKED ISSUE #14646: Problem with `Obscurable` and `CanRotate` traits
- When hidden pieces are rotated, the do not always full side when
  hovering over them. 
  
## The test 

Open the [attached module](https://drive.google.com/file/d/1byFl2Vh7O6BSu2iy_59qOKMU_EIszZpy/view?usp=sharing) (built with [`pywargame`](https://gitlab.com/wargames_tex/pywargame) and attached script: [build.py](https://github.com/user-attachments/files/26841545/build.py)).

There are two rows of pieces. 

- The pieces in row 1 has a peek command defined to `P`
- The pieces in row 2 has no peek command defined 

There are 5 rows, where the `Obscurable` display is set differently.

- Row `A` has pieces with the display set to `Inset` 
- Row `B` has pieces with the display set to `Background` 
- Row `C` has pieces with the display set to `Peek` 
- Row `D` has pieces with the display set to `Image` 
- Row `E` has pieces with the display set to `Image inset`

All pieces have the _Peek on roll-over_ enabled.   Thus, when the
owning side hovers over, one will expect the un-hidden side to be
displayed. 

All pieces has the `CanRotate` trait with keys `A` and `D` set to
rotate the pieces. 

## To reproduce 

1. Open the module. 
2. Switch to side `A`. 
3. Select all the pieces 
4. Press `M` to hide all the pieces. 
5. Still as side `A`, and with all pieces, rotate them by pressing `A`
   or `D`. 
   - Note that the `Plain` piece in `C2` does not rotate. 
6. Move the pointer over the pieces - one by one. 
   - All the rotated pieces - except those in `D1` and `D2` - shows
     the hidden side, which is **_not_ the expected behaviour**. 
   - The pieces `D1` and `D2` show the owner hidden
     side - not the un-hidden side, which is **_not_ the expected
     behaviour**. 
   - The piece in `C2` shows the un-hidden image when the mouse hovers
     over, **as expected**.
7. Now select all pieces again, and rotate them to upright.
   - Note that the piece in `C2` will be rotated, and the un-hidden
     side will be shown. 
8. Then deselect all pieces
   - The piece in `C2` will show the hidden side.
9. Now hover the pointer over each of the pieces - one by one. 
   - All, but the piece in `C2` will show the un-hidden image - **as
     expected**.
   - The piece in `C2` shows the hidden side, which is **_not_ the
     expected behaviour**.  
   - If the piece in `C2` is selected, then the unhidden side is
     shown, **as expected**.
10. Now, select the piece in `C2`, and rotate it to its upright state.
    - Note when the piece in `C2` is selected, it will show the
      un-hidden side, **as expected**. 
      
11. Now switch to side `B`. 
    - All pieces will show the hidden side, **as expected**
    - Hover over each piece will show the hidden side, **as
      expected**. 
12. Now select all pieces and rotate them by pressing `A` or `D`. 
    - All pieces will remain hidden and rotated - including the piece
      in `C2`, **as expected**. 
13. Now hover over each piece - one by one. 
    - All pieces will show the hidden side. 

Yours,
_Christian_

cholmcc:
This patch fixes a problem where a rotated, masked, piece, with `autoPeeking` set to `true` is shown as the owner-obscured version when peeking at the piece via the `CounterDetailViewer`.

The problem is, that `FreeRotator` caches the images used, but does not update when the piece is shown in the `CounterDetailViewer`.  That means the `FreeRotator` will use the on-map image when drawing the piece in the `CounterDetailViewer`.

The fix is to add `isAutoPeeking` to the returned `VISIBLE_STATE` property to `Obscurable`, so that the `FreeRotater` will not necessarily use the same image for both the on-map piece and the piece shown in the `CounterDetailViewer`.

Fixes #14646

Also fixes a deprecation warning.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Obscurable.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Obscurable.java
@@ -21,7 +21,6 @@ import VASSAL.build.GameModule;
 import VASSAL.build.module.ObscurableOptions;
 import VASSAL.build.module.PlayerRoster;
 import VASSAL.build.module.documentation.HelpFile;
-import VASSAL.build.module.map.CounterDetailViewer;
 import VASSAL.command.ChangeTracker;
 import VASSAL.command.Command;
 import VASSAL.configure.BooleanConfigurer;
@@ -221,7 +220,7 @@ public class Obscurable extends Decorator implements TranslatablePiece {
   }
 
   public boolean isAutoPeeking() {
-    return autoPeekRollover && CounterDetailViewer.isDrawingMouseOver() && obscuredToOthers() && !obscuredToMe();
+    return autoPeekRollover && (getMap() != null && getMap().isDrawingMouseOver()) && obscuredToOthers() && !obscuredToMe();
   }
 
   @Override
@@ -316,7 +315,7 @@ public class Obscurable extends Decorator implements TranslatablePiece {
       return obscuredBy;
     }
     else if (Properties.VISIBLE_STATE.equals(key)) {
-      return myGetState() + isPeeking() + getProperty(Properties.SELECTED) + obscuredToMe() + piece.getProperty(key);
+      return myGetState() + isPeeking() + isAutoPeeking() + getProperty(Properties.SELECTED) + obscuredToMe() + piece.getProperty(key);
     }
     // FIXME: Access to Obscured properties
     // If piece is obscured to me, then mask any properties returned by
--- a/vassal-app/src/main/java/VASSAL/counters/Obscurable.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Obscurable.java
@@ -220,7 +220,7 @@ public class Obscurable extends Decorator implements TranslatablePiece {
   }
 
   public boolean isAutoPeeking() {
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/counters/Obscurable.java
```diff
@@ -221,7 +220,7 @@ else if (obscuredToOthers()) {
   }
 
   public boolean isAutoPeeking() {
-    return autoPeekRollover && CounterDetailViewer.isDrawingMouseOver() && obscuredToOthers() && !obscuredToMe();
+    return autoPeekRollover && (getMap() != null && getMap().isDrawingMouseOver()) && obscuredToOthers() && !obscuredToMe();
```
uckelman: Unnecessary parenthesis.

cholmcc: Those were there because I was not sure if it should be 
```getMap() != null && getMap().isDrawingMouseOver()``` 
or 
```(getMap() == null || getMap().isDrawingMouseOver())```

However, I think the first case is the correct logic.  If the piece has no map, then we are not auto-peeking in any shape or form.

PR #14664, opened 2026-04-19
Updates for the developerguide - particularly on release checklist

cholmcc:
By request: https://github.com/vassalengine/vassal/pull/14631#issuecomment-4275964009

Yours,
_Christian_

### DIFF
--- a/developers-guide/developers-guide.adoc
+++ b/developers-guide/developers-guide.adoc
@@ -1,7 +1,3 @@
-
-Refer also to the Wiki for developer documentation:
-link:https://vassalengine.org/wiki/Main_Page#Engine_Development[]
-
 = Developer's guide
 :toc: left
 :toclevels: 4
@@ -175,33 +171,23 @@ Each build step can be skipped by passing certain parameters.
 
 ====== Examples
 * Compile only, build jar in `vassal-app/target`, skip tests, checkstyle, spotbugs, javadoc, asciidoctor, clirr:
-[source]
-----
-mvn clean package -DskipTests=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Dmaven.javadoc.skip=true -Dasciidoctor.skip=true -Dclirr.skip=true
-----
+
+ $ mvn clean package -DskipTests=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Dmaven.javadoc.skip=true -Dasciidoctor.skip=true -Dclirr.skip=true
 
 * Run AsciiDoctor only:
-[source]
-----
-mvn clean compile -Dmaven.main.skip=true -DskipTests=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Dmaven.javadoc.skip=true -Dclirr.skip=true
-----
+
+ $ mvn clean compile -Dmaven.main.skip=true -DskipTests=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Dmaven.javadoc.skip=true -Dclirr.skip=true
+
 
 ==== Regular TODOs
 
 ===== Check for new dependency versions
 
-[source]
-----
-mvn versions:display-dependency-updates
-----
+ $ mvn versions:display-dependency-updates
 
 ===== Check for maven plugin updates
 
-[source]
-----
-mvn versions:display-plugin-updates
-----
-
+ $ mvn versions:display-plugin-updates
 
 ==== Configuration of dependent projects
 
@@ -216,7 +202,7 @@ The project's public maven repository can be found at https://vassalengine.org/m
       <groupId>org.vassalengine</groupId>
       <artifactId>vassal-app</artifactId>
       <version>3.4.1</version> <!--1-->
-      </dependency>
+    </dependency>
   </dependencies>
   <repositories>
     <repository>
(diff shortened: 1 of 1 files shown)


uckelman:
This has made me realize that we need a different release procedure. The release shouldn't be tagged before all the files that go into the release are checked in. We had that prior to adding deb and Flatpak builds.

cholmcc:
> This has made me realize that we need a different release procedure. The release shouldn't be tagged before all the files that go into the release are checked in.
The step 
```
$ git push --atomic origin HEAD x.y.z
```
actually does this for you.   _Both_ pending commits _and_ the tag are pushed as an atomic operation.  This also means that only one workflow is run - namely the `release.yml`.   So in that sense, it is not all that different. 

> We had that prior to adding deb and Flatpak builds.

That was because you made the releases "by-hand" and not through the GitHub actions.  Therefore, the need to push the commits before tagging - which I always assumed you did - is not a consequence of the Debian/RPM/FlatPak stuff, but because you use a GitHub workflow to make the releases. 

What I think _could_ change, would be to drop the `release-x.y` branch, and simple do everything on `master`.  As features and bugs are added in separate PRs, there is, in my mind, little reason to have a separate release branch.   It seems like most of the time that `master` and `release-3.7` are totally in sync.   That would also remove the need to sync those two branches constantly. 

If, at any point, one would need to back-port fixes or features to a previously released tag, then one can easily make a branch for that. 

Did you see the [PR](https://github.com/vassalengine/vassal-site/pull/4) I made against `vassalengine/vassal-site`?  It simply adds a utility script to download artefacts to the web-site in a semi-automated way - just to make life easier :smile:

Also, I think the `release.yml` should be updated to make a `ReferenceManual` artefact - that again will make everything easier. 

I think I will make a PR where I 
- Create the `ReferenceManual` artefact 
- Fix the file permissions in the `flatpak` recipe. 

Alternatively, I could make two separate PRs of that, or even merge that into this one.  Let me know what you prefer. 

Yours,
_Christian_

cholmcc:
> I think I will make a PR where I
> 
> * Create the `ReferenceManual` artefact

Should `snapshots` also be made available as an artefact so that it can be sync'ed to `vassal.org:/var/www/html/site/maven`?

uckelman:
> What I think could change, would be to drop the release-x.y branch, and simple do everything on master. As features and bugs are added in separate PRs, there is, in my mind, little reason to have a separate release branch. 

`master` is for 3.8. Without a separate branch for 3.7, we stop having the ability to do 3.7 releases.

cholmcc:
> `master` is for 3.8. Without a separate branch for 3.7, we stop having the ability to do 3.7 releases.

I don't understand that.   There's hardly ever any difference between `master` and `release-3.7`, so in what way is `master` for 3.8?   That doesn't make any sense.   And why would doing main developments on `master` and removing `release-3.7` prevent any 3.7 releases?   I really don't get it.   I may have missed something, so could you please explain it so that I might get it? Thanks.

If all feature and bug PRs are merged into `master`, then `master` represents the latest developments.  Then, at some point, `master` is tagged and a release is made.   On the other hand, if all feature and bug PRs are merged into `release-3.7`, only to be merged into `master` shortly after, what is then the real difference between `master` and `release-3.7`?

> I think I will make a PR where I
> 
> - Create the ReferenceManual artefact
> - Fix the file permissions in the flatpak recipe.

See https://github.com/vassalengine/vassal/pull/14672 - this also makes the script `dist/linux/integration.sh` executable - an oversight on my part I think. 


> Did you see the [PR](https://github.com/vassalengine/vassal-site/pull/4) I made against `vassalengine/vassal-site`? It simply adds a utility script to download artefacts to the web-site in a semi-automated way - just to make life easier 😄

If you take that PR, then this PR should be updated. 

Also, could you say what you think about 

> Should `snapshots` also be made available as an artefact so that it can be sync'ed to `vassal.org:/var/www/html/site/maven`?

Thanks.

Yours,
_Christian_

cholmcc:
> See #14672 - this also makes the script `dist/linux/integration.sh` executable - an oversight on my part I think.

Example release: https://github.com/cholmcc/vassal/actions/runs/24716750421

uckelman:
> I don't understand that. There's hardly ever any difference between master and release-3.7, so in what way is master for 3.8? That doesn't make any sense.

There are PRs which were merged to master for 3.8 which cannot be in 3.7---specifically 12763 and 12944.

cholmcc:
> There are PRs which were merged to master for 3.8 which cannot be in 3.7---specifically 12763 and 12944.

1. Why where these PRs then merged in the first place - if they are problematic?
2. The two PRs you mention #12763 and #12944 do not seem problematic - they simply add a property each - a global property and a message property for the movement reporter.  They seem rather benign to me, and I don't understand why they could not go into 3.7. 
3. If there are merged PRs that are problematic, then they can be un-done and put back to open PRs, only to be merged when they are no longer troublesome.  

I still don't quite see why you need the two branches - I certainly don't think merged PRs like the ones above warrant that. 

Finally, could you please answer my questions 
- Will you merge https://github.com/vassalengine/vassal-site/pull/4 ? If you do, then _this_ PR need to be updated. 
- Will you merge https://github.com/vassalengine/vassal/pull/14672 ? Again, then this PR need to be updated. 
- Do you think the `snapshots` (or `releases`) directory of a build should be stored as an artefact so that it is easier to pick up and put in `vassal.org:/var/www/html/maven`?  If so, then https://github.com/vassalengine/vassal/pull/14672  should probably be updated too. 
- Will you OK _this_ PR - possibly after edits due to the above? 
- Will you OK the PR https://github.com/flathub/org.vassalengine.vassal/pull/11 so that the FlatPak package may be updated? 

Thanks. 

Yours,
_Christian_

cholmcc:
> I still don't quite see why you need the two branches - I certainly don't think merged PRs like the ones above warrant that.

The best argument I've seen is https://github.com/vassalengine/vassal/pull/12763#issuecomment-1743845370 by @BrentEaston:

> We can't have functionality that a module may use that is present in some versions of 3.7, but not others.

However, that point seems moot with new game library and the discussion we had around the Vassal version number. As you may remember, I argued that the minimum Vassal version requirement should be _major_._minor_.0 - meaning the module will work with all _major_._minor_ releases of Vassal - and as @BrentEaston's comment above also suggests.  As I remember, you specifically argued against this with something like there's no expectation that a module will work with prior Vassal versions to the one it was created with, which would be counter to the above comment. 

If you truly want to keep some sort of compatibility between modules created with 3.7.z and any 3.7 release of Vassal, then it starts to make sense to have an `future-enhancement` branch to collect upcoming improvements.   But then, I think Vassal should bump the minor version far more often so that the `future-enhancement` branch does not grow too deep. 

In any case, the strategy should be made a bit more clear.  E.g., _new_ functionality cannot be introduced on a release branch, but should be targeted at a `future` branch, and also what _new_ functionality means.   When will future enhancements be merged?  Who's to be responsible for the future enhancement are compatible with release changes.  And so on.

In the context of making new releases, it doesn't really impact the basic procedure: 

1. Merge any PR that needs to into the next release 
2. Make book-keeping changes such as updating notes, release number in the metainfo, and similar. 
3. _Either_ push those changes and then tag (and release) via the web-interface, _or_ tag and then push commits _and_ tag atomically to make the release 
4. Update the web-page by downloading artefacts from the release build workflow 
5. Post release announcements from the artefact of the release build workflow 
6. Commit, push, and PR new manifest to FlatHub. 
7. Set up for the next development cycle (update version number to next release number, and so on). 

That seems like a fairly straight-forward procedure.   Some of them could possibly be automatised - such as putting the artefacts on the web-server - but would require a few updates in terms of tokens, keys, etc.   But I think with the scripts added with  https://github.com/vassalengine/vassal-site/pull/4 and https://github.com/flathub/org.vassalengine.vassal/pull/11 will make quite a few of these steps trivial. 

Perhaps step 2 above could also be a bit automatised.  The `make` target `version-set` could for example update `dist/linux/org.vassalengine.vassal.metainfo.xml` too - and, if really fancy, also `debian/changelog`.   Also, step 6 could be somewhat automatised in that the `release.yml` workflow could make the PR against the FlatHub repo.   Step 5 could probably also be automatised, but would require knowledge of the various APIs, access tokens, and the like. 

Yours,
_Christian_

uckelman:
> Why where these PRs then merged in the first place - if they are problematic?

There's nothing problematic about those PRs _for 3.8_. The two PRs in question introduce changes that would make modules saved with them incompatible with any 3.7 release.

> The two PRs you mention https://github.com/vassalengine/vassal/pull/12763 and https://github.com/vassalengine/vassal/pull/12944 do not seem problematic - they simply add a property each - a global property and a message property for the movement reporter. They seem rather benign to me, and I don't understand why they could not go into 3.7.

This is exactly my point. Modules using either of these features would fail to work in earlier versions of 3.7. This is why they must go into 3.8.

> However, that point seems moot with new game library and the discussion we had around the Vassal version number. As you may remember, I argued that the minimum Vassal version requirement should be major.minor.0 - meaning the module will work with all major.minor releases of Vassal - and as @BrentEaston's comment above also suggests. As I remember, you specifically argued against this with something like there's no expectation that a module will work with prior Vassal versions to the one it was created with, which would be counter to the above comment.

I can't tell what conversation you're referring to here.

In general, one would expect a module created with 3.7.x to work with any 3.7.y release, so long as it wasn't relying on a bug that was corrected in a later 3.7 release.

> If you truly want to keep some sort of compatibility between modules created with 3.7.z and any 3.7 release of Vassal, then it starts to make sense to have an future-enhancement branch to collect upcoming improvements. But then, I think Vassal should bump the minor version far more often so that the future-enhancement branch does not grow too deep.

This describes what we've been doing for many years. We've had a `release-3.x` branch whenever `3.(x+1)` and `3.x` are live at the same time.

> In any case, the strategy should be made a bit more clear. E.g., new functionality cannot be introduced on a release branch, but should be targeted at a future branch, and also what new functionality means.

Again, this has long been our practice. "New functionality" _for modules_ typically means something that would prevent a module from working with an older version of Vassal.

> When will future enhancements be merged? Who's to be responsible for the future enhancement are compatible with release changes. And so on.

I'm not setting any kind of schedule for merging PRs. I had been expecting to announce that once all the bug fix PRs open now are dealt with, that I'd begin looking at PRs for 3.8.

cholmcc:
> > Why where these PRs then merged in the first place - if they are problematic?
> 
> There's nothing problematic about those PRs _for 3.8_. The two PRs in question introduce changes that would make modules saved with them incompatible with any 3.7 release.

It would only make them incompatible _iff_ the module actually used those properties, _and_ only with 3.7 releases prior to the release the patches went into.   Certainly not "any 3.7 release". 

> In general, one would expect a module created with 3.7.x to work with any 3.7.y release, so long as it wasn't relying on a bug that was corrected in a later 3.7 release.

That's the sentiment in @BrentEaston's comment, but not what you said back when we discussed the issue in the context of the Game Library.   I'm sure with a bit of digging you can find that conversation if it isn't present for you. 

> Again, this has long been our practice. 

The comment was not about whether one thing or another is the current or not practise - but to make that practise _explicitly_ clear to all concerned.  A `CONTRIBUTING` document could outline these practises and conventions.  As it is right now, there are some who are in the know and others that are not, and that makes it harder for those that don't know to contribute. 

> I'm not setting any kind of schedule for merging PRs. I had been expecting to announce that once all the bug fix PRs open now are dealt with, that I'd begin looking at PRs for 3.8.

This was a comment on the _general_ policy - not on specific targets.   For example, a policy could be 

> When enough enhancement PRs have been merged into the `master` branch, we will typically release a new minor version release (e.g., 3.9.0).   What constitutes "enough" depends to a large extend on the number of features and the complexity of the enhancements, but is typically in the range of 5 to 10 PRs. 

Similarly, 

> When enough bug fix PRs or critical PRs have been merged into the current release branch (e.g., `release=3.8`) then we will typically make a new patch release (e.g., `3.8.2`).   What constitute "enough" PRs depends on the severity and complexity of the PRs, but typically lies in the range of 5 to 10 PRs. 
> 
> The release schedule is of course always contingent on the available man-power, and as such we cannot make any promises.  Vassal is, after all, a volunteer project. 

Vassal seems to have around 30,000 users - give and take - which is not a small user base, but then again - not huge.   I think being relatively explicit and transparent with the user and contributor base is generally a good thing which may lure even more users into the trap - erhm - fold. 

Yours,
_Christian_

uckelman:
> It would only make them incompatible iff the module actually used those properties, and only with 3.7 releases prior to the release the patches went into. Certainly not "any 3.7 release".

There is no facility to detect that kind of incompatibility. Unless someone builds that for V3, the only way to handle it is by checking the version number the module was saved with.

> The comment was not about whether one thing or another is the current or not practise - but to make that practise explicitly clear to all concerned. A CONTRIBUTING document could outline these practises and conventions. As it is right now, there are some who are in the know and others that are not, and that makes it harder for those that don't know to contribute.

I take your point that people who have wandered in more recently are not familiar with the conventions that Brent, Brian, and I had long been following. It's not clear to me that at this point in the life cycle of V3 that it is worth the effort to codify those conventions, because the amount of time they will still be useful is short.

uckelman:
> Finally, could you please answer my questions
>
>   Will you merge https://github.com/vassalengine/vassal-site/pull/4 ? If you do, then this PR need to be updated.

I don't know yet.

>   Will you merge https://github.com/vassalengine/vassal/pull/14672 ? Again, then this PR need to be updated.

This is merged.

>    Do you think the snapshots (or releases) directory of a build should be stored as an artefact so that it is easier to pick up and put in vassal.org:/var/www/html/maven?

~How would that work? The `maven-metadata.xml` file from a single build won't correctly list all the releases.~ 

I see how vassalengine/vassal-site/pull/4 handles this now.

>    Will you OK this PR - possibly after edits due to the above?

I don't know yet.

>    Will you OK the PR https://github.com/flathub/org.vassalengine.vassal/pull/11 so that the FlatPak package may be updated?

This is merged.

uckelman:
> Will you merge https://github.com/vassalengine/vassal-site/pull/4 ? If you do, then this PR need to be updated.

No. The vassal-site repo is intended to contain only files which are part of the site. Util scripts should go into vassal-site-src.

cholmcc:
> > Will you merge [vassalengine/vassal-site#4](https://github.com/vassalengine/vassal-site/pull/4) ? If you do, then this PR need to be updated.
> 
> No. The vassal-site repo is intended to contain only files which are part of the site. Util scripts should go into vassal-site-src.

See https://github.com/vassalengine/vassal-site-src/pull/6

uckelman:
This doesn't render correctly:
image

cholmcc:
The makefile inclusion stuff never seemed to work. All I did was to update the line number to point to the `.PHONY` line.  I'm no expert on asciidoctor so I have no idea how to fix it.

uckelman:
OK. I've removed it.

PR #14672, opened 2026-04-21
Updates for the release workflow

cholmcc:
- Put the generated reference manual in an artefact named `ReferenceManual` so that it can easily be picked up on `vassal.org`.
- Remove `--filesystem=/tmp` permission in FlatPak recipe.  Causes problems with FlatHub linter and isn't really needed.


uckelman:
Where will the Reference Manual artifact be visible after this change?

cholmcc:
> Where will the Reference Manual artifact be visible after this change?

The same place as all the other artefacts (see [updated `developers-guide.adoc`](https://github.com/cholmcc/vassal/blob/197db007dc1fecbc054dab276815474e4050ff35/developers-guide/developers-guide.adoc)).   That is, as artefacts of the `release.yml` workflow run - for example [here](https://github.com/cholmcc/vassal/actions/runs/24735378805) and [here](https://github.com/vassalengine/vassal/actions/runs/24393805519). 

The proposed [`update_from_release.sh`](https://github.com/cholmcc/vassal-site/blob/ab80d868158efe76c500fd8c18b24a64c35ff54a/util/update_from_release.sh) script figures out the location of all these artefacts by exploiting the GitHub API, so that you don't have to go an look for it yourself.   I think this PR, #14664, and https://github.com/vassalengine/vassal-site/pull/4 together will make making a release a fair bit more automatised, and thus free up time to other work - at least, that's the aim :smile: 

Yours,
_Christian_

PR #14676, opened 2026-04-24
Added missing screenshot captions, and releases to 3.2.17 in metainfo

cholmcc:
…info

This adds the missing captions to the screenshot gallery.   This also adds releases down to 3.2.17 (as found on GitBub).

### DIFF
--- a/dist/linux/org.vassalengine.vassal.metainfo.xml
+++ b/dist/linux/org.vassalengine.vassal.metainfo.xml
@@ -31,37 +31,384 @@
   <url type="bugtracker">https://github.com/vassalengine/vassal/issues</url>
   <url type="homepage">https://vassalengine.org/</url>
   <url type="vcs-browser">https://github.com/vassalengine/vassal</url>
+  <url type="help">https://vassalengine.org/doc/latest/userguide/userguide.pdf</url>
+  <url type="faq">https://vassalengine.org/about.html</url>
+  <url type="contact">https://vassalengine.org/contact.html</url>
+  <url type="donation">https://vassalengine.org/about.html#donate</url>
 
   <screenshots>
     <screenshot type="default">
       <image>https://vassalengine.org/images/screenshots/screenshot2-1224.jpg</image>
+      <caption>Vassal playing Twilight Struggle</caption>
     </screenshot>
     <screenshot>
       <image>https://vassalengine.org/images/screenshots/screenshot3-1224.jpg</image>
+      <caption>Vassal playing Stalingrad 42</caption>
     </screenshot>
     <screenshot>
       <image>https://vassalengine.org/images/screenshots/screenshot4-1224.jpg</image>
+      <caption>Vassal playing Command &amp; Colors Napoleonics</caption>
     </screenshot>
     <screenshot>
       <image>https://vassalengine.org/images/screenshots/screenshot5-1224.jpg</image>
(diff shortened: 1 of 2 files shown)


dist/linux/org.vassalengine.vassal.metainfo.xml
```diff
@@ -63,5 +67,194 @@
...
+    <release version="3.3.2" date="2020-07-16">
+      <url type="details">https://github.com/vassalengine/vassal/releases/tag/3.3.2</url>
+    </release>
+    <release version="3.3.1" date="2020-06-22">
+      <url type="details">https://github.com/vassalengine/vassal/releases/tag/3.3.1</url>
+    </release>
+    <release version="3.3.0" date="2020-06-16">
+      <url type="details">https://github.com/vassalengine/vassal/releases/tag/3.3.0</url>
+    </release>
+    <release version="3.2.17" date="2020-04-27">
```
uckelman: 3.2.17 was released 2016-12-31.

uckelman (changes requested):

dist/linux/org.vassalengine.vassal.metainfo.xml
cholmcc: > 3.2.17 was released 2016-12-31.

But, on the GitHub release page it says 2020-04-27 - six years ago to the day :smile:

Also, at `https://vassalengine.org/releases`, the date is **2017-01-01**

Does it make sense to give the older date when the reference is to the GitHub page?   Perhaps the reference URL should be `https://vassalengine.org/releases`?  

Perhaps pre-3.3.0 should all be there too.   See the attached XML file [`releases.xml`](https://github.com/user-attachments/files/27127519/releases.xml). 

Yours,
_Christian_

uckelman: > But, on the GitHub release page it says 2020-04-27 - six years ago to the day 😄

2020-04-27 is the date the GitHub release was created, not the day 3.2.17 was released. We hadn't used GitHub releases prior to that.

> Also, at https://vassalengine.org/releases, the date is 2017-01-01

That's the mtime for the files, which reflects when those files were written to disk, not when the release occurred.

If `releases.xml` is supposed to reflect when the releases occurred, the dates there should be the actual release dates, which are in the change log.

It's not clear to me what the purpose of adding the old releases to the Flatpak metadata is, as they're not available as Flatpaks.

cholmcc: The `.metainfo.xml` is not particular to Flatpak, but is used by `AppStream`.  That framework is used by many different tools to create indexes of software and so on.  I think f.ex. Gnome and KDE's software tools can use this information. 

See updated PR.

dist/linux/org.vassalengine.vassal.metainfo.xml
```diff
@@ -31,37 +31,384 @@
   <url type="bugtracker">https://github.com/vassalengine/vassal/issues</url>
   <url type="homepage">https://vassalengine.org/</url>
   <url type="vcs-browser">https://github.com/vassalengine/vassal</url>
+  <url type="help">https://vassalengine.org/doc/latest/userguide/userguide.pdf</url>
+  <url type="fag">https://vassalengine.org/about.html</url>
```
uckelman: Type should be `faq`?

uckelman (approved):

PR #14687, opened 2026-04-29
Fix problem downloading Windows aarch64 from Liberica build of OpenJDK

cholmcc:
This will fix the problem https://github.com/vassalengine/vassal/pull/14612#issuecomment-4342687310 - hopefully :)

Problem is that Bellsoft changes their prefix on the download URL - kinda annoying, and no real good way to detect it automatically, except by using their [REST Api](https://api.bell-sw.com/#text) - e.g.: 

```
$ curl "https://api.bell-sw.com/v1/liberica/releases?version=26.0.1+10&bitness=64&os=windows&arch=arm&package-type=zip&bundle-type=jdk&output=text&fields=downloadUrl"
https://github.com/bell-sw/Liberica/releases/download/26.0.1+10/bellsoft-jdk26.0.1+10-windows-aarch64.zip
```

This could be put into the `bootstrap.sh` script instead - as  

```
BELLSORT_URL=$(curl https://api.bell-sw.com/v1/liberica/releases?version=${FULL_JDK}+${JDK_BUILD}&bitness=64&os=windows&arch=arm&package-type=zip&bundle-type=jdk&output=text&fields=downloadUrl)
```

(could also be done for Windows 32bit - although it will likely never change).  

A similar tactic could be used for the Temurin builds using their [REST Api](https://api.adoptium.net/q/swagger-ui/#/Binary/getBinary), though I didn't get it to work.

Yours,
_Christian_

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -67,6 +67,7 @@ if [ "x$BUILD" != "x" ] ; then
 fi
 
 # --- Sanitise the JRE version we get in GitHub actions --------------
+FULL_JRE=$(echo "$JRE_API" | sed 's/\([0-9][0-9.]*\).*/\1/')
 JRE_API=$(echo "$JRE_API" | sed 's/\([0-9][0-9]*\)\..*/\1/')
 
 # --- Default architectures ------------------------------------------
@@ -88,9 +89,9 @@ TEMURIN_FILENAME_VERSION=${JRE_API}_${TEMURIN_BUILD}
 TEMURIN_URL="https://github.com/adoptium/temurin${JRE_API}-binaries/releases/download/${TEMURIN_VERSION}"
 TEMURIN_FILENAME_BASE="OpenJDK${JRE_API}U-jmods"
 
-BELLSOFT_VERSION=${JRE_API}+${BELLSOFT_BUILD}
+BELLSOFT_VERSION=${FULL_JRE}+${BELLSOFT_BUILD}
 BELLSOFT_URL="https://download.bell-sw.com/java/${BELLSOFT_VERSION}"
-BELLSOFT_DIR=jdk-${JRE_API}
+BELLSOFT_DIR=jdk-${FULL_JRE}
 
 BELLSOFT_WIN32_API=21.0.10
 BELLSOFT_WIN32_BUILD=10


cholmcc:
Alright, so as stated above, BellSoft provide a [REST Api](https://api.bell-sw.com/#text) to retrieve the URL of a particular release, os, and architecture.   The URL command is something like 

```
$ curl -s "https://api.bell-sw.com/v1/liberica/releases?version=${RELEASE}+${BUILD}&bitness=${BITS}&os=${OS}&arch=${ARCHITECTURE}&package-type=zip&bundle-type=jdk&output=text&fields=downloadUrl"
```
where 
| Field | Meaning | Example |
|--------|--------------|---------------|
| `${RELEASE}` | the release number | `26.0.1`, `21.0.10` |
| `${BUILD}`  | the build number of the release | `10` |
| `${BITS}` | the number of bits | `64`, `32` | 
| `${OS}`  | the operating system name | `windows`, `macos`, `linux`  |
| `${ARCHITECTURE}` | target machine architecture | `arm`, `x86` |

Temurin also has a [REST Api](https://api.adoptium.net/q/swagger-ui/#/Binary/getBinary) which works a little differently.   The URL command is something like 

```
$ curl -s  "https://api.adoptium.net/v3/binary/version/jdk-${RELEASE}+${BUILD}/${OS}/${ARCHITECTURE}/jmods/hotspot/normal/eclipse?project=jdk"  -i | sed -n 's/^location: *//p'
```
where 
| Field | Meaning | Example |
|--------|--------------|---------------|
| `${RELEASE}` | the release number | `25.0.3` |
| `${BUILD}`  | the build number of the release | `9` |
| `${OS}`  | the operating system name | `windows`, `mac`, `linux`  |
| `${ARCHITECTURE}` | target machine architecture | `x64`, `aarch64`  |

Both of these REST Api end-points could be used in `bootstrap.sh` to the proper download URLs.   Note, from the GitHub action `setup/java` we get the release and build that was installed - that is, we get 
```
${RELEASE}+${BUILD}
```
above, and we can simply pass that into the `bootstrap.sh` script to ensure we get the right `jmods` files. 

When running the `bootstrap.sh` in a local repository, we can use the same format to specify the release and build number, _or_ we can pass those as individual options. 

Yours,
_Christian_

cholmcc:
OK, so I have a version of `bootstrap.sh` that uses the Temurin and Bellsoft APIs to get the URL of downloads given the release and build version of OpenJDK.   This version will make the process more robust as we will get the needed information directly from the horses mouth.   

Coupled with the information from the GitHub `setup/java` action, this make the entire build process pretty fault tolerant - wrt. to the JDK used - and automatically adapts to new releases of the JDK as needed.   

You are, of course, still in control of the JDK API version you want to use. 

Do you want that?  If so, then I will update this PR. 

Yours,
_Christian_

cholmcc:
> Do you want that? If so, then I will update this PR.


I made another PR: #14694 - so that you may choose between which of the fixes you like the most.  I would recommend #14694 as it automatises things more and requires less maintenance  by already busy developers :smile:

Yours,
_Christian_

uckelman:
I've merged #14694. Thanks.

PR #14694, opened 2026-05-01
Updates to `bootstrap.sh` to automate JDK downloads

cholmcc:
This patch fixes up `boostrap.sh` so that it will query the Temurin and Bellsoft REST Apis for the appropriate download links given a release and build number.  In this way, one does not need to manually edit the file in case a new JDK is used - the release and build information is provided by the GitHub `setup/java` action in GitHub workflows, and for manual runs, one can specify the release and build numbers on the command line.

Of course, one may want to change the default numbers from time to time.

Note that one can still choose which Java API version to use - in GitHub workflows by providing an option to `setup/java` and manually by specifying the appropriate release and build number on the command line.

Several other changes have also been made to the `bootstrap.sh` script - for example options to turn on verbosity, leave out certain steps, only show what would be done, clean-up the downloads and so on.  These are there to allow a developer to check things out and so on.

This is an alternative PR to PR #14687.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,13 +1,52 @@
 #!/bin/bash -e
 
+# --- Default release numbers ----------------------------------------
+#
+# Edit these for reasonable defaults
+#
 L4J_VERSION=3.50
 LIPO_VERSION=0.8.4
-JRE_API=26
-TEMURIN_BUILD=35
-BELLSOFT_BUILD=37
+TEMURIN_RELEASE=26+35
+LIBERICA_RELEASE=26.0.1+10
+
+# --- Variables used below -------------------------------------------
+#
+# Default values set here
+#
 ARCHS=
 NO_ACTION=0
 CLEAN=0
+WRAPPER=1
+BUILD_DMG=1
+VERBOSE=0
+CURL_SILENT=-s
+WGET_SILENT=-q
+UNZIP_SILENT=-q
+TAR_VERBOSE=
+MAKE_VERBOSE=
+CMAKE_QUIET="-DCMAKE_MESSAGE_LOG_LEVEL=WARNING"
+
+# --- Target directories ---------------------------------------------
+DMGDIR=dist/dmg
+L4JDIR=dist/launch4j
+JDKDIR=dist/jdks
+LIPODIR=dist/lipo
+
+
+# --- Message --------------------------------------------------------
+#
+# First argument is verbosity level
+# Rest are message arguments to echo
+#
+msg() {
+    lvl=$1 ; shift
+    if [ $VERBOSE -lt $lvl ] ; then
+        return
+    fi
+
+    echo "${@}"
+}
+
 
 # --- help -----------------------------------------------------------
 usage() {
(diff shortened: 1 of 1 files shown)


uckelman:
@cholmcc Something's still not working: https://github.com/vassalengine/vassal/actions/runs/25257686561/job/74059688144

cholmcc:
Well, it work in the first go (the merge), as well in all my previous tests. 

It seems like there was a cache hit where it really shouldn't be - in `Cache packaging dependencies` - so that `install packaging dependencies` - or `bootstratp.sh` - is never run.   That means that `make` cannot find the prerequisites to make the release build because the `jmods` are not there.    

Perhaps, to avoid this, the step  `Cache packaging dependencies` should be removed.   Due to the new constrains in JDK 26 that you cannot `jlink` with different builds it now means that `bootstrap.sh` _has_ to be run for the particular built that is being done.      This is what the recent update to `bootstrap.sh` addresses.      But this also means that we cannot use a cache of `packaging dependencies` from one sub-build to another.   

This change would probably need to be done in both `package.yml` and `release.yml`. 

Yours,
_Christian_

cholmcc:
Perhaps a little simpler way of sorting this out - which probably has the same effect as removing the cache - is to be more specific in what is cached - i.e., change 

```
      - name: Cache bootstrap 
        uses: actions/cache@v5
        with:
          path: |
            ./dist/dmg
            ./dist/jdks
            ./dist/launch4
            ./dist/lipo
```
to 
```
      - name: Cache bootstrap 
        uses: actions/cache@v5
        with:
          path: |
            ./dist/dmg
            ./dist/jdks/windows-x86_64
            ./dist/jdks/macos-x86_64
            ./dist/jdks/macos-aarch64
            ./dist/launch4
            ./dist/lipo
```
so that we insist that the proper `jmods` are present before triggering the cache hit. 


It seemed to work: https://github.com/cholmcc/vassal/actions/runs/25271935197, but the thing is that it can depend on when the workers are being run.  However, I think the logic is sound. 

Made a PR: https://github.com/vassalengine/vassal/pull/14706

Yours,
_Christian_

cholmcc:
Just to be clear 

- In the past, caching the result of running `bootstrap.sh` worked because _all_ `jmods` was downloaded without exception.  
- In the current scheme, only the `jmods` relevant to the current installed JDK (via `setup/java` action), are downloaded.  This is because
  - Newer `jlink` will _not_ link `jmods` from a different built (e.g., `Temurin` versus `Liberica`). 
  - We get the release and build of the`jmods` from the `setup/java` action so that we are sure that the `jmods` we get are compatible with the JDK being used. 
  - The current scheme 
    - takes care of the `jlink` problem 
    - ensures consistency between build, package, and runtime when the JRE is embedded in the package.
- `bootstrap.sh` _could_ of course download everything in a go, but then it would have to "guess" which release+build to get for the non-current builds.   That means an additional maintenance step to ensure that the "best guess" is up to date with what is currently available. 
  - The current scheme avoids that altogether by querying the APIs using the information from `setup/java`. 

Yours,
_Christian_

uckelman:
Thanks. 

I look forward to the day when all of this can be retired. `jlink` is such a stupid tool. If it were designed properly, `jlink`  would work for any older version and use whatever jmods you suppied for it... Sigh.

PR #14724, opened 2026-05-09
Replace `-` with `~` when writing RPM `.spec` file

cholmcc:
Should fix RPM build on release


cholmcc:
This PR is good to go. See test: https://github.com/cholmcc/vassal/actions/runs/25589276985

PR #14730, opened 2026-05-10
14729 Cancelling Load and Fast Forward cancels both the load and the fast forward

### LINKED ISSUE #14729: Cancelling of Load and Fast Forward does not cancel fast forwarding of all preloaded log files
Cancelling out of the _File Chooser_ does not cancel the fast forward for both _Load and Fast Forward_, and _Load, Fast Forward and Append_. In my case I was loading multiple logs using _Load Continuation_, which I was going to single step through. Upon inadvertently clicking _Load and Fast Forward_, I cancelled that action at which point Vassal fast forwarded through all the loaded log files. Had to restart loading with the first log file rather than perform a massive number of undo clicks.

The logical behaviour when cancelling an action is a full cancelation of the entire action. Cancelling only the first step is counterintuitive. Cancel should work the same way for all load commands.

However, some users do treat this as a feature to fast forward previously loaded log file(s) as mentioned in the post referenced below. Adding a fast forward button might be a useful feature.

[https://www.gamesquad.com/forums/index.php?threads/linking-vasl-logfiles.174048/]()

rlament:
Cancelling the File Chooser for both _Load and Fast Forward_, and _Load, Fast Forward and Append_ also cancels the fast forward. This differs from the existing behaviour where the fast forward is not cancelable.

Closes #14729

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/GameState.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/GameState.java
@@ -737,7 +737,10 @@ public class GameState implements CommandEncoder {
    */
   public void loadGame(boolean continuation, boolean forceForeground) {
     final GameModule g = GameModule.getGameModule();
+    loadGame(g, continuation, forceForeground);
+  }
 
+  private boolean loadGame(GameModule g, boolean continuation, boolean forceForeground) {
     if (gameStarted && continuation) {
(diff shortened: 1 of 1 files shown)


cholmcc:
This change makes perfect sense - if the load fails - or is cancelled - then do not try to replay it.

PR #14733, opened 2026-05-12
Default all of ListTurnTracker's items to the enabled state

rlament:
The ListTurnTracker has an 'active' flag array to store the state of each list item. In some modules, players can disable individual list items with that state stored in the active array.

When loading a saved game, the _setState()_ function initializes the ListTurnTracker. If the saved game is corrupt or possibly out of date, and since the _setState()_ function defaults the active array to false, all items in the ListTurnTracker can become disabled. This is problematic now that the active flag bug has been fixed because advancing the Turn Tracker will skip over the complete ListTurnTracker list.

- If the module has "Allow players to hide items" unchecked, there is no way for a player to re-enable the list items.
- Within the module editor there is no means to change the active flag state.
- For the average module designer it will be difficult if not impossible to edit a vsav file because of the encryption. This will likely result in recreation of the scenario from scratch.

Both _reset()_ and _setAttribute()_ default the active array to true. _setState()_ initializes to false making it inconsistent with these methods. In fact in debug you can see the active array with all true flags replaced with false flags. Setting the active array to true makes it consistent across all initialization methods.

A case in point is this forum thread [Turn Counter Mystery](https://forum.vassalengine.org/t/turn-counter-mystery/90024).

My suspicion is that a vsav from the prior version was loaded as the basis for an updated scenario. The old module had a slightly different TurnTracker configuration, the decoder was out of sync and thus all list items were disabled.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/turn/ListTurnLevel.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/turn/ListTurnLevel.java
@@ -133,6 +133,7 @@ public class ListTurnLevel extends TurnLevel implements ActionListener {
 
     final String[] s = sd.nextStringArray(0);
     active = new boolean[list.length];
+    Arrays.fill(active, true);
     final int l = Math.min(s.length, active.length);
     for (int i = 0; i < l; i++) {
       active[i] = s[i].equals("true"); //$NON-NLS-1$
--- a/vassal-app/src/test/java/VASSAL/build/module/turn/ListTurnTrackerTest.java
+++ b/vassal-app/src/test/java/VASSAL/build/module/turn/ListTurnTrackerTest.java
@@ -146,6 +146,32 @@ public class ListTurnTrackerTest {
     }
   }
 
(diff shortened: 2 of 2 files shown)


uckelman:
Would it be worthwhile to ask the user to try one of the test builds from this PR?

rlament:
> Would it be worthwhile to ask the user to try one of the test builds from this PR?

This won't have any effect on saved games because the active flags are already false. The user would have to go through the save game update process. I did that myself to prove this out. The steps for Balkan Front are:

1. Download module version 0.80.0 and extract a saved game. I used the blank scenario. I guess one could also load that scenario and immediately save game.
2. Using this PR, run module version 0.90.0 and load that saved game. The year is messed up (reset to zero), but can be corrected by the player. Advance the turn tracker and it should setup through all the phases.

The year is wrong because the TurnTracker configuration is different in 0.80.0 and 0.90.0. The decoder is out of sync and picks up the wrong fields. 

The unit tests cover many cases, but not this one. Thinking about it, a test case could be added to simulate it.

Issue #14559, opened 2026-03-17
[CEP] IllegalBuildException: failed to load class Sombra.SombraDiceButton


uckelman:
| ABR | |
|-----|-----|
| Version | 3.7.20 |
| Reporter |  |
| Error Log | [4857 bytes](https://vassalengine.org/files/gh/908b8d3f86ab203ee59456c0c61bbe769fbaece8) |

<pre>

CEP v1.0 3.7.20

VASSAL.build.IllegalBuildException: failed to load class Sombra.SombraDiceButton
 at VASSAL.build.Builder.create(Builder.java:148)
 at VASSAL.build.Builder.build(Builder.java:80)
 at VASSAL.build.AbstractBuildable.build(AbstractBuildable.java:104)
 at VASSAL.build.GameModule.build(GameModule.java:790)
</pre>

PR #14750, opened 2026-05-19
Report Internationalization setup inconsistencies as Bad Data

rlament:
The support of Internationalization in `AbstractConfigurable` types, requires three arrays ideally all of equal length. Since `AttributeNames`, `AttributeDescriptions`, and `AttributeTypes` are defined in different places it is quite possible for arrays to drift out of sync. The `@Override` declaration on the get methods can further exasperate the situation as the module designer/coder may inadvertently omit the proper chaining to super methods. In addition, enhancements and improvement to the Vassal executable can inadvertently break legacy modules.

A mismatch in array lengths is not a fatal error. A module can still be playable. When the name or description arrays are too short the modules currently terminate with the uncaught `IndexOutOfBoundsException`.

The intent of this PR is to allow modules with translation issues to run rather than terminally fail. This PR catches the exception and reports "Bad Data" allowing players to continue. The "Bad Data" message hopefully prompts the module designer to investigate and address the configuration error.

This is one of the issues preventing the module CEP_v1_0.vmod (#14559) from running using current versions of Vassal. There is an inherent issue with the module itself however it is not fatal and still playable (with mods to DiceButton) if the `IndexOutOfBoundsException` is caught.

A sample warning message looks like this:

Bad Data in Module: Source: Sombra.SombraDiceButton Error: Sistema Sombra - CEP[Sistema Sombra de Dados]: Length mismatch of AttributeNames [15], AttributeDescriptions [15] and AttributeTypes [24]. Array lengths should match.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/i18n/ComponentI18nData.java
+++ b/vassal-app/src/main/java/VASSAL/i18n/ComponentI18nData.java
@@ -28,11 +28,13 @@ import java.util.TreeMap;
 import VASSAL.build.AbstractConfigurable;
 import VASSAL.build.AbstractFolder;
 import VASSAL.build.AutoConfigurable;
+import VASSAL.build.BadDataReport;
 import VASSAL.build.Configurable;
 import VASSAL.counters.BasicPiece;
 import VASSAL.counters.Decorator;
 import VASSAL.counters.GamePiece;
 import VASSAL.counters.PlaceMarker;
+import VASSAL.tools.ErrorDialog;
 
 /**
  * Object encapsulating the internationalization information for a component.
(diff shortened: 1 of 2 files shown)


jonathanrwatts:
That error message does not make it at all clear that the error is in the Internationalization setup.  The user/module owner would need to know that those arrays deal with internationalization to have any clue how to fix the error.

rlament:
Writing concise informative error messages to direct the reader towards a solution are not easy. This is at least my third revision prior to even posting. 

That may be a poor example since it entails some familiarity with the specific module. This message is not targeted to the casual Vassal player or module designer. It is intended for the Vassal programmer i.e. someone adding custom code to a module. It is Java-level changes or additions that would most likely be the source.

To maintain a similar look and feel, the error message is generated using common error reporting functions.

This is how I envision someone understanding and locating the issue using the example provided in the original post:
1. The custom code extends the Vassal _DiceButton_ class and is named _Sombra.SombraDiceButton_. That is the source field.
2. Within the module, the component is _Sistema Sombra - CEP[Sistema Sombra de Dados]_. See figure.
3. The lengths for each array are given in the error message as the numbers in square brackets. No need to count elements, the lengths are reported making the mismatch that much more obvious. Square brackets are chosen to reflect that these refer to arrays.
4. The functions that generate these arrays are named: _getAttributeNames_, _getAttributeDescriptions_ and _getAttributeTypes_. It seems redundant to include the 'get' prefix in the error message. These are the arrays the user should focus on. Knowing the java class and module component should narrow down the location of the problem.

Including the term _Internationalization_ in the message could actually be misleading. It is in that part of the code where the array lengths really matter, but these arrays are also used elsewhere. Perhaps I should retitle of the PR? The change to the codebase is in the i18n section, but the configuration that generates the error is in user code.

CEP

PR #14859, opened 2026-06-30
Fix for `DynamicPropertyTest`

### LINKED ISSUE #14857: `DynamicPropertyTest` fails due to recent change in `DynamicProperty`
See https://github.com/vassalengine/vassal/commit/ad9954d9fa99cdc9b97c4e6c288aa6a4a317fcc6#commitcomment-190803451

Essentially, the code 
```
    // Don't compute the initial value in the editor.
    if (!GameModule.getGameModule().isEditorOpen()) {
      value = formatValue(value);
    }
```
merged into 3.7.24 recently, makes the `DynamicPropertyTest` test fail.   The above change should _either_ 
- be reverted, _or_
- amended to
   ```
    if (GameModule.getGameModule() == null || !GameModule.getGameModule().isEditorOpen()) {
    ```

Yours,
_Christian_

cholmcc:
fix: Fix broken commit #ad9954d

The recent change to `VASSAL.counters.DynamicProperty` (see commit #ad9954d) breaks the test `VASSAL.counters.DynamicPropertyTest` because that commit _requires_ that a module is present, which is not the case for test.

Closes #14857

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/DynamicProperty.java
+++ b/vassal-app/src/main/java/VASSAL/counters/DynamicProperty.java
@@ -202,7 +202,7 @@ public class DynamicProperty extends Decorator implements TranslatablePiece, Pro
     final Map map = getMap();
 
     // Don't compute the initial value in the editor.
-    if (!GameModule.getGameModule().isEditorOpen()) {
+    if (GameModule.getGameModule() == null || !GameModule.getGameModule().isEditorOpen()) {
       value = formatValue(value);
     }
 


rlament:
The problem is with the unit test. This is not the appropriate fix.

The problem is not specific to a version of Java. I reproduced the failure with Java 21, 25 and 26.

Prior to version 3.7.23 the `setValue()` function was not dependent on `GameModule`. Now that it is, the unit test should mock `GameModule`.

If you are running the unit test standalone, `GameModule` is null and the test fails. If you are running the test suite then the order of execution of unit tests may have changed. This could be due to the change of JDK or perhaps the IDE. The test suite has succeeded in the past because one or more unit tests do not properly cleanup leaving an instantiated `GameModule`. Thus `DynamicPropertyTest` may or may not succeed depending on the order of execution. An oversight on my part because the unit tests coincidentally passed (both locally and on the build server).

PR #14875 adds the `GameModule` mock to the unit test.

uckelman:
Concur. `GameModule` should never be null. Merged PR #14875 instead.

Issue #14381, opened 2025-12-04
Dice properties do not update when replaying a log file


rlament:
Given a module with two buttons where one is a dice button and the other an action button. The Dice button is named "dice" and the action button simply reports the contents of the `$dice_result$` property. Record a log file containing a dice roll and an action button press. Note that the action button correctly reports the latest dice roll result. Roll the dice to change the die result. Load the log file and play it back completely. Following that, press the action button noting that that reported result does not match the log file output. Instead it outputs the last local die roll. Also, if the dice button text happens to include the `$dice_result$` property then the button text does not change while stepping through the log file.

This is also seen in online play where the other players' dice properties do not update.

Log playback output:
```
DiceButton version 0.0
*** Dice = 2,6 *** <Ray>
Action Button dice results 2,6
Playback complete, press Action Button
Action Button dice results 1
```

PR #14902, opened 2026-07-18
Fix regression leaking Global Option properties over network

### LINKED ISSUE #14898: Global Option Property is getting converted to a plain Global Property
As described in [this forum post ](https://forum.vassalengine.org/t/tip-using-global-option-user-preferences/72741)Global Option Properties can be used to maintain local state information.

One simple use case is to maintain a formatted player name for report purposes without having to recalculate the formatting each time (e.g. in C&C Ancients module).

From Vassal v3.7.23 this feature is broken. As soon as Set Global Property is used on a Global Option Property its value is shared across the players.

To demo the feature, I have created a simple module (attached). To use the module, start it up and join an online game with two players taking sides A & B. There is a button in the toolbar, initially labelled "Press!".  Players should press this button in turn.

The correct behaviour, under Vassal version 3.7.22, is that the name each player is displayed in the button in their instance, when they press the button and the button in the other players' instances does not change.  However, from Vassal 3.7.23, the name in the button changes in all instances as soon as the button is pressed in any instance.

[TestGlobalOption.zip](https://github.com/user-attachments/files/30131817/TestGlobalOption.zip)

riverwanderer:
Checks if the Property is in the Global Options list before accepting a change into the local instance.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/module/properties/ChangePropertyCommandEncoder.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/properties/ChangePropertyCommandEncoder.java
@@ -58,21 +58,28 @@ public class ChangePropertyCommandEncoder implements CommandEncoder {
      * NB. If there is no containerID in the command, then it is a pre-bug fix command. Legacy
      * behaviour is to execute the change on the first matching property found in any container
      */
-    if (containerId.length() != 0 && !containerId.equals(container.getMutablePropertiesContainerId())) {
+    if (!containerId.isEmpty() && !containerId.equals(container.getMutablePropertiesContainerId())) {
       return null;
     }
 
-    MutableProperty p = container.getMutableProperty(key);
-    if (p == null) {
-      // If this is not an explicit global property
-      // then search for implicit module properties such as Special Dice properties.
-      if (container instanceof GlobalProperties) {
-        p = ((GlobalProperties) container).getParent().getMutableProperty(key);
+      MutableProperty p = container.getMutableProperty(key);
+      if (p == null) {
+          // If this is not an explicit global property
+          // then search for implicit module properties such as Special Dice properties.
+          if (container instanceof GlobalProperties) {
+              final MutableProperty implicitP = ((GlobalProperties) container).getParent().getMutableProperty(key);
+
+              // Filter out any property that is registered as a local user preference option
+              final boolean isPreference = VASSAL.build.GameModule.getGameModule().getPrefs().getOption(key) != null;
+
+              if (implicitP != null && !isPreference) {
+                  p = implicitP;
+              }
+          }
       }
-    }
-    if (p == null) {
-      return null;
-    }
+      if (p == null) {
+          return null;
+      }
 
     return new ChangePropertyCommand(p, key, oldValue, newValue);
   }
--- a/vassal-app/src/main/java/VASSAL/build/module/properties/ChangePropertyCommandEncoder.java
+++ b/vassal-app/src/main/java/VASSAL/build/module/properties/ChangePropertyCommandEncoder.java
@@ -62,24 +62,24 @@ public class ChangePropertyCommandEncoder implements CommandEncoder {
       return null;
     }
 
-      MutableProperty p = container.getMutableProperty(key);
-      if (p == null) {
-          // If this is not an explicit global property
-          // then search for implicit module properties such as Special Dice properties.
-          if (container instanceof GlobalProperties) {
-              final MutableProperty implicitP = ((GlobalProperties) container).getParent().getMutableProperty(key);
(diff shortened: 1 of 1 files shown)


riverwanderer:
@rlament - this fixes the Global Options Property issue with PR #14381 (tested via network using the test module in #14898). Could you review please and maybe re-test the dice fix ? I think that will still work.

vassal-app/src/main/java/VASSAL/build/module/properties/ChangePropertyCommandEncoder.java
```diff
@@ -67,7 +67,14 @@ public Command decode(String command) {
       // If this is not an explicit global property
       // then search for implicit module properties such as Special Dice properties.
       if (container instanceof GlobalProperties) {
-        p = ((GlobalProperties) container).getParent().getMutableProperty(key);
+        final MutableProperty implicitP = ((GlobalProperties) container).getParent().getMutableProperty(key);
+
+        // Filter out any property that is registered as a local user preference option
+        final boolean isPreference = VASSAL.build.GameModule.getGameModule().getPrefs().getOption(key) != null;
```
uckelman: This should go after the null check on `implicitP`, as there's no reason to run this line if `implicitP` is null.

riverwanderer: Thanks @uckelman . Latest update fixes this. Tested for both Global Option and standard Global Properties.

rlament:
This doesn't work for log files. Playing back a log file won't set a Global Options Property. Maybe that's how it always worked?

Edit: I created a log file using the test module. It's two chat messages with a button press in between. Played back the log as the same player using 3.7.17 and button text does not update. The command to update the button text is in the log file, but the decode function filters it out. Placed a breakpoint in the encode function and pressed the button. The oldValue is set as "press!", which confirms the property is not updated. The Global Options Property is therefore not set, which seems to indicate that is the expected behaviour. I would think these properties should be persisted.

What is the point of sending and logging this command if it is filtered out in the decode function? Why not filter it out in the encode function and never send it in the first place?

The dice fix still works.

riverwanderer:
> Edit: I created a log file using the test module. It's two chat messages with a button press in between. Played back the log as the same player using 3.7.17 and button text does not update. The command to update the button text is in the log file, but the decode function filters it out. Placed a breakpoint in the encode function and pressed the button. The oldValue is set as "press!", which confirms the property is not updated. The Global Options Property is therefore not set, which seems to indicate that is the expected behaviour. I would think these properties should be persisted.
> 
> What is the point of sending and logging this command if it is filtered out in the decode function? Why not filter it out in the encode function and never send it in the first place?
> 
> The dice fix still works.

I think I agree. I tested log files with the fix, and that also behaved as per pre-3.7.23 (i.e. the logged change of a Global Option Property was not visible when the log was played back on another computer).  This PR restores behaviour for Global Option Properties to what it was before 3.7.23. You are suggestiing to do the filter instead in another place (encode), providing we can be sure that doesn't cause other problems?

rlament:
Alright, I understand the Global Options Property a bit better now. They aren't persisted in the log but rather as module settings in the prefs folder. Agreed that this fix reverts Global Options Property handling back to pre 3.7.23 behaviour.

Moving the command filtering to the encode function may be risky. Leaving it as you have it is safest. Even though the command is not used, at least it is embedded in the log file and may prove useful at some point in the future.

Experimenting with a Global Options Property I noticed that _Global Key Commands_ don't update the values in the _Preferences_ window. Modifications in the _Preferences_ window do update the button text, but it doesn't work the other way. Not sure if this is the correct approach or the right place to do this but adding the following to the _encode_ function refreshes the _Preferences_ window keeping it in sync with _Global Key Commands_. It also means that the latest setting is stored in the prefs folder whether from an edit in the _Preferences_ window or from a _Global Key Command_. Perhaps best done as a separate PR.

```
    final Configurer configurer = VASSAL.build.GameModule.getGameModule().getPrefs().getOption(cpc.getPropertyName());
    if (configurer != null) {
      configurer.setValue(cpc.getNewValue());
      configurer.fireUpdate();
    }
```

BTW, the entire rational for PR #13904 was for automatic handling of adding player names to buttons or controls. The module designer adds the Startup GKC. Player names appear in the appropriate spot without requiring any manual intervention by the player.

rlament (approved):

riverwanderer:
Thanks @rlament 

On the subject of GKCs updating prefs, there is room for debate here as you suggest. At the moment, I use it to offer the user a convenient way to temporarily override s preference during a game, leaving them to set their long-term preference in the panel.

I don't quite yet see how PR #13904 relates, perhaps I haven't explored that use of the Startup GKC enough.

PR #14907, opened 2026-07-22
Update Preferences window on changes by SetGlobalProperty commands

rlament:
Module designers can add custom preference settings to the global options. If so configured, these settings can appear in the Preferences window. Typically players open the Preferences window to customize the appearance and/or behaviour of a particular module. Since the preference settings are tied to global properties, any game piece with a "Set Global Property" trait can also modify any preference settings.

When a user changes a setting in the Preference window, that change is reflected in the module. However when the setting changes from a "Set Global Property" trait, the value in the Preference window is not updated. Since the contents of the Preferences window is saved to the prefs folder, any change initiated by a "Set Global Property" trait is not persisted.

This PR fires an event to notify the Preference window of a preference setting change.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/SetGlobalProperty.java
+++ b/vassal-app/src/main/java/VASSAL/counters/SetGlobalProperty.java
@@ -27,6 +27,7 @@ import VASSAL.build.module.properties.MutableProperty;
 import VASSAL.command.Command;
 import VASSAL.command.NullCommand;
 import VASSAL.configure.BooleanConfigurer;
+import VASSAL.configure.Configurer;
 import VASSAL.configure.DynamicKeyCommandListConfigurer;
 import VASSAL.configure.FormattedExpressionConfigurer;
 import VASSAL.configure.IntConfigurer;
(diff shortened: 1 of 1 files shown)


riverwanderer:
As it stands - to my understanding - this PR would take away the ability to temporarily override a user preference. The non-persistence of user preference changes via SGP is a feature rather than a bug, so the effect of this PR on existing modules is a risk.

To make it backward compatible - I suggest - this PR would need a checkbox on the SGP trait that is specifically checked if a Global Option Property is to be propagated. For bonus XP, the checkbox could be greyed out / clear itself if the Property is not found in the Global Options list. Alternatively,  a cloned trait based on SGP but specific to setting Global Options in this way (SGP still left as is).

Adjacent to this PR, it would be sometimes useful if the Change-Property component could be attached to a Global Option Property, with ability to check it as "propagating to prefs" or not.

I wonder how many more module devs use this functionality (ie SGP on Prefs) ? I only say that because I've explored it a fair bit and, for me at least, it is working overall ok; even the "change-property" gap isn't something I would want to ask much time spent on, though I'm not surprised to find that I [once requested it](https://forum.vassalengine.org/t/add-change-property-trait-for-global-option-preferences/72696).

rlament:
> this PR would take away the ability to temporarily override a user preference.

That's correct. I'm ok with closing this and leaving things as they are.
 
> To make it backward compatible - I suggest - this PR would need a checkbox on the SGP trait that is specifically checked if a Global Option Property is to be propagated.

I would have one global flag very much like to "leading zeros are preserved" checkbox. User can choose either legacy mode or always update preferences. The challenge would be describing what this flag actually does.

> I wonder how many more module devs use this functionality

Definitely not an obvious feature.

uckelman:
I'm closing this for the reason @riverwanderer gave above.

PR #14928, opened 2026-07-29
Fix `Requires` in RPM spec file, which prevents installation

cholmcc:
- Currently installing the build RPM fails, as detailed in [this post](https://forum.vassalengine.org/t/vassal-3-7-25-released/90361/3).
- The problem is that the _or_ keyword should be lower case `or` - not `OR`, _and_ that the or'ed requirements need to bracketed in parenthesis.
- This fixes the RPM spec skeleton.

I did a local build of the RPM and tried to install the resulting RPM on
Fedora 42 (running as a virtual machine).   That worked.

### DIFF
--- a/dist/linux/vassal.spec.in
+++ b/dist/linux/vassal.spec.in
@@ -9,7 +9,7 @@ Release:	1
 Source0:	https://github.com/@REPOSITORY@/releases/download/@RELEASE@/VASSAL-@RELEASE@-linux.tar.bz2
 URL:		https://vassalengine.org/
 BuildArch:	noarch
-Requires:	java-17 OR java-18 OR java-19 OR java-20 OR java-21 OR java-22 OR java-23 OR java-24 OR java-25 OR java-26 OR java-27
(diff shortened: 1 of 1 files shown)


uckelman:
> The problem is that the or keyword should be lower case or - not OR, and that the or'ed requirements need to bracketed in parenthesis.

That's funny---I got what I did from an example, so the example was also wrong.

cholmcc:
> That's funny---I got what I did from an example, so the example was also wrong.

Without knowing the example, I suspect that you probably forgot to copy the `(...)` too.  

The [RPM documentation](https://rpm-software-management.github.io/rpm/manual/boolean_dependencies.html) seems indicate that the `(...)` _are_ required. 

> Boolean Expressions are always enclosed with parenthesis. [[from the above documentation]](https://rpm-software-management.github.io/rpm/manual/boolean_dependencies.html)

It is not so evident that the _or_ must be lower case, but all the examples use that.  That's why I found it safer to use the lower-case `or`. 

Yours,
_Christian_

PR #15004, opened 2026-08-21
Check random numbers for consistency

cholmcc:
This patch will place markers in a log file, so that an opponent may check that the log has not been doctored in a way to get better outcomes.

The strategy is best illustrated by a sequence of events.

1. Player **A** starts a game
2. **A** starts a log file - say `A1.vlog`
   - This will put the currently used seed - say `sA1` - into the log file
3. **A** generates random numbers
   - _Each_ random number is written to the log file, irrespective of whether it corresponds to a die roll or some other random thing.  In this way, the sequence of random numbers is documented in the log file.
4. **A** ends the log file
5. **A** sends the log file to player **B**

Now it is player **B**s turn

6. Player **B** opens the log file `A1.vlog`
   - Vassal will, as part of reading the log file, see the seed - `sA1` - written down in step 2 above.
   - The Random number generator is seeded with that seed `sA1`.
7.  **B** starts to step through the log file `A1.vlog`
    - When ever **A** made a random number, it is read in from the log file.   Vassal can now check, that given the above seed - `sA1`, that the random number read in is indeed the next random number in the sequence of random numbers - simply by drawing an equivalent random number from the random.
      - If the random number read from the log file matches the generated random number, then all is good.
      - If the random number read from the log file _does not_ match the generated random number, then something is a miss and the problem is reported (an exception is thrown).
8. Once **B** has stepped through the log file, **B** then _immediately_ opens up a new log file - say `B1.vlog`.
   - As above, that means that the current seed should be written to the log.
     - However, there's a bit of a caveat here, as we cannot reuse the old seed - `sA1` (otherwise we'd need to know the full sequence of random numbers, which will no be in the newly started log file).
     - So instead, we draw _one_ random number - say `iB1` (a `long`) using the current seed (the one read from the log file) and write that to the log file.
     - Furthermore, we use _that_ random number  as the new seed - `sB1` - from the random number generator, and we then write that seed to the log.
       - While we made a new seed - `sB1`, the seed we created is derived from the previous seed - `sA1`, and therefore we can check, later on, that that seed is as expected.
       - Had we simply generated a new seed, then we'd be back to the situation where a player may restart logs until favourable outcomes are achieved).
9. **B** now generates random numbers and these are written to the log file, just like above.
10. **B** ends the log file - `B1.vlog` and sends it to **A**.

Now comes the part where we check that things are consistent

11. **A** Opens up its _own_ log file `A1.vlog` first - This reinitialises the Random number generator with the seed in that log file - `sA1`.
12. **A** then fast-forwards through the log file. - As above in 7, this will read in the random numbers and check if they are indeed in the sequence (and in order) of the Random number generator given the seed `sA1`
 13. Once **A** has stepped through the log file `A1.vlog`, the game is back to the state in step 4 above. - But _crucially_ the current random number seed is set to the read seed - `sA1`, _and_ we have replayed the sequence of random numbers to the same point as in step 4 above.
 14. **A** now opens the log sent by **B** - `B1.vlog`
      - This means that the random number generated in step 8 above - `iB1` - using the seed `sA1` is read in.   We can therefore check that that random number is the expected random number - as in step 7 above.
      - Opening the log _also_ means we read in the seed - `sB1` - set in step 8 above, and we seed the Random number generator with that seed.
15. **A** nows steps through the log file `B1.vlog`
      - As in step 7 above, every time we see a random number in the log file - generated from the seed `sB1`, which we are currently _also_ using, we check if that read random number matches the next random number in the current Random number generator sequence.
 16. Once **A** has stepped through all of the log file `B1.vlog`, then **A** _immediately_ starts a new log file - say `A2.vlog`
     - Again a new random number, given the seed `sB1` is generated and written to the log
     - That random number is set as the new seed - `sA2`, and written to the log
17. **A** does its moves, including generating random numbers from seed `sA2`,  which are recorded in the log file.
18. **A** ends the log file `A2.vlog` and sends it to **B**

Now **B** follows the same logic:

19. **B** plays back its own log `B1.vlog`, so that the seed is set to `sB1`, and the sequence up to step 10 above is replayed.
20. **B** then opens the log file `A2.vlog`, which first checks the random number - `iA2` - based on seed `sB1` - written in the log, and then reads in the new seed `sA2`.
21. **B** then replayes the log and on each random number checks that it is the next random number in the sequence seeded in `sA2`.
22. **B** starts a new log - `B2.vlog`, generates a random number based on `sA2`, and sets that as the seed - `sB2`.
23. **B** makes random numbers, based on seed `sB2`
24. **B** Closes the log and sends it back to **A**

In this way, there's a consistency check on every step, that the sequence of random numbers is ultimate derived from the initial seed - `sA1` - possibly via intermittent seeds - `sB1`, `sA2`, `sB2`, `sA3`, ..., which them selves are derived - in a predictable fashion from the first seed `sA1`.   The "trick" is to use the fact that pseudo random number generators are just that - pseudo - and they will always produce the same sequence of random numbers given the same starting point (seed).

The benefit of this approach, in so far as it works and can be implemented in Vassal, is that it is entire self-contained and does not require external services or the like.  That also means that reading and writing log files does not require an internet connection.

Note, if no seed is seen in a log file, then no checks are performed. This would be to support older log files.  Also note, that the logic may carry over to on-line games, as an online game is really just a live replay of a log file.  In an online game, all players will use the same seed and they will all see the same sequence of random numbers so they can check for consistency.

Note that this patch changes the random number generator from a `java.security.SecureRandom` to a standard `java.util.Random`, because the former does not produce the same sequence of random numbers given the same starting seed, while the latter does. The use of a cryptography random number generator is, by all considerations, overkill, and the standard `java.util.Random` is surely safe and random enough for Vassal.

To try this out, pick any module you like and do the following. 

1. Open the module 
2. Start a log - say `A1.vlog`
3. Roll the dice a couple of times 
4. End the log
5. Quit Vassal 

Now for the "next" player 

1. Open the module
2. Fast-forward through the log `A1.vlog` (random numbers will be checked) 
3. Start a log - say `B1.vlog` 
4. Roll the dice a couple of times 
5. End the log
6. Quit Vassal

Back to the original player 

1. Open the module 
2. Fast-forward through the log `A1.vlog`  (random numbers will be checked)
3. Fast-forward through the log `B1.vlog`  (random numbers will be checked)
4. Quit Vassal 

Now, to do a case where inconsistencies are detected 

1. Open the module 
2. Start a log - say `A1.vlog`
3. Roll the dice a couple of times 
4. End the log
5. Quit Vassal 

Now for the "next" - cheating - player 

1. Open the module
2. Fast-forward through the log `A1.vlog` (random numbers will be checked) 
3. Roll the dice a couple of times (so as to get more favourable outcomes)
4. Start a log - say `B1.vlog` 
5. Roll the dice a couple of times 
6. End the log
7. Quit Vassal

Back to the original player who will detect the previous player's fraud

1. Open the module 
2. Fast-forward through the log `A1.vlog`  (random numbers will be checked)
3. Fast-forward through the log `B1.vlog`
    - This will result in an error dialog shown, warning that the random numbers read in the last log file - `B1.vlog` - are not consistent with the random numbers in the log file `A1.vlog`, and hence the previous player may have cheated. 

If, at any point, you'd like to see the content of a log file, use the `vsavdump.py` utility from [`pywargame`](https://gitlab.com/wargames_tex/pywargame). 

Yours,
_Christian_

### DIFF
--- a/vassal-app/src/main/java/VASSAL/build/GameModule.java
+++ b/vassal-app/src/main/java/VASSAL/build/GameModule.java
@@ -157,7 +157,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.NoSuchFileException;
-import java.security.SecureRandom;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -439,7 +438,7 @@ public class GameModule extends AbstractConfigurable
   /**
    * Random number generator
    */
-  private final Random RNG = new SecureRandom();
+  private final VASSAL.build.Random RNG = new VASSAL.build.Random();
 
   /**
    * Server object for online games
@@ -682,6 +681,7 @@ public class GameModule extends AbstractConfigurable
       .append(new RecursiveSingleChildInstance());
 
     addCommandEncoder(new ChangePropertyCommandEncoder(propsContainer));
+    addCommandEncoder(RNG);
   }
 
   /**
@@ -911,6 +911,7 @@ public class GameModule extends AbstractConfigurable
     theState = new GameState();
     theState.addTo(this);
     addCommandEncoder(theState);
+    theState.addGameComponent(RNG);
   }
 
   /**
--- a/vassal-app/src/main/java/VASSAL/build/Random.java
+++ b/vassal-app/src/main/java/VASSAL/build/Random.java
@@ -1,0 +1,503 @@
+/*
+ *
+ * Copyright (c) 2026 Christian Holm Christensen
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License (LGPL) as published by the Free Software Foundation.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU Library General Public
+ * License along with this library; if not, copies are available
+ * at http://www.opensource.org.
+ */
+package VASSAL.build;
+import VASSAL.command.Command;
+import VASSAL.command.CommandEncoder;
+import VASSAL.tools.SequenceEncoder;
+import VASSAL.build.module.BasicLogger;
+import VASSAL.build.module.GameComponent;
+import VASSAL.tools.ErrorDialog;
+
+public class Random extends java.util.Random
+  implements CommandEncoder, GameComponent {
+  protected static final String COMMAND_PREFIX = "RNG\t";
+  
+  protected boolean seedSeen = false;
+  protected long storedSeed = 0;
+
+  static final long serialVersionUID = 0L;
+
+  /**
+   * Get the next random value.  This registers the random value and
+   * puts it to the current log file or clients listening.
+   */
+  @Override
+  public int nextInt() {
+    final int ret = super.nextInt();
+    logRandom("I", Integer.valueOf(ret));
+    return ret;
+  }
+  /**
+   * Get the next random value.  This registers the random value and
+   * puts it to the current log file or clients listening.
+   */
+  @Override
+  public int nextInt(int bound) {
+    final int ret = super.nextInt(bound);
+    logRandom("M", Integer.valueOf(ret), Integer.valueOf(bound));
+    return ret;
+  }
+  /**
+   * Get the next random value.  This registers the random value and
+   * puts it to the current log file or clients listening.
+   */
+  @Override
+  public long nextLong() {
+    final long ret = super.nextLong();
+    logRandom("L", Long.valueOf(ret));
+    return ret;
+  }
+  /**
+   * Get the next random value.  This registers the random value and
+   * puts it to the current log file or clients listening.
+   */
+  @Override
+  public double nextDouble() {
+    final double ret = super.nextDouble();
+    logRandom("D", Double.valueOf(ret));
+    return ret;
+  }
+  /**
+   * Get the next random value.  This registers the random value and
+   * puts it to the current log file or clients listening.
+   */
+  @Override
+  public float nextFloat() {
+    final float ret = super.nextFloat();
+    logRandom("F", Float.valueOf(ret));
+    return ret;
+  }
+  /**
+   * Get the next random value.  This registers the random value and
+   * puts it to the current log file or clients listening.
+   */
+  @Override
+  public boolean nextBoolean() {
+    final boolean ret = super.nextBoolean();
+    logRandom("B", Boolean.valueOf(ret));
+    return ret;
+  }
+  /**
+   * Get the next random value.  This registers the random value and
+   * puts it to the current log file or clients listening.
+   */
+  @Override
+  public double nextGaussian() {
+    final double ret = super.nextGaussian();
+    logRandom("G", Double.valueOf(ret));
+    return ret;
+  }
+  /**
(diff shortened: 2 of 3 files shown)


cholmcc:
The above failure to build the Debian package is to be expected.     current `debian/changelog` has last release as 3.8.0-beta1, and `dch` sees `3.8.0-SNAPSHOT-....` as an earlier release number, and therefore fails.   It is safe to ignore this error.

riverwanderer:
Seems like a good idea, but this caught my eye...

> hence the previous player **_may_** have cheated.

Is that a euphemism or are false positives going to happen due to innocent mistakes or  the normal course of events ?

cholmcc:
> Is that a euphemism or are false positives going to happen due to innocent mistakes or the normal course of events ?

Mostly a euphemism.   

What I mean by that is
* A player may accidentally have messed up before starting the new log, which would be flagged as inconsistent when checked by the opponent.  That is, not really a cheat, but an honest mistake. 
* Without thorough testing, it is hard to say that there will be no false positives - things flagged as inconsistencies, but which are not the result of cheating or honest mistakes.   For example, one could imagine that there could be very particular circumstances where a random number _isn't_ written to the log, or some random number _isn't_ read from the log, which would mean there would be apparent - though not real - inconsistent sequences of random numbers.   Of course, I've done as best I can to make sure that shouldn't happen, by going to the source of the random numbers (the random number generator) rather than some higher level components (e.g., a dice roll). 

So "mostly" because I cannot rule out situations that I've not forseen, and also to not accuse users of cheating if what happened was an honest mistake. 

Yours,
_Christian_

cholmcc:
BTW, if a player does not replay the previous log file, then there will be no checks done.  That is, if the sequence is 

1. Open the module
2. Start a log - say A1.vlog
3. Roll the dice a couple of times
4. End the log
5. Quit Vassal

Now for the "next" player

1. Open the module
2. Fast-forward through the log A1.vlog (random numbers will be checked)
3. Start a log - say B1.vlog
4. Roll the dice a couple of times
5. End the log
6. Quit Vassal

Back to the original player

1. Open the module
2. Fast-forward through the log B1.vlog (random numbers will be checked, but only against itself)
3. Quit Vassal

Because the original player didn't play back `A1.vlog` before playing back `B1.vlog`, it means that there's no check against the seed set in `A1.vlog`.  Effectively, the original player assumes that the second player isn't cheating. 

Also, recent push fixes the case when a user does a _Load Log, fast forward, and append_. 

Yours,
_Christian_

cholmcc:
There's one way of "cheating" that could make this solution unattractive. 

1. **A** generates a log and ships it off to **B**. 
2. **B** opens the log and replays it. 
3. **B** now rolls the dice (or similar) a number of times and notes down the results. 
4. **B** closes the game (no log)

**B** now _knows_ the sequence of die rolls that are coming up. 

5. **B** now opens the game and loads and replays **A**'s log
6. **B** starts a new log 
7. When the time comes for **B** to make a die roll, then **B** consults its notes, and knows which die roll is coming up. 
   * This will allow **B** to choose between using the upcoming die roll where that die roll make be least detrimental to **B**'s objects.  For example, **B** may need to resolve 2 combats - one which **B** can afford to loose and one that **B** cannot afford to loose.  If the first upcoming die roll is in **B**'s favour, then **B** will use that one the crucial battle, and the next die roll (less beneficial) on the not so important battle.  The situation could of course be flipped, in which case **B** resolves the less significant battle first, and then the crucial one. 
  * Note that **B** cannot change the die rolls themselves.  **B** is still bound to the sequence of die rolls, but that sequence may be known. 
  * In games where the defender has some action to take as a consequence of each combat, this will be less of a problem, as **B** will essentially only do one die roll per save. 
    * This points to a possible practise where players _always_ exchange logs when ever a die roll - or similar - is done. 

Yours,
_Christian_

PR #15007, opened 2026-08-21
Fix to VASSAL.sh script and add server option

cholmcc:
This patches makes sure we quite possible file names, so that we correctly parse that out.  Also add the command line option `--server` to start a local Vassal server.

For local Vassal servers, then the option `-URL` will set the URL to report games to.  Also for local servers, the option `-port` sets the port that the localhost server will listen to.

### DIFF
--- a/dist/linux/vassal.6
+++ b/dist/linux/vassal.6
@@ -113,6 +113,16 @@ Create a new module.
 .BI \-\-new\-extension \ module
 Create a new module extension to \fImodule\fP. 
 .TP
+.B \-\-server
+Start a local Vassal server.
+.TP
+.BI \-URL \ URL
+Set the server reporting URL.  Statistics on games will be sent to the
+URL give.
+.TP
+.BI \-port \ PORT
+Set the port number (default 5050) for the server to listen on.
+.TP
 .B \-\-version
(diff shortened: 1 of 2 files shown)


delatbabel:
We're holding a WiF convention next year in Thailand (wifcon.asia). I believe that this change would be beneficial if it were adopted so that the players at the convention could take advantage of the changes to more easily setup a local LAN server.

dist/linux/vassal.6
```diff
@@ -113,6 +113,16 @@ Create a new module.
 .BI \-\-new\-extension \ module
 Create a new module extension to \fImodule\fP. 
 .TP
+.B \-\-server
+Start a local Vassal server.
+.TP
+.BI \-URL \ URL
+Set the server reporting URL.  Statistics on games will be sent to the
+URL give.
```
uckelman, line 121: given?

uckelman:
The quote fix and adding the server options are unrelated, so should be in separate PRs.

cholmcc:
This PR has been replaced by the PRs https://github.com/vassalengine/vassal/pull/15038 and https://github.com/vassalengine/vassal/pull/15037

PR #15033, opened 2026-09-01
Deflate the saved-game command log before obfuscating it

delatbabel:
ObfuscatingOutputStream hex-encodes two ASCII chars per byte, so the ZIP was asked to compress a 2x-inflated 16-symbol stream and could do little with it. Compressing first (Deflater.BEST_COMPRESSION) and obfuscating the small result cuts a 34 MB save to roughly 19 MB.

New-format streams are marked !VCSZ; DeobfuscatingInputStream reads both the new format and the old !VCSK (undeflated) format, and still passes plain text through unchanged. The .vsav remains an ordinary ZIP: the change is confined to the savedGame entry's payload.

One consequence worth stating: saves and logs written by a build with this change cannot be opened by older VASSAL versions (they'll see !VCSZ as neither obfuscated nor valid plaintext). New builds read everything. You may want to gate the new writer behind a module/save version check — that would be your call in review.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -38,7 +38,7 @@ jobs:
         run: sudo apt-get install dos2unix genisoimage ghostscript nsis
 
       - name: Install Java
-        uses: actions/setup-java@v5.7.0
+        uses: actions/setup-java@v6.0.0
         id: setup-java
         with:
           distribution: 'temurin'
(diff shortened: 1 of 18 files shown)


uckelman:
This change will cause save data to be compressed twice, once by the new code, and again when writing the save file. 

I'm not convinced that the space savings is worth it.

delatbabel:
The second deflate is effectively a no-op pass-through with little CPU usage. The overall impact is a signifiant reduction in vsav file size (19MB vs 34MB in our test games) so I think it's worth having.

uckelman:
You've somehow picked up a bunch of commits which appear to be from master.

delatbabel:
Yeah I'm not sure how that happened. The original PR was against master. I rebased it to pick up the fix for the debian build being broken. Let me see if I can recreate it.

PR #15034, opened 2026-09-01
Add File > Download Module from Library to the Module Manager

delatbabel:
Downloads a module and its extensions from the VASSAL game library, given the library page URL (or bare project name) — pasted from the search on https://vassalengine.org/library/projects, which the prompt points at. The flow: pick the project, pick the module when the project publishes more than one (newest release listed first), pick a folder (write access is probed by creating a file, since File.canWrite ignores ACLs on Windows), then optionally pick a .vsav so only the extensions the scenario names are fetched. Extensions land in the conventional <module>_ext sibling folder.

The library talks JSON over its public API (api/gls/v1). The newest copy of each file is chosen per file, not per release — a release is a batch upload, so the same extension can appear in several and most only in old ones. Download URLs get their paths percent-encoded (module filenames contain spaces, which HttpURLConnection passes through raw and the object store rejects with HTTP 400); downloads are written to a .part file, SHA-256 verified, and moved into place. The JSON reader is minimal and hand-rolled to avoid a new dependency for one read-only document.

A scenario's extension names are read by streaming the savedGame entry through DeobfuscatingInputStream and keeping only EXT commands, so memory stays bounded for arbitrarily large saves.

Ported from the vassal_extension_utility project, where this feature and its failure modes (placeholder modules, per-release duplicates, unwritable folders, unencoded URLs) were worked out against the WiF library project.


delatbabel:
This would be a nice to have for 3.8.x

PR #15040, opened 2026-09-01
Set the locale for JOptionPane so the predefined buttons use locale-specific labels

uckelman:


vassal-app/src/main/java/VASSAL/i18n/Resources.java
```diff
@@ -377,6 +378,7 @@ private void setInstanceLocale(Locale l) {
     UIManager.put("OptionPane.cancelButtonText", getInstanceString(CANCEL)); //$NON-NLS-1$
     UIManager.put("OptionPane.noButtonText", getInstanceString(NO)); //$NON-NLS-1$
     UIManager.put("OptionPane.okButtonText", getInstanceString(OK)); //$NON-NLS-1$
+    JOptionPane.setDefaultLocale(locale);
```
rlament, line 381: The `setDefaultLocale` has no effect here. The `UIManager.put` statements prior to it override the  `setDefaultLocale`.

Try `JOptionPane.setDefaultLocale(new Locale("de"));` Start a game and immediately quit it with a File->Quit. The "Save Game" buttons are in the default locale, not German. Comment out the 4 lines of `UIManager.put` statements. Then the "Save Game" buttons are German.

Issue #14489, opened 2026-02-11
MacOS does not open module, save, or log when double-clicked in Finder or similar.


cholmcc:
[Possible solution](https://stackoverflow.com/questions/1582987/java-application-doesnt-know-name-of-file-being-opened-mac-os-x)

The `VASSAL.sh` script does not receive the file name when a Vassal file is double-clicked in Finder, or similar, or when doing 
```
$ open Module.vmod
$ open -a VASSAL --args Module.vmod
```
or similar. 

Yours,
_Christian_

cholmcc:
See also OpenJDK bug: https://bugs.openjdk.org/browse/JDK-8378934

PR #15042, opened 2026-09-01
Update MacOS launch script to allow more options

cholmcc:
- This patch brings the MacOS launch script into a closer match to the Linux launch script.
- It will allow one to pass JVM options via `-D<variable>=<value>`, e.g., to set the Swing Look'n'Feel.
- Allows for direct execution of `Editor`, `Player`, or as server.
- Explicit server option.
- Use full path name of any file passed on the command line.
- Run local server via the option `--server`, and possibly the options `-URL` and `-port`.

For example, one can do

    /Applications/VASSAL.app/Contents/MacOS/VASSAL.sh -l Foo.vmod

and Vassal will correctly find the `Foo.vmod` module, and similar for saves and log.  Also, one can do

    /Applications/VASSAL.app/Contents/MacOS/VASSAL.sh --server

to start a local server.

### DIFF
--- a/dist/macos/VASSAL.sh
+++ b/dist/macos/VASSAL.sh
@@ -1,23 +1,138 @@
-#!/usr/bin/env bash -e
+#!/bin/bash
+#
+# --- Find the path to the installation ------------------------------
+path="$(dirname "$0")/../.."
 
-# get the root directory of the bundle and go there
-APP_ROOT="$(dirname "$0")/../.."
-cd "$APP_ROOT"
+# --- Set default definitions and arguments --------------------------
+defs=("-Duser.home=$HOME"
+      "-Duser.dir=$path"
+      "-Xdock:name=VASSAL"
+      "-Xdock:icon=Contents/Resources/VASSAL.icns")
+args=()
 
-# filter out the -psn_* option added by Finder
-ARGS=()
-for ARG in "$@"; do
-  if [[ "$ARG" != -psn_* ]]; then
-    ARGS+=($(printf '%q' "$ARG"))
-  fi
-done
-
-# determine the correct CPU architecture
+# --- Architecture ---------------------------------------------------
 if sysctl machdep.cpu.brand_string | grep -q Intel ; then
-  ARCH=x86_64
+    arch=x86_64
 else
-  ARCH=arm64
+    arch=arm64
 fi
 
-# fire it up
-exec arch -$ARCH Contents/MacOS/jre/bin/java -classpath Contents/Resources/Java/Vengine.jar -Xdock:name=VASSAL -Xdock:icon=Contents/Resources/VASSAL.icns VASSAL.launch.ModuleManager "${ARGS[@]}"
+# --- Find Java and debugger -----------------------------------------
+java=Contents/MacOS/jre/bin/java
+cmd="$java"
+
+# --- The various entry points and the JAR ---------------------------
+ver_entry=VASSAL.launch.JavaVersionChecker
+mod_entry=VASSAL.launch.ModuleManager
+ply_entry=VASSAL.launch.Player
+edt_entry=VASSAL.launch.Editor
+trl_entry=VASSAL.i18n.TranslateVassalWindow
+srv_entry=VASSAL.chat.node.Server
+entry=$mod_entry
+jar="Contents/Resources/Java/Vengine.jar"
+srv_url=null
+
+# --- Some modus operandi --------------------------------------------
+# Do we do direct execution?, i.e., by-pass the module manager. Are we
+# debugging?  Possible source path for when debugging, e.g.,
+# ~/vassal/vassal-app/src/main/java if the VASSAL sources are cloned
+# to ~/vassal
+drt=0
+src=
+hlp=0
+
+# --- Parse command line ---------------------------------------------
+while test $# -gt 0 ; do
+    case x$1 in
+        x-h|x--help)
+            hlp=1
+            args+=("--help")
+            ;;
+        x-D*)
+            # Defines must come before JAR and entry point 
+            defs+=("$1")
+            ;;
+        x--direct)
+            # If passed, then we by-pass the module manager and
+            # execute the relevant entry point directly.
+            drt=1
+            ;;
+        x--translate)
+            entry=$trl_entry
+            ;;
+        x-l|x--load)
+            args+=("--load")
+            if test $drt -gt 0 ; then
+                entry=$ply_entry
+            fi
+            ;;
+        x-e|x--edit)
+            args+=("--edit")
+            if test $drt -gt 0 ; then
+                entry=$edt_entry
+            fi
+            ;;
+        x-n|x--new)
+            args+=("--new")
+            if test $drt -gt 0 ; then
+                entry=$edt_entry
+            fi
+            ;;
+        x--new-extension)
+            args+=("$1")
+            if test $drt -gt 0 ; then
+                entry=$edt_entry
+            fi
+            ;;
+        x--edit-extension)
+            args+=("$1")
+            if test $drt -gt 0 ; then
+                entry=$edt_entry
+            fi
+                        ;;
+        x--server)
+            entry=$srv_entry
+            ;;
+        *)
+            # If argument is a file ... 
+            if test -f "$1" ; then
+                # ... then store full path name since VASSAL changes
+                # the current directory before opening target files.
+                f=$(readlink -f "$1")
+                args+=("$f")
+            else
+                args+=("$1")
+            fi
+            ;;
+    esac
+    shift
+done
+
+# --- Go to application directory ------------------------------------
+cd $path
+
+# --- Check server ---------------------------------------------------
+if test "x$entry" == "x$srv_entry" ; then
+    args+=("-URL" "$srv_url")
+fi
+
+# --- Run java with defines, entry point, and other arguments --------
+arch -${arch} "${cmd}" "${defs[@]}" -classpath "${jar}" "$entry" "${args[@]}"
+
+# --- Extra help -----------------------------------------------------
+if test $hlp -gt 0 ; then
+    cat <<-EOF
+	In addition to the above options defined by VASSAL proper,
+	this wrapper script allows for a number of additional options.
+
+	Options:
+	  -Dvariable=value	  Set Java variable to value
+	  --direct		  By-pass the module manager and run
+	                          player, editor, ..., directly.
+	  --server                Start local Vassal server
+	EOF
+fi
+#
+# EOF
+#
+


dist/macos/VASSAL.sh
```diff
@@ -1,23 +1,140 @@
...
-# determine the correct CPU architecture
+# --- Architecture ---------------------------------------------------
 if sysctl machdep.cpu.brand_string | grep -q Intel ; then
-  ARCH=x86_64
+    arch=x86_64
 else
-  ARCH=arm64
+    arch=arm64
+fi
+arch=`arch`
```
uckelman: This can't be right. You're assigning to `arch`, then throwing away the value without using it.

cholmcc: The MacOS command `arch` reports the architecture - which is typically `x86_64` or `arm64`, so it is enough to do 
```
arch=`arch`
```
Updated the PR.

uckelman: Have you tested this? It was a long time ago that I wrote the original script, but I believe the reason for using `systcl` to get the CPU architecture is that `arch` doesn't work correctly.

cholmcc: Yes, I've been using a script like the updated one in this PR on a Mac for quite some time - no problems seen.  I've not been able to test it on other chips than the one I have (`arm64`), but looking at the man page of [`arch`](https://www.unix.com/man_page/osx/1/arch/), I see no reason why it shouldn't work. 

Actually, it seems like the call is largely superfluous - the OS should run the best possible version of the universal binary on its own, and it doesn't seem like there's a need to coerce the OS to run a specific architecture.  But, maybe I'm missing something?

Yours,
_Christian_

uckelman: Ah, right: Here is the explanation for why doing it this way is wrong: https://forum.vassalengine.org/t/how-to-bundle-java-in-a-universal-macos-app/77991

cholmcc: [I just tried on my MacOS machine to completely remove the `arch` calls from the script so that what is essentially executed is 

```
cd  /Applications/VASSAL.app
Contents/MacOS/jre/bin/java -cp Contents/Resources/Java/Vengine.jar VASSAL.launch.ModuleMananger
```

and that works without a hitch.  I checked that it is running native as "Apple" - not via Rosetta.  

Also, running `file` on `/Applicatons/VASSAL.app/Contents/MacOS/jre/bin/java` tells me that the binary is "universal" with two architectures `x86_64` and `arm64`.  

Also, what you [linked to above](https://forum.vassalengine.org/t/how-to-bundle-java-in-a-universal-macos-app/77991)  does not argue why it would be wrong to use `arch` to discover the architecture - only that you may need to run 
```
arch -${arch} Contents/MacOS/jre/bin/java  ...
```
to not end up under Rosetta. 

The current PR will still do 
```
arch -${arch} Contents/MacOS/jre/bin/java  ...
```
but use `arch` to discover the architecture. 

Yours,
_Christian_

uckelman: I suspect that `arch` run from a shell script will provide the wrong answer in some cases, because the shell script will already have the wrong answer.

cholmcc: First off, note that 
``` 
sysctl machdep.cpu.brand_string | grep -q Intel
```
isn't fool proof.  You are looking for the string `Intel` in the return, but that does not necessarily distinguish between `x86_64` or `i386`. 

Secondly, I would find it very odd that `arch` would report different values if executed from a script.  The man page of [`arch`](https://www.unix.com/man_page/osx/1/arch/)(1) _does_ say that if the environment variable `ARCHPREFERENCE` is set - which it isn't by default - then that _may_ change the what `arch` will do - but _not_ when reporting the architecture - I just tried:
```
$ ARCHPREFERENCE=x86_64 arch
arm64
```
 The other reason why `arch` would report a different architecture would be if the shell executed program is spawned within the same environment.   I tried - without the explicit `arch` call 
```
$ arch -x86_64 bash
# arch
i386
# /Applications/VASSAL.app/Contents/MacOS/VASSAL.sh (runs as Intel, so does spawned processes)
# open -a VASSAL (runs as Apple - this is what the OS does from GUI launcher)
```
but note that this requires that the shell was executed as a different architecture, which is highly unlikely.   Even without going through the terminal, the JVM will still run as `Apple` (Natively). 

In other words, only under _very peculiar_ circumstances will executing the JVM without `arch` be run non-native.  A user may have very good reasons for doing so.    

Yours,
_Christian_

uckelman: >  You are looking for the string Intel in the return, but that does not necessarily distinguish between x86_64 or i386.

We don't need to distinguish between x86_64 and i386 because there has never been a 32-bit Intel Mac.

uckelman: Wait, no, the very first Intel Macs were 32-bit, but you can't run Java 26 on those, so, 32-bit CPUs still don't matter.

cholmcc: > , 32-bit CPUs still don't matter.

Even so, the query of `mach.cpu.brand_string` isn't a robust way.  

But, whether it is or not, the real point is that there's really no need to call `/Application/VASSAL.app/Contents/MacOS/jre/bin/java` with the `arch` context.  In almost all cases, _except_ in very _peculiar_ situations that the user likely set up intentionally, the OS will do The Right Thing(tm). 

That said, the main point of this PR is not how to call the `java` executable, but rather give users the option to pass more options to the Vassal process and to allow users to specify files in the current directory in a natural fashion.  With that in mind, I have no particular problem - except that it is unneeded and a complication that need to be maintained - with changing the `arch` discovery back to the `mach.cpu.brand_string` query if that is what you want - neither approach solves the much more important issue https://github.com/vassalengine/vassal/issues/14489

yours,
_Christian_

uckelman: > In almost all cases, except in very peculiar situations that the user likely set up intentionally, the OS will do The Right Thing(tm).

We had evidence against this in 2023. Some people were finding that Java was running via Rosetta on Apple Silicon machines.

cholmcc: > We had evidence against this in 2023.

I'm sure you did.  I just couldn't reproduce it unless I imposed very specific circumstances. 

So tell me what you want: 
- Do you want to execute `/Application/VASSAL.app/Contents/MacOS/jre/bin/java` in the context of `arch`?, and if so
- do you want to use `sysctl mach.cpu.brand_string | grep -q Intel` or `arch` to discover the architecture?

Let me know, and I'll update the PR. 

BTW, perhaps `dist/linux/VASSAL.sh`, `dist/macos/VASSAL.sh`, and `dist/VASSAL.sh` should be consolidated into a single script. 

Relevant parts would be something like 

```
# --- Java from environment --------------------------------------
java=${JAVA:-$(command -v java 2>/dev/null)}
jdb=${JDB:-$(command -v jdb 2>/dev/null)}

# --- Application directory ------------------------------------------
path=$(realpath "$0")
path=$(dirname "$path")

# --- OS specific stuff -------------------------------------------------
OS=${uname)
case x$OS in 
    Darwin)
        path="${path}"/../.."
        if test "x$java" = "x" ; then
            java=Contents/MacOS/jre/bin/java
        fi
        pre="arch -$(arch)"
        ;;
    *)
       # --- Check for Java -------------------------------------------------
       if test ! -x "$java" ; then
           cat <<-EOF
	$0: Java not installed.  Try to install it with f.ex.

	   sudo apt install default-jre

	EOF
            exit 1
        fi

        # --- Check for Java version -----------------------------------------
        if ! "$java" -cp "$jar" $ver_entry 2>/dev/null ; then
            cat <<-EOF
	$0: Java installation too old to run VASSAL.  Please upgrade it
        with for example

           sudo apt upgrade default-jre

	EOF
            exit 1
        fi
        pre=
        ;;
esac
cmd="$java"
```

and later on 

```
if test "x$cmd" = "x" || test ! -x $cmd ; then
    if test dbg -gt 0 ; then
        echo "Do not know how to run the Java debugger"
        exit 1
    fi
    echo "Unknown interpreter to run Java code"
    exit 1
fi
```

and finally 

```
# --- Go to application directory ------------------------------------
cd "$path"

# --- Run java with defines, entry point, and other arguments --------
$pre "${cmd}" "${defs[@]}" -classpath "${jar}" "$entry" "${args[@]}"
```

Yours,
_Christian_

uckelman: I do not want to change the way we're detecting the architecture without a compelling reason for doing so. It was difficult to get working, and has worked for years.

I do not want to combine the platform-specific scripts without a compelling reason for doing so. Changes to the run scripts since March introduced bugs, when we had no reported problems with those scripts; I'd prefer not to repeat that without a clear view about the gains we expect from it.

There is a clear case for solving #14489, so make the changes necessary for that.

(There are a lot of Automatic Bug Reports going unaddressed, and those do have a compelling reason to dig into them. They stymie users and contribute to our support burden. That's where I'd put my time were I working on V3.)

cholmcc: > I do not want to change the way we're detecting the architecture ...

I reverted the change so that the architecture is discovered using `sysctl mach.cpu.brand_string | grep -q Intel`

Java is executed in the `arch` context. 

That should address the concerns around _this_ PR. 

> I do not want to combine the platform-specific scripts without a compelling reason for doing so.

The reason would be to minimize maintenance and stream-line the various launch scripts (`dist/VASSAL.sh`, `dist/macos/VASSAL.sh`, and `dist/linux/VASSAL.sh`).  When adding features, etc. they will be available across distributions, for example. 

>  Changes to the run scripts since March introduced bugs 

Those bugs where already there before - they just weren't discovered because of the oddities of the old script.  "If it hasn't got any bugs, it hasn't got enough features" :smile: 

> There is a clear case for solving https://github.com/vassalengine/vassal/issues/14489, so make the changes necessary for that.

I think it is fairly obvious that the problem is a bug in OpenJDK - not in Vassal.  Vassal has the correct code, but the underlying JRE does not do the right thing when Java is executed from a script (see also [the post you linked to](https://incenp.org/notes/2023/universal-java-app-on-macos.html) - the author reverts to a compiled C launcher - exactly so that `java.awt.Desktop` call-backs are triggered correctly. 

There is an open [OpenJDK bug report](https://bugs.openjdk.org/browse/JDK-8378934) - but not much seems to go on with that one. 

> There are a lot of Automatic Bug Reports going unaddressed, and those do have a compelling reason to dig into them.

From scanning over [those bugs]( is:issue state:open label:ABR) it seems to me that many of these stem from outdated custom code - and a lot of them are redunant.  Perhaps a way of forwarding those bugs to the relevant project owners would elevate the burden (the project owners most likely only need to recompile their custom code). 

Yours,
_Christian_

dist/macos/VASSAL.sh
```diff
@@ -1,23 +1,138 @@
...
+                        ;;
+        x--server)
+            entry=$srv_entry
+            ;;
+        *)
+            # If argument is a file ... 
+            if test -f "$1" ; then
+                # ... then store full path name since VASSAL changes
+                # the current directory before opening target files.
+                f=$(realpath "$1")
```
uckelman, line 101: `realpath` wasn't installed by default until MacOS 13. The last version of MacOS 12 was released in 2024; there could still be users who don't have `realpath`.

cholmcc: Sigh!  How can MacOS ever be considered serious.   Ah well.  

Made a new update - in case `realpath` isn't known, use a `poormans_realpath` implementation.  Tested it on MacOS.  Looks good as far as I can tell.

uckelman: https://github.com/WoozyMasta/kube-dump/issues/26 shows how I've seen this handled before.

cholmcc: > [WoozyMasta/kube-dump#26](https://github.com/WoozyMasta/kube-dump/issues/26) shows how I've seen this handled before.

That solution is overkill for the purpose here.   The point is _not_ to get the _canonical_ path with all symlinks resolved.  The point is to get an _absolute_ path to the file, so that when the script `cd`s to the Vassal directory, that the file can still be found.  The current proposal does exactly that, and no more. 

Yours,
_Christian_

uckelman: I believe `readlink -f` will work for giving you an absolute path. The `-f` flag was added in MacOS 12.3, so anyone using MacOS 12.2 or less can upgrade.

cholmcc: OK, just checked it.  `readlink -f` does indeed do what we want - namely give the absolute path.  Note, resolving symbolic links is just a perk :smile:

Did a new push. 

Yours,
_Christian_

dist/macos/VASSAL.sh
```diff
@@ -1,23 +1,138 @@
...
-APP_ROOT="$(dirname "$0")/../.."
-cd "$APP_ROOT"
-
-# filter out the -psn_* option added by Finder
-ARGS=()
-for ARG in "$@"; do
-  if [[ "$ARG" != -psn_* ]]; then
-    ARGS+=($(printf '%q' "$ARG"))
-  fi
-done
```
uckelman, line 13: Why has this been removed?

cholmcc: It does not seem to come up anymore.    That is, I've not seen the script be passed those options at any point.

cholmcc: I mean, it is easy enough to add back in 
```
case x-psn_*)
    ;;
```

cholmcc: The `-psn_` options seems to have disappeared in MacOS 10.9 (see [this thread](https://apple.stackexchange.com/questions/207895/app-gets-in-commandline-parameter-psn-0-nnnnnn-why#207917)).   Since `readlink -f` requires MacOS 12.2 ([Monterey - from January 2022](https://en.wikipedia.org/wiki/MacOS_Monterey)), it seems that we no longer need to take care of such command line options.

PR #15044, opened 2026-09-02
Deflate the saved-game command log before obfuscating it

delatbabel:
ObfuscatingOutputStream hex-encodes two ASCII chars per byte, so the ZIP was asked to compress a 2x-inflated 16-symbol stream and could do little with it. Compressing first (Deflater.BEST_COMPRESSION) and obfuscating the small result cuts a 34 MB save to roughly 19 MB.

New-format streams are marked !VCSZ; DeobfuscatingInputStream reads both the new format and the old !VCSK (undeflated) format, and still passes plain text through unchanged. The .vsav remains an ordinary ZIP: the change is confined to the savedGame entry's payload.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
+++ b/vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
@@ -24,10 +24,13 @@ import java.io.PushbackInputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.zip.InflaterInputStream;
 
 /**
  * A {@link FilterInputStream} which converts a file created with
- * {@link ObfuscatingOutputStream} back into plain text.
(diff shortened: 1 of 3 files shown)


uckelman:
Having thought about this a bit, the problem could be solved in a simpler way: 

Writing the data out in hex after XORing is what doubles the size, but it's also pointless. The purpose of all of this is to prevent _casual_ cheating, and the data is already unreadable---in fact, _less_ readable---before hex conversion.

A simpler solution would be:

* Use a new header. (E.g., `VOBS`. The leading bang is also pointless, and 4-byte magic in headers is more typical.)
* Don't convert anything to hex at all.
* Choose the XOR byte using `rand.nextInt(255) + 1`, so that we never XOR with 0.

A lot of `ObfuscatingInputStream` can then be deleted, the `DeobfuscatingInputStreamImpl` can be renamed to something like `LegacyDeobfuscatingInputStreamImpl` and a new, simpler `DeobfuscatingInputStreamImpl` be made to handle the new output.

delatbabel:
Replaced by #15060

PR #15048, opened 2026-09-03
Consolidate launcher scripts

cholmcc:
- The launcher scripts
  - `dist/VASSAL.sh`
  - `dist/macos/VASSAL.sh`
  - `dist/linux/VASSAL.sh` are consolidated into a single script `dist/VASSAL.sh`

  This is to ease maintenance.

  This PR is a suggestion.  It will need a little bit of testing
  before deployment.


cholmcc:
I ran some tests of the script and found some errors.   The new script should be relatively robust and close to what could be deployed.

cholmcc:
> I ran some tests of the script

I ran the tests on MacOS and Linux.

uckelman:
I'm closing this. Combining the platform-specific run scripts will make them harder to maintain.

cholmcc:
> Combining the platform-specific run scripts will make them harder to maintain.

That we disagree on.   MacOS and Linux (as well as BSD and Un*x) are close enough that it would make sense.  Why not simply change it in to a WIP?

PR #15054, opened 2026-09-06
14906 Fix Truncated text label on hidden pieces

### LINKED ISSUE #14906: Truncated text label on hidden pieces starting with 3.7.16
As reported in the VASL forum: https://www.gamesquad.com/forums/index.php?threads/vasl-draggable-labels-and-hip.208295/#post-2221772

Steps to reproduce:

1. Start a game.
2. Drag a Label from the Draggable Overlays window (the button for which is in the button bar) on to the map. 
3. Edit the label to be very long.
4. Set HIP on the label (which hides the piece).

The breaking commit is e792bc19.

@BrentEaston @riverwanderer you were involved in that commit. Is the behavior change intentional?

rlament:
This is somewhat of a chicken and egg problem. The label size is computed in the `draw()` function. Since the size of the `BufferedImage` is based on the previous rendering, it can be too small for the new image. The proposal here is to recheck the image bounds, and resize the buffer and redraw if the bounds exceed the buffer size.

By moving where the background is rendered, the background now honors the opacity setting.

The adjustments to the translation and scaling remove the pixel shift when zoom is anything other than 100%.

Closes #14906

### DIFF
--- a/vassal-app/src/main/java/VASSAL/counters/Hideable.java
+++ b/vassal-app/src/main/java/VASSAL/counters/Hideable.java
@@ -68,11 +68,6 @@ public class Hideable extends Decorator implements TranslatablePiece {
   protected KeyCommand hideCommand;
   protected String description = "";
 
-  // Cache the Hidden image that is shown to the owner. Only re-generate when Zoom or piece state changes.
-  protected BufferedImage cachedImage;
-  protected double cachedZoom = -1d;
-  protected String cachedState = "";
-
   @Override
   public void setProperty(Object key, Object val) {
     if (HIDDEN_BY.equals(key)) {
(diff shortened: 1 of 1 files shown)


vassal-app/src/main/java/VASSAL/counters/Hideable.java
```diff
@@ -208,51 +203,58 @@ public void draw(Graphics g, int x, int y, Component obs, double zoom) {
...
+          // After drawing, the in-memory image buffer is too small.
+          // Resize the buffer and redraw.
+          // This can happen when a label's length increases.
+          bounds = boundsCheck;
+          cachedImage = ImageUtils.createCompatibleTranslucentImage(wc, hc);
           cg.dispose();
+          cg = cachedImage.createGraphics();
+          cg.translate((int)(-bounds.x * zoom), (int)(-bounds.y * zoom));
+          piece.draw(cg, 0, 0, obs, zoom); // Fully opaque
         }
```
uckelman, line 237: * This appears to be creating an image just to draw it. If we're not going to cache the image, why not just draw directly?

* I presume Brent introduced caching because it made a performance difference for something he saw. Isn't there a way we can maintain it?

rlament: Came up with a simpler solution. Thanks for the probing questions. It's likely the caching is there because `draw()` is called multiple (at least twice) times. Instead of explicitly checking the image bound and redrawing, this version caches the image bounds rectangle. In addition to the other cache checks, if the image bounds change, then the image is regenerated with the new bounds.

vassal-app/src/main/java/VASSAL/counters/Hideable.java
```diff
@@ -208,51 +209,51 @@ public void draw(Graphics g, int x, int y, Component obs, double zoom) {
...
-        t.translate(x / zoom, y / zoom);
-        g2d.fill(t.createTransformedShape(piece.getShape()));
-      }
 
       // Determine piece bounds at current zoom
       final Rectangle bounds = piece.getShape().getBounds();
-      final int w = (int) Math.ceil(bounds.width * zoom);
-      final int h = (int) Math.ceil(bounds.height * zoom);
+      final int w = (int) Math.round(bounds.width * zoom);
+      final int h = (int) Math.round(bounds.height * zoom);
```
uckelman, line 216: Why change from `ceil` to `round` here?

vassal-app/src/main/java/VASSAL/counters/Hideable.java
```diff
@@ -208,51 +209,51 @@ public void draw(Graphics g, int x, int y, Component obs, double zoom) {
...
-      final int drawY = y + (int) Math.round(bounds.y * zoom);
+      if (bgColor != null) {
+        g2d.setColor(bgColor);
+        final AffineTransform t = AffineTransform.getTranslateInstance(x, y);
+        t.scale(zoom, zoom);
+        g2d.fill(t.createTransformedShape(piece.getShape()));
+      }
+
+      final int drawX = x + (int) (bounds.x * zoom);
+      final int drawY = y + (int) (bounds.y * zoom);
```
uckelman, line 256: Why change from `round` to truncating here?

vassal-app/src/main/java/VASSAL/counters/Hideable.java
rlament: Changed it to `round` because of this check immediately following:
```
 // If there's nothing visible at this zoom, skip
      if (w <= 0 || h <= 0) {
```
With `ceil` if the width, height and zoom are non-zero positive numbers that `if` statement is always false. The minimum value of `w` and `h` is 1. At least with rounding if zoomed out enough it will exit the function.

This does affect the size of the `cachedImage`. Not that I've noticed. We could add some padding as the size does not have to be exact. Something like: `createCompatibleTranslucentImage(w+1, h+1)`

vassal-app/src/main/java/VASSAL/counters/Hideable.java
rlament: To be consistent with other drawing functions. In Labeler.draw:
```
   final int labelX = x + (int) (zoom * p.x);
   final int labelY = y + (int) (zoom * p.y);
```
or BasicPiece
`imagePainter.draw(g, x + (int) (zoom * imageBounds.x), y + (int) (zoom * imageBounds.y), zoom, obs);`

This solves the pixel shift that occurred at various zoom levels.

PR #15059, opened 2026-09-07
Allow user to set Java Look and Feel

cholmcc:
This will allow the user to set the Look and Feel (LnF) of the
graphical user interface.   The following changes are done

- The current value of `swing.defaultlaf` is propagated on to child processes. That means, when the `ModuleManager` spawns the `Player` or `Editor` (and other processes), that those child-processes will also get the current setting for the LnF.
- If no LnF is specified by the user, then the default is to use the system LnF.  Note, on Windows, the default was to switch to the system independent _Metal_ LnF, while for MacOS and Linux, the default LnF was always the system LnF.
- Adds a new configurer class `LookAndFeelConfigurer` which are added
  to the shared global preferences.   This allows a user to specify
  the LnF via _File_&rarr;_Preferences_&rarr;_General_&rarr;_Look and
  feel_, even at runtime.  The setting is of course stored in
  `V_Global`, and will therefore be used for subsequent launches.

This PR is somewhat at odds with the PR #14435 and they should not _both_ be applied.  Note that #14435 attempts to enable third-party provided LnFs, while this PR will only use those already provided (could in principle be third party too).

Screenshots of preferences: 

ModuleManagerPrefs
PlayerPrefs

### DIFF
--- a/vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
+++ b/vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
@@ -1,0 +1,170 @@
+/*
+ *
+ * Copyright (c) 2026 by Christian Holm Christensen
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License (LGPL) as published by the Free Software Foundation.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU Library General Public
+ * License along with this library; if not, copies are available
+ * at http://www.opensource.org.
+ */
+package VASSAL.configure;
+
+import VASSAL.tools.ErrorDialog;
+
+import java.awt.Dimension;
+import java.awt.Frame;
+import java.awt.Window;
+import java.awt.event.ItemListener;
+
+import javax.swing.BoxLayout;
+import javax.swing.JComboBox;
+import javax.swing.JFrame;
+import javax.swing.JPanel;
+import javax.swing.SwingUtilities;
+import javax.swing.UIManager;
+import javax.swing.UIManager.LookAndFeelInfo;
+
+
+/**
+ * A Configurer for {@link javax.swing.UIManager.LookAndFeel} values
+ */
+public class LookAndFeelConfigurer extends Configurer {
+  public static class Info extends LookAndFeelInfo {
+    public Info(String name, String clsName) {
+      super(name, clsName);
+    }
+    public Info(LookAndFeelInfo info) {
+      this(info.getName(), info.getClassName());
+    }
+    @Override
+    public String toString() {
+      return getName();
+    }
+  }
+        
+  private JPanel panel;
+  private JComboBox<Info> laf;
+
+  public LookAndFeelConfigurer(String key, String name) {
+    this(key, name,
+         new Info(UIManager.getLookAndFeel().getName(),
+                  UIManager.getLookAndFeel().getClass().getName()));
+  }
+
+  public LookAndFeelConfigurer(String key, String name, LookAndFeelInfo val) {
+    super(key, name, new Info(val));
+  }
+
+  protected Info getInfo() {
+    return (Info)value;
+  }
+  
+  @Override
+  public String getValueString() {
+    return encode((Info) value);
+  }
+
+  @Override
+  public void setValue(String s) {
+    setValue(decode(s));
+    updateLnF();
+  }
+
+  @Override
+  public java.awt.Component getControls() {
+    if (panel == null) {
+      panel = new ConfigurerPanel(getName(), "[]rel[]", "[]rel[]"); // NON-NLS
+      laf = new JComboBox<>();
+      final LookAndFeelInfo[] known = UIManager.getInstalledLookAndFeels();
+      final String sysClsName = UIManager.getSystemLookAndFeelClassName();
+      Info sysInfo    = null;
+      Info current    = null;
+      for (final LookAndFeelInfo info : known) {
+        final Info i = new Info(info);
+        laf.addItem(i);
+        if (value != null &&
+            getInfo().getClassName().equals(i.getClassName())) {
+          current = i;
+        }
+        if (i.getClassName().equals(sysClsName)) {
+          sysInfo = i;
+        }
+      }
+      laf.setSelectedItem(current == null ? sysInfo : current);
+      laf.setMaximumSize(new Dimension(laf.getMaximumSize().width,
+                                       laf.getPreferredSize().height));
+      panel.add(laf);
+
+      final ItemListener l = evt -> setValue(laf.getSelectedItem());
+      laf.addItemListener(l);
+    }
+    return panel;
+  }
+
+  public static Info decode(String s) {
+    final int i = s.indexOf(',');
+    return new Info(s.substring(0, i), s.substring(i + 1));
+  }
+
+  public static String encode(Info i) {
+    return i.getName() + "," + i.getClassName();
+  }
+
+  protected void updateLnF() {
+    // Check for change 
+    if (getInfo().getClassName().equals(UIManager.getLookAndFeel().getClass().getName()))
+      return;
+
+    try {
+      UIManager.setLookAndFeel(getInfo().getClassName());
+      for (final Frame frame : Frame.getFrames()) {
+        if (frame == null)
+          continue;
+        SwingUtilities.updateComponentTreeUI(frame);
+
+        for (final Window window : frame.getOwnedWindows()) 
+          SwingUtilities.updateComponentTreeUI(window);
+      }
+    }
+    catch (Exception e) {
+      ErrorDialog.bug(e);
+    }
+  }
+  @Override
+  public void fireUpdate() {
+    super.fireUpdate();
+    updateLnF();
+  }
+  
+  @Override
+  public void setLabelVisible(boolean visible) {
+    if (panel instanceof ConfigurerPanel) {
+      ((ConfigurerPanel) panel).setLabelVisibility(visible);
+    }
+  }
+
+  public static void main(String[] args) {
+    final JFrame frame = new JFrame();
+    frame.setLayout(new BoxLayout(frame.getContentPane(), BoxLayout.Y_AXIS));
+    final LookAndFeelConfigurer config = new LookAndFeelConfigurer("a", "LaF: "); //NON-NLS
+    frame.add(config.getControls());
+    config.addPropertyChangeListener(evt -> {
+      final Info info = (Info)evt.getNewValue();
+      final LookAndFeelConfigurer fconfig
+        = new LookAndFeelConfigurer(null, null, info);
+      fconfig.setValue(fconfig.getValueString());
+      fconfig.fireUpdate();
+      frame.pack();
+    });
+    frame.pack();
+    frame.setVisible(true);
+  }
+}
--- a/vassal-app/src/main/java/VASSAL/launch/AbstractLaunchAction.java
+++ b/vassal-app/src/main/java/VASSAL/launch/AbstractLaunchAction.java
@@ -39,6 +39,7 @@ import VASSAL.tools.io.ProcessLauncher;
 import VASSAL.tools.io.ProcessWrapper;
 import VASSAL.tools.lang.MemoryUtils;
 import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.SystemUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
@@ -47,6 +48,7 @@ import org.slf4j.LoggerFactory;
 import javax.swing.AbstractAction;
 import javax.swing.JOptionPane;
 import javax.swing.SwingWorker;
+import javax.swing.UIManager;
 import java.awt.Dimension;
 import java.awt.Window;
 import java.awt.event.ActionEvent;
@@ -557,6 +559,17 @@ public abstract class AbstractLaunchAction extends AbstractAction {
       final String vConf = System.getProperty("VASSAL.conf");
       if (vConf != null) result.add("-DVASSAL.conf=" + vConf); //NON-NLS
 
+      // pass on Swing look'n'feel
+      String laf = System.getProperty("swing.defaultlaf");
+      if (laf == null) {
+        // Get the current l'n'f name from the UIManager
+        laf = UIManager.getLookAndFeel().getClass().getName();
+      }
+      if (!StringUtils.isBlank(laf)) {
+        // Pass on the property to child process
+        result.add("-Dswing.defaultlaf=" + laf);
+      }
+      
       // set the classpath
       result.add("-cp"); //NON-NLS
       result.add(System.getProperty("java.class.path"));
--- a/vassal-app/src/main/java/VASSAL/launch/StartUp.java
+++ b/vassal-app/src/main/java/VASSAL/launch/StartUp.java
@@ -28,6 +28,7 @@ import javax.swing.UnsupportedLookAndFeelException;
 
 import VASSAL.preferences.Prefs;
 import VASSAL.preferences.ReadOnlyPrefs;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.SystemUtils;
 
 import org.slf4j.Logger;
@@ -115,29 +116,41 @@ public class StartUp {
           );
         }
 
-        if (!SystemUtils.IS_OS_WINDOWS) {
-          // use native LookAndFeel
-          // NB: This must be after Mac-specific properties
+        // Get the property that sets the look'n'feel
+        String laf = System.getProperty("swing.defaultlaf");
+        final String sysLaf = UIManager.getSystemLookAndFeelClassName();
+        if (laf == null || StringUtils.isBlank(laf)) {
+          // Default to system l'n'f
+          laf = sysLaf;
+        }
+        // NB: This must be after Mac-specific properties
+        try {
+          // System.out.println("Attempting to set LaF to " + laf);
+          UIManager.setLookAndFeel(laf);
+        }
+        catch (ClassNotFoundException |
+               UnsupportedLookAndFeelException |
+               InstantiationException |
+               IllegalAccessException e) {
+          // System.out.println("Failed installing LAF: " + e);
+          // Fall back to system LAF in case of errors
           try {
-            UIManager.setLookAndFeel(
-              UIManager.getSystemLookAndFeelClassName()
-            );
-          }
-          catch (ClassNotFoundException | UnsupportedLookAndFeelException
-            | InstantiationException | IllegalAccessException e) {
+            UIManager.setLookAndFeel(sysLaf);
             ErrorDialog.bug(e);
           }
-
-          // The GTK LaF has a color picker which lacks the ability to
-          // select transparency. We can override that and it doesn't
-          // look too goofy.
-          if ("com.sun.java.swing.plaf.gtk.GTKLookAndFeel".equals(//NON-NLS
-            UIManager.getLookAndFeel().getClass().getName())) {
-            UIManager.put(
-              "ColorChooserUI",
-              "javax.swing.plaf.basic.BasicColorChooserUI"
-            );
+          catch (ClassNotFoundException |
+                 UnsupportedLookAndFeelException |
+                 InstantiationException |
+                 IllegalAccessException ee) {
           }
+        }
+
+        // The GTK LaF has a color picker which lacks the ability to
+        // select transparency. We can override that and it doesn't
+        // look too goofy.
+        if ("com.sun.java.swing.plaf.gtk.GTKLookAndFeel".equals(laf)) { //NON-NLS
+          UIManager.put("ColorChooserUI",
+                        "javax.swing.plaf.basic.BasicColorChooserUI");
         }
 
         // Ensure consistent behavior in NOT consuming "mousePressed" events
--- a/vassal-app/src/main/java/VASSAL/preferences/Prefs.java
+++ b/vassal-app/src/main/java/VASSAL/preferences/Prefs.java
@@ -23,6 +23,7 @@ import VASSAL.configure.BooleanConfigurer;
 import VASSAL.configure.Configurer;
 import VASSAL.configure.DirectoryConfigurer;
 import VASSAL.configure.IntConfigurer;
+import VASSAL.configure.LookAndFeelConfigurer;
 import VASSAL.i18n.Resources;
 import VASSAL.tools.ReadErrorDialog;
 
@@ -66,7 +67,8 @@ public class Prefs implements Closeable {
   public static final String OVERRIDE_DEFAULT_FONT_SIZE = "overrideDefaultFontSize"; //NON-NLS
 
   public static final String TRANSLATABLE_SUPPORT = "translatableSupport"; //NON-NLS
-
+  public static final String LOOK_AND_FEEL = "lookAndFeel"; //NON-NLS
+  
   private static Prefs globalPrefs; // A Global Preferences object
 
   private final Map<String, Configurer> options = new HashMap<>();
@@ -343,6 +345,12 @@ public class Prefs implements Closeable {
     );
 
     globalPrefs.addOption(Resources.getString("Prefs.general_tab"), auditConf);
+
+    final LookAndFeelConfigurer laf =
+      new LookAndFeelConfigurer(LOOK_AND_FEEL,
+                                Resources.getString("General.lookAndFeel"));
+    globalPrefs.addOption(Resources.getString("Prefs.general_tab"), laf);
+
   }
 
   public static String sanitize(String str) {
--- a/vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
+++ b/vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
@@ -61,6 +61,7 @@ General.run=Run
 General.copy=Copy
 General.cut=Cut
 General.paste=Paste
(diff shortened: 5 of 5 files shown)


cholmcc:
Last push fixed a few errors:
- `JDialog` objects would not have their LnF updated.  Now fixed. 
- The `JComboBox` for selecting the theme did not have the current value selected.

uckelman:
Why, at this point, should we do this?

cholmcc:
> Why, at this point, should we do this?

Because users have expressed an interest in it.  

I know your mind is on V4 - and that's all well and fine - but in the meantime, there's really only V3 to play with, so why not make it better and give users a bit more of what they want.   It does not hurt anything else, so the real question is more like why not :smile:

uckelman:
I haven't seen users clamoring for this, and it adds several hundred additional lines of untested code.

cholmcc:
> I haven't seen users clamoring for this, 

Well, the PR  #14435 would suggest otherwise, as well as the fact that [How to set the VASSAL Look’n’Feel](https://forum.vassalengine.org/t/set-vassal-looknfeel/84865) has been read  622 times - quite a bit for such a post - _and_ got one like - also not common for posts.   You yourself [showed some interest](https://forum.vassalengine.org/t/native-laf-uniform-button-heights-different-tile-display/930), and others have too - see [here](https://forum.vassalengine.org/t/swing-look-and-feel/10807) and [here](https://forum.vassalengine.org/t/allow-custom-vm-arguments-for-player-editor/10841/2) for example. 

> ... and it adds several hundred additional lines of untested code.

Do you really think that I do not test this kind of change before submitting it?    I've tested this rather a lot on both MacOS and Linux.   It does exactly as expected.    

You could also run some tests yourself, if you feel that is needed - after all untested code is only untested until it is tested :smile: 

"several hundreds" is a tad of an overstatement.   Yes, the `LookAndFeelConfigurere` is around 100 SLOC, but os straight forward.  The rest of the changes are really simplifications of existing code, or extremely simple and easily audited. 

This is a chance to give the users something that can make their user experience better, and at a fairly low cost.  Even if there where no interest, which there clearly is, then why not give the users that experience. 

Yours,
_Christian_

uckelman:
My comment about configurable LaF was from 2008, at a time when I thought V3 had a future. The other 
two posts are from 2020 and contain much detail about the poor appearance of alternate LaFs.

> Do you really think that I do not test this kind of change before submitting it? I've tested this rather a lot on both MacOS and Linux. It does exactly as expected.

What I mean by "untested" is that there are no unit tests.

cholmcc:
> My comment about configurable LaF was from 2008, at a time when I thought V3 had a future. 

Until V4 materialises, V3 is all we have, and will for a fair bit of time ahead.  Hence, I think it makes sense to enhance the user experience. 

> The other two posts are from 2020 and contain much detail about the poor appearance of alternate LaFs.

What one person may like and another may dislike.  Why not give them a choice?  Who are we to say what people's applications should look like.   Personally, I dislike the `Metal` Look'n'Feel, and the `Motif/CDF` is just horrible.   I much prefer the `GTK+` or `MacOS` LnF.   However, for some users, the `GTK+` LnF makes too large buttons, and would perhaps prefer the `Metal`  LnF.   On Windows, I think the native LnF looks far better than the `Metal` LnF, but currently it is impossible (more or less) to use that LnF.   

It is all about choice. 

> What I mean by "untested" is that there are no unit tests.

How would you make a unit test for this?  It is all about the appearance and not much about code execution. 

That said, the `VASSAL.configure.LookAndFeelConfigurer` defines its own test which you can run with 

```
$ java -cp lib/Vengine.jar VASSAL.configure.LookAndFeelConfigurer
```

It shows the configurer and illustrates the change.   This tests that
- You can choose a different LnF 
- That the UI reacts to it 
- And that the configure point is set and read as it should 

The rest of the changes uses the existing framework which is already tested.  The code for setting the initial LnF is actually a simplification of the existing code.   The code that passes the `swing.defaultlnf` on to the child processes is really simple and easily audited.  

So if you don't think the above is enough to qualify as a test, then I kindly ask you to specify what it is you are looking for, and also a few words on how you would go around implementing a test.  Thanks.

Yours,
_Chrisitan_

uckelman:
> What one person may like and another may dislike. Why not give them a choice? Who are we to say what people's applications should look like.

A reason specifically not to give a choice may be found in the threads discussing previous attempts at using other LaFs: Some of them look bad, and we get blamed for it when they do.

> How would you make a unit test for this?

It's a good question. UI is extremely hard to test in Swing, which is a reason why I would not add more.

Why this and not #14435?

cholmcc:
> A reason specifically not to give a choice may be found in the threads discussing previous attempts at using other LaFs: Some of them look bad, and we get blamed for it when they do.

I users choose a LnF they don't like, then the onus is on them.   If developers choose a theme users don't like - and they can't _easily_ change it - then the onus is on the developers. 

In principle, the `Nimbus` theme allows for full customisation of the LnF.   Essentially, one need to do 
```
UIManager.put(config, colour);
```
See for example [this post](https://stackoverflow.com/questions/36128291/how-to-make-a-swing-application-have-dark-nimbus-theme-netbeans) on SO for how to set a dark theme (something that seems popular). 

These key-value (config-colour) could be read from external file, or via `java.util.Properties`. All values are listed [here](https://docs.oracle.com/javase/tutorial/uiswing/lookandfeel/_nimbusDefaults.html). 

What people have complained about is when a LnF has been _forced_ on them.   If they can _choose_ an LnF, I strongly believe they will not complain. 

> It's a good question. UI is extremely hard to test in Swing, which is a reason why I would not add more. 

This UI component is _tiny_.  Clearly you can easily satisfy yourself that it works as intended.  I don't think the lack of an automated test (which is also the case for many other `Configurerer` components) for this small of a component, is any reason to reject a user-facing improvement. 

> Why this and not #14435?

Let us consider the differences between this PR and  #14435?

| Feature                            | #14435  |  #15059 |
|------------------------------------|---------|---------|
| Respect `-Dswing.defaultlaf`       | &check; | &check; |
| Propagate L'n'F to child processes | &check; | &check; |
| Allow third-party L'n'F            | &check; | &times; | 
| UI to change L'n'F                 | &times; | &check; |
| Live change of L'n'F               | &times; | &check; |
| Third-party dependency             | &times; | &times; |
| LLOC                               | 209     | 207     |
| Test                               | &times; | &check; |

As you can see, there are many overlaps between the two, and places where they differ. 
- #14435 requires _some_ expertise to change the LnF, because one has to pass the `-Dswing.defaultlaf` flag to the JVM.   While that is easily done from the command line on Linux and MacOS, it is not so easily done via the GUI or on Windows.  The effect of changing the L'n'F only happens when you start up a new process.   
- #15059 on the other hand gives the user an intuitive GUI interface to change the L'n'F that requires no expertise.  What's more, the effect is immediate. 
- #14435 in principle allows for third-part L'n'F via an external JAR.  However, I've had mixed experience with that.  The process of getting that to work is certainly an expert operation and likewise is best done via the command line - i.e., an expert operation.
- #15059 only presents the user with build-in L'n'Fs.  Those are, on the other hand, well tested, and not likely to cause trouble. 
  - Third party L'n'Fs could be supported along similar lines to #14435 though it will likely always be an "expert" option. 
  - Another option to add "theme-ability" would be that I outlined above for the `Nimbus` L'n'F (similar could also be done for the `Metal` L'n'F). 

I would therefore suggest to take _this_ PR because
- it is fairly simple
- it puts the choice of L'n'F in the users hands - and not just experts 
- it has its own test


Yours,
_Christian_

cholmcc:
Last push fixes some JavaDoc warnings.

uckelman:
@rlament Would you like to weigh in here?

uckelman:
Switching from GTK+ to anything else and then back to GTK+ adds solid borders to the table cells in the Module Manager. 
image

cholmcc:
> ... adds solid borders to the table cells in the Module Manager.

That's a bug in the GTK LnF, in that it doesn't apply it's settings correctly. The borders will be gone on the next invocation. It could also be a bug in the Swings library.

uckelman:
Is there a workaround? It leaves a poor impression.

rlament:
Why is AnimatedDiceButton here?

rlament:
The reason to support look and feel is for dark mode. Dynamic switching between the canned LnFs is nice but doesn't really add much. My impression is that most requests are for dark mode not to make my Linux box look like Windows.

I like the dynamic change not requiring a restart. As a side effect, does it now do that for the font setting?

My earlier attempt PR #14412" Look and Feel for the Module Manager using FlatLaf and user selectable themes" required a restart. Christian seems to have had a change of opinion though concerning the UI. Back then he wrote: "I'm not sure this PR is really needed ... In essence, what you are trying to accomplish is already possible without any code changes."

For #14435 I went as far as necessary to make it usable. I was hoping for some feedback before spending any more time on it. Not saying it necessary is easy to use, but it is a start. It has the dynamic loading which may be useful.

I had stumbled upon FlatLaf. It seems the most capable out there. It has worked well and is simple to integrate. A nice feature is the capability to load any of the IntelliJ themes https://plugins.jetbrains.com/search?products=idea&includeTags=theme 

I have to look at this more closely, but it looks like a good baseline. I'll be offline the next few days.

cholmcc:
> Why is AnimatedDiceButton here?

I accidentally messed up my branches - fixed here and in cholmcc_animated_dice-3.7

cholmcc:
> The reason to support look and feel is for dark mode. 

As alluded to above, it is possible to set up _themes_ for the various L'n'Fs for dark mode.   Typically, it can be done as simple property files, possibly shipped with Vassal, that are loaded should the user request it. 

> Dynamic switching between the canned LnFs is nice but doesn't really add much. 

The point of that is to make easy for all user - not just experts. 

> I like the dynamic change not requiring a restart. As a side effect, does it now do that for the font setting?

Setting a new L'n'F will also set fonts associated with the L'n'F.  The places where Vassal explicitly chooses a font are likely not affected. 
 
> Christian seems to have had a change of opinion though concerning the UI. Back then he wrote: "I'm not sure this PR is really needed ... In essence, what you are trying to accomplish is already possible without any code changes."

What I was addressing, partially a misunderstanding of your code, was how to set a L'n'F from the command line via `-Dswing.defaultlaf`.  You added another property `look.feel` which should point to a JAR file, and then load the L'n'F file from there.  I, mistakenly, took your property `look.feel` to have the same functionality as `swing.defaultlaf`.   

> It has the dynamic loading which may be useful.

I tried it a bit with the `Darcula` JAR, but never got it to work.   It seems to me, that it is not quite as straight forward as one would like it to be. 

> I had stumbled upon FlatLaf. 

[FlatLaf](https://github.com/JFormDesigner/FlatLaf) seems to work great _if_ it is integrated into the app.  As a loadable plug-in, I found it a bit tricky. 

> I have to look at this more closely, but it looks like a good baseline. I'll be offline the next few days.

I think the simplest way to get to dark mode, is to go with themes.    A way to do that would be to add another configurer - one that allows one to select between light and dark themes, if the L'n'F supports that.  The configurer then reads a `.properties` file from the Vassal JAR and applies the settings with 
```
UIManager.put(key, value)
```

Yours,
_Christian_

cholmcc:
> Is there a workaround? It leaves a poor impression.

I looked a little more in to it, and I now think the problem is in the [`swingx`](https://github.com/wumpz/swingx)  JAR ([`JXTreeTable`](https://github.com/wumpz/swingx/blob/master/swingx-core/src/main/java/org/jdesktop/swingx/JXTreeTable.java) specifically). 

If 
- You start in some L'n'F
- and then switch to `GTK+` or `Metal` 

then the cells in the `JXTreeTable` will have borders, which they shouldn't have in neither the GTK+ nor the Metal L'n'F. 

But, if
- You start in some L'n'F 
- and then switch to `Nimbus`

then the rows in the `JXTreeTable` no longer have alternating background colours as they should have in the `Nimbus` L'n'F. 

And, if 
- You start in some L'n'F 
- and then switch to `CDF/Motif`

then every second column has a white background, but they should all be the same colour. 

I therefore believe that the `JXTreeTable` does something that makes things be messed up.  Note, however, it is only the `JXTreeTable` in the `ModuleManager` that is messed up, and only until the next restart of the `ModuleManager`. 

Yours,
_Christian_

rlament:
> I tried it a bit with the Darcula JAR, but never got it to work. It seems to me, that it is not quite as straight forward as one would like it to be.

The issue may have been that PR #14435 was based off of 3.7.19 and didn't have the latest command line improvements. I've rebased onto the Release-3.7 branch. Flat Darcula should now work from Linux using:
`./VASSAL.sh -Dswing.defaultlaf=com.formdev.flatlaf.FlatDarculaLaf -Dlook.feel=flatlaf-3.7.2.jar`

The FlatLaf jar file should be in the same folder as VASSAL.sh.

Editing VASSAL.sh and appending the FlatLaf jar to classpath also works. In which case the `look.feel` parameter is not required.

delatbabel:
I have just incorporated FlatLaF into vassal_extension_utility based on a quick read through of the code in this PR and it seems to work fine. It's not perfect but much less work than the other alternatives (AtlantaLaF etc) and at least comes with usable light/dark/system options.

https://github.com/delatbabel/vassal_extension_utility/

uckelman:
How large is FlatLaf?

uckelman:
> I therefore believe that the JXTreeTable does something that makes things be messed up.

This sounds like something that can be fixed or worked around.

The issue with the cell background color may be due to this:

https://github.com/vassalengine/vassal/blob/master/vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java#L1990-L1999

Possibly `setShowGrid(false);` needs to be called on the `JXTreetable` after switching the LaF?

delatbabel:
> How large is FlatLaf?

About 1MB

cholmcc:
> > How large is FlatLaf?
> 
> About 1MB

FlatLaf is _not_ available on all target systems - for example, it is _not_ packaged on Debian, which would make it problematic to link against that 3rd party library.   Either it has to be explicitly excluded on that platform (and possibly others), or it should be loaded at runtime if available.  

The Debian package of Vassal has become quite popular (just shy of 50% of all Linux downloads are the Debian package), so it would be a degradation of service if the Debian package cannot be build any more. 

The latter choice could be extended - as in https://github.com/vassalengine/vassal/pull/14435 - to be available for all kinds of 3rd party LnFs. 

Yours,
_Christian_

delatbabel:
> > > How large is FlatLaf?
> > 
> > 
> > About 1MB
> 
> FlatLaf is _not_ available on all target systems - for example, it is _not_ packaged on Debian, which would make it problematic to link against that 3rd party library. Either it has to be explicitly excluded on that platform (and possibly others), or it should be loaded at runtime if available.
> 
> The Debian package of Vassal has become quite popular (just shy of 50% of all Linux downloads are the Debian package), so it would be a degradation of service if the Debian package cannot be build any more.
> 
> The latter choice could be extended - as in #14435 - to be available for all kinds of 3rd party LnFs.
> 
> Yours, _Christian_

It's not very much trouble to bundle the FlatLaf library along with the normal fat JAR file that VASSAL ships. That's precisely what I've done with the extension utility.

For the Debian team, packaging FlatLaf would be extra work. FlatLaf's jar carries seven prebuilt binaries (libflatlaf-linux-x86_64.so, the Windows .dlls, the macOS .dylibs). Debian would have to rebuild those from source rather than ship the upstream jar because of their internal packaging requirements. If we bundle the FlatLaf ourselves then those binaries simply travel along. The Linux native build would gain us FlatLaf's optional custom window decorations, which we don't need.

cholmcc:
So I looked into - in the context of https://github.com/vassalengine/vassal/pull/15073 - whether one can load a JAR at run time, find a `LookAndFeel` - or `LookAndFeelInfo` - class, and then _register_ it with the `UIManager`.   That is, the `LookAndFeel` is _merely_ registered, _not_ set as the current L'n'F.  

The point is, of course, that the UI provided with this PR, would then be used by the user to set the L'n'F. 

That is, unfortunately, _not_ possible. 

The reason is, that to load classes from a JAR, we have to make a new `ClassLoader` (a `URLClassLoader` specifically) that has the JAR (or possibly JARs) as its source of `Class`-es.    That `ClassLoader` will be a _child_ of the application (or system) `ClassLoader`. 

Then, later on, when we know the name of the loaded L'n'F class `<laf-class-name>`, and the user chooses that L'n'F, we would attempt to set the L'n'F via `UIManager.setLookAndFeel(<laf-class-name>)`.   The `UIManager` will then use the application `ClassLoader` to try to make an instance of the class. 

However, `ClassLoader` objects _only_ resolves classes that it knows of itself, or those known by its _parent_ `ClassLoader` object - _not_ those known by its children `ClassLoader` objects. 

That means, that the application `ClassLoader` _cannot_ find the class `<laf-class-name>` because that is only known to one of its children `ClassLoader` objects. 

Thus, the _only_ way to dynamically load a JAR and set an L'n'F from that JAR, is if _either_ 
- The JAR is given as part of the initial class-path (option `-cp` - and similar - to `java`), _or_
- if the L'n'F is set as part of loading the JAR. 

The first option is unattractive, because that can be difficult to propagate to child processes.  Sure the system property ` java.class.path` gives us the current class path, but that could be subject to change.  However, if one relies on that,  as is currently done, then the L'n'F jar should definitely be added to `-cp`. 

The second option is more unattractive, because one can never change the L'n'F later on: The only option a user has is to specify the L'n'F at start-up time. 

For the first option, however, we need a way for the Vassal code to actually register the L'n'F class(es) in the JAR(s) when _starting up_, or possibly at some later point via a UI:

```Java
void registerLnF(String name, String className) {
  try {
    final ClassLoader loader = ClassLoader.getSystemLoader();
    final Class<?>    cls    = Class.forName(className, true, loader);
          
    if (LookAndFeel.class.isAssignableFrom(cls)) {
      final LookAndFeel lnf = (LookAndFeel)cls.getConstructors()[0].newInstance();
      UIManager.installLookAndFeel(name == null || name.isBlank() ? lnf.getName() : name,
                                   cls.getName());
    }
    else if (LookAndFeelInfo.class.isAssignableFrom(cls)) {
      final LookAndFeelInfo lnf = (LookAndFeelInfo)cls.getConstructors()[0].newInstance();
      UIManager.installLookAndFeel(lnf);
    }
  }
  // Sketch handling of exceptions - could be done through ErrorDialog
  catch( ClassNotFoundException  e) {
    System.out.println("Unknown class " + className + ": " + e);
  }
  catch (InstantiationException  e) {
    System.out.println("Failed to make instance of " + className + ": " + e);
  }
  catch (InvocationTargetException e) {
    System.out.println("Failed to invoke CTOR of " + className + ": " + e);
  }
}
```
This could be called through a user interface (a `Configurer`).  Of course, the user would need to know the fully qualified class name to use it.   The UI defined in this PR can then pick up the newly available L'n'F and display that as an option. 
 
We also need a way to add the L'n'F JAR(s) to the class path in the start-up scripts (`dist/VASSAL.sh`, `dist/macos/VASSAL.sh`, and `dist/linux/VASSAL.sh` - Windows is a completely different bag of worms).  That can be fairly simply done. 

Note, the  PR https://github.com/vassalengine/vassal/pull/15073  expands upon (and supersedes) this PR.  In that PR, we can provide Dark and Light themes of the various built-in themes (essentially Nimbus and Metal - GTK+, MacOS, and Windows, follows the Desktop theme, so there's little point in adding themes to them.  CDF/Motif is probably not themeable - or will follow the desktop chosen theme). 

Yours,
_Christian_

uckelman:
> That means, that the application ClassLoader cannot find the class <laf-class-name> because that is only known to one of its children ClassLoader objects.

You can set the class loader for a thread using `Thread.setContextClassLoader()`. We do this in `VASSAL.launch.StartUp`. I believe you could do this for the EDT so that 
LaF classes from a JAR could be found.

cholmcc:
I doubt that `Thread.getContextClassLoader()` will work.  That `ClassLoader` would need to be the one created when loading the L'n'F JAR(s), and that `ClassLoader` will have to be a `URLClassLoader` so that we may point it at non-stant-up JAR(s).     If only `UIManager` would allow one to specify a `ClassLoader` to use for a specific L'n'F class, then this could all be much easier done :-(

uckelman:
`DataArchive` manages to allow classes to be loaded from modules. Setting a class loader for whatever thread the `UIManager` uses should work.

rlament:
Isn't #14435 an example of what you are trying to do? In that case the jar is specified on the command line with a custom definition (_-Dlook.feel_), but the file name could come from the user preference file. It is equivalent to adding the jar to the _-cp_ option. It also loads themes. Turn your themed properties files into _json_ following the _Intellij_ theme format and they should load.

To enumerate LnF/themes use something like `getSystemClassLoader()` or _Reflections_ Library. These look promising, but I have no experience with them.

https://www.baeldung.com/java-find-all-classes-in-package
https://www.baeldung.com/reflections-library

cholmcc:
> Isn't #14435 an example of what you are trying to do? In that case the jar is specified on the command line with a custom definition (_-Dlook.feel_), but the file name could come from the user preference file. 

No, it is not equivalent.   The problem is, that the `ClassLoader` you make will be a _child_ of the application (or system) `ClassLoader`.  And `ClassLoader`s does _not_ use child `ClassLoader`s to resolve `Class`es with - they only use the _parent_ `ClassLoader`. 

In #14435 you use the `ClassLoader` you just made to make an instance of a class that derives from `LookAndFeel`, and then _immediately_ set that as the L'n'F using `UIManager.setLookAndFeel(<object of class LookAndFeel>)`.  That works fine, because the `ClassLoader` that `UIManager` uses - the application `ClassLoader` - isn't in play. 

However, if you don't want to set the L'n'F directly when loading the JAR, but want to defer that to a later user decision, then you would need to have the `ClassLoader` object that you used to load the JAR with.   That is, when you do `UIManager.setLookAndFeel(<string name of LookAndFeel class>)`, then `UIManager` will use the application `ClassLoader` - which doesn't know about the classes in the loaded JAR (because the created `ClassLoader` object - when loading the JAR - is a _child_ of the application `ClassLoader`), and so the `UIManager` cannot find the `Class` and will therefore error out. 

The annotated UML diagram below tries to illustrate the point.
```
     +-----------------+
 ^   | BootClassLoader |
 |   +-----------------+
 |           ^ children
 |           V
 |           | parent
 |   +---------------------+
 |   | PlatformClassLoader |
 |   +---------------------+
 |           ^ children
 |           V
 |           | parent
 |   +-------------------+    +------------------------+
 |   | SystemClassLoader |---o| UIManager              |
 |   +-------------------+    +------------------------+
 |           ^ children       | setLookAndFeel(String) |
 |           V                +------------------------+
 |           | parent
 |   +----------------+ known     +-----------------------------------+
 |   | JARClassLoader |<>---------| <Collection<Class> in loaded JAR> |
 |   +----------------+           +-----------------------------------+
 |
 | <Direction of Class resolution>
```
When the code executes `UIManager.setLookAndFeel(<String name of LookAndFeel class>)`, then a request is send to `SystemClassLoader`, which does not know the class.  The request is then sent _up_ the tree, first to `PlatformClassLoader` and then `BootClassLoader`, neither of which knows the class. 

The `JARClassLoader` _does_ know the class, but that `ClassLoader` isn't reachable from the `SystemClassLoader`. 

See the [implementation of `UIManager.setLookAndFeel(String)`](https://github.com/openjdk/jdk17u/blob/master/src/java.desktop/share/classes/javax/swing/UIManager.java#L625).  It uses [`SwingUtilities.loadSystemClass`](https://github.com/openjdk/jdk17u/blob/master/src/java.desktop/share/classes/javax/swing/SwingUtilities.java#L2034) which in turn may use the [`Thread.getContextClassLoader`](https://github.com/openjdk/jdk17u/blob/master/src/java.base/share/classes/java/lang/Thread.java#L1496) (if not `null`, in which case it uses the `SystemClassLoader`). 

AS you can also see from the [implementation of `ClassLoader.loadClass`](https://github.com/openjdk/jdk17u/blob/master/src/java.base/share/classes/java/lang/ClassLoader.java#L569), the direction of search is up toward parents - not down toward children. 

The long and short of it is, if you want to set the L'n'F at some later point - possibly to a class loaded from a JAR - then the JAR _must_ be loaded by the `SystemClassLoader` (or one of its parents). 

As I said, if one could register a `ClassLoader` for a `LookAndFeel` derived class with the `UIManager`, then the `UIManager` could directly use that `ClassLoader` (`JARClassLoader` in the diagram above), and would be able to resolve the `Class`.  Unfortunately, that is not how `UIManager` is defined. 

Could one wrap `UIManager` so that it would be possible?  Yeah, I think that could be done. 
```Java
class MyUIManager  {
  Map<String> nameLoaderMap = ...
  Map<String> classLoaderMap = ... 

  ClassLoader loadJar(String filename) { 
     Path  path = Paths.get(filename);
     URL   url  = path.toUri().toURL();
     URL[] urls = {url};
     URLClassLoader loader = new URLClassLoader(urls);
     return loader;
  }

  void installLookAndFeel(String name, String className, ClassLoader loader) {
     if (loader == null) {
       UIManager.installLookAndFeel(name, className);
       return;
     }
     nameLoaderMap.put(name, loader);
     classLoaderMap.put(className, loader);
   }

    void setLookAndFeel(String className) {
      if (!classLoaderMap.haskey(className)) {
        UIManager.setLookAndFeel(className);
        return;
      }
 
      ClassLoader loader = classLoaderMap.get(className);
      Class<?>    cls    = loader.loadClass(className);
      Object      obj    = cls.getDeclaredConstructor().newInstance();
     
      if (LookAndFeel.class.isAssignable(cls)) {
        UIManager.setLookAndFeel(cls);
        return;
      }
    }
  }
}
```

> It is equivalent to adding the jar to the _-cp_ option. 

No, it really isn't.  The above shows you why it isn't. 

> https://www.baeldung.com/java-find-all-classes-in-package
> https://www.baeldung.com/reflections-library

I don't think these two libraries changes the fundamental way that `ClassLoader` can be utilised 

Yours,
_Christian_

cholmcc:
>  Setting a class loader for whatever thread the `UIManager` uses should work.

Yes, you have to drag the `ClassLoader` around - whether in a wrapper, or some other way of using that `ClassLoader` (temporarily) via f.ex. `Thread.setContextClassLoader(classLoader)` (resetting it to the previous `ClassLoader` - possibly `null` - afterwards).

cholmcc:
I almost have an implementation that allows dynamic loading of JARs, dynamic selection L'n'F, and so on - all through the _File_&rarr;_Preferences_ interface.  Stay tuned.

cholmcc:
See #15095 which supersedes this.

PR #15060, opened 2026-09-07
Stop writing obfuscated data in hex

delatbabel:
ObfuscatingOutputStream wrote each XORed byte as two hex digits, doubling the size of the savedGame entry and, since the result was a 16-symbol stream, leaving the ZIP little to compress. The hex encoding served no purpose: the point is only to deter casual hand-editing, and the XORed bytes are already less readable than hex.

The new format is the header VOBS, then the one-byte key, then the data XORed with the key. The leading bang is dropped and the header is four bytes, which is more typical. The key is now chosen in 1-255, so we never XOR with 0 and leave the data in plain text.

This deletes the hex table and per-byte pair buffer from ObfuscatingOutputStream, and gives it a real write(byte[], int, int) which XORs through a scratch buffer rather than delegating per byte.

In DeobfuscatingInputStream the old hex reader survives unchanged as LegacyDeobfuscatingInputStreamImpl, still matched on the !VCSK header, so saves written by earlier versions continue to load. A new, much smaller DeobfuscatingInputStreamImpl handles VOBS. Plain text is still passed through unchanged, including input shorter than a header.

A 1 MB payload now occupies 1,000,005 bytes rather than 2,000,007.

### DIFF
--- a/vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
+++ b/vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
@@ -28,12 +28,16 @@ import java.nio.file.Path;
 /**
  * A {@link FilterInputStream} which converts a file created with
  * {@link ObfuscatingOutputStream} back into plain text.
- * Additionally, plain text will be passed through unchanged.
+ * Files in the legacy hex-encoded format are also handled, and
+ * plain text will be passed through unchanged.
  *
  * @author Joel Uckelman
  * @since 3.2.0
  */
 public class DeobfuscatingInputStream extends FilterInputStream {
+
+  /** The header used by the hex-encoded format written before VASSAL 3.8. */
+  private static final String LEGACY_HEADER = "!VCSK"; //NON-NLS
 
   /**
    * @param in the stream to wrap
@@ -42,17 +46,32 @@ public class DeobfuscatingInputStream extends FilterInputStream {
   public DeobfuscatingInputStream(InputStream in) throws IOException {
     super(null);
 
-    final byte[] header = new byte[ObfuscatingOutputStream.HEADER.length()];
-    readFully(in, header, header.length);
-    if (new String(header, StandardCharsets.UTF_8).equals(ObfuscatingOutputStream.HEADER)) {
-      this.in = new DeobfuscatingInputStreamImpl(in);
+    final int hlen = ObfuscatingOutputStream.HEADER.length();
+    final byte[] header = new byte[Math.max(hlen, LEGACY_HEADER.length())];
+
+    int n = readFully(in, header, 0, hlen);
+
+    if (n == hlen) {
+      if (ObfuscatingOutputStream.HEADER.equals(
+            new String(header, 0, hlen, StandardCharsets.UTF_8))) {
+        this.in = new DeobfuscatingInputStreamImpl(in);
+        return;
+      }
+
+      n += readFully(in, header, hlen, LEGACY_HEADER.length() - hlen);
+
+      if (n == LEGACY_HEADER.length() && LEGACY_HEADER.equals(
+            new String(header, 0, n, StandardCharsets.UTF_8))) {
+        this.in = new LegacyDeobfuscatingInputStreamImpl(in);
+        return;
+      }
     }
-    else {
-      final PushbackInputStream pin =
-        new PushbackInputStream(in, header.length);
-      pin.unread(header);
-      this.in = pin;
-    }
+
+    // Not obfuscated; pass the whole stream through unchanged
+    final PushbackInputStream pin =
+      new PushbackInputStream(in, header.length);
+    pin.unread(header, 0, n);
+    this.in = pin;
   }
 
   /**
@@ -62,14 +81,15 @@ public class DeobfuscatingInputStream extends FilterInputStream {
    * @param bytes the destination
    * @param off the offset into the destination array
    * @param len the number of bytes to read
-   * @throws IOException if <code>len</code> bytes cannot be read
+   * @return the number of bytes read
+   * @throws IOException if an I/O error occurs
    */
-  private static int readFully(InputStream in, byte[] bytes, int len)
+  private static int readFully(InputStream in, byte[] bytes, int off, int len)
                                                            throws IOException {
     int count;
     int n = 0;
     while (n < len) {
-      count = in.read(bytes, n, len - n);
+      count = in.read(bytes, off + n, len - n);
       if (count < 0) break;
       n += count;
     }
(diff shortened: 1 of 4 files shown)


vassal-app/src/main/java/VASSAL/tools/io/ObfuscatingOutputStream.java
```diff
@@ -60,31 +64,28 @@ public ObfuscatingOutputStream(OutputStream out, byte key)
...
+    // Obfuscate via a scratch buffer, so as not to alter the caller's array
+    while (len > 0) {
+      final int n = Math.min(len, buf.length);
+      for (int i = 0; i < n; ++i) {
+        buf[i] = (byte) (bytes[off + i] ^ key);
+      }
+      out.write(buf, 0, n);
+      off += n;
+      len -= n;
+    }
```
uckelman: This isn't necessary. The existing implementation of this function already does the right thing.

vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
```diff
@@ -28,31 +28,50 @@
...
+ * Files in the legacy hex-encoded format are also handled, and
+ * plain text will be passed through unchanged.
  *
  * @author Joel Uckelman
  * @since 3.2.0
  */
 public class DeobfuscatingInputStream extends FilterInputStream {
 
+  /** The header used by the hex-encoded format written before VASSAL 3.8. */
+  private static final String LEGACY_HEADER = "!VCSK"; //NON-NLS
```
uckelman: It doesn't make sense for this to be a `String` when we're reading bytes.

Should be
```
private static final byte[] LEGACY_HEADER = new byte[ '!', 'V', 'C', 'S', 'K' ];
```
and then there's no reason to do any conversions.

vassal-app/src/main/java/VASSAL/tools/io/ObfuscatingOutputStream.java
```diff
@@ -31,22 +31,26 @@
...
  *
+ * <p>The output consists of {@link #HEADER}, followed by the one-byte key,
+ * followed by the input XORed byte-by-byte with the key.</p>
+ *
  * @author uckelman
  * @since 3.2.0
  */
 public class ObfuscatingOutputStream extends FilterOutputStream {
-  public static final String HEADER = "!VCSK"; //NON-NLS
+  public static final String HEADER = "VOBS"; //NON-NLS
```
uckelman: Deprecate `HEADER`.  Make a new byte array for the new header.

vassal-app/src/main/java/VASSAL/tools/io/ObfuscatingOutputStream.java
```diff
@@ -60,31 +64,28 @@ public ObfuscatingOutputStream(OutputStream out, byte key)
     this.key = key;
 
     out.write(HEADER.getBytes(StandardCharsets.UTF_8));
```
uckelman: Write the header bytes directly, don't convert.

vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
```diff
@@ -62,29 +81,67 @@ public DeobfuscatingInputStream(InputStream in) throws IOException {
    * @param bytes the destination
    * @param off the offset into the destination array
    * @param len the number of bytes to read
-   * @throws IOException if <code>len</code> bytes cannot be read
+   * @return the number of bytes read
+   * @throws IOException if an I/O error occurs
    */
-  private static int readFully(InputStream in, byte[] bytes, int len)
+  private static int readFully(InputStream in, byte[] bytes, int off, int len)
```
uckelman: What is the purpose of this change?

cholmcc:
Are you sure this should be a 3.7.28 milestone?   Saves written with this code in will not be readable by Vassal without this code, which will break compatibility in an unexpected way.   I think you want to target this change at 3.8.0 instead.

vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
```diff
@@ -28,31 +28,50 @@
...
+
+      n += readFully(in, header, hlen, LEGACY_HEADER.length() - hlen);
+
+      if (n == LEGACY_HEADER.length() && LEGACY_HEADER.equals(
+            new String(header, 0, n, StandardCharsets.UTF_8))) {
+        this.in = new LegacyDeobfuscatingInputStreamImpl(in);
+        return;
+      }
     }
+
```
cholmcc: What if `n != hlen`?  Shouldn't that throw?  If you fail to read 4 bytes, then clearly something is amis.

vassal-app/src/main/java/VASSAL/tools/io/ObfuscatingOutputStream.java
cholmcc: I think the idea is to do a buffered write, rather than one byte at a time.

vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
cholmcc: To append the last byte read off a legacy header to the output buffer.   See line 61 above.   Could probably be done simpler by reading the one byte into the buffer above.

vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
cholmcc: There might be.  In line 55 below, the first four bytes of `header` are compared to the string `"VOBS"`, so you want to compare the first 4 bytes read to the `HEADER = new byte['V','O','B','S']` (or `HEADER = {'V','O','B','S'}` while later on, in line 63, you want to compare the first _5_ bytes to `LEGACY_HEADER = {'!', 'V', 'C', 'S', 'K'}`.   

You probably want to use [`Arrays.equals`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Arrays.html#equals(byte%5B%5D,int,int,byte%5B%5D,int,int)) to compare the byte arrays
```
if (Arrays.equals(header, 0, 4, ObfuscatingOutputStream.HEADER, 0, 4)) ...
if (Arrays.equals(header, 0, 5, LEGACY_HEADER, 0, 5)) ...
```

cholmcc:
> A 1 MB payload now occupies 1,000,005 bytes rather than 2,000,007.

You mean "1MB payload now occupies (1024 &times; 1024 + 7) bytes =  (1048576 + 5) bytes = 1048581 bytes".  :smile:

I know that the _technically_, 1024 MiB = 1024 KiB = 1024 &times; 1024 bytes, but I believe most people use `MB` to mean `MiB`.  1MB = 1000 &times; 1000 bytes is often consider as _salesmans mega-bytes_. 

> The leading bang is dropped and the header is four bytes, which is more typical.

While "typical" - it is certainly not a standard.  The number of magic bytes can be _any_ number, as long as it is unique.  You could opt for the magic bytes `VASSAL_SAVE` - which, arguably, is more descriptive :smile:

Yours,
_Christian_

delatbabel:
> Are you sure this should be a 3.7.28 milestone? Saves written with this code in will not be readable by Vassal without this code, which will break compatibility in an unexpected way. I think you want to target this change at 3.8.0 instead.

Yeah this should be 3.8.0 however I'm not familiar enough with the branching / tagging structure here to know how to target this for 3.8 rather than 3.7

delatbabel:
> > A 1 MB payload now occupies 1,000,005 bytes rather than 2,000,007.
> 
> You mean "1MB payload now occupies (1024 × 1024 + 7) bytes = (1048576 + 5) bytes = 1048581 bytes". 😄
> 
> I know that the _technically_, 1024 MiB = 1024 KiB = 1024 × 1024 bytes, but I believe most people use `MB` to mean `MiB`. 1MB = 1000 × 1000 bytes is often consider as _salesmans mega-bytes_.

It's an order of magnitude estimate in any case -- the actual effect of the deflate / compression can't be guessed at just by knowing the number of bytes in the original input stream.

uckelman:
> Yeah this should be 3.8.0 however I'm not familiar enough with the branching / tagging structure here to know how to target this for 3.8 rather than 3.7

You've already done it. `master` is the 3.8 branch, `release-3.7` is the 3.7 branch.

vassal-app/src/main/java/VASSAL/tools/io/ObfuscatingOutputStream.java
```diff
@@ -31,22 +30,36 @@
...
+   * The header of the hex-encoded format written before VASSAL 3.8.
+   *
+   * @deprecated The hex-encoded format is no longer written, only read.
+   * Obfuscated output is now marked with {@link #HEADER_BYTES}.
+   */
+  @Deprecated(since = "2026-09-08", forRemoval = true)
   public static final String HEADER = "!VCSK"; //NON-NLS
+
+  /** The header marking obfuscated output. */
+  static final byte[] HEADER_BYTES = { 'V', 'O', 'B', 'S' };
```
uckelman: Missing visibility modifier. Should be public.

uckelman:
Once that visibility change is in, I think this is fine to be merged.

cholmcc:
I was more remarking on @uckelman adding a 3.7.28 milestone to the PR. Joel has changed it to 3.8.0 milestone.

vassal-app/src/main/java/VASSAL/tools/io/DeobfuscatingInputStream.java
cholmcc: See #15067

PR #15073, opened 2026-09-13
Allow user to set Java Look and Feel - themeable version

cholmcc:
This will allow the user to set the Look and Feel (LnF) of the
graphical user interface.   The following changes are done

- The current value of `swing.defaultlaf` is propagated on to child processes. That means, when the `ModuleManager` spawns the `Player` or `Editor` (and other processes), that those child-processes will also get the current setting for the LnF.
- If no LnF is specified by the user, then the default is to use the system LnF.  Note, on Windows, the default was to switch to the system independent _Metal_ LnF, while for MacOS and Linux, the default LnF was always the system LnF.
- Adds a new configurer class `LookAndFeelConfigurer` which are added
  to the shared global preferences.   This allows a user to specify
  the LnF via _File_&rarr;_Preferences_&rarr;_General_&rarr;_Look and
  feel_, even at runtime.  The setting is of course stored in
  `V_Global`, and will therefore be used for subsequent launches.
- Furthermore, the PR provides a few themes for some of the L'n'Fs.
  These are stored as `.properties` files in the JAR, or may be
  overriden by the user by having a file with the same name in the
  current directory.

This PR is somewhat at odds with the PR #14435 and they should not _both_ be applied.  Note that #14435 attempts to enable third-party provided LnFs, while this PR will only use those already provided (could in principle be third party too).

This PR expands upon #15059 but supersedes it.


cholmcc:
See #15095 which supersedes this.

PR #15076, opened 2026-09-14
Use plain Swing timer for Map auto scrolling

cholmcc:
This PR replaces the old `org.jdesktop.timing.Animator` use with a plain `javax.swing.Timer` based implementation.

The code is actually simpler than the one using
`org.jdesktop.timing.Animator` and there's less overhead than in that
in that case.   That's because `org.jdesktop.timing.Animator` does a
lot more than what is needed here, but really doesn't add anything of
substance.  The scrolling - off-hand - seems to be smoother.


cholmcc:
See also PRs 

- https://github.com/vassalengine/vassal/pull/15075
- https://github.com/vassalengine/vassal/pull/15076
- https://github.com/vassalengine/vassal/pull/15077

If these PRs are all adopted, then the dependency on `org.jdesktop.timing` can be completely removed.  The code in that JAR is _very_ old (almost 20 years old), and does not seem to be maintained anymore.   Thus, adopting these PRs will result in one less - possibly - problematic third-party dependency. 

The code introduced by these three PRs are at least as simple as the code they replace, and are in all cases at least as effective as the code they replace.   In some cases, it seems like the PRs code is actually more effect and makes the animations run smother.   This is likely due to the overhead that `org.jdesktop.timing.Animator` introduces. 

The PRs can be merged into a single PR, which could then also remove the `org.jdesktop.timing` dependency from `vassal-app/pom.xml` and `debian/control` (and `debian/rules`), if that is desirable.  The PRs were made separately so that they could more easily be evaluated. 

Yours,
_Christian_

PR #15095, opened 2026-09-17
Allow user to set Java Look and Feel

cholmcc:
**TL;DR**:
- User can choose Look'n'Feel in the UI
- Users can load 3rd party Look'n'Feel in the UI
- Look'n'Feels can be themed, by user and distributed

This will allow the user to set the Look and Feel (LnF) of the graphical user interface.

- A new tab is added to the _File&rarr;Preferences_ dialog called **Look and Feel**
- In that tab, there's a drop-down menu that allows the user to choose the LnF.
- Below that is an interface that where a user may add third-party JARs to load.
  - When a JAR is loaded, it is scanned for all available classes.
  - The user is then presented with a list of these classes.
  - When a user select one or more classes from that that, those classes are attempted to be loaded as `LookAndFeel` classes.
  - If the loading succeeds, then the corresponding LnF is added to the drop-down menu.
  - If loading of a class as a `LookAndFeel` does _not_ succeed, then the user is alerted via a pop-up window, and the class is removed from the selection.
- For any known LnF - say `Foo`, if there's a `/VASSAL/theme` resource file, or a file in the Vassal directory, named `Foo.properties`, then that file is loaded, and the colour properties named in that file are attached to the LnF.  When the user sets that LnF, those properties are set on the `UIManager`.  This allows one to theme the LnF.
- Similarly, if there's a resource file or local files named `FooDark.properties` or `FooLight.properties`, then those are loaded too, and a themed LnF named `Foo (Dark)` and `Foo (Light)`, respectively are added to the LnFs available for the user to choose from.
- The new LnF (if any) is installed on the UI when the _Preference_ panel is closed.

Note, since these preferences are stored in the global section - i.e., `V_Global` they are also read in by child processes, and the preference for the LnF is propagated there too.  The converse is also true.

In addition, the PR also fixes a number of issues in `ModuleManagerWindow` that interact poorly with LnFs:

- The text colour of elements in the module listing is retrieved from the `UIManager`, so that dark themes will have light text, and vice-versa.
- Special entries default to the standard colour, or may choose a different colour from the `UIManager` - again to make a contrast between background and foreground.
- In centred columns, the cells' colours are explicitly set to `null` so that the overall table render will choose the colours.  This prevents some cruft while switching between themes.

Note, when using a dark theme, the text in the _chat_ may not have enough contrast to the background.  This can be fixed by the user by configuring the chat text colours.

Note, when switching between LnFs, there's sometimes a bit of cruft left behind.  For example, when switching to the GTK+ LnF, the table in the Module Manager will all of a sudden have borders around the cells.  Similarly, when switching to the Nimbus theme, then the rows do no longer have alternating colours.  These, and similar, issues can be remedied by restarting the application.  These issues seem very much related to the `JXTreeTable` component, and does not seem to pop-up elsewhere.

- This PR incorporates some of the thoughts of #14435, but allows the user to choose via the UI.
- This PR expands upon #15059, but supersedes it.
- This PR expands upon #15073, but supersedes it.


cholmcc:
Some screenshots 

| | |
|-|-|
| LnF chooser | LaFChooser |
| Choice of LnF | LaFChoice |
| Add 3rd party JAR | AddLaFJar |
| Add 3rd part `LookAndFeel` | AddLaFClasses |
| LnF chooser with 3rd party LnFs | LaFChoiceCustom |
| Module manager in Dark LnF | MetalDarkModuleManager |
| Module player in Dark LnF | LaFDarkModule |

cholmcc:
BTW, choices of LnF as well as 3rd party JARs and `LookAndFeel` classes persist over invocations. 

If a 3rd party JAR or `LookAndFeel` class is removed, then it will not affect the current session - the LnF(s) is still available, as we're not unloading the classes.   However, on the next invocation of Vassal, those options will no longer we available.  Had the user specified an LnF that is no longer available because it was removed, then the UI will default to the system LnF. 

If a user tries to load a non-`LookAndFeel` class, then they will see an error dialog like the below: 
LaFClassError

Yours,
_Christian_

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+        UIManager.setLookAndFeel(getClassName());
+
+        Thread.currentThread().setContextClassLoader(old);
+      }
+      catch (ClassNotFoundException          |
+             InstantiationException          |
+             IllegalAccessException          |
+             UnsupportedLookAndFeelException | 
+             ClassCastException              |
+             NullPointerException e) {
```
uckelman: Which statements in the try block can throw which of these exceptions?

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
```diff
@@ -2234,7 +2252,9 @@ private boolean belongsToModule() {
     @Override
     public Color getTreeCellFgColor() {
       // FIXME: should get colors from LAF
```
uckelman: This FIXME is now fixed?

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
```diff
@@ -1991,10 +2005,14 @@ public JPopupMenu buildPopup(int row) {
...
```
uckelman: This FIXME is now fixed?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+            ErrorDialog.show("LookAndFeel.load_class_fail",
+                             className, jarFileName);
+            toRemove.add(className);
+          }
+        }
+        catch (ClassNotFoundException    |
+               InstantiationException    |
+               InvocationTargetException |
+               IllegalAccessException    |
+               NullPointerException e) {
```
uckelman, line 471: Which statements in the try block can throw which of these exceptions?

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
cholmcc: Yes.

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
cholmcc: Yes.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: [`Class.forName(String,boolean,ClassLoader)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Class.html#forName(java.lang.String,boolean,java.lang.ClassLoader)) may also throw

- [LinkageError](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/LinkageError.html) - if the linkage fails
- [ExceptionInInitializerError](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ExceptionInInitializerError.html) - if the initialization provoked by this method fails
- [ClassNotFoundException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ClassNotFoundException.html) - if the class cannot be located by the specified class loader
- [SecurityException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/SecurityException.html) - if a security manager is present, and the loader is null, and the caller's class loader is not null, and the caller does not have the [RuntimePermission](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/RuntimePermission.html)("getClassLoader")

[`Class.isAssignableFrom(Class)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Class.html#isAssignableFrom(java.lang.Class)) may throw 

- [NullPointerException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/NullPointerException.html) - if the specified `Class` parameter is `null`.

[`Class.getConstructors`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Class.html#getConstructors()) may throw 

- [SecurityException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/SecurityException.html) - If a security manager, s, is present and the caller's class loader is not the same as or an ancestor of the class loader for the current class and invocation of [s.checkPackageAccess()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/SecurityManager.html#checkPackageAccess(java.lang.String)) denies access to the package of this class.

[`Constructor.newInstance`]() may throw 

- [IllegalAccessException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/IllegalAccessException.html) - if this Constructor object is enforcing Java language access control and the underlying constructor is inaccessible.
- [IllegalArgumentException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/IllegalArgumentException.html) - if the number of actual and formal parameters differ; if an unwrapping conversion for primitive arguments fails; or if, after possible unwrapping, a parameter value cannot be converted to the corresponding formal parameter type by a method invocation conversion; if this constructor pertains to an enum class.
- [InstantiationException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/InstantiationException.html) - if the class that declares the underlying constructor represents an abstract class.
- [InvocationTargetException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/reflect/InvocationTargetException.html) - if the underlying constructor throws an exception.
- [ExceptionInInitializerError](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ExceptionInInitializerError.html) - derived from `LinkageError` - if the initialization provoked by this method fails.

I know its a lot.   However, the down-stream effect of these are the same - the user cannot use the requested Look'n'Feel, so I think it is correct to treat them all the same and show an error dialog with more information.   One could simply `catch (Exception e)` and avoid the long list of explicit exceptions. 

Yours,
_Christian_

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: These are thrown by [`UIManager.setLookAndFeel(String)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/javax/swing/UIManager.html#setLookAndFeel(java.lang.String)): 
-  [ClassNotFoundException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ClassNotFoundException.html) - if the LookAndFeel class could not be found
- [InstantiationException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/InstantiationException.html) - if a new instance of the class couldn't be created
- [IllegalAccessException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/IllegalAccessException.html) - if the class or initializer isn't accessible
- [UnsupportedLookAndFeelException](https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/javax/swing/UnsupportedLookAndFeelException.html) - if lnf.isSupportedLookAndFeel() is false
- [ClassCastException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ClassCastException.html) - if className does not identify a class that extends LookAndFeel
- [NullPointerException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/NullPointerException.html) - if className is `null`


I know its a lot.   However, the down-stream effect of these are the same - the user cannot use the requested Look'n'Feel, so I think it is correct to treat them all the same and show an error dialog with more information.   One could simply `catch (Exception e)` and avoid the long list of explicit exceptions. 


Yours,
_Christian_

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
uckelman: Please remove, in that case.

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
uckelman: Please remove, in that case.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
uckelman: My concern here is not the quantity of exceptions but the scoping of the try block. One try block over all the function calls makes it unclear what is expected to throw what, and also can catch exceptions we don't expect from statements we don't intend to cover.

Better would be to give each of `UIManager.setLookAndFeel()`, `Class.getConstructors()`, and `Class.forName()` their own try block.

cholmcc: > Better would be to give each of UIManager.setLookAndFeel() and Class.forName() their own try block. 

I miswrote the first reply - see updated reply.  We're talking 4 different `try` blocks here, if we're to pin it out. 

I don't think there's a large chance of being too inclusive here.   After all, the end result is in all cases the same - the Look'n'Feel cannot be installed, and the user is notified of that (with all the context).   Also, the exceptions caught - except `NullPointerException` are really specific and not all general purpose - as say `RuntimeError`, so I think this is OK. 

Yours,
_Christian_

uckelman: For `NullPointerException`, we shouldn't be catching that at all---check for `null` first.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
uckelman: Resetting the class loader ought to be in a `finally` block, also.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: I believe the compiler complains if ` NullPointerException` isn't handled.   The `Class` parameter is clearly not `null`, so there's no need  here to check it for `null` (`LookAndFeel.class` will never be `null`). 

Generally, I try not to handle exceptions, and then let the compiler tell me which needs to be handled.  That's why I'm fairly certain that the compiler will complain if the `NullPointerException` isn't handled. 

Yours,
_Christian_

uckelman: NPE is an unchecked exception. The language never requires you to catch it.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
uckelman: E.g.:
```java
      final ClassLoader old = Thread.currentThread().getContextClassLoader();
      if (loader != null) {
        Thread.currentThread().setContextClassLoader(loader);
      }

      try {
        if (isMetal()) {
          MetalLookAndFeel.setCurrentTheme(properties != null ?
            new CustomMetalTheme(properties) : new DefaultMetalTheme()
          );
        }
        else {
          loadThemeProperties();
        }

        try {
          UIManager.setLookAndFeel(cl);
        }
        catch (ClassNotFoundException          |
               InstantiationException          |
               IllegalAccessException          |
               UnsupportedLookAndFeelException |
               ClassCastException e) {
          // This should very rarely happen
          ErrorDialog.show(
            e,
            "LookAndFeel.set_fail",
            getName(),
            cl
          );
        }
      }
      finally {
        if (Thread.currentThread().getContextClassLoader() != old) {
          Thread.currentThread().setContextClassLoader(old);
        }
      }
```

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+      }
+      return null;
+    }
+    
+    /**
+     * Before loading the LnF, set the theme properties on the
+     * UIManager.
+     */
+    public void loadThemeProperties() {
+      if (isMetal() || properties == null)
```
uckelman: A subclass of `LafInfo` just for Metal would eliminate the need to check whether any particular `LafInfo` is Metal.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+public class LookAndFeelConfigurer extends Configurer {
+  /**
+   * Parse a string that defines a color, either as an RGB value,
+   * given either as hexadecimal number or in HTML style with a "#"
+   * mark.  If neither, assume a named color.
+   * 
+   * @param value String to parse
+   * @return an java.awt.Color or null
+   */
+  public static Color parseColor(String value) {
```
uckelman: Could this be private?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+    catch (NumberFormatException e) {
+      e.printStackTrace(System.out);
+    }
+    return color;
+  }
+  
+  /**
+   * A metal theme that can be configured via properties. 
+   */
+  public static class CustomMetalTheme extends DefaultMetalTheme {
```
uckelman: Private?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+    public Map<String, ColorUIResource> mapping = new HashMap<>();
```
uckelman: Private?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+  protected class LafClassLoader extends URLClassLoader {
+    LafClassLoader() {
+      super(new URL[]{}, getSystemClassLoader());
+    }
+    LafClassLoader(URL[] urls) {
+      super(urls, getSystemClassLoader());
+    }
+
+    @Override
+    public void addURL(URL url) {
```
uckelman, line 317: Is this needed? The docs for `URLClassLoader.addURL` claim that it's a no-op when adding a URL that's already been added. It looks like you could just use `URLClassLoader` without subclassing it.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+      try {      
+        return Paths.get(jarFileName).toUri().toURL();
+      }
+      catch (MalformedURLException e) {
+        return null;
+      }
+    }
+    
+    public List<String> getClasses() {
+      final List<String> ret = new LinkedList<>();
```
uckelman: There are almost no use cases for `LinkedList`, and they all involve frequent removal of items from the middle. `ArrayList` will be more performant and should be preferred.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+          final String fqName = name.replace("/", ".")
+            .substring(0, name.length() - ".class".length());
+
+          // System.out.println("Adding class " + fqName);
+          ret.add(fqName);
+        }
+        return ret;
+      }
+      catch (IOException e) {
+        // Ignore exceptions
```
uckelman: Why should we ignore IOException? Typically we want to raise an error dialog when something can't be read.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+     * Otherwise, we remove it from our selection
+     */
+    void makeAvailable(ClassLoader loader) {
+      if (!isValid())
+        return;
+
+      final List<String> toRemove = new LinkedList<>();
+      
+      for (final String className : lafClassNames) {
+        // Do not attempt at loading a class we already tried
```
uckelman, line 447: Is brute-force iteration over everything in a JAR really the best way to find the classes extending `LookAndFeel`? That can't be the way you're intended to load an LaF, can it?!

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+          ErrorDialog.show(e,
+                           "LookAndFeel.load_class_fail",
+                           className, jarFileName);
+          toRemove.add(className);
+        }        
+      }
+
+      for (final String del : toRemove) {
+        lafClassNames.remove(del);
+      }
```
uckelman, line 481: If we do in fact need to try loading each and every class to determine if it's a `LookAndFeel`, why do we make a list of all the classes and then iterate over it again to find the ones to keep? Couldn't we check them and keep the `LookAndFeel` subclasses in a single pass over the class files?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+    }
+  }
+
+  /**
+   * Configure location of a JAR and select classes from it.
+   */
+  public class LafJarConfigurer extends Configurer {
```
uckelman: Private?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+  }
+
+  /** Declare ahead of time */
+  protected LafConfigurer lafConfigurer;
+      
+  /**
+   * List of jars, and for each jar, a list of classes to add as Look
+   * and Feel classes
+   */
+  public class LafJarsConfigurer extends ListConfigurer {
```
uckelman: Private?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+      super.updateValue();
+      fireUpdate();
+    }
+  }
+
+  /**
+   * Holds information about a look'n'feel, including a possible class
+   * loader
+   */
+  protected class LafInfo extends LookAndFeelInfo {
```
uckelman: Private?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+    
+    public String valueString() {
+      return getName() + "," + getClassName();
+    }
+  }
+
+  /**
+   * Configurer of current Look'n'Feel
+   */
+  public class LafConfigurer extends Configurer {
```
uckelman: Private?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+            setValue(info.getName());
+        }
+      }
+    }
+
+    /**
+     * Update available Look and Feels in the combo box based on all
+     * our known Look and Feels
+     */
+    public void updateAvailableLookAndFeels() {
```
uckelman, line 961: Private?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1137 @@
...
+      updateAvailableLookAndFeels();
+
+      if (panel != null)
+        return panel;
+
+      panel = new ConfigurerPanel(getName(), "[]rel[fill,grow]", "[]rel[fill,grow]"); // NON-NLS 
+      laf   = new JComboBox<>();
+      panel.add(laf, "grow");
+
+      updateAvailableLookAndFeels();
```
uckelman, line 1066: Why do this again? We just called this function 9 lines above.

uckelman:
The table cell borders are still broken when switching back to GTK+, as is some of the selection highlighting.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: `URLClassLoader.addUrl` is `protected`, so it has to be subclassed to make it accessible.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: Isn't an `ArrayList` backed by a plain-ol' array?  If that's the case, then appending to the list, unless there's something I don't know, involves reallocation of memory and copying from memory block to memory block.    In a linked list, append only involves allocation of one element and updating pointers.  True, traversing an array is faster than a linked list, but appending should be much faster (O(1)) compared to an array (O(n)). 

Maybe I got something wrong?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: Because the theme files may simply not be there and there's no reason to raise any error in that case.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: To make the `UIManager` aware of the `LookAndFeel` class, you _must_ _either_ use 
- `UIManager.installLookAndFeel`, _or_
- `UIManager.setLookAndFeel(LookAndFeel)` (not the string version). 

In the latter case, you need to make an object of  that class, which would be wrong here, and there's no reason to set a Look'n'Feel we may not want. 

So the next best thing is to try to load the class, and then check if that class is a sub-class of `LookAndFeel`, which is done here.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
uckelman: Could the entire subclass be this?
```java
private class LafClassLoader extends URLClassLoader {
    LafClassLoader() {
      super(new URL[0]);
    }
    
    @Override
    public void addURL(URL url) {
      super.addURL(url);
    }
  }
```

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: Because the first call is actually a no-op when the `JComboBox` named `laf` isn't set.   (i.e., when `panel` is `null`).  So we need to add the known L'n'Fs to the `JComboBox`.   If this method is called later, when `panel` is _not_ `null`, when we still need to update the `JComboBox`.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
uckelman: `ArrayList` uses array doubling when it gets full, so appending an element runs in amortized constant time. Typically, it's the poor memory locality of linked lists that makes them slow, though---appending to an already-allocated internal array is about the fastest thing you can do.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
uckelman: An `IOException` won't happen here because a file is missing, though---it will happen because the archive is corrupt or there was some sort of system error.

cholmcc: Wait.  Wrong comment.   Sometimes, `ZipFile` throws `IOException` for no apparent reason.  This is to get around that issue.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: No no, we're not iterating over _all_ the classes in the JAR - only the classes that the user selected. 

One _could_ of course loop over all classes in the JAR, try to load them, and check if they are sub-classes of `LookAndFeel`.  If such classes are found, then the L'n'F could be made available to the `UIManager` via `UIManager.installLookAndFeel` with-_out_ the user ever needed to select any class. 

_Edit_: `with` &rarr; `with-out` - kinda crucial.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: Good idea.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: `protected`?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: `protected`?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: `protected`?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: `protected`?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: `protected`?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: See [`URLClassLoader.addURL`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/URLClassLoader.html#addURL(java.net.URL))

cholmcc: You probably still need to set the parent `ClassLoader`.  Otherwise, yes - just did that.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: `protected`?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: protected?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: protected?

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: That is, 
- On the first call to `getComponent`, `panel` is `null` and 
  1. the first `updateAvailableLookAndFeels` is a no-op, 
  2. the second fills in the `JComboBox` 
- On subsequent calls to `getComponent`, 
  1. the first updateAvailableLookAndFeels` fills in the `JComboBox`
  2. the second is never done 

A consequence of the _get-out-early_ approach.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: Right. Done.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: No no, we're not iterating over all the classes in the JAR - only the classes that the user selected.

One could of course loop over all classes in the JAR, try to load them, and check if they are sub-classes of LookAndFeel. If such classes are found, then the L'n'F could be made available to the UIManager via UIManager.installLookAndFeel with-out the user ever needed to select any class.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: The cited code above removes the non-working classes from the user selection.  However, they are kept in the list of classes we already tried to load as L'n'F, so we wont try them again.

cholmcc:
> The table cell borders are still broken when switching back to GTK+, as is some of the selection highlighting.

Yes, and I haven't been able to pin-point the problem.  

At first, I thought perhaps the themes did not properly reset `UIManager` properties, but now I'm not so sure.  

I suspect that the `JXTreeTable` is somehow to blame.   For example, the problem corrected [here](https://github.com/vassalengine/vassal/pull/15095/changes#diff-0ab07530cd583849aff1205e335add63be71a202f8828247f2363417bffeaf1aR1200) is because of the `TableCellRenderer` does its own thing, which somehow messes things up. 

Also, the GTK+ L'n'F isn't the only L'n'F that has this problem, which leads me to suspect something else. 

One way to try to figure it out would be to make a MWE that uses `JXTreeTable` and selectable L'n'F, and then see if one can reproduce the problem. 

Yours,
_Christian_

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
uckelman: The default system class loader is the one you get when you use `URLClassLoader(URL[])`.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
uckelman: That should not happen. What do those exceptions look like?

cholmcc: As I remember, it said - no such file - but I might be wrong.

cholmcc: I agree that they were odd, and I certainly felt the exception to the exception was a bit odd.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: Then, no harm in being explicit ,😀

uckelman: The harm is more unnecessary code.

cholmcc:
Pushed an update. 

I found the reason for the spurious `IOException` - which is now handled and displayed to the user. 

The reason was, that `FileConfigurere` sets the file name to the string `"null"`, which fires off `PropertyChangeEvent`.  Thus, the listening code in `LookAndFeelConfiguerer.LafJarConfigurer.getClasses` tried to open a file named `"null"`.

I now protect against that in the `getClasses` method.  In the property change listener, I also protect against calling `updateClasses` if the filename has not changed - once the `updateClasses` have been run for a specific JAR filename, there's nothing more to do, and so there's no reason to scan the JAR again. 

I also formatted the error messages with HTML so that the JAR and class names are put into `<code>` blocks.

In the first go, I forgot to fix some violations.  Fixed. 

Yours,
_Christian_

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: See message below.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: OK, fixed.   Though in a sense, it can be good to be explicit :smile:

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: I've taken out the catch of `NullPointerException` - you are right that the compiler doesn't complain about that.   Wonder why I put it in in the first place.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
uckelman: Why `protected`? Anything which is `protected` we're stuck with because subclasses can see it.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1167 @@
...
+      catch (ClassNotFoundException          |
+             InstantiationException          |
+             IllegalAccessException          |
+             UnsupportedLookAndFeelException | 
+             ClassCastException              e) {
+        // This should very rarely happen 
+        ErrorDialog.show(e,
+                         "LookAndFeel.set_fail",
+                         getName(),
+                         getClassName());
```
uckelman, line 832: We should return here. If we failed to set the the LaF, then the update code below should not be run.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1167 @@
...
+  /**
+   * Holds information about a look'n'feel - Metal version
+   * loader
+   */
+  protected class MetalLafInfo extends LafInfo {
+    /**
+     * Construct our information
+     */
+    public MetalLafInfo() {
+    }
```
uckelman: This constructor isn't used.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
```diff
@@ -0,0 +1,1167 @@
...
+   */
+  protected class LafConfigurer extends Configurer {
+
+    private       JPanel               panel;
+    private       JComboBox<LafInfo>   laf;
+    private final Map<String, LafInfo> lafs = new HashMap<>();
+    private       boolean              updating = false;
+    private       LafInfo              previous;
+
+    LafConfigurer(String key, String name) {
```
uckelman: Missing `public`.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: fixed

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: fixed

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: fixed

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: Exactly so that someone may sub-class it if there's a genuine case for it.   It makes it much more difficult to do if these are `private` and one has to go back and change to `protected`.

uckelman: I posit that there is no genuine case for anyone outside the project subclassing anything at this point. Every bit of API surface area we add makes repairs harder.

cholmcc:
> > The table cell borders are still broken when switching back to GTK+, as is some of the selection highlighting.
> 
> Yes, and I haven't been able to pin-point the problem.
...
> One way to try to figure it out would be to make a MWE that uses `JXTreeTable` and selectable L'n'F, and then see if one can reproduce the problem.

See the attached code:  [Baz.java](https://github.com/user-attachments/files/32488545/Baz.java)

Compile with 
```
$ javac -cp /path/to/swingx.jar Baz.java
```

Run with 
```
$ java -cp `pwd`:/path/to/swing.jar Baz
```
or, to select the initial Look'n'Feel - say _Meta_
```
$ java -Dswing.defaultlaf=javax.swing.plaf.metal.MetalLookAndFeel  `pwd`:/path/to/swing.jar Baz
```

If you start it up without specifying a L'n'F, it will start up in the system default (e.g., _GTK+_).  A menu bar entry let yuo choose between installed L'n'F. 

If you switch to another L'n'F and then (back) to _GTK+_, you will see the lines in the the `JXTreeTable` (top).   Also switch to _Metal_ or _CDF/Motif_ and see spurious lines in the table.   Switch to _Nimbus_ and note that the rows are not alternating in colours.  

All of these are probably not as intended. 

Now start with 
```
-Dswing.defaultlaf=javax.swing.plaf.nimbus.NimbusLookAndFeel
```
and note that the rows are initially of alternating colours. 

Now start with 
```
-Dswing.defaultlaf=javax.swing.plaf.metal.MetalLookAndFeel 
```
and note that the spurious lines never appears if you switch back-and-forth between some other L'n'F and _Metal_.  

Also very puzzling 

The point is, that it seems to be a more deep problem, and one that can most likely not be fixed in Vassal. 

Again, as you restart the application, the initial L'n'F work as expected. 

Yours,
_Christian_

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
```diff
@@ -1189,6 +1196,11 @@ public CenteringCellRenderer() {
     @Override
     public Component getTableCellRendererComponent(JTable table, Object value, boolean isSelected, boolean hasFocus, int row, int column) {
       super.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, column);
+
+      // Do not set colours - let the tree decide
+      setBackground(null);
+      setForeground(null);
+      setBorder(null);
```
rlament: > The table cell borders are still broken when switching back to GTK+

Have you tried `setGridColor`?
```
      table.setGridColor(getBackground());
      // Do not set colours - let the tree decide
//      setBackground(null);
//      setForeground(null);
//      setBorder(null);
```
This hides the grid for all LaF's, but this could be more selective. This doesn't address the highlight issue.

cholmcc: I tried with `UIManager.put("Table.gridColor","white")` when changing the L'n'F - no effect.  

> This doesn't address the highlight issue.

Not sure what you mean by that one.

cholmcc: BTW, the `setForeground` and `setBackground` seems to be absolutely needed - otherwise you can get stuff left over from one L'n'F to another - in particularly in the customised columns.

cholmcc: `table.setGridColor(null);` works great.   It removes the grid lines after changing L'n'F in all L'n'Fs.  Added to PR.

vassal-app/src/main/java/VASSAL/configure/LookAndFeelConfigurer.java
cholmcc: Changed as many internal classes to be `private static` as possible.  Others to `private`.

uckelman:
The table cell borders appear to fixed when changing back to GTK+ now.

What's still broken is the cell selection highlighting when going GTK+ > Numbus > GTK +.

image

vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java
```diff
@@ -1189,6 +1196,12 @@ public CenteringCellRenderer() {
     @Override
     public Component getTableCellRendererComponent(JTable table, Object value, boolean isSelected, boolean hasFocus, int row, int column) {
       super.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, column);
+
+      // Do not set colours - let the tree decide
+      table.setGridColor(null);
+      setBackground(null);
+      setForeground(null);
```
uckelman, line 1203: The above two lines are preventing selection highlighting for the centered table columns.

cholmcc: Yes, but they also make sure that the old f colours do not spill over, so there's some conflict there. It seems that the custom cell renderer is somehow the culprit. Whether there's a way to fix it without being very explicit, and basically reimplement the default renderer, I'm not sure. 

But please note that this is really transient problem, because it only happens when you switch LnF, and all is good if you restart the app, the problem is gone. As such, I would classify this as a minor problem, however annoying.

uckelman: > Yes, but they also make sure that the old f colours do not spill over, so there's some conflict there.

What steps do I take to see this happen?

rlament: This function is for tweaking the appearance of cells with centered text. To make adjustments after a look and feel, add an override of `updateUI()`.

`table.setGridColor(null)`: Looking at Oracle docs, this is supposed to throw IllegalArgumentException  if gridColor is null. That doesn't appear to happen. Maybe docs are wrong or Java is not compliant. Either way, best to avoid using.
```
@Override
public void updateUI() {
  super.updateUI();
  if (UIManager.getLookAndFeel().getClass().getName().equals("com.sun.java.swing.plaf.gtk.GTKLookAndFeel"))
  {
    // Turn off the table grid if the previous LaF had enabled it.
    setShowGrid(false);
  }
}
```
Other than seeing that the highlight background is the wrong color, I'm also not understanding: _make sure that the old f colours do not spill over_

cholmcc: > This function is for tweaking the appearance of cells with centered text. 

That's pretty obvious  - also you mean `class` right.  There's no functions in Java - only classes and methods :smile: 

> `table.setGridColor(null)`: Looking at Oracle docs, this is supposed to throw `IllegalArgumentException` ...

Yes, [`JTable.setGridColor`](https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/javax/swing/JTable.html#setGridColor(java.awt.Color) should throw in case of a null argument.  Apparently it doesn't. 

About overriding [`DefaultTableCellRenderer.html.updateUI`](https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/javax/swing/table/DefaultTableCellRenderer.html#updateUI()), did you try that?  Perhaps with the above attached [`Baz.java`](https://github.com/user-attachments/files/32488545/Baz.java)?  If so, did that work satisfactory?    Also, why only for the GTK L'n'F?  Why not for all L'n'F?   It seems like @uckelman dislikes the borders all together. 

Are you sure that [`JTable.setShowGrid`](https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/javax/swing/JTable.html#setShowGrid(boolean)) is enough?   The GTK+ L'n'F normally sets this to `false`, so it may not be enough.   

> Other than seeing that the highlight background is the wrong color, I'm also not understanding: _make sure that the old f colours do not spill over_

If the `setBackround` and `setForeground` methods are not sent, then the component may keep the foreground and background colours of the old L'n'F.     This _only_ happens in cells that have an explicit `TableCellRenderer` instance set.   All others will get the right colours.    So somehow, the `DefaultTableCellRenderer` stores the old L'n'F colours.   A way to avoid that, is to set those two colours to `null`.   Of course, that means, as @uckelman points out above, that the selection highlighting colours are also set to `null`, which is _a different_ adverse effect. 

I've made a [snippet](https://gitlab.com/-/snippets/6060630) to play around with these things and try different implementations.  Perhaps you guys could try to make changes there to see if we can come up with a solution. 

I really consider these things as _minor_ issues.   You will _only_ see these issues when you switch between L'n'Fs, and as soon as you restart the application, they will be gone (rows will be highlighted correctly, cell borders will be gone).   In other words, the PR could be applied as is without a lot of issues cropping up for users, and then we can in parallel see if we can find a solution.  User's should of course be made aware that they _may_ see some funny stuff when switching between L'n'Fs and the way to remedy that is to restart the application.   This could be part of the UI. 

Yours,
_Christian_

cholmcc: Note, the implementation of [`DefaultTableCellRenderer.updateUI()`](https://github.com/eagle518/jdk-source-code/blob/master/jdk5.0_src/j2se/src/share/classes/javax/swing/table/DefaultTableCellRenderer.java#L112C1-L124C6) sets the foreground and background colours to `null`. 

Also, [`JTable.updateUI`](https://github.com/eagle518/jdk-source-code/blob/master/jdk5.0_src/j2se/src/share/classes/javax/swing/JTable.java#L2756) sets the `TableUI` _after_ the above messages. 

`JTable.setShowGrid` cannot be sent from `DefaultTableCellRenderer.updateUI()`, because the `DefaultTableCellRenderer` doesn't have a pointer to the `JTable` instance (and probably shouldn't have). 

Yours,
_Christian_

rlament: Sorry, I should have been more specific not realizing updateUI could be overridden in multiple classes. It should be in the `MyTree` class. And, yes I did try it.

> why only for the GTK L'n'F? Why not for all L'n'F?

Depends on where the grid problem shows up. If only GTK then we can filter on it. If on all then that check is not necessary. The point is that any "fixes" can be Look and Feel specific.

cholmcc: >  It should be in the MyTree class. 

Ah, I see. 

> The point is that any "fixes" can be Look and Feel specific.

Preferably not.  

See also the [snippet](https://gitlab.com/-/snippets/6060630).  I think I found the best way to deal with the issue of changing row column background colours, selection colours, etc.  Note, there might need to be some additional overriding the [`ModuleManangerWindow.MyTreeCellRenderer`](https://github.com/cholmcc/vassal/blob/master/vassal-app/src/main/java/VASSAL/launch/ModuleManagerWindow.java#L1159) class  _a la_ what is done in the snippet. 

Yours,
_Christian_

PR #15097, opened 2026-09-19
Make Charts zoomable

uckelman:
This incorporates some of #14579, but adds defined zoom levels instead of bounds. The reason for fixed zoom levels is that floating point imprecision can mean you don't hit exactly the same zoom levels zooming in as zooming out when computing zoom by repeatedly multiplying, which makes tile caching work poorly.


vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
```diff
@@ -388,274 +396,7 @@ public static class LevelConfig implements ConfigurerFactory {
...
-    }
-
-    @Override
-    public void setValue(String s) {
-    }
-
-    @Override
-    public String getValueString() {
-      return null;
+      return new VASSAL.configure.LevelConfigurer(((Zoomer) c).state, key, name);
```
cholmcc, line 399: Minor thing  - you probably don't need the fq-class-name here.

vassal-app/src/main/java/VASSAL/build/module/map/Zoomer.java
```diff
@@ -1396,4 +1144,15 @@ public boolean isMandatory() {
   public boolean isUnique() {
     return true;
   }
+
+  @Deprecated
```
cholmcc, line 1148: Oh, now I see why you need the fqcn above - but why is this kept?  Will it ever be used in the code?

uckelman: I have no intention of using it again, but there could be custom code which already has.

cholmcc:
Looks good, both in the code and in execution.

I was wondering if there was a point in binding a key - default say "1" - to reset the zoom level to the initial value?

Yours,
Christi

vassal-app/src/main/java/VASSAL/build/widget/Chart.java
```diff
@@ -52,25 +73,123 @@
...
+
+    @Override
+    public List<Double> getLevels() {
+      final List<Double> l = new ArrayList<>(levels.length);
+      for (final double d : levels) l.add(d);
+      return l;
+    }
+
+    @Override
+    public void reset(List<Double> l, int i) {
```
rlament, line 166: This class should have some range checking to ensure `cur` and `initial` stay within the array bounds.

Shouldn't this function update `cur` to remain within array bounds? Otherwise `getZoom` will throw index out of bounds.

uckelman: `cur` is set now. Thanks for catching that.

`LevelConfigurer` already ensures that `initial` is within the array bounds.

uckelman:
@cholmcc Would you mind taking a new screenshot for the docs, as #14579 had?

cholmcc:
ChartEditor

uckelman:
@cholmcc That's missing the Initial hotkey. Is this a screenshot from current master?

cholmcc:
> @cholmcc That's missing the Initial hotkey. Is this a screenshot from current master?

No, from what I took down yesterday for testing.   Update:
ChartEditor

cholmcc:
BTW, why the change to `Ctrl-` prefix on the keys?   I think it's quite nice to be able to simply do `+`, `-`, and `1`, for example.

uckelman:
Ctrl is standard across many applications for changing scale, as is 0 for full-size.

rlament:
> Ctrl is standard across many applications for changing scale, as is 0 for full-size.

It would be nice to also use the numpad plus and minus keys: `Ctrl+Num+Plus` and `Ctrl+Num-Minus`. Perhaps a second set of zoom in/out keys?

Zoom keys should be common to all charts in a module, not per chart. If the module designer wants to change hotkeys it should be in one place not in every chart.

Typo: "Initial zoom key" change _key_ to _hotkey_.

Unnecessary quotes around the sentence _An '*' indicates the initial zoom level_.

cholmcc:
> Zoom keys should be common to all charts in a module, not per chart. If the module designer wants to change hotkeys it should be in one place not in every chart.

Perhaps, if that is implemented, then it should be set through the _File_&rarr;_Preferences_. 

_Edit_: There's little chance of conflicts with Module defined keys, because keys typed in a `Chart` window will not propagate to `Map` or similar windows.  In that sense, it is perfectly safe to allow the user to set general keys for zooming _all_ `Chart` windows.   

The Levels should probably still be set per-`Chart` 

Yours,
_Christian_

uckelman:
> It would be nice to also use the numpad plus and minus keys: Ctrl+Num+Plus and Ctrl+Num-Minus. Perhaps a second set of zoom in/out keys?

Ideally there would be a single place to associate actions with keys, and they'd be user-configurable. I don't see a tidy way to do that in V3.

> Zoom keys should be common to all charts in a module, not per chart. If the module designer wants to change hotkeys it should be in one place not in every chart.

I would go further and say they should be common to all Charts _and_ Maps.

> Typo: "Initial zoom key" change key to hotkey.

Fixed.

cholmcc:
> Ideally there would be a single place to associate actions with keys, and they'd be user-configurable. I don't see a tidy way to do that in V3.

What's wrong with _File_&rarr;_Preferences_?  There could be a separate tab - say _Short-cuts_ - which would configure such things as zoom, undo, ..., and so on.  Now, some modules may already set up specific keys for some of these, and then one would have to find a way to accomodate that, but for manyt things, it would be straight forward, as there's already a general setting.   A setting could also be _Respect Module settings, if defined_. 

Yours,
_Christian_

rlament:
> What's wrong with _File_→_Preferences_?

_File_→_Preferences_ is common to all modules. Module designers are free to associate actions to almost any key or key combination. A definition in _File_→_Preferences_  may work for most modules while conflicting with others. _Global Options_ already defines some keys to specific actions. Adding zoom here makes it module specific. Unfortunately it is not an obvious place to look for it.

> Zoom keys should be common to all charts in a module

Right, and maps too. Something I might try: Declare the zoom keys static in Zoomer. Changing the definition in one place changes them all.

uckelman:
> Declare the zoom keys static in Zoomer. Changing the definition in one place changes them all.

Mutable static members should not exist at all. I won't merge anything that adds one.

uckelman:
I'm not convinced it's worth pursuing this further. However, should someone wish to:

1. Moving the hotkeys from Chart to Chart Window would be an improvement.
2. Moving the hotkeys from Chart out to the module preferences would be an improvement over 1.
3. Moving the hotkeys from Chart out to app preferences would be an improvement over 2.

I have a hard time believing many people will change the Chart hotkeys from their defaults, so I expect the impact of doing any of these to be low.

PR #15103, opened 2026-09-22
Game Library Browser UI

cholmcc:
User facing:
- This adds a UI to browse the VASSAL Game Library, and download files from the library to the local machine.
- The data is structured according to the Game Library structure.
  - Project
    - Package(s)
      - Release(s)
        - File(s)
- A user may select files and mark from for download in the far left column, by checking the check-mark.
- If a user marks a full release, package, or project, then _all_ files under that will be marked for download.
- When a user has selected the files to download, then the user can press the _Download files_ button.
  - The user will then be prompted to select a download directory.
  - A progress bar will show the current progress.
- Modules that are already registered in the _ModuleManager_ are _not_ available for download.
- Downloaded modules are registered with the _ModuleManager_.
- The interface is available as _Library browser_ in the _Tools_ menu of the _Module Manager_.
- Note, loading the initial list of projects may take some time.
- Progress bars will be shown when retrieving data and downloading files.

Technical:

- While the list of _all_ projects are downloaded initially, the individual project detailed information is downloaded lazily.
- Downloaded modules are recognised by their file basename.
- Note, that there's no way of detecting already-downloaded modules if they have been renamed on the users computer.
- Similarly, there's no reliable way to recognise which module extensions belong to.  The user must therefore themselves add the extension to the module.
- Loading the initial list or projects takes some time (3000+ projects, done 10 at a time).  However, getting module details is fast.
- Note, the module manager will overwrite the written module list if it is running at the same time.  There should probably be a way to re-read in the module list.
- The UI can have many improvements:
  - A _File_ menu
  - A _Close_ button
  - Configurables:
    - Number of projects to get - initially - at a time.
    - Can we download already known modules?
    - ...


cholmcc:
Start the UI with 

```
$ VASSAL.sh --library-browser
```
(Linux and Mac), or from _Tools&rarr;Browse library..._ in the _Module Manager_. 

LibraryBrowserToolsMenu


When loading the project, a progress bar will be shown - loading takes a few seconds 

LibraryBrowserLoading

Once the project list has been loaded, we have all the projects listed.  Note that the size is set to `?` because we do not know - yet - how much space is taken up by the projects

LibraryBrowserStart


When we expand a project, then the project details are loaded and can be displayed.  The `Size` column accumulates sizes of its children, and the `Last update` column takes the latest date of its children.   Selecting a file will mark it for download.   Files (match by file name) that are already know to the Module Manager are already marked and will not be downloaded. 

LibraryBrowserProject

Yours,
_Christian_

cholmcc:
Previous force push had wrong commit message - updates.

vassal-app/src/main/java/VASSAL/library/FileName.java
```diff
@@ -0,0 +1,63 @@
...
+ * You should have received a copy of the GNU Library General Public
+ * License along with this library; if not, copies are available
+ * at http://www.opensource.org.
+ */
+package VASSAL.library;
+
+/**
+ * A file name
+ */
+public class FileName {
```
uckelman, line 22: Is there some reason for creating this class instead of using `java.nio.file.Path`?

vassal-app/src/main/java/VASSAL/library/JSON.java
```diff
@@ -0,0 +1,540 @@
...
+ */
+package VASSAL.library;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A simple JSON parter.  The parser return an Item object which may
```
uckelman, line 25: We should not be writing our own JSON parser without a good reason. There are several battle-tested JSON parsers to choose from in Java, e.g., Jackson.

vassal-app/src/main/java/VASSAL/library/GameLibrary.java
```diff
@@ -0,0 +1,323 @@
...
+    }
+    return total;
+  }
+  
+  protected String getJSON(URL url)  throws IOException {
+    final HttpURLConnection connection = getRequest(url, "application/json"); //NON-NLS
+    
+    try (InputStream in = connection.getInputStream()) {
+      final ByteArrayOutputStream out = new ByteArrayOutputStream();
+      copyResponse(in, out);
```
uckelman, line 133: `InputStream.transferTo()` is the fast way to dump an `InputStream` into an `OutputStream`.

Since you want a `String`, even simpler would be
```
return new String(in.readAllBytes(), StandardCharsets.UTF_8));
```

vassal-app/src/main/java/VASSAL/library/GameLibrary.java
```diff
@@ -0,0 +1,323 @@
...
+      return out.toString(StandardCharsets.UTF_8.name());
+    }
+    finally {
+      connection.disconnect();
+    }
+  }
+
+  protected void getFile(URL url, File outname, String sha256, long size)
+    throws IOException {
+    final HttpURLConnection connection = getRequest(url, null);
```
uckelman, line 144: The try...finally which closes the connection needs to start immediately after the connection is opened to ensure that it's closed on all execution paths.

vassal-app/src/main/java/VASSAL/library/GameLibrary.java
```diff
@@ -0,0 +1,323 @@
...
+    connection.setInstanceFollowRedirects(true);
+    connection.setRequestProperty("User-Agent", AGENT); //NON-NLS
+    if (accept != null && !accept.isBlank()) 
+      connection.setRequestProperty("Accept", accept); //NON-NLS
+
+    final int code = connection.getResponseCode();
+    if (code == 404)
+      throw new IOException("not found (404): " + url);
+    if (code / 100 != 2) 
+      throw new IOException("HTTP " + code + " from " + url);
```
uckelman, line 101: These two throws will cause the connection to fail to be closed.

vassal-app/src/main/java/VASSAL/library/GameLibrary.java
```diff
@@ -0,0 +1,323 @@
...
+    catch (ParseException ignored) {
+    }
+    return (new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSX")).parse(s); //NON-NLS
+  }
+
+    
+  /**
+   * Do a GET request
+   */
+  protected HttpURLConnection getRequest(URL url, String accept) throws IOException {
```
uckelman, line 87: We deprecated the only thing using `HttpURLConnection` (`VASSAL.tools.HTTPPostBuilder`) in 201 in favor of the Apache httpclient5 library. If you search for `client5` in the code, you will find some examples of where that's used.

uckelman (changes requested):

vassal-app/src/main/java/VASSAL/library/SemVer.java
```diff
@@ -0,0 +1,239 @@
...
+
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
+import java.text.ParseException;
+
+/**
+ * A semantic version.  This is fully parsed by the class, and we
+ * can sort releases.
+ */
+public class SemVer implements Comparable<SemVer> {
```
uckelman, line 27: There must be a semver parsing library that we could use instead of implementing it ourselves.

uckelman:
> Loading the initial list or projects takes some time (3000+ projects, done 10 at a time). However, getting module details is fast.

A better approach would be to ship a project list that's complete at the time to populate a cache, query only for projects updated since the previous check, and to ask for more than 10 at a time. (IIRC, the default limit is 50.)

cholmcc:
> A better approach would be to ship a project list that's complete at the time to populate a cache, query only for projects updated since the previous check, 

I don't think the public API allows for such a query - i.e., projects that have an modification time after some point. 

> and to ask for more than 10 at a time. (IIRC, the default limit is 50.)

How would you specify the number of projects to get at a time?   I pass no query parameter, so the 10 I get must be the default. 

Yours,
_Christian_

vassal-app/src/main/java/VASSAL/library/SemVer.java
cholmcc: For example: https://github.com/zafarkhaja/jsemver

vassal-app/src/main/java/VASSAL/library/JSON.java
cholmcc: OK.  Used `jackson-databind`

The main reason to use a home-made implementation (both of this and SemVer) is to avoid (yet-another) third-party dependency for something that is fairly trivial. 

We don't need the full bells and whistles of `jackson` or the fancy parsing of https://github.com/zafarkhaja/jsemver.   We need something that can handle the limited data we get from the Game Library and which has exactly the fault tolerance we need.  No more, no less.

uckelman:
> I don't think the public API allows for such a query - i.e., projects that have an modification time after some point.

`sort_by=m` will give you the items sorted in order of descending modification time.

> How would you specify the number of projects to get at a time? I pass no query parameter, so the 10 I get must be the default.

`limit` is the parameter name. The [search page](https://vassalengine.org/library/projects) fetches 50 at a time.

vassal-app/src/main/java/VASSAL/library/GameLibrary.java
cholmcc: This only makes sense for the JSON queries.  When download files, we want to the slightly round-about way to calculate the message digest as we go.

vassal-app/src/main/java/VASSAL/library/FileName.java
cholmcc: Main reason was to investigate the file ending.   Probably a bit of overkill - `String` will do just fine.

vassal-app/src/main/java/VASSAL/library/GameLibrary.java
uckelman: For the file hashing, you might want a `FilterInputStream` or `FilterOutputStream` subclass for that. However, it might turn out to be faster to reread the files from disk to hash them. YMMV.

uckelman: I had forgotten what they're called: `DigestInputStream` and `DigestOutputStream`.

cholmcc:
> > I don't think the public API allows for such a query - i.e., projects that have an modification time after some point.
> 
> `sort_by=m` will give you the items sorted in order of descending modification time.
> 
> > How would you specify the number of projects to get at a time? I pass no query parameter, so the 10 I get must be the default.
> 
> `limit` is the parameter name. The [search page](https://vassalengine.org/library/projects) fetches 50 at a time.

The maximum `limit` parameter value seems to be `100`.   Would probably speed up load time if this could be set higher (say 500). 

With the increased number of returns per query, the time to get the projects is quite manageable - a little over 8 seconds.

Also, I think if the query is to be cached, then a better choice than to ship a file with Vassal, would be to write a file in `vassal.confdir` - say `projects.properties` - which is then updated after each query.   In that way, there's no need to manage the additional information in Vassal, as the user will be in charge of the data.   Of course, the first invocation will take longer to do then subsequent ones. 

I will update the PR without caching.  If you really think it is important, then I may do the above - cache in a user file. 

Yours,
_Christian_

cholmcc:
Last few pushed were f**ked up because I tried to merge `master` and got it all wrong.  Hopefully this last push will fix the problem.

vassal-app/src/main/java/VASSAL/library/LibraryFile.java
```diff
@@ -0,0 +1,135 @@
...
+public class LibraryFile {
+  protected final String   filename;
+  protected final URL      url;
+  protected final String   sha256;
+  protected final Date     date;
+  protected final long     size;
+  protected final Version  version;
+  protected final Version  requires;
+  
+  LibraryFile(String   fn,
```
uckelman: The ctor should be public if the class is.

vassal-app/src/main/java/VASSAL/library/Package.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.ArrayList;
+
+/**
+ * A package.  Packages are named and contain releases of files.
+ */
+public final class Package {
+  public final String name;
+  public final List<Release> releases = new ArrayList<>();
+  
+  Package(String name) {
```
uckelman: The ctor should be public if the class is.

vassal-app/src/main/java/VASSAL/library/Package.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.List;
+import java.util.ArrayList;
+
+/**
+ * A package.  Packages are named and contain releases of files.
+ */
+public final class Package {
+  public final String name;
+  public final List<Release> releases = new ArrayList<>();
```
uckelman: These members either should be private or there should not be getters for them.

vassal-app/src/main/java/VASSAL/library/Release.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+import java.util.ArrayList;
+import com.github.zafarkhaja.semver.Version;
+
+/**
+ * A release.  A release contains a number of files.
+ */
+public final class Release {
+  public final Version number;
+  public final List<LibraryFile> files = new ArrayList<>();
```
uckelman: These members either should be private or there should not be getters for them.

vassal-app/src/main/resources/VASSAL/i18n/VASSAL.properties
```diff
@@ -890,8 +890,51 @@ LaunchRequest.extension=Edit a module extension
...
+LibraryBrowser.extension=Extension
+LibraryBrowser.log=Log
+LibraryBrowser.aux=Auxiliary
+LibraryBrowser.name=Name
+LibraryBrowser.type=Type
+LibraryBrowser.last_update=Last update
+LibraryBrowser.size=Size
+LibraryBrowser.url=URL
+LibraryBrowser.inconsistent_read=Read %1$s bytes, expected %2$s
+LibraryBrowser.inconsistent_checksum=SHA-256 %1$s, does not expected %2$s
```
uckelman: Does not match?

vassal-app/src/main/java/VASSAL/library/GameLibrary.java
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  protected Date parseDate(String s) throws ParseException {
+    try {
+      return (new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX")).parse(s); //NON-NLS
+    }
+    catch (ParseException ignored) {
+    }
+    return (new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSX")).parse(s); //NON-NLS
+  }
```
uckelman: Thee above four functions could be static. There might also be some existing classes in V, in the JDK, or in httpclient5 which handle assembling URLs.

vassal-app/src/main/java/VASSAL/library/GameLibrary.java
```diff
@@ -0,0 +1,368 @@
...
+    try {
+      if (sha256 != null && !sha256.isBlank())
+        digest = MessageDigest.getInstance("SHA-256"); //NON-NLS
+    }
+    catch (Exception ignored) {
+    }
+    
+    final InputStream  in   = getRequest(url, null);
+    final OutputStream out  = Files.newOutputStream(outname.toPath());
+    final long         read = copyResponse(in, out, digest);
```
uckelman: Could be replaced by wrapping the `InputStream` in a `DigestInputStream`.

vassal-app/src/main/java/VASSAL/library/Project.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Date;
+
+/**
+ * A whole project.
+ */
+public final class Project {
+  public final String name;
+  public final String title;
+  public final Date   date;
+  public final List<Package> packages = new ArrayList<>();
```
uckelman: Why public members and getters?

uckelman (changes requested):

